PR #12, opened 2017-10-05
Improve cache

project-am:
By utilizing an object, you actually can reference the images directly, which could be optimized to a Hashmap by the JS VM, resulting in O(1) ;)

Also the source code is less complex, shorter and has a better overview!


Ben987:
I love it!  It looks very simple, I'll try it soon :)

PR #17, opened 2017-10-11
Natalie fixes for evil end and a spank error

wildsj:
That has the images working right for me now during evil end, plus attitude change from spank now works.

During stages where the images should only ever be determined by the script.js, should I have a just in case image in Intro.csv or leave those cells blank?


Ben987:
I usually always put an image in Intro.csv just in case but that's up to you.  It depends how your code is done.

Thanks for the fixes, I'll try them soon.  Is everything ready to publish on your side?

wildsj:
Provided there are no more bugs, I think it's ready. There are odd places where I feel the flow and options could be improved a little, but are probably not worth it.

Ben987:
Alright!  I'll do a few more tests and I'll publish your character pretty soon.

Thanks again for you hard work :)

PR #18, opened 2017-10-11
[FR Translation] Small corrections chapter 2, 3 and egg item

gottspende:


gottspende:
Small corrections before chapter 4 !

Ben987:
Merci beaucoup :)

Issue #10, opened 2017-10-03
Clean source code


project-am:
Iirc, you wrote that you would like to use this project to learn JS. Since I alrady have a decent amount of JS experience, I would like to help you by providing my opinions and maybe even cleaning up some of your source (if that is ok with you), though as I wrote, I won't have a lot of time over the next few weeks.

I have already started becoming a bit familiar with your source code, so I see a few quick-gains in terms of architecture and a few improvements concerning your draw-related code. Before doing all that stuff, though, I want to check if you actually want to have such PRs...

Most of the mentioned fixes are good for mobile data volume, battery life and performance, while they should also greatly improve the code architecture, making the program more generic, hence easier to extend later on (in case you want to add a feature, or make the day repeatable, maybe randomly putting together classes)

## Quick improvements

- [x] cache downloaded files instead of re-downloading them whenever you need them (mobile data volume!)
    - [x] images
    - [x] CSV files
- [ ] switch all scripts to [strict-mode](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Strict_mode)
- [ ] use double-buffering
- [ ] instead of creating functions and variables in the global scope, use a namespace for your application
- [ ] put all functions into objects (by chapter?) and just use `window.app.c003.routines[dynamicCall]();` instead of creating a function by string concatenation and creating a new function object
- [ ] get a linter (for example eslint) with a sane preset (I like the airbnb one)
- [ ] use a build-step to bundle and optimize the code (for example using webpack)
- [ ] use a tree for you interface, in which you register areas, so you can find out the clicked area very quickly in a generic manner (at the moment you run the check on every single thing)
- [ ] remove double-toggles. For example, you have `C002_FirstClass_Classroom_MildredSubdueSuccess` and `C002_FirstClass_Classroom_MildredSubdueFailed`, which might lead to the ambitous situation of both being true. Instead, make an enum-like structure and fill in a number for the status. Alternatively, use Monadic Options (so you can have `Some(true)`, `Some(false)` or `None`) or in the worst case `null`.
- [ ] store actor-info on the actor instead of using separate variables for each chapter
- [ ] (if someone has terrifying performance problems only!) use Numbers to work with actors and items instead of doing string comparisons everywhere
- [ ] use instantiated objects with explicit fields for actors instead of an array
- [ ] make the game a [progressive app](https://developers.google.com/web/progressive-web-apps/)

## Big changes

[...]

PR #22, opened 2017-10-19
implement simple CSV caching

project-am:
for #10


Ben987:
Very nice!  I've been thinking of doing it since I was caching images.  Thanks :)

PR #28, opened 2018-01-06
Kinbaku Club Part 1

wildsj:
This should have the intro and club room layout, with Erica complete. I left the shortcuts I use in the intro, as there is also the blind-mans buff game if the player blindfolds themselves in the club room and no other way to get a blindfold at the moment.

One thing I noticed when uploading was the fight.js is no longer in the top level of your master directory, I haven't investigate yet whether that is an issue, no way is Erica going to just let someone get the better of her without a fight.


wildsj:
I rushed adding the changes to the drawing.js with copy and paste from an old set of files I had been working with and changed the scaling inadvertently and only realised when the doing a pull request. I tried to correct it but might have got that wrong as well. I have no intention of changing the scaling, only another part of that line of code for the collar and blindfold.

PR #30, opened 2018-02-04
Kinbaku Club Lauren

wildsj:
New character to complete the blind mans buff and a load more changes to the Kinbaku Club.

Changes include save system to give correct Intro for the club and links to follow on after Drama and Revenge chapters.

Cuffs now have an escape option for using a shim and option for when the player wouldn't be able to reach their key. This should fix the escaping cuffs bug with Natalie at lunch break.


wildsj:
I just downloaded my fork to test it and there is a mayor bug with it, probably a mistake with uploading the files.

PR #34, opened 2018-02-13
Fix Bug and Update Translation on Stage_CN.csv

earthquake02:
Fix Bug and Update Translation.
And then..The first time I make some mistake but it has recovered.
If it works this time,I wanna continue translate more things.


Ben987:
It works great!  Thanks a lot, I hope you can do more :)

The only thing is to make sure you put the correct capital/lower letters for functions

I had to change UnCollar to Uncollar

Everything else seems perfect, though I cannot read Chinese

PR #35, opened 2018-03-08
[Bondage College] French Translation - Chapter 4

gottspende:


Ben987:
This is huge!  Thanks a lot gottspende, Chapter 4 isn't a small task to translate.

I had a great time playing in French, it will be integrated in the next release.

Merci beaucoup :)

PR #36, opened 2018-03-12
[Bondage College] French Translation - Chapter 5

gottspende:
I confirmed it was more easier. ;)


Ben987:
This is great, another chapter in French!  And thanks for reviewing the text I've translated :)

PR #61, opened 2018-06-11
Chapter 12 full german and chapter 11 a bit german

fleisch11:
i think i made a mistake with the last pull request, so please ignore the last one


Ben987:
This is great!  Congrats on your first pull request.

The accents don't seem to work though.  All the foreign letters shows as question marks.  I'm not sure how you push foreign characters in GitHub but many people were able to do it so there must be a way.  You can ask Gottspende, maybe he can help you.

Also, have you noticed the pipes " | " characters in the English version?  These pipes are used to split the phrase in two lines.  You will need them so the text don't overflow.

Good luck!

gottspende:
You need to keep the encoding UTF-8 to display the accents, you can use an editor as notepad++ for example on your file and force the encoding. I see some files of yours and the encoding is "ANSI" that may be the problem.

PR #66, opened 2018-10-08
German Chapter 3

adokilume:


Ben987:
Wow!  You're really fast, this is impressive.  The game will be fully available in German in no time, congrats :)

If you plan on doing the full game, I would love to include you in the game profit sharing from Patreon with the other main developers and translators.  It's not a huge amount but I want to motivate and thank anyone who works hard on the projects.

Thank you!

adokilume:
Wow, thank you so much for offering, I wasn't really expecting that! I'm not sure if I'd be comfortable getting money for this sort of thing. Mainly because it's just fun to do this, but also because I don't know if that would be considered a side job so I'd have to deal with taxes and stuff, which I'd rather not have to worry about for now.
It's funny, I actually had a thought in the opposite direction and joined your Patreon yesterday, it kind of would have felt werid not to after basically having access to the cheaters version for free via the source code. :P But I also really wanted to take part in the polls.

Just to be clear, your appretiation already is way more than I was hoping to get out of this. :)

Also, the later chapters might take a bit longer to translate, but at this point I'm still determined to finish this. I also have another idea I could contribute other than translations, but first I'll have to see if that would actually work.

Ben987:
It's fine.  If you do want to go ahead with the next chapter and change your mind about the money, I'll be glad to include you in the profit sharing.  Yeah, since it's a revenue you might need to declare it and pay taxes.

Thanks a lot for your Patreon support, seeing the poll will also give you an overview of all the possible scenarios for the game :)

No rush for the later chapters, some of them are text heavy so they will certainly take longer to translate.

Danke!

PR #67, opened 2018-10-10
German Chapters 4 & 5

adokilume:
Had some free time today, so I was able to finish this one a bit sooner than expected. I hope this doesn't feel spammy, I could also do larger commits/PRs with more time between them if that would be better.

A few minor questions came up for me in this chapter:

1. Does the line "call me Julia" imply that Julia wants her students to be on a first-name-basis with her (if you can even call it that since nobody seems to have a last name)? For now I assumed that it does. It kind of matters because in german you adress people with different words depending on how formal your relationship is. (This is actually going to be a challenge when submissives call other characters "Miss" more often, because their way of adressing them would probably have to switch back and forth, but I guess that's something for me to worry about when I get there.)

2. Is Julia supposed to have an accent? I was under that that impression when I was first playing, but now I realized she talks rather normally, just with a few italian words thrown in. It wouldn't change much in the translation, but still might be good to know.

3. Is there a list of all the "gag speech" in this game, and what it's actually supposed to mean? I get that most of it is probably just generic mumbling, but "Eep ah hoze eew wuwii!" looks like "[Something] new pupil" and I just can't decipher the first part. It would be nice to be able to "translate" this into gag speech of the german phrases, if that makes sense.

**Edit:** Added chapter 5, since it's rather short.

### DIFF
--- a/C004_ArtClass/HiddenEgg/Intro_DE.csv
+++ b/C004_ArtClass/HiddenEgg/Intro_DE.csv
@@ -1,0 +1,4 @@
+Stage,LoveReq,SubReq,VarReq,IntroText,Image
+0,0,0,Common_PlayerNotRestrained,Hinter der roten Farbe scheint|etwas Pinkes versteckt zu sein.,HiddenEgg.jpg
+0,0,0,Common_PlayerRestrained,"Hinter der roten Farbe ist etwas Pinkes versteckt,|aber mit deinen Fesseln kommst du nicht ran.",HiddenEgg.jpg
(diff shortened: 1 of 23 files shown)


Ben987:
This is great!  Thank you so much :)

Yes, Julia says she prefers to be called by her first name but it's up to the player to call her like that or not.  It will change how she sees the player but that's it, there's not that much behind it.

Not really, it's hard to simulate an accent without actual voices.  You can put a few words that sounds a little more Italian if you want :)

Some gagged speech do mean something but most of it is gibberish.  I'm not sure how it will work in German, you can leave it as it is I guess.

Thanks again!  This is awesome.

PR #69, opened 2018-10-14
German Chapter 6

adokilume:
This'll probably be the last chapter I can get done this fast, the next ones are pretty big.


Ben987:
This is great!  Another chapter in German,  

Yes, chapter 7 and up will take much longer to translate, they are more text heavy.

Danke!

PR #71, opened 2018-10-23
German Chapter 7

adokilume:
I didn't expect this one to take more than a week, but at least it's done now. Since Natalie's scene has a few lines where the meaning of gag talk is actually somewhat important, I decided to borrow an idea from the Chinese translators and make some notes in the _Translation folder, assuming that's ok. I'll probably add to them from time to time, it will be useful for future reference and in case anyone else wants to join (back) in.

Also there's two things I noticed and wanted to ask about:

If you go to lunch with Jennifer and you order either cheesecake or a chili dog, the result is "That's pretty unhealthy.|(She orders the same.)", which seems odd to me. Why would she order the same if she disapproves? Is this intentional? For now I just translated it like that, but I could add another commit to change it.

The other question is actually about Chapter 8. If the player puts on the costume for her role and then decides to strip again, it says "You strip down from your|school outfit to your underwear." It's the same line as when stripping the school uniform, so it looks like it might be a copy-paste error. I'm just asking to be sure I'm not misunderstanding things.


Ben987:
This is unbelievable, thank you so much!  I'm about to release a new version.  German players will be thrilled to be able to play so much of the game in their native language.

Yes, you probably found copy/paste errors, I will correct them soon, thanks!

adokilume:
I hope German players will actually like it. I'm rather good with language in general, but I don't actually have much experience about how Germans talk to each other in bondage scenarios (like what certain items are called or what dommes call their subs and vice versa). I did some research, but a few things might still be off, so I'm hoping maybe someone with more experience in the lifestyle might give some feedback. This might be an odd thing to bring up after translating this much, but most things should be okay.^^' I'll go back to the earlier chapters eventually because there are some phrases I don't like that I now have better alternatives for.

Since you're apparently ready to release the next public version: Is there still time to quickly fix Jennifer's two lines in German? I'm not sure if I should make another pull request just for this, and the branches are conflicted at the moment, so I'd probably have to make a new branch as well. Or if you'd prefer to just do it yourself, you can see what to change here:
https://github.com/adokilume/Bondage-College/commit/ff3a03573e8b3cd3580ec47ccc48a66a85df079b

PR #72, opened 2018-10-24
German Chapter 7 quick fix

adokilume:
Just a quick correction of those two Jennifer lines. I would have just included this with the next chapter, but maybe this way it can still make it into the upcoming release.


Ben987:
Perfect!  I'm installing right now so the timing is excellent.

PR #74, opened 2018-11-03
Basic achievement system

adokilume:
Sooo... here's an idea I had a while ago, but decided to hold back because I didn't want to add something so experimental right before a public release. It might not be a very good idea or a good match for this game, so I'll totally understand if you don't want to accept this for any reason. I just needed to get it out of my system, because the night I thought of it, it literally wouldn't let me sleep.

Basically, I thought that achievements might be a nice way to let the player know what they might have missed in the game, without telling them exactly how to do it the way a walkthrough would. That's why for the ones I added, the description hints are kinda vague if the achievement is still locked (unless it's more of a challenge that requires more information). Ideally, there could be an achievement for every major missable event, plus a few extras thrown in for challenge and fun. I've written down a few specific ideas in the file _achievementIdeas.csv, but only implemented 5 of them so far for testing. Those 5 are all pretty simple and trigger at one specific event (unless there's more than one way to do it). More complex ones could probably also be done by using the game log, but I haven't looked into that yet.

A few more notes: there are 3 pages in the achievement menu, 2 of which are empty at the moment. I only included those to test and showcase how navigating the list could work. Maybe there could also be a way to display how many achievements have been unlocked so far, for the completionists out there.
Also, the image that shows up for locked achievements is just a placeholder I quickly put together with my limited art skills, so it doesn't look great. It's supposed to be a victory cup with a ball gag strapped inside. Maybe you or someone else can make something better looking?
Lastly, the only way to check which achievements you have at the moment is to reload the page and go to the achievement menu. Adding a notification during gameplay probably would'nt be too hard, but I'm not sure if that would be helpful or annoying.

This is mostly just a suggestion for now. I haven't made copies of the files for other languages yet, because it might be hard to keep track of which line is which if the achievement list grows and gets reordered. I'm honestly not sure if I'd be willing to write achievements for the entire game (my initial hype for this idea has died down a bit), but maybe you like the idea. And if it's not ready by the time the next public release comes around, I think this could be disabled rather easily by removing the option from the menu.


Ben987:
That's a great idea!  I thought about doing some trophy system back then but never got to it.  I would have to code and test all the achievements for all chapters before going live with that change though, and it might take a while since I'm working on the Bondage Club.  I will include the change and remove it from the menu for now.  I'll see if I can find the time to add more achievements soon.  Thanks!

adokilume:
Sure, I had a feeling the Club was keeping you busy. :) Really curious to see where this is going, I did read your concept announcement but I'm having trouble imagining how it's all going to work at the moment. Part of me also really hopes you're still going to add more content to the main game, but I'm looking forward to whatever happens in the Club too.

I tried to keep th

Ben987:
Yep, I'm pretty much working full time on the Bondage Club and only correcting bugs in the College.  But I'll be back for more contents later :)

Oh yes, I've checked your code and it's really basic, it would be easy to add more stuff later on and it will help players to find specific paths that they could have missed.

PR #77, opened 2018-11-14
[Bondage College] French Translation - Chapter 9 and 999 - ADD_MINUTES:X - Revision

gottspende:
As agreed this PR contains ADD_MINUTES:X and smalls revisions for chapter 9 and the chapter 999
I saw a Achievements folder but I don't translate much I think It's in progress.

I quickly looked on to chapter 8 and I didn't see any bug in particular
I will spend more time on when I review the old chapters but now it's chapter 10's turn.

:)


Ben987:
This is great!  As usual, "un gros merci" for your awesome work Gottspende.

No need to translate the achievements for now, it's still a work in progress by Adokilume.

PR #78, opened 2018-11-22
German Chapters 10 & 11 (+ Chapter 9 Update)

adokilume:
To avoid confusion with the halfway-translated chapter 11, I decided to squeeze both chapters into one PR. Plus a little something extra.

Chapter 9: When playtesting the translation for this chapter, it kinda bothered me that there's still English text on the image of the framed "Silence is Golden" rule. But I figuerd I couldn't do much about it, so I just added a translation for it to the corresponding intro text and called it a day. Later (after making the PR for that chapter), it ocurred to me that it should be totally possible to translate that image without affecting it for other languages, since each language's intro file references the images individually. So here it is, I tried to make it look as much like the original as possible, but I'm a bit of a noob when it comes to Photoshop, and I couldn't find the same font (I tried asking BlazingKnightHero, who made the original image and was very willing to help, but unfortunately he couldn't remember what the font was called either). I hope adding this is okay and doesn't break something I didn't think about. If there's interest, maybe I could upload my project file somewhere so the other translators can make their own versions more easily.

Chapter 10: Not too much to say about this one. Except that I wasn't expecting this project would lead to me reading 5 different articles about pig noises because I've never seen "Soooowee" before.

Chapter 11: Making drastic changes to someone else's translation makes me feel bad, but most of the lines where just too long for the game or sounded slightly too unnatural. I hope fleisch11 doesn't take it personally.

Not much more left to do for now. Maybe I'll finish chapter 12 before the end of the year (not sure though, it's huge, but so far at least I'm not bored), after that it's just the Kinbaku Club and whatever is finished of the Bondage Club by then. :)


Ben987:
Once again, this is HUGE!!!  You're a machine :)
And yes, the trick to use another file in another language is pretty slick, well thought.
I'm sure fleisch11 won't mind, you will need to do the same job in chapter 12, it was translated too early so you'll probably need to start back from English in some cases.
Bondage Club is far from ready to be translated, you'll be able to take a break when you get there.
A thousand "Danke"

PR #80, opened 2018-11-29
Polish fix

adokilume:
Since I happen to know some polish and the encoding errors in these files has been bothering me for a while, I figured I'd just fix them. And also add some missing lines, hopefully without making any mistakes with the numbers. My polish isn't practiced enough to do a full translation (which is also while I left most of the untranslated lines as they are), but at least I could tell what the words were supposed to be and which letters were missing. I'd be interested to see the rest of this translation if anyone decided to continue working on it.

Just as a sidenote, since it kinda fits here: I wrote a little editing tool specifically for the CSV files of this game. I made it mostly for myself to have a live ingame-like preview of the text while I'm translating it, but when I find enough time I could try fleshing it out and cleaning it up for others to use as well. It would automatically take care of stuff like encoding and formatting, which many seem to have trouble with. Not sure when it's going to be ready to publish, but I just wanted to let you know ahead of time. I guess by now it's kind of obvious that I'm a bit obsessed with this entire project lately. ^^'

### DIFF
--- a/C000_Intro/ChapterSelect/Intro_PL.csv
+++ b/C000_Intro/ChapterSelect/Intro_PL.csv
@@ -1,4 +1,5 @@
 Stage,LoveReq,SubReq,VarReq,IntroText,Image
-0,0,0,,"Witaj w Akademii Bondage|Baw siê dobrze, zagl¹daj we wszystkie zakamarki i daj siê ponieœæ!.",
-10,0,0,,"Wybierz rozdzia³ gry.|Jeœli nigdy wczeœniej nie gra³eœ, zacznij od rozdzia³u pierwszego.",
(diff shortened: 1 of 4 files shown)


adokilume:
Wow, that's a funny coincidence indeed. I don't think I'd be much help with the actual translation, like I said I'm really out of practice and only know the language on a rather basic level. Enough to communicate, but it wouldn't really sound natural to a native speaker. When it comes to technical help though, like using git or avoiding formatting errors, I could probably offer some advice. Though I'm guessing you could explain those things just as well if not better.

Ben987:
Excellent!  He might get trouble with the non-English characters that switches to " ? " when you save the CSV file or with doing pull requests.  If that's the case, you'll be great to help.

Thanks again!  This is awesome as usual.

PR #96, opened 2019-02-07
new cheat for bondage club

fleisch11:
A cheat which allows the ones that can use it to control when a random kidnap starts by clicking a button that will appear in the main hall when the cheat is activated. (with the cheat random kidnaps will only start when the button is clicked)
(i don't know why there are so many changes, i only changed something in the mainhall and the cheat folder)


Ben987:
Thanks for the cheat, it's a good idea.  Though you've uploaded waaaaaaaaay to much stuff.  Make sure you only push the cheat and nothing else in your pull request.  If I pull the request it will break lots and lots of code.

PR #98, opened 2019-02-08
New Cheat + look for trouble

fleisch11:
look for trouble starts a random kidnap, new cheat disables the random kidnaps when moving from the mainhall


Ben987:
That looks great!  I'll try it right now :)

PR #100, opened 2019-02-12
Fix missing ADD_MINUTES

adokilume:
I recently realized that when I made PR #70, I completely forgot about the common folder and all those (1 minute) annotations were apparently broken ever since. Sorry! This should fix it, I used search and replace accross the entire game folder, so I'm confident I got them all this time.

As a quick update, the Chapter 12 translation is going pretty well after taking a couple weeks off, I think it should be finished sometime this month.


Ben987:
Thanks for the fix!  It will make the translations more coherent with the English version :)

PR #103, opened 2019-02-20
German Chapter 12

adokilume:
Finally. This took way longer than I expected. Which is mostly because I took a break of a little over a month, but still. I'm glad to finally be done with this chapter for now.

There are two small things I noticed in the meantime:

1.) When Sarah is about to leave for the Bondage Club and invited the Player to come along, it still says the Bondage Club isn't available in the current version (this only happens if you wait for her to go, not if you ask her about it). I'm guessing this is probably an oversight?

2.) This is a very minor thing, but I noticed that when you try to enslave Jennifer, line 105 ("It's Dominant magnetism. You cannot escape it.") calls the RefuseSubmit function, but the following, much harsher line doesn't. Maybe that's how it's supposed to be, but it doesn't really seem to fit with the rest of her dialogue reactions and it felt odd when I read it. Pointing it out just in case.

Anyway, on to the Kinbaku Club!


Ben987:
Wow!  This is huge!  Congrats, that's quite a feat.  Yes, I might have to fix these errors you've found, I'll check that soon.  Thanks again :)

You can do the very small Chapter 13 before the Kinbaku Club if you want an easy one.

Congrats again!

adokilume:
Good point, I'll probably do Chapter 13 when I get home. My plan was to go in release order (even though the Kinbaku-Club isn't finished), but taking care of that one first does make more sense now that I think about it. I might include some adjustments to earlier chapters as well, I wanted to do that for a while and this is a good opportunity. I'd say you can expect it a couple of hours from now.

Also, I hope I'm not being annoying or anything when I point out potential mistakes like that. I know that's kind of the point of open source projects, but I tend to be nitpicky so I'm sometimes not sure it's too much. ^^'

Ben987:
Excellent!  You're amazing.

For the Kinbaku Club, you might want to check with wildsj before doing the translation.  Maybe he's re-writing parts of it and you could end up working on something that won't be in the final release.  Send him a deviant art note just in case.

I love having bug reports like that.  The games needs lots and lots of testing.  We programmers are too lazy too test properly :)

adokilume:
Yeah, that might be a good idea. I was going to try keep track of all changes and react accordingly, but in case larger parts get rewritten it might be best to talk it through first.
And I totally get what you mean by the testing thing. Even though I've mostly stayed on the surface here so far, I'm actually a programmer myself. ;)

Ben987:
I've corrected the two bugs you've found, can you check the German version?  Thanks :)

PR #105, opened 2019-02-22
German Chapter 13 + early translation tweaks

adokilume:
As discussed, here's the translation for chapter 13. I noticed that last time I didn't update the chapter numbers in the main menu language selections, but I'm also not entirely sure what to do with them now (does chapter 13 count if it doesn't appear in the menu and leads into the untranslated Bondage Club?), so I'll leave that up to you.

Also, some adjustments in the earlier parts, so most of the stuff I wasn't quite happy with before reads better now. Many (hopefully all) typos in chapters 1 & 2 are now also gone thanks to LanguageTool, I'll check for more in the later chapters another time.


Ben987:
This is great!  You're so fast.

And yes, I've kind of guessed that you were a programmer :)

PR #106, opened 2019-02-23
Rr frope spanish/bondage college

RRFrope:
Hi. 
This is chapter 2 translation into Spnish. It also include translation of some of common files.
I have been translating an older version, maybe it is not important but use only Spanish files (_ES).
Hope it works.

Rafa


Ben987:
This is great!  Thanks a lot :)

But have you downloaded the latest version of the game before translating?  There are many conflicts with other Spanish translations.  Chapter 2 was already translated but not fully tested so it creates many conflicts with the current branch.  Can you compare your files with the actual translation?  Thanks

RRFrope:
I was being translating the version I had some months ago. I didnt´know that there is another one translating into Spanish. I dont know how to fix the conflicts. So it is better not to merge the files.

Ben987:
My advice would be to download the latest version in a separate directory, check the translation that's in there and copy the missing pieces in the current version.  Your version has more files translated than the current one, it would certainly be useful for Spanish players.  I will close this pull request for now.  Thanks again for your work :)

adokilume:
Personal recommendation: use https://www.diffchecker.com/ (or something similar) to compare the files in the current version with yours, see which parts are better in which version and keep those.

Ben987:
Good idea Adokilume.  There's probably lots of stuff in that pull request that could be merged to the main branch for Spanish players.

PR #107, opened 2019-02-23
German Chapter 12 fix

adokilume:
Just those 2 recent fixes for the german version.

Edit: Plus a few more corrections and improvements for the early chapters.


Ben987:
This is great!  You're a machine :)

PR #108, opened 2019-02-26
Russian translation

fluend:
Russian translation of chapter 1. I changed the common file and added the Russian translation version to the files of other languages.
If I did something wrong somewhere, please show me.


adokilume:
Hi, I don't speak Russian, and I'm not the one who decides what gets put into the game, but I can see two small problems with this:

1. This isn't really a big deal, but you have the German chapter numbers switched from 13 back to 11, probably because you downloaded the files before those chapters were translated. No need to worry about it though, that should be an easy fix.

2. This one is a bit more important, some of your CSV files have formatting errors. If there's a comma anywhere in a text field, the whole field needs to be enclosed in quotes. Otherwise the game won't know how to deal with that, which you'll probably notice when you try playing with the translation. If you want to keep editing the files by hand, I recommend using Visual Studio Code with the Rainbow CSV extension, which will highlight each column in a different color and makes it easier to see which parts need to be fixed. Or you could use a tool like Excel to edit the files instead, I believe it does that stuff automatically.

Other than that, nice work!

fluend:
Hello, regarding the German translation, well, I understood, I will be at home and correct.
But at the expense of errors in the encoding, I do not even know, like everything was rechecked several times. but since you are talking about them, I will check again. Thanks for the advice on the program, I myself use notepad ++, because Excel does not make tables for me anyway, but only changed the file encoding. Ok, by evening i can check for errors

adokilume:
I don't mean the character encoding, everything seems to be fine there. I'll try to explain what I mean with an example.
Take a look at [this file](https://github.com/fluend/Bondage-College/blob/master/C001_BeforeClass/Sidney/Stage_RU.csv) inside GitHub. At the top, it says:

> We can make this file beautiful and searchable if this error is corrected: Missing or stray quote in line 6 

Now, line 6 looks like this:
`0,0,0,,Какой скверный характер!,"Лучше заткнись, пока нос не сломала."|(Она поворачивается и смотрит на другую девушку),10,-2,0,`
CSV seperates the columns at every comma. For this line, the result would be:

Stage: `0`
LoveReq: `0`
SubReq: `0`
VarReq: _[Nothig, since there's nothing between comma 3 and comma 4]_
Interaction: `Какой скверный характер!`
Result: `"Лучше заткнись`
NextStage: ` пока нос не сломала."|(Она поворачивается и смотрит на другую девушку)`
LoveMod: `10`
SubMod: `-2`
Function: `0`
_[And then there's another comma, which is confusing Github (and probably Excel too) because it looks like there are 11 columns when there should be 10]_

As you can see, what you intended to be the `Result` column got split into `Result` and `NextStage`, and everything after that was shifted one column away. Why? Because in the code, there's a comma after `"Лучше заткнись`. Of course it probably needs to be there, and the CSV format has a way to deal with that: quotes. The corrected line would look like this:

`0,0,0,,Какой скверный характер!,"Лучше заткнись, пока нос не сломала.|(Она поворачивается и смотрит на другую девушку)",10,-2,0,`

Note that all of `Лучше заткнись, пока нос не сломала.|(Она поворачивается и смотрит на другую девушку)` is now enclosed in `"` quotes. This tells a computer that any commas inside the qutes is part of the text, and should not be split into columns. There are still 10 commas, but the computer now only "sees" 9, thanks to the quotes.
This only works if the ENTIRE string is quoted, any punctioation like `.` will also need to be inside. If you've corrected this for every line, you'll know because GitHub will display the file as a table, as seen [here](https://github.com/Ben987/Bondage-College/blob/master/C001_BeforeClass/Sidney/Stage_EN.csv).

(Hint: if you ever actually need to use quotes inside of text, they need to be doubled `""` and also require quotes around the whole thing, for similar reasons.)

I hope I was able to help! I've seen many people having trouble with this, and I didn't know about it either when I started. So maybe this could help others too if they see it.

fluend:
Ok, i got you. I thought that there was not much difference, select only a comma, or the whole replica with a comma. In my opinion, simply, since the comma is not counted in "", then I will do it "," or "text, text" - there is no difference. Thanks for the explanation

PR #111, opened 2019-03-04
German Kinbaku Club

adokilume:
It. Is. Done. Well, for now at least. :)
As you suggested, I wrote a message to wildsj asking if there are any parts I should stay away from for now. Unfortunately, I never got a response, but I think it's fine. This way, everything in the current version will be translated, and if anything changes, I'll just have to react accordingly. There were also a few lines of gagged talking I wasn't able to decipher so if wildsj ever does respond (or happens to see this), I'll probably ask about those as well.

Now, about the Bondage Club: I recently had a discussion with Dallun511 (which you probably saw), to see which of us would be translating which parts, if there's going to be two translators now.
...Then I remembered you saying that the Club is probably not ready to be translated yet. So the question is: Is this still true for all of it? I was thinking the Introduction Class and Maid Quarters would be a good start since they seem rather stable and finalized. Dallun511 wanted to do the Shop and Fight System. Would those be good ideas, or should I / we still wait before doing this?

Also, on a different note: I might have mentioned this before, but I coded a tool for myself to make translating easier. I've been using it for the last couple of chapters, and now finally got around to polishing it up and uploading it so others could use it as well. You can find it here, if you wanna take a look and maybe spread the word (if you think it's good enough): https://github.com/adokilume/BC_TranslationAssistant
I'm not sure if everyone will find this as useful as me, I mostly wanted to be able to see how the text I'm typing is going to look in-game, without needing to go through all the scenarios. It helps for finding the best spot for line breaks, and to know if some of the text is too long. It doesn't work with files from Bondage Club yet, but that could change soon.
The current version was made mostly with translators in mind, but I could add a mode that enables editing for all the columns and lets you add or remove lines. Maybe then it could even be useful for developers like you or wildsj? Anyway, I'm very curious to hear what you think about this.


Ben987:
Wow!  You truly are a machine.  Yes, I will offer your translation assistant to anyone that wants to do translation, maybe they will find it useful.

For the Bondage Club, I wonder if there would be a better way to do translations.  I still have the CSV files just like in the College but there will still be new lines of texts added here in there from time to time.  

Since it's not chapter based, all modules can evolve.  Yesterday I've added a few lines in the introduction room and maid quarters to greet the player when she's a club Mistress.

It would be problematic if I need to reach all translators to add a single line of text.  I would need a way to show that new line in English if it wasn't translated yet.  Maybe I could have some kind of dictionary of all possible words and phrases and have the software translate everything that he can when the files are loaded.  Anything that cannot be translated stays in English.  What do you think?

adokilume:
Thanks! I suppose now I could say, I'm not really a machine, I just built one to make the job easier. ;)

I'm not sure if I fully understand what you mean with the translation system. From what I understood, whenever you added new lines in the College you just copied them over to all the translated files that existed at the time. At least I think that's what you did when you added the option to find a collar in gym class. As long as the translators aren't available, those lines appear in English (like they still do in the Chinese version), but they're still there, so they don't break the game. Wouldn't the same thing work for the Club as well?

When I started these translations, I wasn't really sure if your system with the CSV files was such a good idea. They're basically scripts, since they don't just contain the text, but also gameplay-affecting numbers and function calls. This seemed dangerous to me, because if someone changed something other than the text by accident, the game would suddenly play differently or stop working completely, in just that one language. If someone reported such a bug after playing, it would be really hard to track it down if that person didn't think to mention what language they were playing in (it wouldn't really be obvious to a player that the language setting changes how the game behaves). I'm not an expert, but I think games usually tend to separate logic form content for this reason.

However, I also found some upsides with this way of doing things. Having all that extra information right there with the text helps provide context, which can be really important if the translation is supposed to make sense. Whenever I was confused by the meaning of a certain line of text, all I had to do was look at the stage number of that line, then search for lines that lead to that stage (or find the number in a JavaScript function and then find the lines that call that function), and I would immediately know what part of the scene I'm in and what that line really means. If I only had the line of text without any of those clues, that would have been much harder, especially since there are so many branching paths. (Also, being able to edit the other columns turned out to be useful that one time I wanted to localize the image with the sign in the library.)

I didn't have much opportunity to work with the Club CSVs yet, but from what I saw, they make this even easier by combining all the context from "Intro", "Stage" and "Text" into a single file. (The only thing I don't get is why you got rid of the column headers, but once I understood which column does what, it wasn't really a big deal.)

I guess I'm saying that both ways have pros and cons. As for your dictionary idea, I'm not sure if I understand it correctly. But if you're planning to assemble sentences out of phrases and single words, that makes me a bit worried. Some words might always be the same in English, but have different forms in other languages depending on where and how they're used in the sentence. This could be solvable by providing different forms for each word (even if they don't change in some languages), but that kind of sounds like a pain in my opinion. It also sounds like it would be way harder to understand the context if everything is split into fragments. I could be totally wrong about that, it probably depends on the execution.

Sorry if this got a bit lengthy, these are just some thoughts I gathered to hopefully give some useful advice. I obviously don't want to keep you from trying new things. If you can find a way to improve how translations work, that would be awesome. It's just that from my point of view, it's already pretty damn good. But of course I have no idea how stressful it is for you to coordinate this project in its current form.

Ben987:
Yes, since there's code in the CSV file, there could be bugs in a specific translation that's not in the other translations.  I love having some code in the CSV files, it saves me a lot of time while creating the game. 

That's why I'm thinking of some kind of dictionary but I'm not sure how it could work.  Maybe have a file with the _DE extension that comes in the same folder as the _EN but that file would be text only.  The first column would be the word/phrase in English and the second column would be the translation.  For example:

Hi, my name is DialogPlayerName.       Hallo, ich heiße DialogPlayerName.

That way I could add or edit the English version without worrying about causing bugs for foreign speaking players.

adokilume:
Ah, I see. That sounds like a pretty good solution. I actually translated another game some time ago that did things in a very similar way, I just didn't think to mention it because I got so used to how things work here. The only problem I can see tight now is that this removes the context I was talking about, but it really shouldn't be that much of an issue since it can simply be looked up in the main English file if needed.

I think this could also make things a lot easier for translators who aren't also programmers and/or aren't familiar with CSV. Many seem to have trouble with it at first. If you wanna go all out on that, it might even be a good idea to ditch the CSV format entirely (for the translation files, at least) and instead just separate the strings by newlines, like this:

```
Hi, my name is DialogPlayerName.
Hallo, ich heiße DialogPlayerName.
I like Bondage.
Ich mag Bondage.
```

That's how the other game I worked on did it. The benefit is that this way, translators wouldn't even have to worry about quoting strings with commas in them (since many do seem to use plain text editors rather than tabular tools). The drawback is that the English lines have to match exactly, and accidental whitespace at the end would break them. Though it shouldn't be hard to code around that with the trim function or something.

Ben987:
Yes, since CSV can scare a few translators, we could use a very basic line by line translation.  You're right about needing to use the exact text though.  If I correct a typo in the English text, I'll need to remember to correct it in the translations also, it's not that bad but it might be necessary to have a good text proofing before starting any translation.  I can trim whitespaces on the javascript side, so it's not a big issue.

For currency symbol, I'm using $ right now but should we also translate it?  Should it become the euro symbol for German, French, Spanish players?  Or should it be related to the server you're playing on?  $ for North America and euro for Europe?  Maybe a Yuan symbol if I start a server in Asia?

adokilume:
Right, I didn't think about typo corrections, but I don't think that's a huge issue. A script could be used from time to time to check for any lines in the translation files that don't match anything in the originals, and then those could be corrected after the fact.

I don't really have any strong opinions about the currency symbol, personally. I can't speak for other countries, but most big American games from that get localized to German usually keep $ as their currency, and we just assume it takes place in the US. That's also why I didn't bother translating words like "College" or "Miss" (also because there's not really a German word for "Miss" that you can use without attaching a name to it), it's a rather common practice in popular media to just keep terms like that. (As an aside: Japanese games that got an American release which changed Yen to Dollars, still tend use Dollars in the German version. Probably because it's closer to what we know.) I actually rarely ever see the € symbol in video games, unless it's about microtransactions with real money. Maybe I just haven't played the right games though, so I'd recommend asking some more people, or possibly making a poll.

Ben987:
Yeah, it shouldn't be too bad.  We can also forget about currency symbols for now if you're used to see $ in Germany.

I'll come up with a translation system pretty soon.  Can you prepare a first file?  Make it a TXT file instead of CSV and I'll see if I can make it work.  Pick a character dialog file to start with please, go with that format:

Hi, my name is DialogPlayerName.
Hallo, ich heiße DialogPlayerName.
I like Bondage.
Ich mag Bondage.

Thanks again!

adokilume:
Sure, I did the introduction maid for now. I noticed the file ends up looking kinda messy since it contains dialogue for both the maid and the player, so I thought it might not be a bad idea to add some comments, to at least make it clear which lines are said by whom. It would be great if you could take this into account when coding the system. I used `###_` as a prefix for comments to make them stand out, but a single `#` should be enough to recognize them.

Here's the file: 
[Dialog_NPC_Introduction_Maid_DE.txt](https://github.com/Ben987/Bondage-College/files/2935641/Dialog_NPC_Introduction_Maid_DE.txt)
Or, if you can't work with comments or just want to test what we discussed for now, here's the same file without comments: 
[Dialog_NPC_Introduction_Maid_DE_NOCOMMENT.txt](https://github.com/Ben987/Bondage-College/files/2935644/Dialog_NPC_Introduction_Maid_DE_NOCOMMENT.txt)

Also, this is the script I used to generate the base file, just to save time in the future. 
[ClubTranslationGen.zip](https://github.com/Ben987/Bondage-College/files/2935679/ClubTranslationGen.zip)
 I thought it might make sense for you to have it too, in case you want to provide quick templates. Just paste the complete CSV file into the box, click convert and the result should appear in the box.

By the way, I noticed some duplicates that probably aren't necessary. The line "That's about everything.  Remember that your arms need to be free to restrain someone." appears twice, one has two spaces and the other just has one. There are also the lines "Can you teach me more about rope bondage?" and "Can you teach me more on rope bondage?" which I don't think really need to be different by a single word. It's not really a problem right now, but things like this could cause confusion later on, since they do need to be exact matches in the translation file.

Also, the maid can't seem to decide if there are eleven interactible body parts or just eight. :)

Ben987:
Great.  I'll check the file and see if I can make a translation system for it pretty soon :)

Issue #110, opened 2019-03-03
Translating actor names / placeholder


adokilume:
While playtesting my progress on the german Kinbaku Club translation, I noticed the word "Unknown" in the spot where a character's name is usually displayed, if you don't know that character's name yet (I'm not sure how often that happens in the game, I just never noticed or thought about it before). This confused me at first, because as far as I know I've been through just about every translatable file in the game, until I found out that the "Unknown" string is hardcoded in Drawing.js.

Obviously this isn't a _huge_ deal, but it feels kinda weird to have that one English word there when everything else is translated.

This also made me think about name translations in general. It's not really a problem in my case since German uses the same letters as English, so all the character names are the same, but for languages like Russian or Chinese it might be a different story.

Would it be possible to add a global CSV somewhere that contains the English names as tags and the translated ones as content? Would that work if it's not really tied to any chapter?

Ben987:
The unknown tag are for the twins, when you still don't know who is who.

Do you know where the all the "Unknown" shows?  We could ask wildsj to pull the tags from a file instead of plain text in code.

adokilume:
I wasn't sure if it's only for the twins, I thought it might appear on ohter occasions when you haven't learned someone's name, but you probably have better knowledge on that. I just checked that it was indeed wildsj who added that line, but the thing is that it's not part of the Kinbaku Club itself.

It happens right inside the `DrawActorStats` function in Drawing.js, if the current Actor's `ActorHideName` is true. Since this is a global game function, I'm not sure if reading from a file would be possible without changing the structure or logic of that system. My first thought would have been to place the file somewhere inside the C999_Common folder, but Drawing.js isn't a part of that so I'm not sure if that's a clean way to do it.

A possible solution I'm thinking of right now might be to add a new value type to Actor.js, like `ActorDisplayName` or something like that (since the normal name is needed to identify the actor in the code, from what I can tell). Then, if that value is requested, the name could be looked up from a file, depending on current language and if the name should be hidden. Would that make sense? Sorry if I'm rambling, I'm still unsure about the details and there might be some things in the code I don't know about, plus I'm a bit tired at the moment.^^'

Ben987:
Yes, you could certainly keep that translated value somewhere and show it instead of "Unknown".  The value could be pulled from the beginning of the game in one of the first files or some kind of new dictionary.

Is this something you would want to code?

adokilume:
[...]

PR #112, opened 2019-03-06
Translatable Actor Names

adokilume:
This should take care of issue #110, hopefully without creating any new problems. For now I added CSVs for the languages that were already present in the common folder, though FR and CN are just default English templates and will need to be adjusted.

Everything seemed to be working fine when I tested it in different languages, but please check the code just to make sure I didn't do anything the wong way. For example, I wasn't sure if I should use `CurrentLanguageTag` (which sisn't display any names when I selected a language there wasn't a file for, even though I thought `ReadCSV` would fallback to English in that case) or `GetWorkingLanguage()` (which seems to work better, but will probably change which language it uses depending on wheter the chapter has been translated already - I think that works out well though, so I went with that).

You might notice the line `if (ActorToLoad == "") return;` that I added to `ActorLoad`. I did this because of a strange occurence when talking to the RopeGroup in the KinbakuClub, which added an "empty actor" to the list of Kinbaku Club relationships. It was just a line with stats but no character name, or the "MISSING TEXT FOR TAG" string after my changes. Apparently this is because the script called `ActorLoad` with an empty string to unload the current actor, but instead it created a new actor without a name. I didn't want to change anything in wildsj's code without asking or knowing the reason, so I did this instead. I hope there's nothing wrong with it now.

(Edit: Forgot about Natalie, added her to the list.)


Ben987:
Awesome!  Have you tested it with all characters to make sure they all show properly?

adokilume:
As of right now, yes.^^ I assumed they'd all work because I copied the main characters' names from the player script (including Sophie, just in case), and I checked the Kinbaku Club before. But yes, I quickly went through all of them again in English and in German and they al seem to work. Unless I forgot someone from the Kinbaku Club, like Natalie at first. But that was mostly because she wasn't in that chapter, so I didn't have her fresh on my mind.

adokilume:
Actually, it seems I spoke too soon, sorry! The names don't show up in Spanish or Russian if you're playing a translated chapter. Probably because GetWorkingLanguage() assumes the files should be there, since the game doesn't actually enter Chapter 999 to fetch the names. Any idea how to fix this? Or would it simply be required for people to translate that file right away when doing a new language?

Ben987:
Correct, chapter 999 isn't translated in all languages.  Could you add something in your code to only use that new part of code if the language is English, French, German or Chinese?

adokilume:
I could, but wouldn't that just create another list of languages that nees to be updated all the time? After thinking about it a bit, maybe a better way to get the same result would be to add an optional parameter to GetWorkingLanguage, so I could pass chapter 999 to it, then it would just default to English as normal. I'd do it in a way so it would still use the current chapter if nothing is passed.

Ben987:
I'm fine with it also.  Just make sure it works with every current languages :)

Thanks!

PR #113, opened 2019-03-07
Actor Names Fix

adokilume:
Alright, this should make the translated actor names work for all languages in all chapters. Hopefully nothing else is broken, that would be really embarrassing by now. But I tested chapters 1, 2 and 101 in English, German, Spanish and Russian, and everything seemed to be working fine.

While I was at it, I changed the translated chapter list to make it a bit more compact and slightly easier to edit in the future. I was hesitating for a pretty long time if I should do this, because it might cause merge conflicts if someone is currently translating a new chapter and isn't aware of these changes. But I think that would have happened either way because I needed to change the function as a whole. I hope this is okay, if there's anything wrong with it, I can change it back to the previous form before you accept the pull request. Functionality-wise, it works, like I said above.

### DIFF
--- a/Actor.js
+++ b/Actor.js
@@ -52,7 +52,7 @@ function ActorSpecificGetValue(SpecificActorName, ValueType) {
 
 // Return the current actor's localized name
 function ActorGetDisplayName() {
-	if (ActorNamesText == null) ReadCSV("ActorNamesText", "C999_Common", "ActorNames", "Text", GetWorkingLanguage());
+	if (ActorNamesText == null) ReadCSV("ActorNamesText", "C999_Common", "ActorNames", "Text", GetWorkingLanguageForChapter("C999_Common"));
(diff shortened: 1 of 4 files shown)


Ben987:
Have you tested it throughout the game?  I'm always more scared to do changes on the core stuff.

Thanks!

adokilume:
I totally understand that, it would make me uncomfortable as well if someone messed with my code. I've gotten more involved with this project than I thought I would, so please stop me if I'm doing things I'm not supposed to do.

Turns out that the Spanish and Russian Chapter Select files had a few minor errors that made it impossible to select certain chapters from the main menu. I had to fix them so I could test the chapters efficiently.

Other than that, after checking every chapter, translated or not, with each language setting (and trying multiple paths each time), I couldn't find any problems. I also played a few routes from start to finish without issue. I'm 99% sure that everything's fine.

Ben987:
Excellent!  Thanks again, you're a machine :)

PR #117, opened 2019-03-15
German Bondage Club Beginnings + Translation Tools

adokilume:
This contains the german translations for the Introduction Class, the Maid Quarters and the Main Hall. I also fixed a small formatting error in Dialog_NPC_Introduction_Sub.csv which would have affected the translation, hope that's ok.

I also included 2 HTML tools that I think might be useful for translators. The one to generate templates is the one I already showed you, I just changed it a bit to make it slightly more user-friendly. The other one is for checking if all the translation files are up to date. I plan to add some more features soon, but for now it should automatically detect if text gets added to a file that already has a translation, or if a line gets changed so the translation for that line will no longer work. This way it should be easier to keep track of everything as the game evolves.

~I'll probably try doing the Shibari Dojo and Club Management rooms next.~

Edit: I'm liking this new system, it really speeds things up! :D I guess the private room would be next...

### DIFF
--- a/BondageClub/Screens/Character/Login/Text_Login_DE.txt
+++ b/BondageClub/Screens/Character/Login/Text_Login_DE.txt
@@ -1,10 +1,44 @@
-﻿Account Name
-Kontobezeichnung
+﻿Welcome to the Bondage Club
+Willkommen im Bondage-Club
+Enter your name and password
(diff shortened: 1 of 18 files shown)


Ben987:
Thanks!  That looks great, you're a machine :)

PR #120, opened 2019-03-21
German Bondage Club: Private Room

adokilume:
Whew. After this big chunk of text, the rest should be a breeze. Almost up to speed. :D

I know you added some lines to these and some other files in the last few days, so I'll adjust those translations after this is merged.

I haven't really decided which parts to tackle next. I'm trying to stay away from the Shop and the Kidnapper's League for now, since Dallun511 said they wanted to do those, but there's still plenty of other stuff to translate.

PS: I'll probably need to revisit some lines, a few of them might be too long. I haven't found a good way to automatically check for that yet (there's the exit button to consider), but I'm working on it.

Edit: TranslationStatus.html now checks for line lengths (and indeed exposes a few lines that are longer than they should be, I'll take care of those soon (Edit: Done.)).


adokilume:
By the way, quick question: I noticed that in the MaidDrinks Minigame folder, the files Dialog_NPC_Customer_Left.csv and Dialog_NPC_Customer_Right.csv have dialogue inside them, but I don't think it's ever used. Should those be translated?

PR #123, opened 2019-04-08
German Bondage Club: Kidnapper's League + UI

adokilume:
And with that, it's done (for now). :) Unless there's something I missed. I was in a bit of a rush to get this done before you do the next public release. Even though the Female3DCG translations might not be working yet, but at least they're there. I might need to revisit them later, because some of them are probably too long.

The UI translations are based on the files Dallun511 posted in #99, just transformed into the new format, slightly adjusted and updated.

I'll probably be taking a small break (unless you add some more lines before the update), and then I'll get back to the College achievement system I started a while ago.


Ben987:
This looks great!  I'll add it to the live version, congrats :)

PR #124, opened 2019-04-08
Bugfix Stable

GIT4NICK:


Ben987:
Thanks!  I will add this fix to the live version :)

PR #126, opened 2019-04-20
German Bondage Club: Chat + Pump Gag

adokilume:
Just a translation update to include the new online features and the pump gag (both are really fun by the way^^), as well as some adjustments to existing translations so that certain things make a bit more sense. I wasn't able to playtest this (having some trouble with node.js and CORS), but I think it should be fine.

A few things I noticed:

- The character customization menu still doesn't translate completely. When I set a clothing option to nothing, it says "None" in english, even though it's translated in the file. The variant options (color names etc.) don't seem to be translatable either. Not a big deal, but might be slightly confusing.
- In case you haven't seen the error report on Discord yet, item names are displayed in the user's language in chat messages when they're used. Thinking about this more, I now realize that this probably applies to all automatic notifications in the chat, it's just that so far, the item names were the only things that were translated. I'm not sure if this necessarily _needs_ to be changed, though it might break immersion in some cases. I guess we could simply expect players to create/join rooms of their own language (and call them something like "[DE] <room_name>").
- Lastly, I'm not sure if listing all the words separately in `Text_ChatRoom.csv` is the best idea, from a translation perspective. Languages that have a vastly different grammar structure might have some trouble with that (though I'm mostly speculating here). Maybe template strings like `DialogPlayerName uses ItemName on DialogCharacterName` would be a better idea in the long run? Though that's probably not a priority at the moment. For German, it works reasonably well.


Ben987:
This all looks great!  Congrats.  There's a few more words and phrases if you want to look at them before I release the next version.

Yes, I'll have to track the "None" and maybe set room languages later on.  And maybe do full phrases by replacing the names but it wouldn't work great since the replacement is done when we build the dialog, it would have to be more dynamic.  The solution right now should work fine with European languages but might be problematic with Asian languages, we'll see with time.

adokilume:
Just to clarify about the room languages: I was thinking this is something the users could do themselves in the room name, not something that would need to be coded. Except maybe to create a nicer way of displaying the language. Not sure if it would be a good idea to restrict users/rooms by language, if that's what you were thinking.

PR #127, opened 2019-04-21
German Bondage Club: new lines, some fixes

adokilume:
Other than adding those few new lines, I also shortened some of the previous ones (because that last change you made allowed me to test locally again and I noticed they were too long). Good idea with those tooltips in the Main Hall by the way, I was actually about to suggest something like that. :D

### DIFF
--- a/BondageClub/Screens/Online/ChatCreate/Text_ChatCreate_DE.txt
+++ b/BondageClub/Screens/Online/ChatCreate/Text_ChatCreate_DE.txt
@@ -5,7 +5,7 @@ Raumname (Buchstaben & Ziffern)
 A short description or what's going on.
 Eine kurze Beschreibung oder was hier los ist.
 Private (Need exact name to find it)
-Privat (Genauer Name zum Finden erforderlich)
+Privat (Zutritt nur mit genauem Namen)
 Select the background
 Hintergrund wählen
 Create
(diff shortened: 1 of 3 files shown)


Ben987:
Excellent!  Yes, you can play without setuping a local node JS server now, the live server is up to date.  Thanks for your quick translations.

adokilume:
I'm a bit confused/concerned right now... When I did the translation for this pull request and tested it, all the item usage messages worked and flowed nicely for the most part. However, currently in the live version, they're a mix of English and German again. In fact, the whole structure seems to have changed to also contain the body part the item is being used on (when did that even happen? I didn't see it in any of your commits). This is now a bit weird, first of all because the items and body parts are transformed in lowercase (which they shouldn't be in German), and also because with this sentence structure, we're suddenly dealing with wrong grammatical genders and conjugations. Maybe it'll still work out once I see how the messages are actually being built now, I'm just mostly confused because the live version seems to be something that hasn't been on github yet. ^^'

Ben987:
Excellent, yes, I took your suggestion and added two new phrases to lock and unlock.  

I'm not sure how you can make it fit in German but I guess the lowercase would still work and you can translate " 's " to " " if it doesn't exist at all in German. 

Cheers!

PR #129, opened 2019-04-24
German Bondage Club: Updates to chat etc

adokilume:
Not sure what rhythm to do these pull requests in, but here are the translations for the latest changes. (I ignored the magic theater for now, it'll be easier to do all at once when it's a bit more complete.)

Unfortunately I kind of had to revert back to the old style for the text messages, and do some minor tricks to make it work. Translating grammar is really complicated for dynamic strings like these^^'. I also had to leave out the body parts in the translation for now, having their names in their base form only just wouldn't work well. But the new system still allowed me to improve the word order, so it's definitely better than before in my opinion.


Ben987:
Excellent!  This all looks great, thanks again :)

PR #133, opened 2019-05-01
New Zone for Nipple Items

Lucritius12:
This is a small modification that splits the "ItemBreast" group into two subgroups: "ItemNipples" and "ItemBreast".
The "ItemNipples" category has clamps and piercings, the "ItemsBreast" category has has the chastity bra as well as crops and whips.

The reason for this change is to be able to lock clamps and piercings below a chastity belt, similar how vibrating eggs and buttplugs can be locked below a chastity belt.

The shop has also been updated to work with the new item group, and the maid in the introduction class will teach about the new zones.


Ben987:
This all looks great!  That change was on my list for a long time but I've never did it.  Thanks a lot.

I'm merging it now and I'll try it later, there might be more maid text to change.  

Congrats and cheers :)

PR #134, opened 2019-05-01
German Bondage Club: Nursery + PasswordReset

adokilume:
Next batch. Turns out there's another untranslatable string (though not technically), namely the "Escape" button in the Nursery. This one actually seems like an easy fix since it could just be placed inside Text_Nursery.csv , but I wasn't sure if it would be a good idea to make that change myself. It's not terribly important, though.


Ben987:
This is great!  Once again, you're the machine :P

wildsj is still working on the nursery right now, you might want to send him a note before you modify the source.  Maybe ask him to put the "escape" label in the text file himself, it sounds easy.

Take care!  Thanks again.

PR #135, opened 2019-05-04
Chat Styles (experimental)

adokilume:
I remembered that colors in chat were a somewhat heavily requested feature on Discord, so I just decided to give it a shot. This is mostly intended to be a demo; you might not want to merge this, it might cause some problems I didn't think about. However, it did work rather nicely when I tested it.

Currently, this colors people's names before their chat messages (currently it just uses the character's hair color, could be changed to be customizable), as well as parentheses-actions and asterisk-emotes (the latter also being made italic). To make it a bit easier to read if someone has white hair, I added a text shadow, which is probably not perfect, but it's something.

One potential issue with this approach: since the characters (and therefore their colors) are picked by name, this will not work properly if there are 2 or more people with the same name in the room. Maybe the server could somehow send the account identifier along with the message?

I hope this code can be useful in any way, even if only as inspiration.

### DIFF
--- a/BondageClub/Screens/Online/ChatRoom/ChatRoom.js
+++ b/BondageClub/Screens/Online/ChatRoom/ChatRoom.js
@@ -11,8 +11,8 @@ function ChatRoomCreateElement() {
 	if (document.getElementById("InputChat") == null) {
 		ElementCreateInput("InputChat", "text", "", "250");
 		document.getElementById("InputChat").setAttribute("autocomplete", "off");
-		ElementCreateTextArea("TextAreaChatLog");
-		ElementValue("TextAreaChatLog", ChatRoomLog);
+		ElementCreateDiv("TextAreaChatLog");
+		ElementContent("TextAreaChatLog", ChatRoomLog);
 		ElementScrollToEnd("TextAreaChatLog");
 		ElementFocus("InputChat");
 	}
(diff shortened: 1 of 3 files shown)


Ben987:
Thanks!  I might use some of this code later when I code the chat colors but I was thinking of doing it in a different way.  I'd like the player to be able to pick their color and send the color over with the chat message to the server, which will redistribute that color to the clients.  I still need a way to show that color on the clients though, so this will be useful.  

Cheers!

adokilume:
Sure, the hair color was at best only meant to be a temporary thing anyway, just to have anything identifiable at all. Being able to set the color yourself would be a lot better of course, but the main thing I was trying to do here was to find a working system that allows different colors to be displayed at all. So I'm glad if this ends up saving you some time. :)

Ben987:
Sure!  I'll probably use the same color picker as for clothes when I get there, your structure will be useful :)

adokilume:
Hi again, since you haven't actually closed this PR yet, quick question: would it make sense for me to redo this without the name colors? This code does still color emotes and actions to make them slightly more grey, which quite a few people on Discord seemed to appretiate on a screenshot I shared (I'll attach it here just in case you haven't actually seen what it looks like yet: Screenshot_780). I won't try to rush it to get this feature into the upcoming release; I think it could use a longer testing period anyway, just to make sure there aren't any flaws. For example, I recently realized that with the way I've written the code here, it might be possible to inject HTML into the browsers of other players (which would be especially bad with a `<script>` tag), so that would definitely need to be prevented.

So my plan would be to rework this shortly after the next update, unless you'd rather wait until you're ready to implement a way for the custom name colors to work. Or maybe you'd just prefer doing all of this yourself? I just thought I'd ask beforehand this time. :P

PS: In case you're wondering where the translations are, I'll probably do a pull request in about an hour or so. It unfortunately won't include the Sophie scene however, I seriously underestimated how much there is to it.

PR #136, opened 2019-05-05
Remote Control for the Vibrating Egg

Lucritius12:
This mod adds a new Remote Control item that can adjust the vibrating egg. The item is purchasable in the shop for $80 under the "ItemVulva" category. There are a few features that come with it:

- When clicking the remote in your inventory while focused on a character that has a vibrating egg, an extended item menu will open. In that menu, the intensity of the vibrating egg can be set similarly to how the pump gag works. There are five intensities: Disabled, Low, Medium, High and Maximum.

- Whenever an egg is active, it's icon will vibrate on the inventory screen, making it easily visible that it's active.

- An egg set to an intensity higher than "low" will make the wearer stutter. How strongly they stutter depends on the setting of the egg, with higher settings resulting in more stuttering. This applies to both online chat and offline dialogue.

- If the player who has the egg is wearing a chastity belt (and "ItemVulva" items are therefore inaccessible), a new "Use Remote" button appears in the spot where normally the struggle or unlock buttons would be that allows players to use a remote anyways. That button is only accessible when unbound.

This mod also fixes a minor bug that previously allowed you to use chastity keys directly on the breast and pelvis in online chat, resulting in a message like "(TestCharacter used a chastity key on her breast.)"
I've changed it so that only items that don't have an effect tied to them (Which currently includes crops, whips and vibrating wands) trigger the default usage message.


Ben987:
It all looks great!  I can't wait to try it.  Congrats :)

PR #137, opened 2019-05-09
Shock collar + vibrator remote fixes

Lucritius12:
Included in this package is a new shock collar item with a fitting remote. The collar has the following features:
- It can be bought for $250 at the shop. When you do, you also get a fitting remote.
- Clicking the shock collar allows the user to set an intensity for the collar's strength. It can be set to "low", "moderate" or "high"
- Due to it being requested, the extended item menu also features a checkbox that hides the chat message when changing the collar's intensity. This checkbox is only visible while in online chat rooms.
- Using the remote while the dialog is focused on a character will activate that character's shock collar, if they have one. As for now, it doesn't have any additional effects other than displaying a chat message.

I've also fixed a small bug with the vibrating egg that would prevent it from stuttering text correctly due to a typo in the code.


Ben987:
It all looks great!  I'll try it very soon, thanks :)

Could the "TriggerShock" effect be generic and usable for other item types?  Let me know.

Lucritius12:
Yes, it should be easily usable for other shocking items.

PR #140, opened 2019-05-11
A few more items done for next release

wildsj:
Including vibrating nipple clamps, fishnet stockings and a strait leotard


Ben987:
There are lots of conflict in this pull request.  Are you sure you've used the latest version?

Can you do another one with only the new files?  You seem to be pushing old files with it, which creates the conflicts.

Thanks!

PR #142, opened 2019-05-12
Shock collar fix

Lucritius12:
A small fix for the shock collar that fixes the javascript issue caused by equipping a shock collar online and then using it without altering its intensity first. I've also fixed a few comment lines which I forgot to update when pulling code from the pump gag


Ben987:
Perfect!  I'll try it soon.

PR #143, opened 2019-05-14
Collar priority fix

Lucritius12:
This is a small item that changes the priority of collars to be higher than Clothes and Arm Restraints (such as the straitjacket leotard).
The priority is set to be just below other head items, so that things such as the pump and duster gag are still displayed with a higher priority than collars.
Outfits that are supposed to hide the collar have received the Hide: ["ItemNeck"] tag. I've gone through every single outfit and decided for each case which one looks best.

Finally, I've edited the collar sprites that are rendered on the Character a bit. They had a brown border before which didn't look good when being rendered above other outfits.


Ben987:
Thanks a lot!  I will try it later today.  

Cheers.

PR #148, opened 2019-05-18
A few tweaks and fixes

Lucritius12:
In this package we have a few fixes and tweaks to the code:
- New InventoryItemHasEffect(Item, Effect) function in Inventory.js, which checks if an item has a specific effect. For this, it checks both the item's asset and property attribute. If no Effect is specified, it just checks if the item has any effects at all. The new function has been integrated into the code wherever applicable, replacing the old code.
- Fixes a bug for the shock collar that would cause a Javascript error when using the remote on a different type of collar
- Tweaks to the code of vibrating eggs. Now any item with the "vibrating" effect will shake in the inventory
- Tweaks to the vibrating nipple clamps, including some fixes with the dialog text
- Changed the priority of nipple items. This is due to them rendering under harnesses before which looked kind of weird. Since bras still render below harnesses and bras are supposed to cover nipple items, a Hide: ["ItemNipples"] tag has been added to most bras so that nipple items don't appear in front of bras.

### DIFF
--- a/BondageClub/Assets/Female3DCG/Female3DCG.js
+++ b/BondageClub/Assets/Female3DCG/Female3DCG.js
@@ -166,23 +166,25 @@ var AssetFemale3DCG = [
 	
 	{
 		Group: "Bra",
-		Priority: 14,
+		Priority: 13,
 		ParentGroup: "BodyUpper",
 		Underwear: true,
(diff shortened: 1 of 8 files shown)


Ben987:
This all looks great!  Thanks a lot, cheers.

PR #149, opened 2019-05-19
Cloth gag merge

Lucritius12:
This is a tweak that combines all the four cloth gag types into one item, primarily to de-clutter the inventory a bit.
The combined cloth gag uses an extended item menu similar to the leather cuffs. In order for the different variations to be displayed on the character, Appearance.js has been updated to allow for variations of the same item. Which variation is displayed on the character is set through the item's Property.Type attribute.


Ben987:
That's a great idea!  It will thin out inventory a little.  I'll try it right now.

PR #150, opened 2019-05-23
Chat colors and tweaks

adokilume:
Improved and updated version of my previous Pull Request. :) This includes the following:
- Colors the chat message sender's name in their preferred label color (defaults to gray if not set)
- Turns actions gray to help distinguish them from normal messages
- Same for emotes, except they're also italic
- Displays the local time the user recieved the message next to it (this was requested on Discord)
- When a new message arrives, the chat only scrolls down if it was already at the bottom before, making it easier to scroll up if you want to reread something

Note that any of these styles are easily adjustable by changing the CSS in index.html.

Minor problems/concerns:
- I noticed there's no message when a player enters the room anymore, and I think I know why. The message seems to arrive before the player gets added to the chat room's array, so when the function checks for the player, they aren't actually there yet. This is something I noticed before while I was experimenting, not sure about how to fix this elegantly.
- Using a DIV with inner HTML might be a bit more performance-heavy on the browser than just a regular text area. I'm not sure at which point this becomes a problem, but if it does, it might be a good idea to delete the oldest messages as soon as they reach a certain limit. Maybe allowing each player to set that limit for themselves.

I was also going to include the following line of code to `ChatRoomCreateElement()`, submitted by barnmaddo#7290 on Discord:
`document.getElementById("InputChat").addEventListener( 'blur', e => {e.target.focus();} );`
It's supposed to fix the virtual keyboard popping in and out on mobile devices. But unfortunately I wasn't able to get the game working on my phone, so I couldn't test it and decided to leave it out. If you hvae the time, maybe you could see about adding it in?

PS: Here's a preview screenshot I shared on Discord: 
Screenshot (899)


Ben987:
This looks great!  I'll try it later today, thanks a lot.  Prost!

PR #152, opened 2019-05-24
Gag Effects Tweaks

Lucritius12:
This is a small change requested on Discord. People were pointing out that no matter how gagged you are, you should still be able to hear sentence structure and tone, and I agree. This change makes it so that ".", "!" and "?" characters are always preserved no matter the gag level.


Ben987:
That's fine with me :)

PR #153, opened 2019-05-24
InventoryItemHasEffect Fix

Lucritius12:
A small fix to the new InventoryItemHasEffect function. The function now only checks an item's Property attribute if you specify it in the function parameters. This should prevent abuse by hackers like using a collar with the "Lock" and "Freeze" effects on a player, which would otherwise trap them in the main hall with no way out.


Ben987:
Thanks a lot, I'll try it later today.

PR #154, opened 2019-05-24
Kitty blindfold

Lucritius12:
Apparently we are getting a new Kitty Blindfold to match the new Kitty Mask Gag. Also made by Verity!


Ben987:
Perfect!  Or should I say: Purfect.

PR #156, opened 2019-05-25
Nursery cost fix + Bunny set + Pigtail ribbons

wildsj:
This contains the fix to stop the nursery charging players when the ask to leave.
It also contains various clothes and a carrot gag for bunny girls
Only contains one style of hair ribbon for pigtails


Ben987:
Perfect!  It all looks good, I'll test it later today.  Cheers!

PR #157, opened 2019-05-25
Kitty paw mittens

Lucritius12:
Verity gives us yet another cat-themed item, this time it's paw mittens!


Ben987:
Very good!  More kitty stuff for players :)

PR #159, opened 2019-05-27
Add the possibility to send private messages.

NoneNoname:


Ben987:
Thanks!  This will be very useful.  There's one thing I'd like to change though, it's to make it more "User Friendly".  Instead of having to type /whisper or going in the player menu, I'd like a button or a drop list or something convenient directly in the main chat room screen.  Maybe a button next to each girl that would make you enter in "whisper mode" for that girl, or something next to the chat box.  What do you think?

NoneNoname:
Yes, i was not sure how to implement this in a user friendly or in a way you would like to see this. But yes clicking on a girl than entering "whisper mode" might be the right way and than add button to exit the mode. Adding something visual to the girl you are whispering to should also help I think.

adokilume:
Just as a note, please don't touch the translation file (Dialog_Online_DE.txt) unless you understand how it works and/or you're able to supply the translation yourself. The way you added the line, it would cause `(Ban her from the room.)` to be translated to `(Whisper to her.)`, and all the translations coming after that would stop working.

NoneNoname:
I see I have not worked with those files, I also changed up a bit so no more /whisper i will add a new commit shortly and I will remove that line from there as well

Ben987:
I'll use parts of your code in a few weeks, when I'm ready to include private messages.  I'll design some kind of buttons in the UI to make it easier for players to send private messages.

Thanks again!  This will be useful :)

Ben987:
I've implemented private messages / whispers in the latest version.  It's not exactly like you did it but you gave me good ideas, thanks a lot.

PR #162, opened 2019-05-27
German Bondage Club translation updates

adokilume:
Just to get the base translation up to date (mostly UI stuff), hopefully before the next release. The Sophie Chapter translatio unfortunately still isn't finished, but we might have another volunteer for that, as you probably saw.

**Edit:** I also tweaked the translation template generator a bit so it can now automatically fix translation files that are out of date, and used it for Sarah's and Amanda's dialogue to include their new lines (not translated yet), and then I also included my current progress on the Sophie file. This is mostly so that Drussler1 can start working on it, since they expressed interest.


Ben987:
Once again, you've outdid yourself!  This is great, thanks a lot :)

Sorry for the delay, I wanted to correct the bugs first before adding new stuff.

PR #171, opened 2019-06-11
German Bondage Club: Sophie completion

adokilume:
Took way longer than it should have, but the Sophie scene is finally done! All of these parts that weren't there before were translated by @Drussler1 , not me. Thanks again for the help! All I did on this one was proofreading and making sure it's somewhat consistent with the rest of the translation in trems of style.
A few of the lines might still be too long, though I'm not sure if that's even still a concern with the new Dialog code anymore. I haven't really looked into its limits yet. If the remaining 8 lines are still too long, I'll change them as soon as I can. Next I'll also take care of the remaining new strings that were added to the game.


Ben987:
Thanks a lot Drussler1 and Adokilume, this is great.  Danke!

Drussler1:
Thank you very much. With @adokilume it was really fun to translate the texts! Ben both of you are a good team! I'm looking forward to your ideas every week ^^

PR #181, opened 2019-06-18
Mini Wardrobe

NoneNoname:
I added a new wardrobe that does not load an additional 12 characters to be usable (18 with a fully loaded private room), it is available from the appearance selection menu and capable of saving loading other character clothes as well. This update also adds names to wardrobe characters.

Small fix for CharacterAppearanceHeaderText on others.


Ben987:
That looks pretty good.  How does it handle dressing up other players?  Like in online play or when dressing your submissive in your private room?  There's no bug in that case?

NoneNoname:
WardrobeFastLoad first removes all non-appearance item and in case of an online or an NPC it also keeps the non-clothing items, then it loads the items from the wardrobe with a reverse filter so any appearance item or clothing items only, then load the canvas and sync it with the server.

WardrobeFastSave gets all clothing items form NPCs or online characters than concatenate this array with the player non-clothing but appearance items, in case of Player the first gets all appearance items, then if the normal wardrobe is loaded it loads this new bundle on that character, then it saves the wardrobe

WardrobeLoadCharacters only load 1 character if the mini wardrobe is generated the first time and randomize only that character preventing 11 characters

I tested it but bugs could have remained, I tested on myself and NPCs so as online players the only problem that could think of if a default item is not loaded for some reason like item name has changed or removed the item which I just fixed.

The new mini wardrobe opens from Appearance menu so adding price to the change on NPCs should not be a problem.

This should solve any problem arising form wardrobe adding a lot of character on slow PCs or mobiles

Ben987:
Very well, I'll try it right now and let you know.  Thanks a lot!

PR #197, opened 2019-07-04
Add /me command to Chat

TheGnarp:
Also allow /me command to send emote messages (working exactly like '*') - for those users that have used /me far too often in the old IRC days or with Skype, etc.


adokilume:
I just had a quick glance at the code, maybe consider removing the `g` flag from the regular expression? Otherwise, if someone writes "/me" somewhere in the middle of an emote, it would get deleted. Removing the flag would only delete the "/me" at the beginning, leaving the rest of the message intact (assuming the player didn't use *, might actually want to use separate ifs for this). Probably an edge case and not super important though.

Ben987:
Thanks!  I'm not sure how many people will use /me but if it can help a few players then it's good for me :)

PR #198, opened 2019-07-05
Optimize invisible assets

NoneNoname:
Asset Visibility was not used, this commit makes unwearable items invisible so as the two invisible assets the egg and the black plug, as of now the unnecessary images are not removed.


Ben987:
Simple and faster than what we've had before.  I love it :)

PR #199, opened 2019-07-06
German translation update

adokilume:
German translations for all the new strings since the last update, plus a few minor fixes.


Ben987:
This all looks great.  Danke!

PR #202, opened 2019-07-10
backward and forward split for character appearance item selection and chat room selection

NoneNoname:


Ben987:
How does this works exactly?  The button is split in two?  Left side / Right side ?

NoneNoname:
yes if you click on the left side you go backward otherwise you go forward the button is split at the middle

Ben987:
Some players are easy to confuse, We'll get many bug reports stating that the menus are bugged.

Can you redesign both screens to have two buttons instead?

NoneNoname:
the chat room one is possible as for the appearance one that is a bit problematic since space is limited and text fills the hole button, maybe a different hover color can be done to the two sides but that won't be available on mobile, as of now if you miss clicking it only goes to a different direction, it could also be solved by giving the player to turn this on or off

Ben987:
I'll think about it, maybe there's an easier solution.  There's no rush anyway, there's only 3 screens to navigate in.

Can you make a pull request with the chat room background selection only?  maybe make two button and show the room name in the text label?  Up to you.

NoneNoname:
sure I make this pr to only change the chat room for now but I might leave the backward option for the next item for future use

PR #203, opened 2019-07-11
new wings and a 2 raccon tails

gatetrek:
just a few new wings and two raccoon tails


Ben987:
Thanks!  I'm sure that players will enjoy them :)

PR #205, opened 2019-07-13
Wiffle Ballgag added

B7Rad7:
I also formatted it to work like the harness ballgag in terms of coloring. I might look into making it possible for the straps and balls to be picked separately. For example you could pick HarnessBallgag and then select the solid or wiffle ball, and vice versa for the single strap.


Ben987:
That looks great!  Though can you use the "InventoryAvailable" function that's already there instead of adding a new "InventoryContains"?  We probably don't need to have two functions for that.  Thanks.

B7Rad7:
Oh no problem. I just didn't find the function the first time. I've got one more commit to finish adding the Exposed property to some of the assets. I'm still new to github so I didn't properly start the maid stuff in its own branch sorry.

B7Rad7:
Should be all set now. let me know if you have any other issues.

PR #208, opened 2019-07-14
Added Kitty Harness Panel Gag from Verity

B7Rad7:
New gag courtesy of Verity for the assets. Also made the kitty blindfold lockable since it is a leather blindfold in the preview.


Ben987:
It looks nice!  Thanks both of you :)

PR #209, opened 2019-07-15
Just a few toys

gatetrek:
Added: 
            Weighted piercing: is two ring piercing with a weight on each for the breast nipples
            Chain clamp two clamps on the nipples with a long chain linking them
            Anal beads will doesn't need an explanation name gives it away there
           Mermaid rope tie to the feet area


Ben987:
It looks great!  Thanks a lot, I will try it soon :)

PR #212, opened 2019-07-17
Two new masks

AlexBlakeKincaid:
I added two new masks to the game, a domino mask and a Venetian mask. They are purely aesthetic and are included in the glasses section.


Ben987:
Great!  But why are you including a file called Domino Mask 1.png?  Is that file necessary?  Thanks :)

AlexBlakeKincaid:
Sorry. That was some accidental flak when I was deciding where to save unfinished files. Would you like me to reupload without that file?

Ben987:
Just remove the file from your pull request and it will be fine then :)

PR #213, opened 2019-07-17
new mouth 5 and new dazed eye expression

B7Rad7:
Also added layers to the ring gag to separate out coloring the inside of someone's mouth, which was weird looking


Ben987:
Thanks!  That's a great idea.  I'll try it soon :)

PR #216, opened 2019-07-17
German Translation Updates

adokilume:
Just a couple more translations for the new lines in the private room, as well as a few other places (this isn't 100% up to date, I'm mostly just doing this now to lessen the workload later, and to update the german version of the line discussed in #201 ).
(This also includes a fix for a minor formatting error in the translation updating tool, and I removed the length checking for NPC lines from the status tool, since overflow isn't really a concern anymore.)


Ben987:
Awesome!  Some texts might be changed a little when I proof-read though.  I'm sure your tool will be able to catch the changes.  Thanks again :)

PR #220, opened 2019-07-20
German translation updates

adokilume:
This should be every new line in the game currently.


Ben987:
Once again, you're the machine :)

PR #221, opened 2019-07-20
Change res structure, adding ez/nm charts, add translation support.

sevensonix:
Made this mini-game available for mistress as well.
Revert the res folder structures to original format for easier organizing.
Adding the easy and normal difficulties variants of the existing song. The new song is still being worked on, not sure if it is going to be up before the next update.
Adding translation support, not sure if it works. (Note that some text such as 'PERFECT', 'MISS', I decide not to make them translatable, as these game terms are not usually translated.)
Adjusted the scoring function where 'PERFECT' hits have higer scores, so that player could earn higher bonus.


Ben987:
It looks great!  I will try it right now, thanks :)

adokilume:
I kinda think the translatability of "MISS" and "PERFECT" and such should be decided on a per-language, basis, but it's not a big deal

PR #222, opened 2019-07-21
missing null check causing error

B7Rad7:
noticed in the beta my code wasn't working right. I was missing a null check.


Ben987:
Thanks.  I'll update the beta pretty soon.

PR #226, opened 2019-07-22
Muzzle Gag and New Blindfold Type

Lucritius12:
Akita from the Discord made some nice looking new restraints for the game. These are pretty simple from a code perspective, no need for changes to extended code, but they do look pretty nice.


Ben987:
That's great, thanks!  I've modified "ItemMoth" to "ItemMouth", you can do the same change on your side.

PR #227, opened 2019-07-23
missing code

gatetrek:
the panties had the lock code part missing to make them keep locked after using the remote


Ben987:
Excellent!  Cheers.

PR #234, opened 2019-07-27
some more tape for the head and mouth

gatetrek:
added: Akita new tape art for the head and mouth
made wrap tape into part of an extension for the head instead of its own item


Ben987:
This is great!  the Mouse X & Y position doesn't seem right though.  Can you review the coordinates that you've put?  Thanks.

gatetrek:
on which item?
if it the face cover it been moved upwards as to not cover the others
http://prntscr.com/ok9pbo

Ben987:
Check MouseX and MouseY in InventoryItemMouthDuctTapeClick()

It doesn't seem to be correct from my side.

gatetrek:
yeah was only the Y's that was wrong and been fixed now
thanks for pointing it out

Ben987:
Excellent!  I'm doing the hot fixes for the current version but I will review it soon.  Thanks :)

Ben987:
That looks great!  I'll test it soon.  Thanks :)

PR #236, opened 2019-07-27
Added deafen effect

GermanPerson:
This is just a quick and dirty deafening effect. I saw it in the deviantArt comments and decided to implement it.

My code is probably questionable at best since I'm not familiar with the project structure. Feel free to suggest fixes. Currently, the leather hood has a deafening effect.


Ben987:
Thanks a lot!  I can't include it now since I'm too close to release.  But I'll review it in the next days :)

Ben987:
Instead of playing with the chat screen color, could you use the speech garble functions?  So any regular incoming text would be garbled, emotes and actions would be shown regularly and text that you received before being deaf would still be shown.  The light hood could garble a little and the thick hood could garble a lot.  What do you think?

GermanPerson:
Sounds like a good idea, although I think the thick hood needs a garble
function that barely let's anything through, similar to a fully inflated
pump gag. That would make it a full sensory deprivation hood.
Would that be acceptable?

Ben987 <notifications@github.com> schrieb am Di., 30. Juli 2019, 20:53:

Ben987:
Sure.  You could use the heavy garble function for the thick hood and the light garble function for the basic hood.  The heavy garble makes it impossible to understand anything that's a little complex.

GermanPerson:
Okay. I will create a new pull request for this.

PR #238, opened 2019-07-29
redo on the image size of the weights

gatetrek:
someone reported the size the weights being off so redid them to match all the normal-sized breasts


Ben987:
Great!  Thanks.

PR #240, opened 2019-07-31
Added deafening effect to hood, number 2.

GermanPerson:
I now use the standard speech garble functions, although I had to create a function combining both.
I didn't feel confident to modify the normal SpeechGarble() functions since a lot of stuff seems to depend on them.

What do you think?

### DIFF
--- a/BondageClub/Assets/Female3DCG/Female3DCG.js
+++ b/BondageClub/Assets/Female3DCG/Female3DCG.js
@@ -630,11 +630,11 @@ var AssetFemale3DCG = [
 		Left: 150,
 		Top: 50,
 		Zone: [[150, 35, 200, 125]],
-		Asset: [ 
+		Asset: [
 			{ Name: "ClothBlindfold", Effect: ["BlindLight", "Prone"], Hide: ["Glasses"], Value: 20, Time: 5 },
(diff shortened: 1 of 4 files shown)


Ben987:
That's good.  But there must be a way we can re-use the standard garble functions isn't it?  So we don't duplicate the code.  Also, the deaf effect should be applied when NPC speaks to the player.

GermanPerson:
I didn't spot a function that easily handles all text displayed. One way to
do that would be applying a gag effect to all NPCs when the players ears
are plugged, but that's a pretty ugly hack.
Do you have any suggestions?

Ben987 <notifications@github.com> schrieb am Mi., 31. Juli 2019, 23:14:

GermanPerson:
Okay, I found a way that worked and removed all duplicate code.
Please review.

Ben987:
That looks good, have you tested it with NPCs?  Does the dialog boxes/flow still works fine?  Thanks

GermanPerson:
I tested it with a few NPCs I could find and didn't spot any issues.
However, I don't know where to find all them, so it might be a good idea if
someone more familiar with the game tests this further.

Ben987 <notifications@github.com> schrieb am Fr., 2. Aug. 2019, 00:08:

Ben987:
Alright!  I'll merge and test it a little later, thanks a lot, cheers :)

PR #247, opened 2019-08-01
straps

gatetrek:
added tail straps
there still tails in the buttplug area but not all want their tails stuff into their butts

planned act
when code can allow having layer images have different priorities 
well be adding a belt that places in front around the waistline of the body to make it realise as a possible


Ben987:
Does it work well with other items at priority #3 ?  It doesn't create any conflicts to repeat a number?

PR #250, opened 2019-08-03
ear plugs and a collar

gatetrek:
kirsty earplugs her images
someone ask if could add a batcollar


Ben987:
That looks great!  Thanks :)

PR #254, opened 2019-08-07
German Club Slaves, Market etc

adokilume:
German translations for the new Club Slave and Slave Market content, as well as (probably) everything else that changed recently.

This also contains a typo fix in `Dialog_Player.csv` ("ONT" to "OTN"), I asked gatetrek if it was fine before I made this change.

I also noticed a possible typo in `Dialog_NPC_Management_RandomGirl.csv` (on line 5, "This slave is here to serve**r** you Miss."), which I didn't want to change without consulting you first, but I added two versions of that line in the translation file, just in case.


Ben987:
Yes, this is certainly a typo, I haven't proofed-read the text yet, I hope there won't be too much rework for you.  Thanks again, you're still the machine :)

PR #256, opened 2019-08-09
new Leg rope tie and ear plug in there own zone

gatetrek:
the upper part of the mermaid rope tie as well the ear plugin there own zone so people can have blindfolds at the same time


Ben987:
Yes, adding an extra zone for ears makes perfect sense, I'll try it soon.  Thanks a lot :)

PR #258, opened 2019-08-09
Added inventory: Irish 8 Feet Cuffs

AshKitty-bd:
Just trying to add inventory content. 
Cannot find store and price code, All other seems not broken.
Irish 8 Feet Cuffs game mechanics equal to Leather Belt. Just visual changes.


Ben987:
Thanks a lot!  If 8 feet isn't the correct word we can change it later.  I guess 8 means the shape of the cuffs.  Cheers!

PR #260, opened 2019-08-12
add OwnerOnly items, refresh on color change

NoneNoname:
this commit add possibility to owner only items, I added slave collar as an example which now can be changed by the owner, I also fixed the refresh on the color change which was missing


NoneNoname:
the pull request is split to multiple parts so this one is closed

PR #264, opened 2019-08-15
Typo Fixes

klorpa:


Ben987:
This is great, thanks a lot for checking the typos.  Are you sure about "Colour" instead of "Color"?

I rarely see "Colour".  I'm from North America though, you might be from the UK.  A quick search on google gives a lot more results with Color than Colour.

Thanks!

adokilume:
"Skillfully" is also more common than "skilfully" in American English, according to this: https://grammarist.com/spelling/skilful-skillful/

The rest looks good though.

klorpa:
I assumed the game preferred British spelling over American; there are several pre-existing examples of "colour" for example.

Ben987:
There's no preferred version of English setup, and every developers will write in their preferred version.  Try to use the most popular version I guess, we can use google to start with.  Color has more results than Colour and Skillful has more results than Skilful.  We can go with that.  The rest looks great.  Thanks!

Ben987:
I've merged most of your changes in this commit:
https://github.com/Ben987/Bondage-College/commit/b4fbfbd72e206c432d0f07ed553f42436e125918
Thanks a lot for your help!

PR #265, opened 2019-08-15
Friend list in chat search

wildsj:


Ben987:
That should be useful!  Thanks :)

PR #266, opened 2019-08-15
fixes and layers for the plug gag

gatetrek:
added layers to the plug gag so tongue doesn't colour and the muffle Effect is right now
added missing image lock to the irish8 cuffs
added another of Kirsty background image the chill room


Ben987:
This looks great!  Cheers :)

PR #275, opened 2019-08-20
refresh on color change

NoneNoname:


Ben987:
Good catch!  I will test it later today.

PR #280, opened 2019-08-22
quick fix to 2 bras

gatetrek:
added priority to bra style 3 and bandeau to hide some the torso items due big-chested ladies was not right


Ben987:
Thanks!

PR #283, opened 2019-08-23
Alpha channel, more code from defcab (ON HOLD), Alpha to shocks and gloves

NoneNoname:


NoneNoname:
this code do not change any color or image unless an alpha is provided, there was a filter part that changes colors but that is not implemented in this change

PR #286, opened 2019-08-25
German translations

adokilume:


BondageClub/Screens/Character/Player/Dialog_Player_DE.txt
```diff
@@ -11,6 +11,8 @@ Remove the chain to the floor first
 Vorher die Kette zum Boden entfernen
 A collar must be fitted first to attach accessories to.
 Für Zubehör muss zunächst ein Halsband angelegt werden.
+Must be collared and not suspended first.
+Muss ein Halsband tragen und festem Boden sein.
```
GermanPerson: Wouldn't "Muss ein Halsband tragen und festem Boden stehen." be a bit more correct here?

adokilume: Thanks, nice chatch.^^ Fixed it to add the missing word (I chose "sein" rather than "stehen" to also account for kneeling).

PR #288, opened 2019-08-26
fix layer has type false

NoneNoname:


NoneNoname:
so previously if a layer had HasType: false then the Type was cleared and no next layer could get it

PR #296, opened 2019-08-31
Diamond harness rework

gatetrek:
fixed the Prone on the tape that goes on the head|eyes it didn't stop you using the wardrobe before 
merged the normal hemp harness and the new diamond harness into a menu instead having to buy it


Ben987:
It's a great idea to merge it with the current harness.  Thanks!

PR #298, opened 2019-09-02
cage items

gatetrek:
New item devices images, stand up cage and kennel, created by Kirsty
Both items block access to all the bindings once the player is placed inside, also has hide the tails and wings due to clipping outside of the cage items
made a new Prerequisite to stop an item being equipped when kneeling and when suspended, was used on the stand-up cage


Ben987:
It looks nice!  I'll try it soon, thanks.

PR #300, opened 2019-09-02
Magic.js: changed ToC array to FromC array in the last if statement

Dracodeus:
Fixed bug in the magic show that caused an undefined error in the `MagicRestrainCopyTransfer(FromC, ToC) function`. The cause was the appearance array was trying to be accessed with an index > array length


Ben987:
Thanks a lot!  Cheers!

PR #303, opened 2019-09-03
Changes wording and a little fix

gatetrek:
(GTester uses a stand up cage on her body.)
instead
(GTester uses a stand up cage on her devices)

broke the Prerequisite due when was suspended you couldn't untie yourself even with hands-free and when clicking the Suspension it pops up saying needed to be standing and not suspended first >.<


Ben987:
Yes, that should work better.  Thanks.

PR #305, opened 2019-09-05
quick rename fix

gatetrek:
rename the mistress lock and key preview so their image shows up 
not sure if the metal belt image should be the leather belt preview but leave that up to you beny


Ben987:
Thanks!  I haven't finished cleaning up the belt stuff but this will be one less thing to do :)

PR #307, opened 2019-09-06
New belt UI enchantments, fixes (ON HOLD - Move code from character to chastity belt script)

NoneNoname:


Ben987:
It looks great!  I'll try it now.

PR #311, opened 2019-09-07
Belt wand

gatetrek:
a belt that has a vibe wand, might have it able to be swapped out to put a vibrator as a choose 
the socks on the boots show only the mistress one had 2 that didn't go well one only a line and the other when kneed would vanish so hid it instead


Ben987:
That looks great!  I'll try it right now.  Thanks :)

PR #315, opened 2019-09-09
new leg cuffs added

B7Rad7:
the body lower changes were just me noticing a small amount of grey not cropped out in the thigh gap.


Ben987:
Thanks a lot!  I'll try it right now.  Cheers :)

PR #316, opened 2019-09-10
Bolero straitjackets

wildsj:
New straitjacket that can be worn over clothes or just leave the occupant fully exposed.


Ben987:
Great!  I'm curious to try it.  Thanks!

PR #317, opened 2019-09-10
minor code cleanup: exit button for usable items

EveFerox:
removed occurrences of
DrawButton(1885, 25, 90, 90, "", "White", "Icons/Exit.png");
in all usable items.

Added exit button drawing in Dialog.js instead.


Ben987:
Nice idea!  Less code is always good :)

PR #318, opened 2019-09-11
UI work on the shock collar,pump gag and butt pump

gatetrek:
redid the items:
shock collar
Butt pump
Pump gag
did the increase and the decrease the same as the vibes UI
added dialogue on the pumps gag and butt.


Ben987:
Great idea!  Thanks a lot.

PR #322, opened 2019-09-13
belt invalid function and spankingtoys null fix

NoneNoname:
I removed some duplicated code as well


Ben987:
Thanks a lot!  Two bugs solved in one PR.  I love the way you simplified the spanking assets.

PR #323, opened 2019-09-13
full item lock on locked item

NoneNoname:
this allow an item to be unchangeable until it is locked


Ben987:
What does it fix?  Which item should have it's properties locked?

NoneNoname:
CollarNameTag needs this since the lock prevents the UI to change the Property.Type this code is to prevent hacks for the item while it is locked

PR #328, opened 2019-09-16
New Collar Name Tag item

FeuertinKira:
*new collar attachment for collars that allows to display up to 36 different pre-selected words, ranging from endearing and cute to humiliating and daring

*inventory script and button layout to support the name tag functionality by Nythaleath

*adjustments to make a selected name tag lockable and avoid switching it while locked by NoneNoname


Ben987:
That sounds like a good idea.  I'll try it right now.  Thanks.

PR #332, opened 2019-09-17
Small fixes

FeuertinKira:
-fixes a slight visual glitch on the small wooden box, removing a stray line that was visible on the left of it
-adjusts difficulty of collar name tags up, making them less trivial to remove when locked on, while not having the approriate key


Ben987:
Great idea!  Thanks again.

PR #334, opened 2019-09-17
wooden horse

gatetrek:
wooden horse and pose for it
some the cloths are hidden till can find time to do them


Ben987:
This is great, really awesome!  I'm too close to release date though, this will have to wait for the next version.  It will give you time to fix the remaining clothing options.  Cheers!

PR #337, opened 2019-09-18
Block fix to leg binder

gatetrek:
added a block to the leg binder to stop the suspension


Ben987:
Thanks!  I'll do a "hotfix" right now.

PR #339, opened 2019-09-19
spankingtoys fix type change on other characters

NoneNoname:
I also made IsRestraint asset specific instead of group specific


Ben987:
There are lots of changes in there, and some are touching the game core files.

Are you able to split them up so I can review and test them one by one?

Thanks a lot.

NoneNoname:
slit to multiple parts

PR #340, opened 2019-09-19
Tiny fixes

FeuertinKira:
-smoothed the bottom of the straitdress/legbinder belt images, so they line up with the smoothing changes on the straitdress/legbinder themselves
-cleaned bolero images a tiny bit after my previous edit


Ben987:
Thanks!  I'll do a "hot fix" right now.

PR #341, opened 2019-09-20
add Tags and Description properties to characters

EveFerox:
-text input boxes in character info sheet (InformationSheet.js)
-sync Tags and Description changes in rooms
-pushed player permissions field down a bit to make room


Ben987:
This looks great, I've been wanting a multiplayer profile screen for a long time.  But could you move it to a new screen, that be accessed from the information sheet screen?  Maybe have a new "Online Profile" screen or something like that.  Later we could also add a BDSM preferences list in there so players can specify what they like or don't like.  Also, what's the use of tags?  I'm not sure I understand the purpose of that.  Thanks a lot.  Cheers!

EveFerox:
replaced with https://github.com/Ben987/Bondage-College/pull/443

PR #344, opened 2019-09-21
wooden horse

gatetrek:
wooden horse and pose for it
only need do is socks, jeans, shorts and pj's but that's when I have the time to do them, so for now, post what I got for now
also, fix up the teacher outfit around the hands from this http://prntscr.com/p9326c to http://prntscr.com/p932bq


Ben987:
Could that new pose also be used for something else like spreader bars?  Could it be more generic?  Thanks.

PR #345, opened 2019-09-23
Added Buttons to RoomAdmin to moveLeft and moveRight

MrsHoudini:
Buttons for corresponding actions. I just added them below the kick and ban options. I have no better idea where to put them


Ben987:
That's a good idea!  I'll test it pretty soon.  Thanks :)

PR #346, opened 2019-09-24
New Hobbleskirt and Open crotch StraitDress

FeuertinKira:
New Hobble Skirt and Open crotch StraitDress:

-added new hobble skirt item for the leg slot

-added new open crotch strait dress item for the arms slot

*both items images based on the original StraitDress images by Kirsty

-changed the "Top" position of all leg images to allow for the hobble skirt to actually cover up to the  character's waist

-adjusted all leg item images accordingly to work with the new "Top" position correctly

-added a new pose "StraitDressOpen" to accommodate the new open crotch StraitDress, preventing clipping in any situation while still allowing to show the naked crotch and any existing clothes like pants, skirts and pantyhose correctly underneath


Ben987:
Why do you need a new "StraitDressOpen" pose?  Poses should be physical human poses, not item specific.  Thanks.

FeuertinKira:
For the exact same reason the "Bolero" has a new pose. To avoid clipping in various situations and wearing various clothing.

Example. character has small upper body and xlarge lower body, this dress exposes the crotch so I cannot just hide the lower body, while kneeling there will be massive clipping on the legs and knees making it look terrible. The new pose allows me to merge the crotch part onto the upper body images, hence hiding the lower body isnt an issue anymore. Same goes for wearing pants or pantyhose, the exposed crotch part can now still show the pants because of the new pose. Otherwise they would either clip heavily on the sides or have to be hidden entirely.

Tl:dr it is neccessary for proper implementation, same as it was with the Bolero

Ben987:
Yes, it's the same problem as the Bolero.  And I don't like the bolero pose either, someday we need to find a better system.  Poses should be physical poses, not item related.  Since we don't have much choices now I will add it.  Maybe someday we'll have a stronger engine that won't require bogus poses.  Thanks a lot for your hard work, I'm sure it will be stunning.

PR #347, opened 2019-09-24
Refactor

NoneNoname:
double InventoryGet -> InventoryIsWorn
InventoryLock add auto asset, auto lock
InventoryExtendedItemGetCurrentCharacter
"OwnerRule" return false if no owner
format Common.js


NoneNoname:
split

PR #351, opened 2019-09-24
IsRestraint on Asset

NoneNoname:


Ben987:
I like the fix you've did for the spanking toys and the IsRestraint on the item level, I've been wanting to do it for a while now.  Cheers!

PR #352, opened 2019-09-24
double InvetoryGet -> InventoryIsWorn

NoneNoname:


Ben987:
I like that change, but have you tested all the possible path it changes?  Or at least most of them.  Thanks.

PR #354, opened 2019-09-24
"OwnerRule" return false if no owner

NoneNoname:


Ben987:
In what case does an owner rule can be added without an owner?  What scenario triggers that?

Ben987:
Not needed, since this scenario is less than likely to happen.

PR #362, opened 2019-09-26
Change drinks to include mocktail

Sociocratic:
Hi,
First PR so let me know if I've not followed any conventions or broken anything!

I just thought there should be a non-alcoholic option for the drinks!


Ben987:
Having non-alcoholic drinks is a good idea.  But you could expand the menu instead of removing an option.  You could create a sub menu for free drinks, regular drinks and expansive drinks, and add non alcoholic drinks in all of them (water, orange juice, coffee, tea, virgin ceasar, etc.)  What do you think?

Sociocratic:
Happy to give that a go, was worried people wouldn't like too many clicks to get to the drinks

Sociocratic:
Will close this PR as I replaced it with #369

PR #363, opened 2019-09-26
Tape Expansion

FeuertinKira:
-Added many New DuctTape images for legs and feet
-Changed DuctTape for legs and feet to have 4 stages, ranging from the original sparse appliance to full coverage of thighs, calves and feet
-added according inventory scripts to support the functionality


Ben987:
That's a great idea!  I'll try it right now.  Thanks.

PR #367, opened 2019-09-28
German Translations (Slave Training, Asylum, general updates)

adokilume:
It was about time I got some work done on these again. This should as of now be fully up to date, unless there are new CSV files hidden somewhere that I didn't see.


Ben987:
This is huge!  Once again, you're the machine.  Danke.

PR #368, opened 2019-09-29
CharacterAppearanceFullRandom fix AllowNone

NoneNoname:
when the body is hidden it still must be part of the appearance


Ben987:
Good catch!  I will try it soon.

PR #371, opened 2019-09-30
extended item reduce code

NoneNoname:
unify vibration intensity
unify exit/image/description


adokilume:
While I do generally like and agree with the idea of having less text to maintain and translate, there's a (very) minor problem with this that I'd like to point out. For some items, the proper grammar is ``DestinationCharacter's vibrating egg stops vibrating.`` (singular), while for others, it should be ``vibrating nipple clamps stop vibrating.`` By combining everything into one template, you lose that distinction, and I'm not sure how it could be preserved and still work in other languages, which may have different rules for this. Not sure which is more important, keeping the amount of lines down, or having proper grammar.

NoneNoname:
I could add singular and plural to the text since it is a usual difference in most of the languages and that would work until an item keeps is singular/plural state in every translated language which I think it should (falling back to a single case is always possible), there are many ways to implement this the most simple is to add two versions of the full text and bind a singular/plural tag to the asset, other ways are also possible like a more complex string like something like this: "DestinationCharacter's {0} #(0,stops,stop) vibrating." whit this option multiple items with different singular/plural state can be correctly loaded on a single text.

NoneNoname:
I made the change for the more complex option `DestinationCharacter's {0} {0#stops#stop} vibrating.`

NoneNoname:
#380 will mess this pull request up and it will need fixes until that pr is not pulled I close this one

PR #372, opened 2019-09-30
better text for CollarNameTag

NoneNoname:
better text for translation


Ben987:
Absolutely.  I've been wanting to clean that one also.

PR #373, opened 2019-09-30
SlaveCollar color pick

NoneNoname:


Ben987:
The "Red" color indicates that the slave is owned, it's part of the club standards.  We would need to change some dialog also.  Also, couldn't we use the last color that was picked in the color selector and simply apply it?

NoneNoname:
as usual, I am not sure what you want in-game and what would be game-breaking it is quite easy to write this part of code rather than asking the idea, and this case you only need to add or reject it depending on if it fits, as for the color picker menu it makes it easier to select the color since the click is responsive and you get a color right after the click like for a cloth

gatetrek:
I know a lot of people would like to choose there own colour for the collar

Ben987:
I'll think about it, I would need to rewrite some dialog lines but it wouldn't be a big deal.  Maybe use the currently selected color would be simpler though.

Ben987:
I'll try it like that and see the dialog to change later.  Thanks!

PR #374, opened 2019-09-30
unify SetType, bugfixes

NoneNoname:
SetPose -> SetType with AllowType bundle check,
some minor bugs were fixed since Restrain sometimes were mixed with Type which caused selected type background would not be darker

new Type syntax with #: "<TrueType>#<FalseType>" in the image only the <TrueType> value is calculated in path


Ben987:
This is a pretty big change and I don't want to introduce bugs like we've did before.  Could you simply do the bug fix with what we have now?

NoneNoname:
the idea will need to be reworked

PR #376, opened 2019-10-01
2new dress and long strap Sandals

gatetrek:
2new dress and long strap Sandals with some fixes
added an alpha mask to the cages so people can have there tails and wings and also to the horse so close don't go through it


Ben987:
Why is the item name "Sandels", shouldn't it be "Sandals" ?  Can you rename it and rename the images?  Thanks

Ben987:
One last thing.  Could you name the item "Dress1" instead of "Dress".  I'm pretty sure that "Dress" will be used again.  Thanks a lot.

gatetrek:
sure but any chance I can do that after it's been pulled so be easer to redo them and send up?

Ben987:
Sure, just do another pull request for Dress1, I'll merge that one right now.  Thanks again!

PR #377, opened 2019-10-01
SetPose -> SetType bugfix only

NoneNoname:
I extracted the bugfix part from the other commit


Ben987:
Should you leave the NULL validations?  It could lead to code crashes if you remove it.

Ben987:
And what does this fix exactly?  What bug is there in the current version that I can recreate and validate?

NoneNoname:
in the modified DuctTape.js the Item.Property use Type in the code while the UI use Restrain which do not work

Ben987:
What doesn't work?  Can I see the bug now if I play?  Will I get an error message?  I want to make sure it actually fixes the bug.

NoneNoname:
other pull request fixes the same thing (Asset Type Info)

PR #380, opened 2019-10-02
Asset Type Info

NoneNoname:
Type Info and Type Selection is standardized and Property:(Effect/Block/SetPose/Hide/...) is moved to type information and no longer can be set individually in these cases further reducing the possibility of an undesired state of an item, the many draw/click function are combined, a possibility for hidden type by '_' added which do not set new image only adds possible type information

new Type formula: "ImageType_HiddenType"


Ben987:
I like the concept, but it could lead to many bugs and would require tons of tests.  I'm afraid to open the Pandora's Box before beta.

PR #382, opened 2019-10-03
Chat ui redesign

MrsHoudini:
UI changes to the chatroom to accommodate more buttons.
I'll start to build the room admin screen next. We'll see how it goes


Ben987:
Just let me know when the admin screeen is ready, the new buttons won't be needed when it's done.  Thanks!

MrsHoudini:
At least one will be needed. True the lock and private ones aren't really necessary if there is a proper screen, I put them like a sort of quick access action, And it gave me room to display the room name for non-Admins. I can remove them and add only the room one. 
I would still like to still split them in two rows, there will be 6 of them then, and keep the bigger input <textarea>. Got to figure out what the room button would do for non-Admins, Probably just display name, description, and maybe a list of admins?

Ben987:
Yes, only one button should be needed to access the room admin screen.  The screeen could be "read only" if you're not an admin.  If I remember correctly they are 5 buttons there.  You could do 2 rows of 3 buttons.  What do you think?

MrsHoudini:
Yes there are 5. With room it would make it 6, easy 2x3. I'll start doing it now. No idea how long will it take me, hopefully not too long.
I'll let you know when its finished.

Ben987:
Take your time, I'll be curious to see the final result :)

Ben987:
This looks great!  I'll test it right now.  Thanks a lot.

PR #383, opened 2019-10-03
better logic for less blocking when not chained together

B7Rad7:


Ben987:
Thanks!  I will try it soon.

PR #384, opened 2019-10-03
more reworks and clothes added

gatetrek:
reworked the flower dress so the flowers don't get coloured 
added a leather choker 
added pet collar to slave option image is Kirsty
added two new boots images was from Kirsty with some editing from me
added Fingerless Gloves images were from Kirsty with some editing from me
fixed one the masks was slightly off to the right, can be seen best when short hair


Ben987:
BondageClub/Assets/Female3DCG/Glasses/Mask1.png seems to be in conflict.  It could be due to another pull request.  Can you review please?  Thanks.

FeuertinKira:
BondageClub/Assets/Female3DCG/Glasses/Mask1.png can be left out of this pull request. 
I actually fixed the same issue in my changes, plus the resizing I did to allow the ButterflyMask to be displayed without being cut off.

PR #385, opened 2019-10-03
Ankle Socks added

TheGnarp:
My first effort to work with Git (Sourcetree) and only via Github, so please go easy on me.

I added some ankle socks since my character would never wear anything longer even on the North Pole.
Mostly as a pove of concept for my workflow, but I think it is a cool addition to the standard wardrone.


Ben987:
This looks great, I'll try it right now.  Thanks!

PR #386, opened 2019-10-04
more reworks and clothes added#2

gatetrek:
reworked the flower dress so the flowers don't get coloured
added a leather choker
added a bone gag 

added pet collar to slave option image is Kirsty
added two new boots images was from Kirsty with some editing from me
added Fingerless Gloves images were from Kirsty with some editing from me


Ben987:
Awesome!  I'll try it right now.  Cheers :)

PR #387, opened 2019-10-04
Lace Babydoll

gatetrek:
added another cloth LaceBabydoll
move the bone gag due to being called sleeping pill


Ben987:
This looks great.  Make sure you name the files "XLarge" and not "Xlarge" though.  It will create issues on Linux servers and the main production server.  Thanks.

gatetrek:
>.< why do I always do that

PR #388, opened 2019-10-05
Chat Display Settings

adokilume:
A new menu as a preference subscreen that allows players to adjust the look of their chat. To access it, go to the preference menu and click the chat icon on the right. On first use, the settings default to the style that's currently active in the game for all players. These settings can also be changed without issue while you're inside a chat room.

I added a bit of code to allow styling messages differently that inform about players joining or leaving a room, since a few people said those could get annoying sometimes. If possible, maybe this could be changed in the future? So that the server would send them as their own message type, ideally allowing these strings to be translated as well.

I probably didn't test all possible scenarios you can create by playing around with the settings, but I did test a good number of them and didn't run into any trouble. Hopefully everyone will be able to find a style combination they like. :)

(Also removed some obsolete lines from the ElementScrollToEnd function, which I believe should fix problems with the mobile keyboard, as discovered by a Discord User a while back. If it hasn't been fixed another way already.)

### DIFF
--- a/BondageClub/CSS/Styles.css
+++ b/BondageClub/CSS/Styles.css
@@ -51,6 +51,44 @@ canvas {
 	font-style: italic;
 	color: silver;
 }
+#TextAreaChatLog[data-ColorTheme=Dark i] {
+	background-color: #111;
+	color: #eee;
+}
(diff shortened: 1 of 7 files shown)


Ben987:
This is huge!  Thanks a lot, I will try it soon.

PR #390, opened 2019-10-06
German Translation Updates

adokilume:
Just a few updates to account for the recent changes. I also altered the way the background image selection for chat rooms works, so they can also be translated, as well as have proper English names (that has been bugging me for a while). I hope you're fine with that change.


Ben987:
This looks great!  I will try it right now.  Danke.

PR #391, opened 2019-10-06
Enable selfbondage on rope harnesses

GermanPerson:
In real life it's one of the easiest rigging tasks to tie a harness by yourself.
This commit enables selfbondage with both types of rope harnesses.

### DIFF
--- a/BondageClub/Assets/Female3DCG/Female3DCG.js
+++ b/BondageClub/Assets/Female3DCG/Female3DCG.js
@@ -671,8 +671,8 @@ var AssetFemale3DCG = [
 		Top: 200,
 		Zone: [[50, 250, 100, 150], [350, 250, 100, 150]],
 		Asset: [
-			{ Name: "NylonRope", DefaultColor: "#909090", SelfBondage: false, Value: 15, SetPose: ["BackBoxTie"], Effect: ["Block", "Prone"], Time: 15 },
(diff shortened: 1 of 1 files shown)


Ben987:
This isn't the rope harness, this is the full arms bondage.  The rope harness can already be done on yourself, check the ItemTorso section.

GermanPerson:
Sorry, I was actually referring to the full arms bondage. I always refer to that as harness too.
My point still stands: a simple rope boxtie that both of the rope options are is trivial to tie by yourself and does not require a great amount of rigging skill.
It should be possible to tie by yourself.

adokilume:
Hope you don't mind if I add a comment here. I don't really have any rigging experience, but I kinda doubt tying your own hands behind your back would be "trivial", without much skill. Maybe it could become possible after you reach a certain bondage skill level, just like in the College game.

GermanPerson:
A nice tutorial I found [here](https://www.boundanna.com/html/en_tech_selfbondageboxtie.html) shows a stripped down version of a boxtie. It uses a small tweak to make it easier, but I think being a  tiny bit inaccurate isn't that bad.

GermanPerson:
Requiring a certain skill would in general be a fun mechanic.

Ben987:
Without lots of knowledge, it's pretty much impossible to tie yourself like that.  Just check the link that you've sent, it's far from simple and you need a metal ring + zip ties.  Someday I might do more complex self bondage but there would have to be a skill requirement.

PR #393, opened 2019-10-07
just some more clothing and hair

gatetrek:
added a new pair panties 
added new socks
added four new back hairstyles 
added one front hairstyle


Ben987:
Wow!  This is huge.  Thanks a lot.

In Female3DCG.csv, you're adding twice "HairBack8".  Is that a mistake?  Thanks again.

gatetrek:
yes was meant to be a 11 >//<

PR #397, opened 2019-10-08
Add Saddle Stand and Leg Spreader, allow shorthand hex color codes

TheGnarp:
v1.0 of the Saddle Stand and Leg Spreader. There are still ways to mess things up a little if you apply resrtains that conflict with the pose, but this should only be a minor issue and will be improved in future updates. The assets and there, we'll get it to work. Please feel free to improve rules, effects, prereqs, etc.

Also added a small addition to the hex color code validation (which should have been in its own branch/pull request) to allow short hand hex color codes like `#05A` for `#0055AA`.


adokilume:
(Nice work, just one small thing I'd like to say: In the future it would be nice if you didn't edit the translation files yourself, unless you actually know the language well. Us translators have a tool that lets us automatically find new text that was added or changed, and translate it when we have time. Makes it easier to keep things consistent. You don't need to change it back, I just wanted to mention it. :) )

Ben987:
That looks greats, thanks a lot.  I will try it later today.  Thanks for the short hex also, but next time try to do two different pull requests.

TheGnarp:
Thanks, Ben and adokilume!
Still learning and always trying not to mess things up too much or annoy too many people. ^_-
Good thing you told me about the translation tool. I'll have to look into this since I can at least take care of two languages without accidentially saying something rude.

PR #401, opened 2019-10-08
Hood Rework

FeuertinKira:
-added three new hoods:
*Sensory Deprivation Hood - does what it says, takes away all senses basically
*Open eyes and mouth Latex hood with hair - very light, mostly aesthetic
*LeatherHoodSealed - In between, basically like the open mouth hood, but without the open mouth

-"nerfed" deafen effect on old hoods, reducing Deafen effects from DeafenHeavy -> DeafenLight and from DeafenLight to none

-increased gag effect of Leather Hood from Light to Normal

-realigned all "ItemHead" images to fit the new Left - Top positioning to avoid cutting off the new hoods images prematurely


Ben987:
Thanks!  I will try it right now.

PR #402, opened 2019-10-08
frill panties/hair fix and other fixes

gatetrek:
in the right spot now
added the code for the missing Prerequisite "LegsOpen" for the metal leg spreader


Ben987:
Awesome!  I will try it right now.

PR #403, opened 2019-10-09
New Room Background Images

FeuertinKira:
-Added six new online room background images


Ben987:
Excellent, I will try them right now.

PR #404, opened 2019-10-09
Crotch ropes

TheGnarp:
Added nylone and hemp crotch ropes
Now with no unintentional other commits since we have a separate branch.


Ben987:
Very nice!  I'll try it soon.  Thanks.

PR #407, opened 2019-10-09
Added lock image for metal leg spreader

TheGnarp:
My test character didn't have any locks... so she couldn't test that.
Yeah, let's blame her!


Ben987:
Yep, let's blame her :P

PR #409, opened 2019-10-10
Fixes

gatetrek:
moved mistress bottoms into the right spot
could only fix 2 sizes of the LatexPants1 the small and normal
did this when you kneeled https://prnt.sc/ph776d
but couldn't fix the Large and XLarge  the person who made them might be able to fix those
and fixed a mistyped on one the locks 
fix to the saddle to keep in in that place instead of going into the roof when put boots on like so 
http://prntscr.com/ph7ik2
and some other fixes https://prnt.sc/ph7vyl


FeuertinKira:
Thanks for adding the kneeling images for the latex pants. Noticed earlier I forgot them... Will save me some time though, cheers!

Ben987:
Sweet, thanks a lot.  I'll test it soon.

PR #411, opened 2019-10-10
Chopstick gag

wildsj:
Gag made from chopsticks


Ben987:
Good idea!  Cheers :)

PR #413, opened 2019-10-11
Refactoring of SpeechGarble

witcher2:


Ben987:
Have you tested it with all types of gags strengths and both in online chat and with NPC?  I've had bad experiences with custom refactoring before, I don't want to introduce new bugs.  Thanks

witcher2:
I tested it with all 4 gag strenghts in online chat and with NPC's, and it works just like it used to work.

PR #414, opened 2019-10-11
Chopstick tongue

wildsj:
Fix to prevent tongue being recolored with the chopstick gag


gatetrek:
>.< but green tongues for all

wildsj:
Bad luck gate, these chopsticks are now certified free from colorings and preservatives.

Ben987:
Good catch!

PR #416, opened 2019-10-11
Chat Style Adjustments

adokilume:
Two small changes to the new chat styles, based on feedback from Discord:

- adjsuted the code a bit to make the non-default styles work in the Edge browser
- changed the color of whispers when using dark mode, to make them more distinguishable from regular chat


Ben987:
Awesome!  I'll prepare a second beta with it soon.

PR #417, opened 2019-10-11
Bad comma fix with a police woman hat

gatetrek:


Ben987:
Awesome!  I'll prepare a second beta with it soon.

PR #418, opened 2019-10-12
Bamboo Gag

TheGnarp:
First of several new bamboo shibari bondage items - just to start things off with something simple.


Ben987:
That's a great idea.  More shibari items will be a nice touch.

PR #419, opened 2019-10-12
Lovers system and lover locks

EveFerox:
-Players can chose one other player as their lover via 'manage your relationship'
-Added lover locks that can be used on characters that set you as their lover

-Changed how character sheet displays lovers to support both text and new lovers object type
-Owners can able to remove lovers locks

***requires applying server PR (https://github.com/Ben987/Bondage-Club-Server/pull/10)


FeuertinKira:
This is the most important pull request ever made.

adokilume:
Shouldn't adding someone as a lover require confirmation from the other person?

FeuertinKira:
Adding someone as lover doesnt do anything by itself, apart from giving permission -to- the person you set as lover to use Lover's locks on you. So no harm can be done with it, unless you want to count displaying someones name as their lover on their character sheet without the other person agreeing... Which someone could already do in the game as of current if they really wanted for some odd reason.

Ben987:
That's impressive.  I'll start a discussion with the gang first to see how they want the lovers to be setup.  It will have to wait until the next full release is ready.  Cheers!

tetris245:
Almost 2 months since this big contribution has been submitted and still nothing in the official code. It's very disappointing. Besides, a bad decision has been taken concerning a small item related to this big contribution: the lover collar has been excluded from the options for the slave collar. It's absurd, because a lover can also be a sub, and i consider my unofficial lover as my first sub. Fortunaly, there's a solution: no more full collaring for my subs, and I will invite my lover to start a new trial. It will be a 'permanent' unfinite trial, and I will be able to use an owner lock on a lover collar.

Kitsune-hellfire:
the lover leather collar was actually taken and added to the slave collar selection with out permission from the creator of the collar

Ben987:
@tetris245, this change cannot be included as it is right now, there's lots more to do for a lover system to be deployed in game.  Here's a quick list:
- Have 3 lovers level (girlfriend, fiancée, wife), each level must have some time between each other.  And the wedding should cost some $$$ to both lovers.
- Allow getting NPC lovers in the private room with the same 3 levels system.
- The system must be consensual and breakable like the owner/sub system.  So an NPC somewhere must be written to be able to break the bond, and when it's done, both lovers must be updated as having no lover.

I'll probably code all of it in the next months, but there is lots of work to do and lots of testing.

EveFerox:
Atasly done a better job :]

PR #420, opened 2019-10-12
Teddy bears anyone and a fix to the saddle kneeling

gatetrek:
hold your teddy bears today
there a puppy, kitty, bear and fox to hold and when tired they sit next to you well when your tired and kneeling it kinda runs off. the teddies are Kirsty I just did some reworks on them and made more for the pose's 
added freeze to the saddle so cant kneel and so cant be equipped while kneeling
mask1 still was like this so fix it as well
before: https://prnt.sc/pihtub
after: http://prntscr.com/pihxku

if feel the price of these teddies is too high you can change the price if you like
if wildsj says yes can add them to the Nursery
added OverrideHeight to the mistress and pony due if wear them and BalletHeels/BalletWedgesyou shoot to the roof
added a bigger harness ball gag


Ben987:
Thanks a lot, a fun extra feature for the official release :)

PR #421, opened 2019-10-12
Couple Fixes, New Clit Piercings & another Halloween addition

FeuertinKira:
-Added Missing PolishedMittens_Lock image

-Changed SaddleStand to not allow kneeling

-Changed LatexHoodOpenHair to show glasses

------------------------------------------------------------------------------------------

-Greyscaled Fishnet Stockings, this way the game can actually change their color, opening many more options than just black fishnet stockings

-Fixed a typo

-added two more "Hide"exceptions to avoid clipping on ItemBoots

--------------------------------------------------------------------------------

-Changed hidden item entries on "Cloth" items around so only cases that cause bad clipping are hidden, opening up more outfit variety

-------------------------------------------------------------------------------------------

-Adjusted the HideItem entries on Cloth to include Skirt1 and TennisSkirt1 for all dresses

-Added New Clit Piercings based on the ones available for the nipples:

*New Straight Clit Piercing
*New Round Clit Piercing
*New Weighted Clit Piercing

-Added new Chastity Clit Piercing

-------------------------------------------------------------------

-Added new Pumpkin Gag; image of the Pumpkin Gag by Tyna - another little addition for Halloween, hehe

-Added new Maid Collar; image of the Maid Collar by Tyna

-fixed uneccessary spacing on HarnessBallGag1

PR #422, opened 2019-10-13
German Translation Updates

adokilume:
Once again, keeping the translation files up to date and adding new ones. I already made it so this will be compatible with FeuertinKira's typo fixes from #421 .


Ben987:
Perfect!  Thanks a lot.

Ben987:
I haven't done my proof reading though, so there might a few changes, but we can cover them later, cheers!

adokilume:
That's alright from my side. Revising the translation files after you've done the proof reading won't be a problem. I'd rather do these more frequently and change things multiple times than wait a few weeks and have thousands of lines to do at once. :)

Ben987:
Very well.  There's usually not that many changes after proof reading anyway,

PR #423, opened 2019-10-13
Chat Color Fix

adokilume:
Just a quick fix for a problem that was pointed out to me on Discord. If a player hasn't specified a color for their label, the default color for actions and emotes had no transparency and looked strange when using the dark theme.


Ben987:
Good catch!

PR #424, opened 2019-10-13
Extend private room beyond 8 people

MrBondt:
Added ability to extend the private room to have room for more than 8 people. To navigate through multiple rooms more easily, I've also added a counter and a previous button.


Ben987:
Thanks a lot.  But 8 is already a big crowd.  I'll think about that one.

adokilume:
Being able to buy a single room extension is somewhat believable, like buying a room next to yours and they're connected with a door. Being able to buy an indefinite amount of them though, until you're able to habe hundreds of guests, would make me start really questioning this club mansion's layout.^^'

MrBondt:
I've been experimenting with this further and also came across a performance issue with a lot of people when you enter your private room. I also agree with you both that it breaks immersion.

I had fun finding out how the code works and making my changes though, but for now let's not do this :)

PR #425, opened 2019-10-13
Make Spreader Bar more compatible with Saddle Stand

TheGnarp:
Updated the image and changed the priority of the spreader bar to make those two items work together even better.
Already saw the other fixes for those toys and I'm happy that we help each other out! Feels great.


Ben987:
Thanks a lot!  Cheers.

PR #426, opened 2019-10-14
Remove character difference checking in ChatRoomSync and CharacterLoadOnline

NoneNoname:
less work on each ChatRoomSync by knowing which character changed in the room


Ben987:
That's a good idea, but did you test it a lot on your side to make sure it doesn't create any problem?  Thanks.

PR #427, opened 2019-10-14
Re-add accidentially overwritten fix ^_-

TheGnarp:
Removed the spreader's priority (which was needed when the saddle stand was not yet a device) but it is hard to get rid of as it seems :P


Ben987:
Alright!  Thanks.

PR #429, opened 2019-10-14
three new items

gatetrek:
spanking toy whip and two new teddies images are from Kirsty
http://prntscr.com/pjl9zy whip
http://prntscr.com/pjlaif teddies


Ben987:
Great!  I'll slide it for the full release.  Thanks a lot.

PR #433, opened 2019-10-15
Hair accessories extension

Nythaleath:
i was requested by a few people in the club to see if i could get multiple hair accessories to work at once
what i did:
-renamed HairAccessory to HairAccessory1
-added second hair accessory clothing slot HairAccessory2
-moved the bow, ribbons, and flower to hair accessories
-made hair accessories be bought for both categories when buying one


Ben987:
That's a good idea.  What if the player already bought hair accessories from the shop?  She will lose them and have to buy it again?

Nythaleath:
yes she would unfortunately have to but this is also intentional as buying them again will buy them for both slots, and i couldn't see how to have the same effect any other way

gatetrek:
wouldnt leaving it called HairAccessory,  let people keep them and the other instead of being called 2 but 1 instead?

Ben987:
Sweet!  I'll test it right now.  Thanks a lot.

PR #434, opened 2019-10-15
Chopstick nipple clamps

TheGnarp:
Added Chopstick nipple clamps to go with Wildsj's Gag.
Only "issue" I see at the moment is the ability to add clothing over the clamps (bras and upper body clothes) which will make them vanish. For some clothes, this makes sense, for others (skimpy bras) this looks like magic. But hey, when somebody is wearing those clamps, she should be restrained and naked anyway, right? ^_-
Preview courtesy of Wildsj


Ben987:
Awesome!  We have more and more Asian items :)

PR #435, opened 2019-10-16
Love belt fix

NoneNoname:
Sophie could remove the love chastity belt with this update she is no longer able to do so

typo fix: SophireReleaseEveryoneButSarah -> SophieReleaseEveryoneButSarah


Ben987:
Thanks.  Have you played the full Sophie scenario with and without the belt to make sure it's fluid and everything works?

Ben987:
I've changed Sophie dialog to prevent going the submissive route if wearing any chastity item in a previous commit.  Thanks for pointing out the problem.  Cheers!

PR #436, opened 2019-10-16
Missing images fix

gatetrek:


Ben987:
Good one.  I'll do a hotfix right now.

PR #438, opened 2019-10-17
Fix HeightModifier

NoneNoname:
fix typo OverrideHeight -> Height on assets
only visible items should modify height


Ben987:
Sweet.  Have you also tested the locking boots + Mistress boots combo?  I'm guessing that the "visible asset validation" takes care of it.

Ben987:
Thanks a lot, I've fixed it with this commit: https://github.com/Ben987/Bondage-College/commit/e7af23a7091ba1da71f3f9b7bc46a94866d06e4d

It's the same as yours with the visible condition reversed.

PR #439, opened 2019-10-17
reduce the CharacterAppearanceSort calls, remove AssetReload

NoneNoname:
in some cases the sort was done in three times
assets should never be invalid state there by AssetReload is pointless


Ben987:
That's a great catch.  Have you done lots of tests to make sure it doesn't create any bugs?  Thanks.

NoneNoname:
yes, item order only matter in CharacterAppearanceBuildCanvas function which only called from CharacterLoadCanvas

Ben987:
Alright, I will try it later today.  Lets hope it doesn't create any strange situations.  Thanks a lot for that optimization.

PR #442, opened 2019-10-19
add JSDoc to Asset.js Character.js Inventory.js

NoneNoname:
adding JSDoc maybe helpful to other developers this is a minimal start for that

the Typedefs.js does not need to included in index.html and it should only contain JSDoc

it might be worth considering removing all comments in the actual web page


Ben987:
It's probably not needed for now, programmers are able to work with the current code.  Thanks for the idea though.

PR #444, opened 2019-10-19
Capture Ball

Nythaleath:
was talking to some people about making a pokeball like item so here it is... can be explained in RP person is just hiding somewhere as opposed to actually being inside of it

what i did
-added a CaptureBall item


Ben987:
The idea is funny, but the club doesn't have much "supernatural" or "fictional" yet.  I'm not sure if players are ready for that or if it would break the mood.  I'll do a quick survey to see if players would like that.  Thanks.

PR #449, opened 2019-10-22
[ON HOLD] ChatRoomStatusManager v0.1 - add player is typing status

EveFerox:
For reference or when classes are welcome :(

on server: https://github.com/Ben987/Bondage-Club-Server/pull/13

quick gif of how it looks like: https://gyazo.com/d917e32546ada35fd9142ddcdc72a029


Ben987:
Sure.  Some of that code will probably be used someday when I get to it.  Thanks!

PR #450, opened 2019-10-23
Added russian (and cyrillic based) language support to gag talk feature

FunnyEnthusiasm:
Now gag talk feature works for Cyrillic based languages.


Ben987:
Thanks a lot!  I'll do the same for Latin accents, it's a good idea.

PR #455, opened 2019-10-28
Stable dialogue revision

adokilume:
Expanding on the dialogue fixes that @GermanPerson started, as I mentioned earlier. Unfortunately I had less time than I thought, so this is still just the trainer dialogue for now. I will do the same for the pony girl and the other rooms soon (I messaged Git4nick and got the OK from him to do this).


Ben987:
Great!  I'm sure that git4nick will appreciate.  Cheers :)

PR #457, opened 2019-10-30
Rope Item Merge

Nythaleath:
-Merges Mermaid and basic ties on Legs and Feet
-Merges Crotch Ropes into the Hemp and Nylon Harnesses
-Makes all rope items of the same type in the same buy group


Ben987:
That looks great.  I'll test it later today.  Cheers!

PR #464, opened 2019-11-01
remove to fix

gatetrek:
the pet collar just the leather choker but used to be in the slave thingy so no need to be able to buy it


Ben987:
You should put -1 as value if you don't want players to buy it.

PR #465, opened 2019-11-01
wrist shackles and one bar prison fixes

Nythaleath:
-fixed dialog error with wrist shackles
-fixed alpha mask of the boots cutting off the base of the One bar prison


Ben987:
Thanks a lot!

PR #466, opened 2019-11-01
Small fixes

FeuertinKira:
-Fixed hands clipping over skirts again and for good this time...

-Changed Sandals to only hide socks that actually clip and changed their priority accordingly


gatetrek:
the Priority to the Sandals so they go under the pants not over them due pants be clipping
and won't talk about socks and Sandals >.<

Ben987:
Thanks!  I'll try it soon.

FeuertinKira:
I tested it and they still went under the pants, so not sure what you mean. And the socks are all hidden still from the sandals, only enabled stockings.

PR #468, opened 2019-11-02
fix and another collar item

gatetrek:
CattleProd was being sold as a racket so change it
added a new collar not so sure on the name yet so might change later


FeuertinKira:
Not sure about the priority change for the full blindfold. It will look a bit better with the larger coverage gags, but it will also look weird with a few other of the gags. E.g. tape, ring gag, spider gag, cloth stuffing, chopstick, carrot will all look a little bit weird when sitting above the blindfold.

gatetrek:
that blindfold looks like be touching right to the face so be odd that a ball gag gets covered by it

Ben987:
Good catch.  Thanks for the pull request.

FeuertinKira:
Well now it will make it so that the mouth portion of gags like the ring gag will appear to sit on top of the blindfold which is even more odd. technically it also clips slightly into the balls of the ball gags now

PR #469, opened 2019-11-03
Tiniest fix

FeuertinKira:
-Changed Priority on sandals to 21 from 22, so now there is no chance of them being above the pants, regardless of which one is put on first.


Ben987:
Thanks!

PR #473, opened 2019-11-03
Magic & Prison dialogue revision

adokilume:
Some more dialogue changes, this time focusing on the prison and the magic theater. Also fixed a few oversights in the photo room, and removed the content from `Text_Magic.csv` because, as far as I could tell, it was just copy-pasted over from the stables, and was never actually used.


Ben987:
Excellent!  This is huge, thanks a lot :)  I'm sure git4nick will be thrilled.

Issue #467, opened 2019-11-01
Player Information difficult to read


Dallun511:
@Ben987 can you do something, for better readability attempted?

![grafik](https://user-images.githubusercontent.com/40240507/68060036-0c049e00-fcff-11e9-8737-f6f61c4330d6.png)
***
![grafik](https://user-images.githubusercontent.com/40240507/68060056-1c1c7d80-fcff-11e9-83b1-fbf4a9989816.png)
***
![grafik](https://user-images.githubusercontent.com/40240507/68064104-ab358f80-fd17-11e9-8e42-9f7afd9b2ca6.png)
***

Regards
Dallun511

Ben987:
You could put abbreviations so it fits better in German.  Or change the buttons size in the love belt.  Good luck with whatever solution you prefer.

Dallun511:
I will come up with something

adokilume:
@Ben987 Would you be okay with simply using the code that automatically makes the text shrink to fit the buttons, just like the item labels do? I think that would look a lot nicer than any kind of abbreviation, and would be easier to work with in general.

Ben987:
Yes, I could code the DrawButton function to shrink the text if it helps you guys.  I'll add it to my to do list.

adokilume:
Alright, thanks, applying that to the button drawing funtion is actually way easier than what I had in mind. :) I think it would simply make a lot of sense, this way all buttons could fit any length of text in any language, without needing to adjust any of their widths.

Dallun511:
![grafik](https://user-images.githubusercontent.com/40240507/68097505-ac052780-feb7-11e9-9822-5bc38fbb1358.png)
@adokilume Was sagst du zu dieser Formulierung?

adokilume:
@Dallun511 Finde ich gut. 👍 Sollte für die Meisten auch klar sein, was das spieltechnisch bedeutet.

PR #474, opened 2019-11-04
Fix one problem #467

Dallun511:


Ben987:
Danke!

GermanPerson:
I don't think "to get to know her better" conveys as much as information as "to learn her traits". 
Are you sure this is a good change? Objectively, the current translation is more accurate.

adokilume:
@GermanPerson The argument for this change was to make it fit onto the character sheet better by shortening it. The previous one was overflowing it slightly, as seen in #467 .

PR #479, opened 2019-11-05
Dog Hood

Nythaleath:
adds Dog Hood & collar


Ben987:
Great!  I'm sure there will be plenty of "pets" that will like it.

PR #480, opened 2019-11-06
Fix For toe items

Nythaleath:
- Prevented use of toe items on items that should otherwise keep your legs open


Ben987:
Good catch!  I will prepare a hot fix.

PR #481, opened 2019-11-06
Two missing images & small priority adjustment

FeuertinKira:
-added two missing images for piercings

-fixed an issue where 'VulvaPiercings' would sometimes appear behind 'Pussy'


Ben987:
Great!  I'll push it to the live version right now.  Cheers.

PR #482, opened 2019-11-06
Energy drinks

wildsj:


Ben987:
Why do you add this line in the character information sheet?
SkillGetLevel(C, "Evasion");

It seems out of place.  Thanks a lot.

wildsj:
That was a fix, a crude one I admit.

Without it, if the a player looked at the information sheet before doing anything else to trigger that command, while they actually still had an active modifier then the modifier would not display.

Just having it like that didn't seem to cause any errors when I tested it.

Ben987:
Alright then.  For the global variable names, could you name them based on your module, so put the "Cafe" prefix before them.  I'm thinking about these variables:

var EnergyDrinkPrice = 5;
var GlassMilkPrice = 5;
var CupcakePrice = 5;
var AskedFor = null;
var Price = 0;

So instead of: Price, you would have: CafePrice.  Just to make sure they don't clash with other variable names in other modules.

Thanks a lot.

wildsj:
I have corrected the variables names.

There are a couple of things though I was wondering about regards the skills modifier. At the moment a player with a skill level 6 or higher can still buy up to a +5 modifier, but their skill is still capped at 10. Should players with high skill levels be able to boost them over 10 or should I add some way to discourage them from buying modifiers they cannot actually use?

The other thing is how much should they cost and is a skills boost of up to +5 just from buying energy drinks reasonable amount?  For thinks like the Pony training it would mean players would only need to do a fraction of the training.

Ben987:
Yes, you could change the code to allow going up to 15 in a skill.  So 10 from experience and +5 from boost.

You could change the GetSkill code from:
			if (Level > SkillLevelMaximum) Level = SkillLevelMaximum;
			if (Level < SkillLevelMinimum) Level = SkillLevelMinimum;
To something like that:
			if (Level > SkillLevelMaximum + MAXMODIFIER) Level = SkillLevelMaximum + MAXMODIFIER;
			if (Level < SkillLevelMinimum - MINMODIFIER) Level = SkillLevelMinimum - MINMODIFIER;

And set skill SkillLevelMinimum to zero.  Let me know if it works for you.

Should the pony dressage be affected by the boost?  It might mess the logic coded by git4nick.

wildsj:
It seems reasonable that all skills should be boosted, but +5 might be a bit much of a good thing. I tried Pony training with a fresh profile and with +5 boost it only took three training sessions to get the base skill up level 1 and then be able to take the exam. Players may still want to train to a higher level though to have a good change of passing the final stage of the exam.

Ben987:
Yes, +5 is reasonable.  But we don't want to break the logic that git4nick built.  Could the boost only affect the "Bondage" and "Evasion" skills and nothing else?  Maybe show the skills number in a different color if they are altered.

wildsj:
I try making it only work for those two.  I guess a stimulant like an energy drink boosting willpower seems a little odd.

Ben987:
Sure, do it for "Evasion" and "Bondage" for now and we can expand for other skills later.  Let me know when all of this is ready to test.  Thanks a lot!

wildsj:
I have implemented those changes.  It works but I'm not so sure about my approach to it now.  I wanted to add the amount of modifier on the end normal skills stat, but kept getting into errors.

I wonder now about adding a modifier level and duration to each skill so there is a Player.Skill[S].Modifier and Player.Skill[S].Duration. Then skills can have their own modifiers and it might improve the code.  However I think that can wait for now.

The modifier no longer affects Dressage or Willpower.

PR #486, opened 2019-11-07
remove from choice

gatetrek:
wasn't meant for owned players to wear more leaning to the people who are collarless and semi lovers


FeuertinKira:
Could have atleast replaced it with different one then, to fill the 8th slot.....

gatetrek:
nope cause prefer to ask the person if was ok to use there's

PR #487, opened 2019-11-07
Russian Translations by MAKeevchanin

adokilume:
Discord user MAKeevchanin has translated the first few files into Russian, and since they're new to GitHub, they asked me to submit them here.

**Edit:** By now, it's become a lot more than just the first few files, as you can see.^^


Ben987:
Wow!  This is awesome.  Thanks a lot you two.

I won't be able to test it much but I'm sure that Russian players will appreciate.

PR #488, opened 2019-11-07
New Bondage Bench

FeuertinKira:
-Adds a new bondage bench characters can be restrained to

Functions relatively uniquely compared to other items, characters are placed on the bench and then seperately restrained with leather belts.

-Added new slot "ItemAddon" for item attachments or addons to allow for items to work more dynamically, basically allowing an item to change it's state and also be displayed behind aswell as in front of the character --- This should be useful for other future items too. Example: Could be used to make things like the cages or boxes have an actual background image, and then a foreground layer that could act as the lid of the box or the door of the cage, both visually and functionality wise.
People can come up with even more creative items this way

This was pretty much a joint effort between both EveFerox and me


Ben987:
That's impressive!  I will test it soon.  Thanks a lot you two.

PR #489, opened 2019-11-08
Short back hair

wildsj:
Visible back hair for short and just touching shoulder length.


Ben987:
Awesome!  Thanks a lot.

PR #490, opened 2019-11-10
little change

gatetrek:
new collar added.
add the way for people to switch their collars at the club management but tree are missing from the owner from doing it maybe a next button to switch like the spanking toys could be useful there


FeuertinKira:
Odd, what happened to "prefer to ask the person if was ok to use there's"?

gatetrek:
if old stuff of mine never cared unless something just added what like be for lovers out there without a collar placed on them by there mistress.
and thought do it to you like semi payback I guess could call it >w<

FeuertinKira:
Actually, it's called hypocrisy.

Ben987:
Very nice!  Doesn't having 937 for MouseX position interfere with the clickable character zones?  I'll have to test it.  Thanks a lot.  Cheers.

PR #496, opened 2019-11-23
Beep Sound

Atasly:
Plays a sound when a beep is received. 
Note : will only play a beep sound if there is no other beep sound playing
needs testing on mobile version


Ben987:
Very nice.  Could you put some kind of parameter with it?  

Not everyone will want sound in the game so it would be nice to be able to activate or deactivate it.

Atasly:
I've begun working on an another feature that would need to save a setting so I'll look into that.

Ben987:
Excellent, having a parameter for it will be a lot more convenient.  We could also use that parameter later when we have moans, whip sounds, etc.  Thanks.

Atasly:
Created a a new set of variables (inspired by adokilume's work) for futures settings on first screen of preferences. Audio is now togglable and turned off by default.
still needs testing on mobile

Ben987:
That looks great!  I will test it soon and let you know if everything works on my side.  Cheers!

PR #497, opened 2019-11-23
Added a height setting

Atasly:
Adds a height setting for each player and will display so in rooms, chatrooms, dialog and wardrobe
Currently allows to go from 0.9 to 1.0 of maximum size.
Possible to add more place where characters are resized through the call of the DrawCharacter function
untested on mobile version

known issues : 
-No translation support
-Possible to add more lower height settings but might break the suspension rope

### DIFF
--- a/BondageClub/Assets/Female3DCG/Female3DCG.js
+++ b/BondageClub/Assets/Female3DCG/Female3DCG.js
@@ -464,9 +464,7 @@ var AssetFemale3DCG = [
 			{ Name: "0.975", Visible: false },
 			{ Name: "0.95", Visible: false },
 			{ Name: "0.925", Visible: false },
-			{ Name: "0.9", Visible: false },
-			{ Name: "0.875", Visible: false },
-			{ Name: "0.85", Visible: false }
+			{ Name: "0.9", Visible: false }
 		]
 	},
 
--- a/BondageClub/Assets/Female3DCG/Female3DCG.js
+++ b/BondageClub/Assets/Female3DCG/Female3DCG.js
@@ -455,7 +455,6 @@ var AssetFemale3DCG = [
 		Asset: ["HairFront1", "HairFront2", "HairFront3", "HairFront4", "HairFront5", "HairFront6", "HairFront7", "HairFront8", "HairFront9", "HairFront10", "HairFront11", "HairFront12", "HairFront13"]
 	},
 
-	//todo add translation and use better names than "0.9"
 	{
 		Group: "Height",
 		AllowNone:false,
--- a/BondageClub/Assets/Female3DCG/Female3DCG.js
+++ b/BondageClub/Assets/Female3DCG/Female3DCG.js
@@ -456,6 +456,19 @@ var AssetFemale3DCG = [
 	},
 
 	{
+		Group: "Height",
+		AllowNone:false,
+		AllowColorize:false,
+		Asset: [
+			{Name: "0.95", Visible: false},
+			{Name: "0.975", Visible: false},
+			{Name: "1.0", Visible:false},
+			{Name: "0.9", Visible: false},
+			{Name: "0.925", Visible: false}
+		]
+	},
+
+	{
 		Group: "Eyes",
 		Priority: 8,
 		AllowNone: false,
--- a/BondageClub/Assets/Female3DCG/Female3DCG.js
+++ b/BondageClub/Assets/Female3DCG/Female3DCG.js
@@ -455,6 +455,7 @@ var AssetFemale3DCG = [
 		Asset: ["HairFront1", "HairFront2", "HairFront3", "HairFront4", "HairFront5", "HairFront6", "HairFront7", "HairFront8", "HairFront9", "HairFront10", "HairFront11", "HairFront12", "HairFront13"]
 	},
 
+	//todo add translation and use better names than "0.9"
 	{
 		Group: "Height",
 		AllowNone:false,
--- a/BondageClub/Screens/Character/Wardrobe/Wardrobe.js
+++ b/BondageClub/Screens/Character/Wardrobe/Wardrobe.js
@@ -71,11 +71,11 @@ function WardrobeRun() {
 	DrawText(TextGet("SelectAppareance"), 1375, 60, "White", "Gray");
 	for (var C = 0; C < 12; C++)
 		if (C < 6) {
-			DrawCharacter(WardrobeCharacter[C], 500 + C * 250, 100, 0.45);
+			DrawCharacter(WardrobeCharacter[C], 500 + C * 250, 100, 0.45, true);
 			if (WardrobeSelection == C) DrawEmptyRect(500 + C * 250, 105, 225, 440, "Cyan");
 		}
 		else {
-			DrawCharacter(WardrobeCharacter[C], 500 + (C - 6) * 250, 550, 0.45);
+			DrawCharacter(WardrobeCharacter[C], 500 + (C - 6) * 250, 550, 0.45, true);
 			if (WardrobeSelection == C) DrawEmptyRect(500 + (C - 6) * 250, 555, 225, 440, "Cyan");
 		}
 }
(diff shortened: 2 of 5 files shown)


Ben987:
Very nice, I've been wanting to add a feature like that for a while.  

Why do you add a parameter in DrawCharacter to allow the height change or not?  Shouldn't it always be allowed?

Ben987:
Also, why do you sort it .95 .975 1 .9 .925 ?   You could sort it from tallest to smallest or smallest to tallest, it would be easier for players.

Have you tried going all the way to .8 ?  Would it be too small?

Cheers!

Atasly:
well there are instances where we don't want to zoom (the face in the wardrobe menu for example) hence the argument for the function. I'll correct it so that it's enabled by default and only disabled for exceptions then.

I wanted 0.95 to be the default size (medium). but Heights are sorted correctly. you choose your height in the wardrobe menu so it cycles through the list (it just doesn't look good in the code).

as for the min size, It looked kinda strange to me to go less than 0.9 but that's easy to add (will have to fix the suspension ropes to work properly with values under 0.9 so may take some time), here is a preview of 0.8 : 
Capture
Thanks for the feedback !

Ben987:
0.8 doesn't look too bad, I'll have to test it myself, maybe 0.85 should be the smallest, we'll see.  

Yes, try another version where DrawCharacter doesn't have that extra parameter.  The size should affect all versions of the character, even the zoomed head in the dressing room I guess.

FeuertinKira:
No offense, but tbh 0.8 looks way too small. I think 0.9 is about where the limit should be.

Ben987:
@FeuertinKira, Yep, maybe 0.9 will be the way to go.  I'll have to test it in game.

Shiranui-Izayoi:
Personally while not a contributor I think that 0.8 is fine and should at least be an option as long as it doesn't cause in game issues.

adokilume:
@Shiranui-Izayoi There could be an issue, not about the game mechanically, but legally. It's only a matter of time before people start thinking of characters that small as children, and/or roleplaying as such.

Shiranui-Izayoi:
@adokilume I guess that could be true but at the same time we have no idea how tall the characters are in game. So assuming that the characters are the average height for women in the US aka 5'4" or 162.6cm then at 0.8 scale they would be 4'3" or 130.08cm.

Personally I don't think that it is that bad but up to the devs and you guys I guess, just my opinion on the matter

Ben987:
Let's say 0.9 becomes the next average height, we could label it as 160 cm in game if it's the average women height.  Then we could have a rough range from 0.8 to 1 like this:

140 cm (0.8)
145 cm (0.825)
150 cm (0.85)
155 cm (0.875)
160 cm (0.9)
165 cm (0.925)
170 cm (0.95)
175 cm (0.975)
180 cm (1)

Do you think we could get in trouble for having characters at 140 cm?  That seems pretty reasonable to me.

Atasly:
Ordered heights from 1.0 to 0.9
DrawCharacter function still needs a bool argument as the height on the zoomed face or in the maid cleaning mini game looks off, but it's true by default now. Have tested as many occurences as possible but might have let some slipped away.

As for the discussion regarding 0.8, for now it simply doesn't work with suspension ropes, so I haven't included it for now. I'll keep working to implement it (just cause it bugs me why it doesn't work) but that will definitely take more time.

Shiranui-Izayoi:
I think 140cm is fine since its 4'7" which is completely fine and yer if there is issues with the items then I think we should keep it out for now again just a fan of the game giving their opinion :)

FeuertinKira:
The problem isn't converting the values between 0.8 and 1.0 to real values. In my opinion the problem is the visual difference when you compare 1.0 to 0.8.... The chracter with 0.8 doesnt even reach up to breast height of the 1.0 character. Even if there can be short adult people occasionally IRL, I shudder at the though of having a horde of 0.8 characters running around in the game. Also height difference can become even more extreme if the 1.0 character puts on mistress boots or the like.

As adokilume mentioned it's a risky topic. Personally I find it creepy to say the least. I'm all for height differences, but mostly as a subtle thing to distinguish characters from one another

Ben987:
@Atasly, let me know when everything is ready and tested on your side.  I'll check to see how to label the height tags and the minimal value.  Thanks!

Shiranui-Izayoi:
> 
> 
> Let's say 0.9 becomes the next average height, we could label it as 160 cm in game if it's the average women height. Then we could have a rough range from 0.8 to 1 like this:
> 
> 140 cm (0.8)
> 145 cm (0.825)
> 150 cm (0.85)
> 155 cm (0.875)
> 160 cm (0.9)
> 165 cm (0.925)
> 170 cm (0.95)
> 175 cm (0.975)
> 180 cm (1)
> 
> Do you think we could get in trouble for having characters at 140 cm? That seems pretty reasonable to me.

@Ben987 I just wanted to say building on this comment you made which I have quoted above if 0.9 is made the default height assuming that 0.8 can be made to work then it would likely fix the looking too small issue. 

Since the people with 1.0 scale would be characters who ingame are taller than normal and thus it would make sense if the 0.8s aka smaller than normal look very small compared to them, like comparing a basket ball player to a average person.

At least that is my take on it.

FeuertinKira:
That still leaves the issue of 0.8 characters barely reaching up to chest height of a 1.0 character... regardless of what the 'default' is...

Ben987:
It's hard to assume the best minimum without trying it in the game.  When Atasly is ready, we'll tests the possible configurations and set the best minimum.

Shiranui-Izayoi:
@FeuertinKira Well that is like saying that a normal person barely reaching chest height on a basketball player is an issue. If one person is taller than normal and another is shorter than normal it is basically guaranteed. 

To put it in perspective look at this https://www.mrinitialman.com/OddsEnds/Sizes/compsizes.xhtml?Figure 1~female~180_Figure 2~female~160_Figure 3~female~140 

Assuming that 0.8 is possible which if it isn't and causes problems it should be removed no questions asked the 0.8 scale comes up to neck height on the new default height and the default comes up to 1.0. I think you are focusing a bit to much on the comparison of the between the extremes instead of the new potential middle to the extremes.

FeuertinKira:
There is no point in comparing the 'default' to the extremes... You have to compare the extremes since those are the ones that will look the weirdest. And I guarantee you the extremes will be a lot more commonly used in this game then you would encounter such extremes IRL. In fact I think you comparing 1.0 to a basket ball player makes no sense, considering it was attributed to be 180cm before...

Shiranui-Izayoi:
While yes the basket ball player example isn't the most apt comparion it still works as a example of how taller than average people make even people of average height look small by comparison which is only increased by smaller than average.

I really don't get why you find it so strange and are against it? Everything is within normal ranges some people are tall others are short some places have lots of tall and short people with few inbetween. 

I do understand that there can be legal problems if the characters look to childlike, but I don't think as it is proposed atm would cause any resonable person to think that they are.

Atasly:
I kinda agree with Kira here, people will tend to choose extreme options rather than more usual ones, and 1.0 vs 0.8 looks weird, maybe 0.85 is okay.
For testing you can check it on your own, just add whatever size you want in female3DCG.js around lines 460 but the suspension rope won't display correctly on heights lower than 0.9

I'll get back to debugging for now (didn't think it'd be such a controversial pull request >< )

Atasly:
ok it should be good to go
you'll find a preview of 1.0 vs 0.85 in this comment
please let me know if anything's wrong
Capture

adokilume:
Yeah, I don't think that extreme height differences are a good idea. No matter how much you'll try to justify it with actual heights and their realism for adults, people are gonna see this as encouragement to roleplay as kids.

Shiranui-Izayoi:
That is true but it is their choice and it can't really be stopped without extreme measures and we already have the nursery so. 

Not to mention their is a big difference between RPing as a kid aka ageplay and actually saying your character is underage, and if people do do that then yer they are wrong and should be punished for it.

Still in the idea of better customization I still think it is a good idea.

FeuertinKira:
Yeah trust me, the limit should between 0.9 and 1.0 at most... With the mistress boots in that picture aswell... Not a fan at all... That is still a ridiculous size difference. It's not just the height, the width shrinks just as much... It's pretty awful.

adokilume:
Of course there's a difference between ageplay and actually saying your character is a child. Thing is, ageplay and the nursery don't change how tall you are, which makes it clear that you're an adult pretending to be a child. With smaller sizes, it becomes more inviting to say your actual character is a child, which is simply not okay.

Shiranui-Izayoi:
I still don't see much of an issue but I am just a player not a dev/contributor so meh I have offered my thoughts by now so I won't comment further.

Atasly:
fixed a small bug unnoticed and decided to limit heights at 0.9. Still easy to change that for lower values.

Ben987:
Great!  I will test all of this right now, thanks.

PR #498, opened 2019-11-24
Yoke pose

wildsj:
Adds a yoke pose and yoke that holds the wrists up and out at neck height.

Contains yoked versions of all affected cloth and item assets except for: T-Shirt, Sweater, Nomadic dress and Fur gloves, which I did not found matching mods for.

The original pose was created by Nythaleath

It will need updating for assets created since work started, namely the Leather and padded leather mittens.

Note to other contributors.  I have added a new pose versions of some assets added by other people to fit. Such as lengthening some bra straps and a made a version the metal mittens raised up based on the original, etc.  I haven't changed the existing assets, except for adjusting some of the asset image areas.  See what you think and feel free change the new poses of assets you added.


Ben987:
Oh... my... god!  This is huge, thanks a lot, I will test it later today.

Ben987:
For missing items, could you put fully transparent PNG files so there's no errors triggered in the console?  Thanks.

PR #501, opened 2019-11-30
Rework of All Item Priorities

FeuertinKira:
-Reworked all Item priorities; Idea is to give a bit more breathing room between different categories. So there is less overlap between certain edge cases sharing the same priorities and a more clear order between them.

Could use some more feedback from other people, there is a lot of potential item combos. I spend two hours testing as many as I could think of, but there's a chance I didnt get every little edge case. In which case let me know and I'll make adjustments accordingly.


Ben987:
This is big!  Thanks a lot.  Players will probably find combos that doesn't fit but we will correct them if it's the case.  Cheers.

PR #502, opened 2019-12-01
Improvements for sensory deprivation experience

Atasly:
Added some improvements to the sensory deprivation experience, they are listed below.
As it is not to everyone's taste, there are two separate toggleable settings in the preferences menu :
- - garble the names of senders
   - hide enter and leaves messages.
   - replace the name of the sender in emote by "someone".
   - does not apply for whispers or player's messages/emotes. Applies only when heavy blindfolded and with heavy ear plugs on.
- disable examining other players by clicking on them when heavy blindfolded


FeuertinKira:
I appreciate what you're doing with this.... And I understand a toggle for options like this is good...
But at the same time I feel like it's kinda weird to have inconsistent item mechanics like this, that change potentially quite drastically how someone is interacting (or not interacting) with other people, especially without any direct way for other people to even know if that person in the hood over there has it toggled on or off. 
Just seems odd to see the same hood on two people's heads, but one of them is more sensory deprived than the other... weird.

I think it would be even worse than the issue earplugs already have right now. Since earplugs have no visuals, people will often not notice that someone is deafened. Only after a while of usually not getting a response, or not the response they were expecting, or the deaf person just telling them, will they notice they have ear plugs in. I think this feature will be worse than that, since there is absolutely no way for someone to know if the hooded person chose to toggle it on or off without specifically telling people.

tl;dr about the toggle: having the choice to opt in or opt out is great, but at the same time it will make gameplay or RP a lot more bizzare, if there is no consistency between the affected items and different people using them

Atasly:
I understand your point, inconsistency of gameplay (information here) can be troublesome for RP, but I don't think it will affect too many people. (note : only when heavy blindfold + heavy ear plugs).
Anyway, one way to solve that problem would be to show on characterscreen which settings the user has, maybe with small icons (sound muted for example) so that it would be easy to add more icons for more potential settings.
As for the ear plugs I don't think they will ever be visible so we'll have to get used to check that item slot I guess.

PR #503, opened 2019-12-05
Owner rule : remotes

Atasly:
similar to the owner rule for keys but for vibrators' remote
owners can block their subs from buying remotes, they can confiscate their remotes (meaning they'll lose every remotes they currently have) and they can allow them to buy them.


Ben987:
I love it!  More control for owners :)

PR #505, opened 2019-12-06
New Collar Name Tags & Attachments, some other item upgrades & fixes

FeuertinKira:
-Added a lot of new name tags with new shapes

-Changed pet themed name tags around and moved them to  their own name tags

-Added Chloroform Cloth

-Added Latex Posture Collar to also be available as gag

-Added Leather Corset Collar to also be available as gag (again)

-Adjusted some items to hide LeatherCuffs

-Added ability to ButtPlugLock and AnalHook for attaching chains in various ways

-Changed 'Top' value for ItemButt and adjusted every item in it accordingly to display the chain attachments correctly

-Adjusted BitchSuit to hide more clipping items

-Adjusted StraitDress to hide more clipping items

-Adjusted StraitDressOpen to hide more clipping items

-Adjusted Yoke Priority slightly

-Fixed long-standing inconsistency of the PlugGag; changing from GagNormal to GagTotal and back to GagHeavy --- Now goes from GagNormal to GagTotal back to GagNormal

-Adjusted certain items to hide OrnateCuffs

-Added the new Collars I made to the SlaveCollar

-Added six new collar attachments; Sun, Lapis, Cross, Pentagram, Flower and Rose


Ben987:
Wow!  This is huge!  Thanks a lot, I'm sure that players will love it.

Dwscdv3:
There's a minor bug in the neck corset gag.

Applying a gag effect from neck items may cause a problem since all rescuers are currently ignoring this slot. And thus can't talk to the (imbalanced) photographer, too.

~~On the other hand, it's the surprisingly ultimate way of being at another one's mercy if using a lock. That will be interesting, but still a bug.~~

PR #507, opened 2019-12-08
2 new winter scenes

Nythaleath:
added 2 new backgrounds that are outside so you can now take your subs on a humiliation walk outside, assets heavily aided by kirsty


Ben987:
Very nice!

PR #509, opened 2019-12-08
Check at login for stable items

Atasly:
-Will check at login that the player has indeed passed the exam and adds stable items in her inventory (bit gag and pony boots), also add lockable pony boots in her inventory if the case.
-Won't remove restraints currently used, but players will still have them deleted from their inventory.

disclaimer : locking pony boots and clothing pony boots can seem a bit redundant for the moment, but it won't be when there will exist some itemboots not hiding shoes.


Ben987:
That's a good idea.  Thanks a lot.

PR #511, opened 2019-12-08
Fix breaking whispers when using HTML characters

adokilume:
It has been brought to my attention that when using whispers in the chat, certain character sequences (like a broken heart, `</3`), would mess up the chat's HTML structure. On the senders end, all following messages would appear like whispers and be further indented, even if they exit whisper mode. These lines should fix that, although I wasn't sure if they should only be in effect when whispering, or always.


Ben987:
Thanks!  I will add the fix to the beta.  Cheers!

PR #512, opened 2019-12-08
Tiny fix for Bondage Bench

FeuertinKira:
-tiny fix to make sure the bondage bench actually works, since Priority: 50 is currently not recognized


Ben987:
Yep, priorities goes from 0 to 49 right now.  Thanks for the fix.

PR #513, opened 2019-12-08
Tiny fix for Blush

FeuertinKira:
-Blush ended up behind Upper Body and so nobody can ever see it, should fix the issue


Ben987:
Yep, thanks for correcting it, I was about to do it myself.

PR #514, opened 2019-12-09
Added missing Parameter to new ItemTorso items & misaligned ItemPelvis Harness

FeuertinKira:
-Added AllowLock: true to ItemTorso items, so they can actually be locked like they were supposed to be

-Fix for misaligned Harness on ItemPelvis when wearing handcuffs

-Added two missing images in "ItemVulvaPiercings\Kneel" and one in "ItemHead"


Ben987:
Awesome!

PR #515, opened 2019-12-09
add the thingy

gatetrek:
added the new collars to the management pick


Ben987:
Perfect!  Thanks a lot.

PR #516, opened 2019-12-10
Fix ButtPlugLock Removal

FeuertinKira:
-Added entry so maids can properly remove ButtPlugLock, should fix the issue of getting stuck with a chain and chastity on top.


Ben987:
Absolutely!  Thanks :)

PR #518, opened 2019-12-10
Strip seperate clothing in appearance menu +fix

Atasly:
In appearance menu, adds a button for each item slot which allow for quick undress of specific slot.
fix : locking pony boots are now added after a pony/trainer exam, and not only on login (untested)


Ben987:
Good idea, I've been wanting to do something like that for a while.  Thanks!

PR #519, opened 2019-12-11
Update Female3DCG.js

wildsj:
fix to make duster gag appear over the yoke.  Checked to make sure duser gag doesn't interfer with fronthair or hoods.

There is another clipping issue with the duster gag and LatexHoodOpenHair, but there does not seem to be a simple fix for that.


Ben987:
Thanks!

PR #521, opened 2019-12-11
Seperate collar restraints from accessories

wildsj:
New group for collar restraints.
Moved the leash and chains out of the accessories group.
Updated the maid's zones speeches in the introduction room.
Reduced the height of breast zones a fraction to make the collar zones a little bigger.


Ben987:
That's a cool idea.  Have you tested it a lot?  I'm getting closer to release date so I'm getting nervous with big changes.

wildsj:
Yes, I have, Apart from being in separate groups and the accessories no longer considered restraints

PR #522, opened 2019-12-12
3rd collar group

wildsj:
Tested a lot more, found and fixed a bug with collar restraints so they are now removed too if the collar is removed. Commiting that last fix though, having already fixed a conflict through github.com, caused the github desktop to terminate the last PR.

Only minor issue I can find now is if a collar accessory is owner locked to a slave collar and the owner won't remove it (or isn't around to) then the only way to remove it is to terminate the ownership and have the slave collar removed. In that circumstance though, ending the ownership contract might be the best cause of action for the player. Either way the player will not be left restrained by the accessories.


FeuertinKira:
We've been over this. The talk happened just before you brought seperating the chains and leash up. You can just add "IsRestraint: true" to the items and the maids will remove them on rescue. Then there wont be any issue with getting stuck with them.

wildsj:
It is just the non binding accessories, such as the tags and shock collar that won't be removed by NPCs such as the maids or photographer now. Part of the reasoning behind this change was so that players will keep there tags while doing jobs like serving drinks when the maid removes all the "IsRestraint: true" items on completion.

Ben987:
Excellent!  If everything is tested on your side I'll test it later today.  Thanks again!

PR #523, opened 2019-12-12
Santa hat

Ulith:
Split into Fabric and fur, and edited slightly to recolour in the same way the other furred items do. (dress,gloves,socks)


Ben987:
Nice!  Thanks a lot.

PR #524, opened 2019-12-14
Fix for Teddies

Ulith:
Minor fix to stop teddy bear set plushies from floating in front of the chest when Yoked


Ben987:
Great idea!  Thanks.

PR #526, opened 2019-12-15
reimage

gatetrek:
abit more res image for the xl bone gag 
asked nyth first before uploaded this and was ok


Ben987:
Thanks a lot!  I will put it in the full release I'm preparing right now.

PR #529, opened 2019-12-16
Tiny spelling error fix and ItemNeckRestraints entry

FeuertinKira:
-fixes a  spelling error on name tag lock image

-adds missing entry for ItemNeckRestraints

-swapped priorities for ItemNeckAccessories and ItemNeckRestraints, so the chains and leash hook onto the collar below the accessories and stops them from obfuscating name tags etc.


Ben987:
Good catch!  Thanks.

PR #530, opened 2019-12-16
Itemneckrestraints fix

wildsj:
Added line so leashes are now removed from "collar" instead of "itemneckrestraints"


Ben987:
Good catch.  Thanks!

PR #531, opened 2019-12-17
harness yoked fix

gatetrek:


Ben987:
Thanks for the quick fix, but I did the exact same fix in my batch of fixes, so it's not needed anymore.  Cheers!

PR #532, opened 2019-12-21
Expression

gatetrek:
someone asked me if could add drool in the Facial Expression area


FeuertinKira:
That's unfortunate... I'm currently working on implementing new eye expressions, tears, and reworking mouths entirely. Is it really neccessary to waste the precious limited space of the current expressions menu on two simplistic thin lines of drool? I think it would make more sense if they were part of the mouth expression catgeory instead of it's own.

Ben987:
That's a good idea, I'm sure lots of player will like it.  @FeuertinKira , when your stuff is ready, we'll figure out a way to make it all work together.  Thanks!

FeuertinKira:
The problem is, I already have drool implemented in my expressions too (Amongst all the other stuff).
I implemented drool together with tears as "Fluids" in it. Plus I redid mouths, so these drooling images wont line up with that anymore. Only thing here that can be merged with my expressions overhaul is the drooling on struggling code using drooling images I added.

Wish gatetrek would bother communcating with me, we could have probably figured something out beforehand, but doesn't seem like there was any interest beside having a 3rd party deliver an insult toward me. :/

Ben987:
Is there a way for you to include the drooling we have now?  It's pretty basic.

If not, let me know when your version is ready and we'll compare what you have vs what we have now.  We could even do a special channel on Discord so you, me and gatetrek can talk about it, and get the best final solution.

FeuertinKira:
To be honest, the drooling images I made are more detailed than these ones. I don't see much reason to keep these basic ones, if I have more detailed ones. I don't want gatetrek's work to be wasted, but some of it will be obsolete with my overhaul.

gatetrek:
3rd party deliver an insult toward me.?
what you on about? and why I haven't bothered talking with you was due 
" Is it really neccessary to waste the precious limited space of the current expressions menu on two simplistic thin lines of drool?"
when I read that I just couldn't be bothered with you, but not seeing where this insult coming from?

FeuertinKira:
The insult came from someone on Discord, 'general' channel. You can check there yourself if you want. Probably shouldn't lay it all out here again. Github ain't the place for that.

And well I apologize, I formulated that part of my reply probably too harshly. But part of it is still important, there is only space for about ~7 rows in the expression menu. Using one of those rows just for drool seems like a waste. In my overhaul I now combined tears and drool (which I first had as just a mouth expression) into one row together.
I have Blush, Eyebrows, Eyes, Fluids (tears/drool), Mouth, Emoticons right now. And I was playing around with poses (arms behind back etc.) as the 7th row of expressions. If the expressions menu can be compacted a little more, maybe even an 8th row for body markings (such as rope marks, or whip marks). So with those plans in mind, the space on the menu was quite precious.

gatetrek:
well I don't read general so I'll have a look but I know for sure haven't asked anyone to insult anyone, I can easily do that myself when I want.
the drool, not 100% bothered about it as was something asked of me if can do for them and that's all. I didn't put that much into it as wanted to see if people wanted it and then would've moved onto making it look better.

Ben987:
Just let us know when your version is ready and we'll keep the best drool image.  If we need more expressions, we can squeeze for a 8th place or do something to scroll.  Cheers!

PR #533, opened 2019-12-23
Mistress/Owner Timed locks

Atasly:
Added two new timed locks with lots of options. One for mistresses, and one for owners, they both use their same respective key.
options :
-auto remove the restraints when time is up, only accessible to lock number member
-hide the time shown in UI/being shown in chat, only accessible to lock number member
-allow everyone access to the add more time feature, only accessible to lock number member
-add more time whenever you want, capped to 4h for mistress locks and one week for the owner one, only accessible to mistresses/owners
-unlock the restraints with your key whenever you want, only accessible to mistresses/owners

Thanks to FeuertinKira for the pictures


Ben987:
That's a great idea.  But how does the Club Mistress acquire the item?  Is there some piece to give it to her automatically or when she becomes a Mistress?  Cheers!

Atasly:
any club mistresses gets it automatically when logging in/becoming a club mistress (in LoginMistressItems function)

Ben987:
Excellent!  I will test it later, thanks a lot.

PR #535, opened 2019-12-28
Ornate Collar added

gatetrek:
I got a request from Nythaleath if I could add the Ornate Collar to the slave mix for them.


Ben987:
Sure!  That's a good idea.

PR #14, opened 2017-10-06
LunchBreak with Natalie

wildsj:
I think this has got all the files added to make Natalie work. I'm still getting my head around how GitHub works.

PR #536, opened 2019-12-29
Translation of actions and sensdep rework

Atasly:
Translation :
Actions are translated per tags linked to them in ChatRoom.js
a little overview of some terms I decided to use for sending/retrieving information :
- Tag : the term that will be replaced, some of them have specific meaning
  - SourceCharacter : the sender of the message/action
  - DestinationCharacter : the destination of the message/action will automatically translates it to "her" or add a "'s" depending on the Sender
  - DestinationCharacterName : the destination of the message/action. will automatically add a "'s" depending on the Sender
  - TargetCharacter : the destination of the message/action. will automatically translates it to "herself" depending on the Sender
  - TargetCharacterName : the destination of the message/action.
- MemberNumber : in order to compare who the message is coming from (we don't want to send whole Characters objects for each message/actions)
- TextToLookUp : will look up in Dialog_Player for the string attached to it.
- AssetName : will look up AssetName's description.
- AssetGroupName : will look up AssetGroupName's description.
- Text : the text the tag will be replaced with if not a special case

I also kept the translation files up to date with that new system, hence the numerous changes, but due to the high quantity of text (and the fact that I doesn't speak every 4 languages) it's possible that errors slipped by.

Sensdep :
- Normal sensdep setting now garble every name in chat
- More extreme sensdep setting : no color, no membernumber

Probably some other fixes I forgot about. 

Moved two settings from that pull request into the [#537](https://github.com/Ben987/Bondage-College/pull/537) one
Need Server Pull Request [#14](https://github.com/Ben987/Bondage-Club-Server/pull/14) to work


Ben987:
This is huge!  Thanks a lot.  Can you split these changes so I can review and test them one by one?  It becomes very tricky when there are lots of changes at the same time and some of them have bugs and others don't.  Thanks.

Atasly:
well the new sensdep settings require the dictionary, and it's not really possible to split the dictionary pull request.
the two new settings (volume audio and toggle auto-remove at login) should work correctly. could make a separate pull request for them, but since I modified Preference.js I felt like adding them to it at the same time, is that what you want me to split ?

Ben987:
The smaller you can split it, the easier it is for me to test it and undo it if there are bugs that are too major.   Try to do as many smaller pull requests as you can and I'll see what I can do.  Cheers!

Atasly:
split and updated description. I'd recommend accepting the commit fecdcfe if my pull request #537 is accepted as well as this one, since it's merged with it already.

Ben987:
That change makes me nervous, it impacts lots and lots of items for a situation that isn't a huge problem.  Have you tested all of these items to make sure they all still work?  Also, why are you removing the " 's " everywhere?  Are you adding them later?  Will that work well with other languages?

Thanks a lot.

Atasly:
no I haven't tested all the items, they are many items after all and many settings for each of them, but if a "SourceCharacter" or something appears in chat it's really easy to track down the issue.

the `'s` is now added in a consistent way (if DestinationCharacter is different than SourceCharacter then we add `'s` after DestinationCharacter's Name else it's `her`) and how it should affect other languages is simple : `'s` is being replaced by whatever it's translated to. if its translation equals nothing then great, it simply removes it.
I tried to convert the translations to this new system to the best of my ability but I don't speak all four languages so can't guarantee that either.

Ben987:
Are you able to do a quick tour of all the affected items to make sure they still work?  Some big change requests like that had unsuspected impacts before.   Once you've tested it, I'll merge and test it on my side also.  Yes, we'll ask Adokilume and MAK to test the translation results.  Thanks a lot.

Atasly:
tested all items and their action messages and found a few forgotten items, now fixed, also fixed the love belt's butt lock message infamous error.

Ben987:
So much stuff, you're a machine!  Thanks a lot, I will test it soon.  And thanks for your extra testing.

Atasly:
please keep in mind that the pull request https://github.com/Ben987/Bondage-Club-Server/pull/14 must be accepted as well in order to see messages correctly.
(which might cause some issue for the beta as everyone shares the same server. and there will be people, either the beta testers or the one without it, who are going to see messages with "SourceCharacter" or "ActionUse")

PR #537, opened 2019-12-31
Two new settings

Atasly:
- Volume audio control, goes from 0% to 100% and increments by 10%
- Toggle to keep all restraints that would usually be removed at login because some people like to keep them.


Ben987:
Excellent, I will test it soon.  Thanks a lot.  

Let's hope this will not trigger too many questions such as: "I'm stuck in the cage when I log back in."  The more options we add, the more confused players get.

Cheers!

PR #539, opened 2020-01-01
New items

BunnyMadMan:
A group of new items for the game~

Under Tails and ItemButt
2 Fox Tails/Tailstraps
2 Kitty Tails/Tailstraps
3 Wolf Tails/Tailstraps

Under Ears
2 Fox Ears
2 Kitty Ears
2 Wolf Ears
1 Puppy Ears

Under ItemMouth
Foxy Harness Panel Gag

Under ItemHead
Foxy Mask - Small warnings; after many tests, the mask is currently appearing underneath the hair; this is intentional, since it matches the best from all the tests. On top of that, the Allow: Lock effect is added in the code, but since it would look weird, the lock image wasnt added for this particular Item. You can change this at a later time, but in my personal opinion, the way it is now is the best.


Ben987:
This looks great!  Thanks a lot.  I will try it right now.

PR #540, opened 2020-01-01
Fix tailstraps

BunnyMadMan:


Ben987:
That was fast!  Thanks :)

PR #544, opened 2020-01-05
Rework timer locks

Atasly:
reworked public timer locks after some feedbacks/discussions :
- less potential bugs
- actually added mistress timer padlock when the player becomes one
- ability to remove time from timer locks (for mistresses/owners (depending on the lock))
- if the option to enable other people adding time is selected (now renamed EnableRandomInput), those are the options available once for everyone : add time, remove time and random. the club number of the player is then saved and the player won't be able to add/remove/random any more time, except of course if she is a mistress/owner (depending on the lock)

known bug : changes to locks aren't updated directly per client which could result in someone adding 10min and someone else adding 10min at the same time but only one of the two changes (the last one) is saved.

I have some other plans to expand those locks but wanted to change how they work before the next patch.


Ben987:
Sure!  That's a great idea.  Thanks a lot.

PR #545, opened 2020-01-06
Fix for better compatibility between mouth expressions and gags/gag layers

FeuertinKira:
-Hid the mouth from gags where it made sense, so you dont end up with silly things like having two tongues or the tongue sticking out from underneath tape etc.


Ben987:
Great idea!  Thanks :)

PR #546, opened 2020-01-06
Fix for gag effects not being correctly applied

Atasly:
as the title says, when an item with multiple effects (hoods for example) only their first effect was checked
should work properly now


Ben987:
Good catch, the effect level should stack like that.  Thanks!

PR #547, opened 2020-01-07
Gags Bug Fix for Magic Theater

FeuertinKira:
-added ItemMouth2 and ItemMouth3 to the appearance swapping so that they also get swapped between characters if player equips them on npc to prevent scene from getting stuck


Ben987:
Good catch there.  Thanks.

PR #549, opened 2020-01-08
Bug Fixes

FeuertinKira:
-Removes muffling from chat room text for moving someone left or right

-Lets maids in maid quarters also remove ItemMouth2 and ItemMouth3


Ben987:
Good idea!  I'll prepare a "hot fix".  Thanks :)

PR #551, opened 2020-01-08
Show chatroom space in beeps and friends list

ace-1331:
This pull request goes with the pull request of the same name for the server side.

-When a user is in the asylum, it will be shown in the beeps and the friends list to avoid confusion when trying to find a room. Will work if more "spaces" are added in the future. This is ignored if the user is in a private room

I have tested it locally with different languages and with characters in different public/private rooms and found no bugs thus far.

Since this is my first PR, please do let me know if I did anything wrong so I can correct it asap.


Ben987:
That's a good idea.  I can't review it yet though, I'm still in "bug squashing mode" from the previous release.  Thanks!

PR #552, opened 2020-01-09
Some more fixes

FeuertinKira:
-Fixed some more instances reported by other players getting themselves stuck by starting certain activities with gags in ItemMouth2/ItemMouth3

-Added Cloth Gags from ItemMouth2 & ItemMouth3 to the introduction check


Ben987:
Nice fixes, bugs like that were sure to happen with such a huge change.  I'll prepare another hot fix.

PR #555, opened 2020-01-12
Preference option to hide tip messages

TheGnarp:
The new tip messages offer good advice and state clear rules and boundaries, but they might be a bit immersion breaking for some visitors of the club.

Just imagine being in a first-class "adult établissement" and constantly being reminded by the staff or by signs that you could be robbed by the ladies that sit on your lap, or catch diseases, that you must be over the age of 21 (for the umpteenth time although you are clearly over 40), or that you should be polite to the working ladies (even if you are the nicest person on the planet).

Also for users that find the tips useful – and they all are – seeing them every time you visit the main hall could be just a little too often for some guests.

### DIFF
--- a/BondageClub/Screens/Character/Preference/Preference.js
+++ b/BondageClub/Screens/Character/Preference/Preference.js
@@ -37,6 +37,11 @@ function PreferenceLoad() {
 		ForceFullHeight: false
 	};
 
+	// If the user never set the visual settings before, construct them to replicate the default behavior
+	if (!Player.VisualSettings) Player.VisualSettings = {
+		HideTipMessages: true
+	};
+
 	// If the user never set the audio settings before, construct them to replicate the default behavior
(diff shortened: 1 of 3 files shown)


Atasly:
why have you copied the block line 35-38 to 40-44 ? it's redundant
I'd write something like that : 
```javascript
	// If the user never set the visual settings before, construct them to replicate the default behavior
	if (!Player.VisualSettings || (typeof Player.VisualSettings.HideTipMessages !== "boolean")) Player.VisualSettings = {
		ForceFullHeight: false,
		HideTipMessages: false
	};
```
You can also move your line 71 from MainHall.js into your condition.

Ben987:
Lets force the warning messages for now, some of them are very important and might help us on the legal side if someone ever wants to take the game down for moral reasons.

TheGnarp:
I understand.
For me, it feels like putting warning stickers and advertisements on the Mona Lisa, but I see your point.

PR #556, opened 2020-01-12
Less imperative tip messages

TheGnarp:
_Please proof-read the changed tips, since I'm not a native English speaker._

My suggestions for tip messages from my experiences with UI/UX and psychology studies (warning and info signs in various fields like news, commercials, emergency warning).

The drawback is obviously the added third line for some messages, but this will probably happen for any other language than English anyway.

The benefit is a less "harsh" or "imperative" tone or mood.
Just think of a dinner in a fancy restaurant (that nothing compares to...) and the waiter stating obvious restrictions like "Do not use your cutlery as percussion instruments. It will annoy other guests!".

**Side note: Tip No. 0 and 1** are so important and obvious to the club that it shouldn't be a "tip" imho. It could be something a new user agrees to when creating an account or something the welcoming staff points out! That however, is probably a little beyond my experience with the code of the Club and something I would rather suggest than create a pull request for.


Ben987:
Some of the edits sound strange in English, I'll do a softer version of the messages myself another time.

They need to be kept short so they are easy and quick to read.

Thanks a lot!

PR #558, opened 2020-01-12
Fix: Added pony bit gag to the stable check login function (layer2/layer3)

ace-1331:
The pony bit gag was there for layer 2 and 3, but was never given to the players


Ben987:
Thanks a lot!  Good catch :)

PR #559, opened 2020-01-12
Message when changing clothes in chatrooms

Atasly:
Just a simple message sent when clothes are changed in a chatroom.
might spam a little the chatbox though.
also possible to add a C.MemberNumber as a target to the message sent to avoid that side effect, but I'm not a big fan of that solution.


Ben987:
Yes, this will reveal who are the "ninja strippers".  Thanks a lot!

PR #560, opened 2020-01-13
Couple prerequisites fixes

FeuertinKira:
-Fixes some odd combos with horse + bench or spreader bar + bench

-Fixes "NotChained" prerequisite  from not working


Ben987:
Great!  I'll check to apply the fix right now.

PR #563, opened 2020-01-14
Racoon Ears

Ayesha-678:
Made some Racoon ears for all those Racoon girls out there~


Ben987:
Sure.  We need to think about all the raccoon fans out there :P

PR #564, opened 2020-01-14
fix chatroom space 

ace-1331:
A little bug slipped through my attention in the previous accepted PR which I noticed as I was testing the change on the live server.

- Changed line 360 in Server.js and line 80 in Friendlist.js to fix the broken beep history log (back to what it was and to properly pass the received data in order to show it in the history)

Happily, everything else was working as expected on the live server, nothing was different then when I tested it locally so it should be good to go


Ben987:
Good catch.  Thanks a lot for that.

PR #565, opened 2020-01-14
Command for alternate emote

Atasly:
Added a command in chat to use an emote without the name of the sender in front of it for more diversity in RP and descriptions.
the command is `**` or `/action`
it will still be possible to distinguish different messages as emotes are in italic and actions are in parenthesis.


Ben987:
Sure!  But I hope it won't get too complicated for players to remember all these commands.  Maybe we can put a help button to list all the commands in-game someday.  Thanks!

Atasly:
good idea, might add that one when I have time

PR #567, opened 2020-01-15
Black Out Lenses

Ayesha-678:
Blackout Contact Lense Blindfold.
Possibly in the future they could be edited to allow blinking with them on.


Ben987:
That's a little creepy, but I'm sure some players will enjoy them.  And players can always block them from the permission screen.  Thanks!

Ayesha-678:
I agree they are a bit creepy, but a good friend of mine wanted to see them in, so I thought I’d do it for them ^^

PR #568, opened 2020-01-15
latex lockdown suit 

gatetrek:
mine and BufaloAcquatico contribution.
: added a latex lockdown suit with openable zippers and an attachable wand that may have other toys in future.
: fix to the one bar when equipped couldn't get off with the player or another character. 
: added new nipple to choose from.


Ben987:
This is impressive!  Thanks a lot you two.  I will test it soon.

PR #569, opened 2020-01-15
/help command

Atasly:
Added a `/help` command listing all commands available in chat. Not describing them since that would easily take half the screen (and they are all pretty self explanatory, except for non english speakers)
the list of commands will only be sent to the one who typed it.


Ben987:
Thanks!  I will also add a tip to the main hall so players will know that they can type "/help".  Cheers.

PR #572, opened 2020-01-16
dictionary fix : spanking toys

Atasly:
the previous fix for spanking toys wasn't a proper one (and was a bit rushed too), this one covers everything (hopefully)


Ben987:
Thanks a lot!

PR #573, opened 2020-01-16
add chatroom background preview

NoneNoname:


Ben987:
I like the concept but it touches lots of code, so I'll probably cherry pick it later.  Thanks!

PR #574, opened 2020-01-16
Suspension rotate instead of mirror (Name Tags)

NoneNoname:


Ben987:
Good catch!  Thanks a lot.

PR #577, opened 2020-01-17
Fixed various item prerequisites/effects

ace-1331:
- Made it so the short chain attachment on the buttplug lock has prerequisite to prevent things like being impaled, slammed or thrown in the air while being mounted, suspended or in related items that should not allow the kneeling state. (adds the possibility for more prerequisite or to also block the long chain if we want that at some point)

- Made it so the blocked items once on the saddle stand are similar to the one bar, this prevent all the problems involving your legs being torn apart by rope and cuffs or being drilled to the ceiling

- Made it so the display frame would act as a "mount" item to prevent things like being slammed to the ground by a short chain (collar or buttplug lock)


Ben987:
Good idea!  I'll try it soon.  Cheers :)

PR #578, opened 2020-01-17
ServerMove functionality

EveFerox:
requires https://github.com/Ben987/Bondage-Club-Server/pull/20


Ben987:
That's a good idea.  But instead of going into the second player's menu, could we swap the players directly by clicking on that second player.  So clicking on the second player would do the swap.  What do you think?

EveFerox:
Good idea, I changed it.

Ben987:
Thanks!  I will test it soon.  I'll probably rename the transaction to "Swap" though.  Cheers :)

PR #579, opened 2020-01-17
leather items

gatetrek:
added a
Leather Breast Binder
some cleanup
added two kinds of straps to the armbinder for people to choose if so wished


Ben987:
How does the ItemHidden group works?  It's a group that you cannot click?  It's only used to show extra images for the item?

gatetrek:
yeah it's a hidden group you cant reach that has the strap image to be used if people liked to have straps with there armbinder.

Ben987:
Excellent!  Thanks a lot :)

PR #583, opened 2020-01-19
wand fix

gatetrek:
InflatableBodyBag had the wand showing if had the suit under it now hides the wand now


Ben987:
Nice catch!  Thanks :)

PR #584, opened 2020-01-19
Sound tying-untying: hemp

aldarris:
Two new functions in the Dialog.js.
The first is executed when dialogue progress (tying/untying) starts, the other when it ends, either by success or user clicking elsewhere.


Ben987:
That looks great, I'm sure players will like.  Instead of an "Immersion" sub preference screen, could we have an "Audio" preference?  And move everything that's audio based in there?  Also, are you using the volume from the default preference for your sound effects?  Thanks a lot, most games have an audio setting screen, it would only make sense to have one also.  Cheers!

PR #585, opened 2020-01-20
DictionaryFix

NoneNoname:


Ben987:
That looks great!  Can't have hackers include HTML tags.  Thanks :)

PR #586, opened 2020-01-20
Suspension effect for flower dress

TheGnarp:
Since my character ended up upside down in the shibari dojo with a gravity defying dress, I though I should try to fix that.

I hope the original creator of this dress doesn't object!!

Any improvements or changes are of course very welcome and please tell me if it conflicts with any restraints I haven't tested.


Ben987:
That's a great idea!  I'll try it soon.  Thanks :)

PR #589, opened 2020-01-23
OwnerLock validation

NoneNoname:
fixed `C.Appearance[A].Asset.LockedBy` -> `C.Appearance[A].Property.LockedBy `

the timed owner lock is now secure if `EnableRandomInput  == false`

Color == "Default" could still cause missing body since no Default body color exists, moved to Asset.Group.ColorSchema[0] which in the most case "Default" except the body parts


Ben987:
Yes, validating Asset.LockedBy didn't made much sense.

I'm not sure about validating properties one by one though, I guess it could be used against hackers.

Thanks!

PR #590, opened 2020-01-23
Room preview v3

NoneNoname:
I moved around the changes so it shows less line change from the previous one, functionally similar to the last one.


Ben987:
I love the idea but I'd like to have this background selection in a separate menu, so I can use it elsewhere like in the private room.  You can leave it here so I can cherry pick it later to have a screen that we could re-use everywhere.  Thanks!

NoneNoname:
Sure, I moved it into a new screen

Ben987:
That looks good, I'll be able to re-use it for other screens.  I'll try it later today.  Thanks!

PR #593, opened 2020-01-24
Add missing exits

NoneNoname:


Ben987:
Do we really need to put the exit code in a separated function?  What's the gain?

Sekkmer:
Yes, this makes the escape key work in those pages.

NoneNoname:
Yes, this allows escape key to work in these screens, other screens already have this functionality.

PR #594, opened 2020-01-24
ChatRoomSyncSingle 

EveFerox:
add ChatRoomSyncSingle to update a single character in the room

requires https://github.com/Ben987/Bondage-Club-Server/pull/21

### DIFF
--- a/BondageClub/Screens/Online/ChatRoom/ChatRoom.js
+++ b/BondageClub/Screens/Online/ChatRoom/ChatRoom.js
@@ -547,6 +547,17 @@ function ChatRoomSync(data) {
 	}
 }
 
+function ChatRoomSyncSingle(data) {
+	if (data == null) return;
+	if (typeof data !== "object") return;
+
+	for (var C = 0; C < ChatRoomCharacter.length; C++) {
+		if (ChatRoomCharacter[C].MemberNumber == data.Character.MemberNumber) {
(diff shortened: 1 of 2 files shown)


Ben987:
So the code doesn't do anything right now?  It's only a transaction we could use in the future?  I've been wanting to create something like that for a while now.

EveFerox:
It does:
1. Prevent clothing edit reset when someone enters/leaves the room
2. Sends less data

Ben987:
I'll wait until the next full release is done before checking that one, server side changes are always scarier and harder to deploy since every player & version must migrate at the same time.

Atasly:
with the recent update and the lag it brings with, I think it should be interesting to consider that PR

EveFerox:
I agree, this could be a life saver at a every low cost.

Ben987:
Have you seen an actual speed increase with that one?  In theory, the client already validates if there's an update needed on the character or not and refreshes that character only if it's needed.

Since these transactions are critical for chat rooms, I'm always scared of changing them.  Have you tested this PR a lot?

EveFerox:
I can't really test the speed increase since I can't replicate 200+ players on my localhost server.

It will make the server send way less data to clients in the room so there certainly will be an decrease on server load.

I tested enough to validate it just works.

PR #596, opened 2020-01-25
add AfkTimer

EveFerox:
A very basic AfkTimer in a separate script that would set players emoticon to 'afk' after set idle duration.


Ben987:
Have you tested if MainRun() is called when a browser is minimized?  It could be skipped and the afk icon wouldn't be triggered.  Also, should we set a parameter for that?  Some players might get angry if we put an automatic emoticon on them.  Thanks!

EveFerox:
Actually, it was very wrong to put the AfkTimerInit in MainRun, i fixed it.
Regarding letting players choose - do you mean in preferences in UI?

Ben987:
Can you remove the gitignore from the pull request?  Also does having 5 new event listeners impact the game performance?  I don't want to slow anyone down for an AFK timer.  Thanks!

EveFerox:
-removed changes to .gitignore file.
-some fixes after merge.
-better performance from Noname

Ben987:
What's the performance impact on that one?  I'm not a big fan of adding 5 new event listeners for such a minor change.  Thanks!

EveFerox:
With help from Noname, there's no performance impact at all

Ben987:
Thanks!  I'll test it pretty soon.  Do you think it could also be useful to add a /afk command in the chatroom to activate the AFK emoticon?

PR #597, opened 2020-01-25
Login validation: BuyGroups

ace-1331:
This new function will prevent issues with any items added to a group after it was acquired by a player. It will be helpful for the future and it is backward compatible.

 Consequently, I have fixed the values of all the items that appear more than once in the shop to prevent them from clogging the shop screen. (The gags no longer appear 3x for example)

- Check if any items in the inventory is missing an item for its BuyGroup and gives it to the player on login
- Removed duplicate items from the shop


Ben987:
Absolutely!  I'll probably let the zip ties enabled on the legs though, lots of players don't have all items anyway.  But cutting the triple gags and hair accessories is a good idea.

PR #599, opened 2020-01-27
New backgrounds

aldarris:
First half, mostly fantasy, next one will have sci fi and industrial modern.


aldarris:
Industrial and sci-fi.
There are more on the list, but not this week.

Ben987:
Awesome!  It will give more diversity.  Thanks :)

PR #600, opened 2020-01-27
BackgroundSelection filter and texts

NoneNoname:
Since more backgrounds are coming I added a filter so users can search by name of the background,

Added few texts at the top like currently selected room and info about the filter


Ben987:
That's a good idea!  I'll try it right now.  Thanks a lot :)

PR #601, opened 2020-01-28
Fixes + New items

ace-1331:
- added collar cuffs, insertable anal beads and a new type of vibrator
- fixed a bug where when you add a shock collar/unit, the first input would publish an action but would not save the state

for the anal beads, I felt uninspired, I couldn't find an appropriate name since it was already taken. if you find a better name, feel free to change it

Edit: I took the shock collar as a template, while testing I found a bug with it which I fixed while fixing it for the anal beads


Ben987:
This is great!  Lots of new fun toys.  I'll try it right now.  Cheers :)

PR #602, opened 2020-01-28
FR translation

ace-1331:
Here are a few more files for the FR translations. This task was way longer than I first thought so I tried to go for consistency (Like going with every menu first, then asylum, etc.), there is a lot to do still 

Thanks to Mizielli for providing some of the typo corrections from the existing translation and the missing cut scene files


Ben987:
Wow!  This is huge!  Merci beaucoup :)

PR #603, opened 2020-01-28
AssetTypeInfo v2, immediate type selection

NoneNoname:
Currently, SpankingToys and DuctTapeGags can be used like this, later more items can be converted.

SpankingToys Dynamic functions were changed so it works with a character now, that allow DynamicDescription to work in ChatRoomMessage

All items that are converted dynamically get the Draw and the Click methods and the unique js file can be eliminated thereby reducing the code size and makes it more simple to make a new typed item for new contributors (for this more items need to be converted)

As of now, not all features are implemented in TypeInfo.js but all typed items can be converted as shown in #380 

For now, immediate selection do not send a custom message but custom messages are simple to implement (SpankingToys do send the equipped item name since DynamicDescription but it is the usual use message)

If `AssetTypeSet(C, Item, NewType)` is called it should correctly set all properties of the Item, allowing simple ways to implement Type for NPCs


Ben987:
Since this changes a lot of stuff, I'll wait until the next full release is done before reviewing that one.  I don't want to introduce too many last minute bugs.

Ben987:
All of these changes are required to popup the extended menu before applying the item?  Maybe there's a much easier way to do that.  I'll think about it and come back to it later.  Thanks.

NoneNoname:
Obviously not just the popup, I specifically mentioned way more things that this commit accomplish like duplicated code reduction simplified way to add new typed items, reduce possible bugs that new typed items may produce since having a single code only give one possible place to make mistakes, better code in NPC rooms like:

```js
	// Screen/Room/Sarah/Sarah.js   line: 546
	var Cuffs = InventoryGet(C, "ItemArms");
	Cuffs.Property = {};
	Cuffs.Property.Restrain = "Wrist";
	Cuffs.Property.SetPose = ["BackBoxTie"];
	Cuffs.Property.Effect = ["Block", "Prone", "Lock"];
```

would look like this:

```js
	AssetTypeSet(C, InventoryGet(C, "ItemArms"), "Wrist");
```

and right now to see what a type does on an item you have to go through separate folders and files, hope that they have AllowBlock while you setting a Property.Block, same for the Effects, this change meant to make type on items simple to implement

An example implementation of the LeatherCuffs:
(besides these lines no further code required for the type to work)

```js
        LeatherCuffs: {
            Allow: [null, "_Wrist", "_Elbow", "_Both"],
            Mofifiers: {
                _Wrist: { Difficulty: 2, SetPose: ["BackBoxTie"], Effect: ["Block", "Prone"], },
                _Elbow: { Difficulty: 4, SetPose: ["BackElbowTouch"], Effect: ["Block", "Prone"], SelfUnlock: false },
                _Both: { Difficulty: 5, SetPose: ["BackElbowTouch"], Effect: ["Block", "Prone"], SelfUnlock: false },
            }
        },
```

PR #606, opened 2020-01-29
Suit

gatetrek:
add a suit in their own slot so can have equipment used like harness and stuff, plan to have an unzip ver soon


Ben987:
Instead of "AccessBreast1", could you put a prerequisite name that fits with the item?  Something that tells us right away that the suit zipper must be unzipped.  Thanks!

gatetrek:
ok changed >.< after doing the wrong one and fixing that it's in the right spot this time

Ben987:
This is impressive!  I'll test it right now.  Thanks a lot!

PR #608, opened 2020-01-29
Fix: Dog collar

ace-1331:
- addressed a concern where the dog collar seemed pixelated/floating on some clothes (removed the white pixels around it from what I believe was the original background of the asset to improve its quality, it is still the same image)
- addressed a concern where the lock on the dog collar was the wrong size therefore showing on the nose like a nose piercing


Ben987:
Thanks a lot!

PR #610, opened 2020-01-29
fix : mounting items

Atasly:
fixed the issue where Mounted Effect conflicted with the prerequisite NotMounted, making it impossible to dismount.
also added locks to stand saddle and one bar prison, because why not.


Ben987:
Sure!  Thanks :)

PR #612, opened 2020-01-30
working zips and exposed versions

wildsj:
Petsuit updated to use the zip system from the latex suit and a new version with large cut outs for exposed breasts and crotch.


Ben987:
Very nice.  But why is there two versions of the item?  Doesn't one version simply opens?  Also, why is there a "- Copy" image?  Thanks!

wildsj:
That copy shouldn't be there, I will fix that.

The one version is for the existing pet suit, now with working zips. The other version is a different pet suit with large cut outs instead of zips, so the breasts hang out and crotch is always fully exposed.

Ben987:
This looks great.  I need to prepare a second beta, I'll include the items in it.  Thanks a lot!

PR #614, opened 2020-01-30
Fix: some priorities

ace-1331:
- addressed a concern where the ornate cuffs would overlay the hobbleskirt
- addressed a concern where switching from the wooden horse to the sybian would make you fly

EDIT:
- addressed a concern where the hemp rope would not carry it's effects and blocks as intended


Ben987:
This PR creates a conflict in the asset master file.  Can you verify it?  Thanks.

ace-1331:
yeah the conflict viewer was not very clear on that one (err one lines uh), just resolved it (it was just about adding ornatecuffs to the hideitems of hobbleskirt and adding nothorse on the sybian prerequisite list)

Ben987:
Great!  Thanks a lot :)

PR #615, opened 2020-01-30
Latex boots

ace-1331:
- added new latex heels boots as shoes and restraints

Thanks again to Rui for the assets and the great work she did getting them to fit the characters nicely.
Might not be in time for this update, which I understand.


Ben987:
This is great, I need to prepare a second beta anyway, I'll include them in it.

PR #618, opened 2020-01-30
Lovers feature

Atasly:
based on the work of EveFerox and FeuertinKira
tried implementing everything that was asked in the discussion of the original PR, here is the result.

requires server PR [#22](https://github.com/Ben987/Bondage-Club-Server/pull/22)

- three levels relationship (girlfriend, fiancée, wife)
  - each level takes 7 days to progress through
  - last level asks 100$ from each lovers
  - both players have to agree to be each other's lover
  - unique relationship, infos available through management
  - breakable at any point through management without any cons
   - will update both players
   - will remove any lovers locked items on them
- lovers locks
- timer lovers locks
- new permission for owner and lover only, are considered as whitelisted otherwise
- love leather's collar is now only equippable on lovers, thanks to gatetrek's authorization
- NPC support, they can be lovers too, with only one level. not implemented at the moment (only available through console)
- changed timer locks pictures to better suit each family of locks (red = lover, gold = mistress, etc)
- removed tip in main hall about not being able to get a lover

known issues : 	
- rare case where multiple people can love one person (and not the other way around). but it requires a setup difficult to replicate by chance and shouldn't be possible to go past level 1. can detail it if needed.
- when getting a private owner or lover, if going online without logging off, people will be able to propose the player to be their submissive/lover
- if the player is owned/in love with a npc, the duration of their relationship won't be shown on player's information screen


Ben987:
This is huge!  Thanks a lot.  I'll wait until the version is fully released to review it though.  Cheers :)

Ben987:
I've reviewed your PR and it looks awesome, there is tons of work in there, players will love it.

It creates a conflict in Dialog.js now, can you check it?  It's probably nothing big.

Also, this line in Private.JS seems incomplete:

"if ((CurrentCharacter.Love < 60) || (CurrentCharacter.IsOwner()) || (CurrentCharacter.IsOwnedByPlayer()) || CurrentCharacter.IsLover || CurrentCharacter.Is || (parseInt(LoveFactor) < 0))"

Can you validate it for me?  ".IsLover" should be a function and what is ".Is" ? 

Thanks a lot.  This is huge!

Atasly:
good catch, must have forgotten to correct that line in private.js ><
have tested the code one last time and everything seemed to work

PR #598, opened 2020-01-27
Enhanced color picker

vsanvsan:
A enhanced color picker based on HSV color model (like color picker in photoshop)

PR #619, opened 2020-01-30
Revert "Enhanced color picker"

Ben987:
Reverts Ben987/Bondage-College#598


Ben987:
I've found a few bugs with the new color picker:
1, The color slider can go beyond it's limit on both left and right, and give the #NaNNaNNaN color code.
2. Sometimes when I select a color like a green/blue #00707C and then go back in the color picker, the color palette as turned red.  Not the palette I had before.
Thanks for reviewing it, it looks very promising.

PR #620, opened 2020-01-31
Highlight zones with items

EveFerox:
Pretty small improvement that highlights zones that have items equipped.


EveFerox:
added red color for blocked zones

PR #622, opened 2020-01-31
Enhanced Color Picker (BUGFIX)

vsanvsan:
Original Feature PR: https://github.com/Ben987/Bondage-College/pull/598
Bug Report PR: https://github.com/Ben987/Bondage-College/pull/619

Also fixed a bug which color picker won't work on single click/touch event without move the mouse/finger


vsanvsan:
BUG: player cannot input colors to input element

waiting to be fixed

NoneNoname:
I think the event listener for the start should only be active if the color picker is drawn and not constantly, so instead of the window load event it could be added when the color picker is selected to be active and removed on exit

vsanvsan:
> I think the event listener for the start should only be active if the color picker is drawn and not constantly, so instead of the window load event it could be added when the color picker is selected to be active and removed on exit

True, that'll be more elegant.

Ben987:
Thanks!  I'll test it right now.  Cheers :)

PR #623, opened 2020-02-01
Unlock item to remove locks

EveFerox:
Unlocking an item will now remove the locks only without removing the item.
*only for items that are locked with padlocks


Ben987:
This is a good idea but it might slow down roleplay for some.  Could you split the "Unlock" button into two buttons?  One to unlock only and another one unlock and remove.  Thanks!

EveFerox:
done!

Ben987:
Excellent!  I'll test it in the next version.  Thanks :)

PR #624, opened 2020-02-01
new titles

Atasly:
- [Lady Luck](https://en.wikipedia.org/wiki/Lady_Luck) at reputation 100 for gambling
- [Nawashi](https://en.wikipedia.org/wiki/Nawashi) at skill 10 for bondage
- [Houdini](https://en.wikipedia.org/wiki/Harry_Houdini) at skill 10 for evasion


Ben987:
Sure!  Thanks a good idea.

PR #625, opened 2020-02-01
Better dark factor for backgrounds

NoneNoname:
replace dark backgrounds with a transparent black rectangle

my tests have shown no performance hit on drawing the rectangle


Ben987:
Thanks!  I'll try it right now.

PR #628, opened 2020-02-01
missing asset + catsuit assets

ace-1331:
- changed the name of a missing asset
- addressed an issue where all assets of the catsuit were not spelled properly (CatSuit vs Catsuit)

I decided to change the name in the .js file since it was easier than to rename every single assets


Ben987:
Already fixed in latest commits, but thanks!

PR #631, opened 2020-02-02
ColorPicker: sync picker position with input element's value

vsanvsan:
Now color picker will automatically adjust hue and s/v picker position to reflect user pasted/typed hex color code in the binded html input element.

captured


Ben987:
That's a good idea!  I'm getting too close to release date but we could include in the next version for sure :)

PR #632, opened 2020-02-02
Item color realtime preview

vsanvsan:
Player can change color of items in place.

**Please Note:** this PR has commits of https://github.com/Ben987/Bondage-College/pull/631 included

captured


Ben987:
We've already talked about that one and I'm not sure if I want it.  Realistically, it's not possible to change the color of a rope when your victim is already tied up.  In a game it could work though, I'll do a survey with the gang to see how they feel about it.

vsanvsan:
> We've already talked about that one and I'm not sure if I want it. Realistically, it's not possible to change the color of a rope when your victim is already tied up. In a game it could work though, I'll do a survey with the gang to see how they feel about it.

I see your points and I agree with them. 

When I mis-choose the wrong color on character I had to put it like off and on and off ... which bothers me a lot. So there must be gains and looses.

It's always a good idea to do a survey. And what ever it goes, please feel free to tell me.

Issue #633, opened 2020-02-03
Cannot close color picker when requesting desktop mode on mobile devices


vsanvsan:
Reported by Sybil

Way to reproduce:

1. Open Chrome/Safari on mobile devices
2. Turn on "Request Desktop Site"
3. Enter game
4. In Main Hall, click on character, select any part of body, then opens the color picker
5. Now the confirm/close button lost function when player clicks on it (hover interaction not affected)

vsanvsan:
I'll fix that bug and create a PR asap

PR #634, opened 2020-02-03
Color picker bugfix for mobile browsers

vsanvsan:
See issue #633


Ben987:
Thanks!  I will prepare a "hotfix" right now.

PR #636, opened 2020-02-04
Date.now()

NoneNoname:
`Date.now()` is faster than `new Date().getTime()` and works abolutaly the same

(the old one create an object each time with a frozen Date of the Date.now())


Ben987:
Sure, most modern browsers should support Date.now()

PR #638, opened 2020-02-04
Dialog Item Order

NoneNoname:
Move blocked items at the end of the list.

Move items that do not met the prerequisite after the usable items.

Only add remote if there is something to use it on.

With this is should become obvious what is usable and these item should be accessed more easily at the beginning of the list

Faster DialogInventoryBuild using Map, less calculation in the Draw function since color is calculated beforehand


Ben987:
Yes, moving the blocked items to the end of the list is a good idea.  Your code is hard to read for me though, I've did it in another fashion, go check my latest pull request.  Thanks a lot!

PR #642, opened 2020-02-05
Add extra palletes to indicate current and previous colors

vsanvsan:
There will be two extra pallete slot which holding the initial color and the picked color.

The pallete slot has sufficient area to make color previewing on mobile devices much more friendly.

Also, player can quick switch between the two colors by clicking at the slot.

captured


Ben987:
Wonderful!  I'll test it soon.

PR #643, opened 2020-02-05
Update Email

Atasly:
requires Server PR [#24](https://github.com/Ben987/Bondage-Club-Server/pull/24)

allow players to update their email linked to their account.
useful for people who : 
- forgot to/didn't setup an email address
- have stopped using the email address
- wants to delete some digital footprints


Ben987:
For security reasons, we don't send the email back to the client when he logs, so the field will always be empty by default.

Wouldn't it create confusion for players to see an empty field all the time?  I would think that saving the email doesn't work myself if it always comes back blank.

Thanks!

Atasly:
added a "Your email has been updated !" message in the input box after clicking on the button

Ben987:
What I meant is: When players will open the screen, they will always see a blank email, and they will think that their email isn't saved on the server.

And they will report bugs telling us that their email isn't saved on their account.

EveFerox:
Maybe have a "email linked" or "no email linked" indication somewhere?

Ben987:
Yes, something like that.  The player must know that an email is linked to his account and we do not show that email for security reasons.  Try to make it as clear as possible so we don't get tons of questions and bug reports.  Thanks :)

Atasly:
added "No email saved, write here to add one."
and "An email is linked, write here to update it."
depending if an email is linked, or not, to the account

Ben987:
Should we add another validation to prevent hacking?  Like requesting the entry of the first email before overwriting with the second one?

Atasly:
reworked a little how it works, there is now two input fields. one for current email and one for the new one. 
if no email is detected, a message "No email saved, write here to add one" is in the second box.
once the button update email is clicked, if the current email is the same as the one saved, then we overwrite it by the new one. and send a message to the client about the success, or the failure, of the operation.

PR #645, opened 2020-02-06
Sailor outfits and ClothAccessories.

wildsj:
White based sailor outfit with only the main color parts recolor and separate ClothAccessories slot for scarfs or bow.

Hands layer changed to come slightly further up the arms to avoid a clipping issue with the skirts.

Changes to Kidnapping a college to remove ClothAccessories and refit if necessary.


Ben987:
That looks great!  Thanks a lot.

I don't see any assets for the hogtied pose though, have you thought about those?

Also, could you rename "IsOccasional" to "Random", we already have "Random" on the item property, it would be natural to have it on the group.

Cheers :)

Ben987:
That looks great!  I'll try it soon. Thanks.

PR #646, opened 2020-02-08
fix : stables resetting backhair's color

Atasly:
just a little fix : the stables won't reset the color of back hair again.
untested in stables, but the function called at that moment works great.


Ben987:
Thanks a lot!  I'm sure git4nick would like it.

PR #647, opened 2020-02-08
Plain dildo

wildsj:
For Vulva and Butt.

Ideal under chastity for subs that have earn't stuffing, but not buzzing.


Ben987:
Sure!  Thanks :)

PR #648, opened 2020-02-09
Fix for short chain

wildsj:
Allows short chain to be used with petsuits again and new Prerequisite so it cannot be used with items such as the fullLatexSuit which cannot be used while kneeling.


FeuertinKira:
wouldn't it be easier to just add images for kneeling for the fullLatexSuit or am I missing a detail here?

Ben987:
Good catch.  But there's already the "CanKneel" function on characters.  Couldn't we use that instead of adding a new effect?  Thanks.

wildsj:
Kneeling Images for the suit would be good, but until someone makes them, it will have to remain standing only.
I'll check that "CanKneel"

wildsj:
I don't think using the existing "CanKneel" would be that popular.  I have often seen a player bind another while stood up in a way that CanKneel() would return false for the second person (e.g. ties there toes and ankles.) and then use the short chain to force them into kneeling.

The issue with the Items like the FullLatexSuit is just that it doesn't have assets for the kneeling pose and the current prerequisite used to fix that has blocked other things that players are used to doing.

I can rename the prerequisite I made to avoid confusion.

Ben987:
Lets use this route then.  And yes, I might rename the "CannotKneel" to "BlockKneel" or something like that, but it still works.  Thanks a lot!

PR #651, opened 2020-02-11
Strapons as a device

wildsj:
Strap on intended to appear over most items and many clothes.
Two versions, one smooth the other not.


Ben987:
This will be useful!  Cheers :)

PR #652, opened 2020-02-11
Blindfold Tweaks

Ayesha-678:
Grey scaled the blackout lenses, and recentered the sleep mask to be more in the middle of the face, and got rid of a small black mark on the sleep mask image.


Ben987:
Thanks a lot!

PR #653, opened 2020-02-12
Fixes item names

wildsj:
Fix some item names and also a few other small details to those items.


Ben987:
Awesome!  Thanks a lot, it will make it easier later on.

PR #655, opened 2020-02-12
reimaged on glove and upper

gatetrek:
reimaged the gloves and upper suit so line be gone and fixed the lower case h in tapedhands


Ben987:
Excellent!  Thanks a lot.

PR #656, opened 2020-02-12
Skimpy Naked Apron

Ayesha-678:
A Skimpy Version of the Naked Apron.


Ben987:
Same here.  Can you unify this will the other maid clothes?  Thanks a lot.

PR #657, opened 2020-02-12
Maids return hand, leg and feet items

aldarris:
Not accepting work when restrained with owner's locks is already implemented.
Added a new structure to hold all the items that need to be stashed and returned.


Ben987:
Thanks!  So all it does is dress you back up as you were when you finish the maid mission?

Why are you adding a "//where is it marked as restrained??" line in the code?

aldarris:
Sorry, that was a note to self, removed.
Yes, the legs, feet and arms are in addition to the dress and hat.

Ben987:
Alright!

PR #658, opened 2020-02-13
Chain Bondage

Kitsune-hellfire:
Chain Bondage for Arms, Legs and Feet

Follows the style of the Hemp Rope, so there are poses for suspension as well and chains are lockable (only issue found is that the lock dissappears when switching between posistions, which i cant find a fix for at this time)


Ben987:
This is great!  I've been wanting to add chains for a while now.  Thanks a lot, I will test it soon.  Cheers :)

PR #659, opened 2020-02-13
Wedding Dress & Bridesmaid Dress

FeuertinKira:
-Added two new Wedding Dresses; one with shoulder straps, the other shoulderless

-Added a Bridesmaid Dress

-known issue: Since I made images for basically every single pose with the wedding dresses, there is a few pose combinations that might look slightly odd since they will overwrite which pose the wedding dress uses ---> example: Yoke + kneeling will use images from kneeling, so the shoulder straps look a bit odd, basically lower body specific poses can overwrite upper body specific poses


Ben987:
It looks awesome!  I will test them soon, I'm sure player will love them.

PR #660, opened 2020-02-13
Chains fixes

FeuertinKira:
-Fixed chains locks vanishing

-Changed chain pose to not be changeable when locked (went with the simple implementation for now; could probably be done fancier/more elegantly as discussed before)

-Added a few missing lock images


Ben987:
Thanks a lot!  I'll try it right now.

PR #661, opened 2020-02-13
2 Gowns, bits and different Left + Top within groups 

wildsj:
Puffball and Hobble gowns.
Bondage bouquet
Individual assets within a group can be given their own Left and Top attributes.
Slight change to preview pics fo to penis dildo and plug.

One known possible issue.
The Bondage bouquet only cuffs the persons wrists in front of them and should really prevent them from doing things, like the FullBodyShackles. For example once the bridesmaids are fully dressed with their bouquet, they can still help prepair the bride (When the wedding part is ready mittens can be applied if necissery). They do not "Block" the player. This means though that the rescue maid in the main hall will not notice the bouquet on it's own or remove it when they remove other bondage.


Ben987:
Awesome!  I didn't expect a "bondage bouquet" :)

PR #665, opened 2020-02-14
Maid Apron Fixes.

Ayesha-678:
Fixed a missing 1 in the maid apron name, and fixed a bug that made it invisible.


Ben987:
Yep!  Good catch.

PR #666, opened 2020-02-15
Exposed Maid Apron

Ayesha-678:
An exposed version of the Maid Apron~

UPDATE
Added a missing name in the csv file


Ben987:
Thanks a lot!  I'm sure that maids will love it.

PR #667, opened 2020-02-15
ABDL dress and covered diapers.

wildsj:
1 dress and 2 types of covered diaper.


Ben987:
Awesome!  Thanks a lot :)

PR #668, opened 2020-02-15
Improved dice roller

nikoleis:
Allows to roll XdY on top of the previous method


Ben987:
Sure!  That sounds like fun :)

PR #669, opened 2020-02-15
Added Lockable Full corset

Ayesha-678:
Added a the full breast corset as a lockable item.


Ben987:
Alright!  More locking items :)

PR #670, opened 2020-02-15
Gown and accesseries

wildsj:
Items added
Another gown.
Silky gloves to match previous silky/satin underwear.
Bonnets, Hair Ribbons and Crowns

Changed priorities round or Hat and one of the HairAccessory slot, so that a HairAccessory can be put in front of or behind the had.

Fixed image size for the hair flower and ribbons when supended.


Ben987:
Absolutely!  Crowns are a good idea.  Cheers :)

PR #672, opened 2020-02-16
Fix for InflatableBodyBag with Catsuit

wildsj:
The legs of the cat suit could appear outside the BodyBag, when the body bag was inflated. Suit and SuitLower added to the list of assets hidden by the BodyBag.


Ben987:
Good catch!  Thanks.

PR #673, opened 2020-02-17
Collar items.

wildsj:
Added:
Collar tags based on livestock ear tags
Slender steel collar including for slaves.

Fix:
Bouquets appear over the top of collar chains and leash.


Ben987:
Very nice!  Thanks.

PR #675, opened 2020-02-18
Apron as an accessory

wildsj:
Added an apron I had been working on as part of a cafe style maid outfit, now as just a separate accessory and acquired with all the other maid clothes.

Fixed asylum nurse to never wear ClothAccessories.


Ben987:
Great!  I will test it soon.  Thanks :)

PR #676, opened 2020-02-20
Sounds for the new items

aldarris:
It is a problem that when rope (or the new chain) is changed type (from regular to hogtie), the server does not send the sender info like in others.  Another thing to code around.


Ben987:
Yes, the info sent is different when changing pose.

Thanks for the new sound effects!  I'll try them soon.  Cheers.

PR #677, opened 2020-02-20
Remove Blush & reset bug from  chain/rope hogties

FeuertinKira:


Ben987:
I've wanted to change the timed facial expression for a while.  Check my latest pull request, it merges both expression functions and prevents it from being called if there's an expression already setup.  Thanks for the pull request though.  Cheers!

PR #682, opened 2020-02-22
Chain Leash

Kitsune-hellfire:
Chain Leash, pretty standard with lock and variations for taken and holding (tho like the normal leash, not usable yet)


Ben987:
Great!  I will test it today.  Cheers :)

PR #687, opened 2020-02-25
all fours

gatetrek:
new pose all fours for pet crawler also added the pose to the chains|ropes.  
know issue when that pose the other zones are inaccessible so no pet in a kennel for now
two, new suits that are see-though that one has zips and the other doesn't


FeuertinKira:
That "PetCrawler" looks oddly familiar. Almost like the petsuit I made for Skyrim, hhmmm yes it's exactly that.

Ben987:
This is huge!  Thanks a lot.

What happens with clothes right now?  Can you post a few screenshots?

gatetrek:
right now they would look like 
http://prntscr.com/r7km6c
well need to angle them to fit the body and as there so many clothing right now
was planning when next have time to do the clothes

PR #689, opened 2020-02-29
Seamless Catsuit

Kitsune-hellfire:
Seamless catsuit, alteration of orginal just with no zippers on this one, left the price as the same as catsuit


Ben987:
Awesome!  It creates conflicts with the latest pull requests though, can you check them out?

Also, why are you adding a comma on these lines of the catsuit: 
{ Name: "Zip", AllowColorize: false },

Since the array ends, there should not be an extra comma there.

Thanks!

Kitsune-hellfire:
okay, think that fixed it, as for the comma, must have been a typo on my side, sorry for that

Ben987:
There's still an extra comma on line 218 and 220.  Can you check these?  Thanks.

Kitsune-hellfire:
should be sorted now

Ben987:
The extra comma is still there but I'll remove it in another commit.  Thanks a lot!

PR #690, opened 2020-03-03
menu rework

gatetrek:
both the chain and hemp rope menus have been reworked on request 
gloves for the all four pose 
item added:
Clit&Dildo Vibrating belt,  that can vibe either the clit or the crotch or both


Ben987:
Great!  But why is your pull request adding a "SpeechStutter" function?  Thanks.

gatetrek:
due if have both running on the same Intensity when turn one on both would turn on so made it a copy what let with vibe to turn on with off it touching the other

Ben987:
No, we cannot have two functions with the same name that does almost the same thing, we'll need to add the "Buzz" code to the current function we have.

Since I'm currently coding the arousal meter system, we could do a survey on stuttering to players, to see what should or should not stutter.  I'll do a survey really soon and tag you on it.  Thanks!

Ben987:
This looks great, you code very quickly.  I'll test all of this soon, it should work great with the new stuttering system.  Cheers!

PR #691, opened 2020-03-05
Crotch Chain

Kitsune-hellfire:
more chains, was suggested so i figured id add them, its just a simple chain crotch bind.


Ben987:
Great Idea!  Thanks a lot.

PR #692, opened 2020-03-05
Locking Heels

Kitsune-hellfire:
was given the suggestion of some normal "discreet" locking heels, so have used the normal heels and made some into a pair.


Ben987:
That looks good, I'll test it right now.  Cheers.

PR #693, opened 2020-03-05
Display Cases

Kitsune-hellfire:
have some fun locking a sub in a display case, watch them in the case like a piece of art or just as a trophy.
normal size and small size, lockable as well, similar to the cage and crate, but more visible while still contained.


Ben987:
Very nice, I'll try it now.  Thanks!

PR #694, opened 2020-03-06
Display Case Tweaks

Kitsune-hellfire:
small tweak, removed the "BlindNormal" effect as its a glass case and allows you to see out as well as in.
Added a "DeafLight" effect so that the glass muffles what the player hears, and works with the "GagLight" effect to muffle the player inside to those outside


Ben987:
That makes perfect sense.  Thanks!

PR #696, opened 2020-03-07
moved up

gatetrek:
moved up as requested 
added some skin to the chain menu pic


Ben987:
Thanks a lot!  I will test it now.  Cheers.

PR #698, opened 2020-03-09
New owner remote rule

Atasly:
new owner rule to prevent the use of remote on oneself


Ben987:
Sure! That's a fun rule, I'll try it soon.

PR #700, opened 2020-03-12
Taped Vibrating Clit Egg

Ayesha-678:
Added a taped vibe egg in the clit piercing area


Ben987:
Looks good!  I will try it right now :)

PR #704, opened 2020-03-16
New Slim Masks and fix for leather mittens hiding gloves

CthulhuDancing:
3 variants for the mask
Full face,
open mouth
open eyes
no alpha zones


Ben987:
Thanks a lot!  I'll test it really soon.  Cheers :)

PR #705, opened 2020-03-16
Email Update improvements

Atasly:
Added placeholders for email status
Moved email input fields to their own security tab


Ben987:
Thanks, but why are you adding code from other sub screens and comment it?  Shouldn't we remove that?

Atasly:
oh- sorry I should have forgotten to remove that part when creating the new page >< seen that you removed it later, thank you

PR #708, opened 2020-03-17
Boxes with head exposed

ace-1331:
Here is a variant of the boxes with the head exposed (and ability to have the hands outside of it as well)

As a mean to keep someone else's work as is, I didn't alter the existing ones to make it so they are all under one menu. (Technically they are not the same size anyway)

_Credit goes to Klatchi for the modified assets_


Ben987:
Great idea!  I'll test it really soon.  Thanks a lot.

PR #709, opened 2020-03-17
adjustments

Kitsune-hellfire:
small adjustments to the assets, making the characters fit with out gaps between tails and the edge and not allowing boots to clip out the bottom


Ben987:
Great!  Cheers :)

PR #710, opened 2020-03-17
mermaid suit tweak

Kitsune-hellfire:
added a few more things to hide, this stops clipping of items worn under the restraint


Ben987:
Thanks a lot!

PR #713, opened 2020-03-19
fix boxes

ace-1331:
-addressed a concern where the small box would prevent you from being unlocked while your hands were outside (clash between current pose and prerequisites)
-blocks from pulling the hands in/out if they are tied


Ben987:
Excellent!  Thanks a lot :)

PR #714, opened 2020-03-19
mouse ear tweaks

Kitsune-hellfire:
tweaks to the mouse ears, removed a gap between the ears and the head on certain hairstyles


Ben987:
Puuuuuuuuuuuuuurfect!

PR #715, opened 2020-03-20
Update to the Slim Masks to prevent them from hiding hair accessories and hats

CthulhuDancing:
the slim masks were not supposed to hide anything on the hair.


CthulhuDancing:
hmmmm, i dont think i did this pull request right. Let me try to delete it and do it correctly

PR #716, opened 2020-03-20
Fix for Slim Masks hiding hair accessories and hats

CthulhuDancing:
They shouldn't be doing that!


Ben987:
Good catch!  This makes a lot of sense.

PR #717, opened 2020-03-20
Various fixes

Atasly:
- modified pictures of seethroughsuits to avoid having visible seams near the wrists and the hips
- removed changeClothes message for oneself


Ben987:
Excellent!  Thanks a lot :)

PR #718, opened 2020-03-21
fixes default blush/eyes/drool/eyebrow bug when affected by arousal

Atasly:
default blush/eyes/drool/eyebrow weren't avalaible when using "affected by arousal" setting. this fixes it


Ben987:
Good catch!  Thanks.

PR #719, opened 2020-03-21
modified pink overlay near orgasm

Atasly:
reworked the size/position of the pink overlay near orgasm to avoid having it on top of dark background or UI elements.


Ben987:
Sure. There's no need to change the buttons when an orgasm is building.

PR #722, opened 2020-03-21
Admiral

Kitsune-hellfire:
admiral uniform to go with the captains hat, hoping to add it to this update if possible.

if added, please credit it under my club name "Kirsty"


Ben987:
That's a fun idea!  Thanks a lot :)

PR #724, opened 2020-03-22
Fix for Slim mask hiding hair accessories and hats + Shinobi mask for glasses slot

CthulhuDancing:
Added a new item for the glasses slot, a shinobi mask, based on the open eyes mask.
This will let people use it for RP costumes since it looks good as a ninja mask and not have to worry about the gag effect or blocking regions. The mask itself lays under the gags and actually looks really good with most of them.


Ben987:
That sounds like fun.  Thanks a lot!

PR #726, opened 2020-03-23
More seam and opacity fixes for see-through suit

Atasly:
like the title says
known issue : missing visuals for all-four pose  causing visuals to misalign


Ben987:
Thanks a lot!  Cheers

PR #727, opened 2020-03-23
Few Quick Fixes

Kitsune-hellfire:
Fixes to the crotch chain, just erased the flecks that seemed to carry over, fixed a slight rise in the admiral top that was mentioned to me.


Ben987:
Thanks a lot!  I'll prepare a hotfix.

PR #728, opened 2020-03-24
Leather mittens Hiding Gloves again

CthulhuDancing:
I had to put the gloves back the way they were but i might take this time to see if we can work out a way to clip items under other items. For now I had to put the glove hiding setting back so that the gloves fingers don't poke out from under these mitts (wasn'ta  problem with the cloth mitts cause the cloth mitts extend past the fingers


Ben987:
Thanks, but why are you pushing the shinobi mask again in your request?  Are you using an old branch?

CthulhuDancing:
oh, i must be. I keep a forked branch that i update from the master so let me recreate the pull request without all the unnecessary stuff.

PR #730, opened 2020-03-24
Gloves Fixes

CthulhuDancing:
Added TapedHands pose to gloves assets and fixed some clipping items so that we don't have to hide gloves anymore when wearing mittens that set TapedHands


Ben987:
Thanks a lot!  I'll prepare a hotfix.

PR #731, opened 2020-03-27
Additional Property Check

NoneNoname:
Remove impossible combinations of properties


Ben987:
Sure!  Better patch that little exploit.

Shiranui-Izayoi:
While this is a little awkward to ask I kinda don't want this to be patched. Since I recently started using the ability to set properties to kinda "save and load" settings for a console sctipt I made and use, they only effect me and no one else so it doesn't harm anyone.

Also I can't really see how setting properties could negativily effect anyone I mean they don't do anything to my knowledge.

Maybe I am naive of its potential harm and if so go ahead but if it doesn't do any harm I fail to see why it should be patched.

PR #639, opened 2020-02-05
Reduce CharacterRefresh calls 

NoneNoname:

PR #735, opened 2020-03-27
#639 CharacterFullRandomRestrain

NoneNoname:


Ben987:
Have you seen speed improvements when using these new function calls?  This changes lots of code in some of the most used functions.  Thanks.

NoneNoname:
CharacterRefresh causes the hole character canvas to be redrawn basically it cost a lot for the game to do this, for example, this `CharacterFullRandomRestrain` can cause 6 of them, `InventoryRemove` would straight up cause as many as restrains you have on yourself.

The Login screen slight freeze is caused by on average 14 `CharacterRefresh` calls when the new character is generated, this update can cut it down to 2 where the second one is just from the image loading

Every function is changed in a way that it works the same in case the new `Refresh` parameter is left out by that allowing any previous code to work the same

`if (Refresh || Refresh == null) CharacterRefresh(C);`

Ben987:
CharacterFullRandomRestrain isn't called that often, but removing the small delay in the login screen could be nice.  I'll review the other changes and test that one after.  Thanks!

PR #740, opened 2020-03-28
Various assets changes

Atasly:
- changed leg binder, bitch suit and full latex suit default color to match up their preview image
- changed priority to allow seeing piercings under suit
- cleaned some unnecessary priorities
- tweaked hobble skirt to allow seeing boots, similar to how leg binder works
- adjusted alpha zone for boots and shoes (to avoid having a too big portion of the bench being cut)


Ben987:
Thanks a lot!  This should fix a few visual bugs we've had for a few versions.  I will test it soon.

PR #741, opened 2020-03-28
Various small quick fixes

Atasly:
- Activity messages are formatted like Action (same CSS style)
- Stable won't change hairs of Player trainers anymore
- the heart piercings won't be removable through a suit anymore
- added ability for lovers to unlock any lock on their lover except owners locks
- someone with her arms tied could unlock someone else (missing CanInteract check)


Ben987:
Should the lover key become a master key like the owner key?  I'll do a quick survey on that one.

Atasly:
I was actually pondering on that one myself, wanted to include that as a rule for owners to forbid lovers key usage (possible for lovers to unlock by default)

Ben987:
Since this created a heated debate, we'll wait until Patreon voters choose if they want Lover keys to unlock Mistress locks or not.  The poll result will come up on April 3rd.  Thanks!

PR #742, opened 2020-03-29
Optimize AssetGet

NoneNoname:
As more and more assets are added to the game array iteration is getting longer and longer, Maps are significantly faster than arrays, this pull request aims to replace `AssetGet` to a faster version and update some iterations to use this function instead


Ben987:
What in-game speed increase I should be able to see/monitor with this pull request?  Give me an in-game example I could test.

NoneNoname:
image
image
image

Ben987:
I'll run some tests on AssetGet to see how fast or slow it goes, I'll let you know on my results.  Thanks!

Ben987:
I've ran some tests on AssetGet with many different assets and it's really fast on my side, it doesn't seem to be an issue.  All AssetGet I've tried took less than 1 millisecond to run.  Are you sure this is needed?

PR #743, opened 2020-03-29
ChatRoom enter leave

NoneNoname:
when a character enters or leaves everyone gets updated this mess up appearance edition and slows down the update process, this pull request makes is so that only one character is updated

### DIFF
--- a/BondageClub/Screens/Online/ChatRoom/ChatRoom.js
+++ b/BondageClub/Screens/Online/ChatRoom/ChatRoom.js
@@ -706,12 +706,28 @@ function ChatRoomMessage(data) {
 function ChatRoomSync(data) {
 	if ((data != null) && (typeof data === "object") && (data.Name != null)) {
 
+		var Joining = false;
+
 		// Load the room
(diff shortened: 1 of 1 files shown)


Ben987:
How can I recreate that bug?  I need someone in the appearance screen and someone else entering the room?   What bug does it create?

Sekkmer:
well it is not really a bug it just how the game works if someone edits a character in the appearance menu `CharacerRefresh` will set everything back which in a busy room can happen a lot, with the new Single update this happens much less, but enter leave works with the full Sync

Recreate: edit anyone appearance in a chat room, enter or leave with another character

NoneNoname:
well it is not really a bug it just how the game work since appearance menu edit is on a character `CharacterRefresh` will override all changes which previously happened on every update now since SyncSingle is added the two most common change that remains are the 'enter' and the 'leave', other actions are rarely happen and do not worth an extra check

Recreate: edit anyone's appearance while enter/leave with another character, any change that you have made will be reseted

Ben987:
Great!  I was able to recreate the bug, I'll see if it fixes it.  Thanks :)

PR #744, opened 2020-03-29
Bugfix: Collar removal in kidnap minigame

elliesec:
* Ensures that collar accessories are removed when removing a collar during the kidnap minigame (this would occurr when entering a minigame whilst already wearing a collar with accessories).


Ben987:
Good catch!  Thanks a lot, I will test it soon.

PR #745, opened 2020-03-29
Remove Zone from TailStraps

NoneNoname:
TailStraps belong to the Appearance, Zone should only exist on Items


Ben987:
Good catch, that doesn't belong there.  Thanks!

PR #746, opened 2020-03-30
setting : sounds are played only if player involved

Atasly:
added a setting in the audio tab preferences to play sounds only if player is involved


Ben987:
Sure!  It will give more flexibility to how players configure their experience.  Cheers :)

PR #747, opened 2020-03-30
Appearance v2

NoneNoname:
https://www.youtube.com/watch?v=1y6cN_tlbQY

Added features:
Unified groups: Head, Upper/Lover Body
Block Items
Improved selection
Import/Export
View other characters' appearance in view-only mode
more color functionality (HUE/luminescence up/down, copy/paste )
Undo option
color is preserved in back/next if the user goes over the None option

#733 is merged in this pr

(The video was made by GLDraw since the old engine is less responsive)

### DIFF
--- a/BondageClub/Screens/Character/Appearance/Appearance.js
+++ b/BondageClub/Screens/Character/Appearance/Appearance.js
@@ -766,9 +766,8 @@ function AppearanceWardrobeRun() {
     else AppearanceMode = "";
 
     if (navigator.clipboard) {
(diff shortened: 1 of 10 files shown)


Ben987:
Sorry, I probably won't take this one, it changes the code structure way too much and will make it much harder for me to track and correct bugs.

There's no class, or let or const right now, plus you're introducing many new code concepts that we don't have and will make the code much harder for new developers to join the project.

I'll think about it and review it again later.

EveFerox:
noice.

PR #748, opened 2020-03-30
GLDraw Single Canvas

NoneNoname:
Previously GLDraw used 2 canvases with double amount of texture, this has been fixed.
Normal and blink characters are drawn side by side on the same canvas.


Ben987:
Sure, I'll test it soon and let you know if I get better results.  Thanks!

PR #749, opened 2020-04-04
Hemp Rope & Chains arms enhancements

Nina-1474:
Reworked the type-selection page for Hemp Rope & Chains on arms based mainly on the SpankingToys page. Allows multiple pages of options and easier addition of new types in future.
Added Wrist Tie and Wrist & Elbow Tie types.


Ben987:
That's a good idea!  Should the wrist tie become the new default then?  The one that requires no bondage skill level?

Also, have you tested the Shibari Dojo to make sure the rope interactions still works?  Thanks.

Nina-1474:
I've made Wrist Tie the default Hemp Rope and Chains arms options, and rearranged the options in ascending order of difficulty, slightly upping the difficulty for most.
Added a small fix to display a message when the chains are locked.
I've set the NPC/non-ChatRoom hemp rope arm ties to use box-ties instead of wrist-ties where possible for the sake of consistency and using the more visually interesting option.
I've tested the dojo and other rooms and it all looks ok!

Ben987:
True, having the BoxTie as default will be better, since it's visual.  Changing the rope scripts everywhere will not be perfect, since NPCs also applies random bondage and kidnapping matches will always show the wrist tie, showing no ropes.  

Could we have this setup then:
- BoxTie (default), req. 0 bondage
- WristTie, req 0 bondage
- Elbow Tie, req 2 bondage
- Hogtie, req 4 bondage
- All Four, req 6 bondage
- Supsended hogtie, req 8 bondage

So with each 2 skill levels, you will get a new bondage position.

What do you think?  Thanks.

Nina-1474:
@Ben987 I did consider the random-tie impact but thought it might be worth the loss for the benefit in the chat rooms. Since it makes more sense roleplay-wise to start with the minimal option then add more ropes, rather starting with box-tie and potentially subtracting ropes immediately. Do you reckon it's still worth keeping boxtie default for the sake of the single-player?
In either case, yes I can space out the skill levels that way.

Ben987:
Yes, please keep the box-tie as default.  I'm sure there will be lots of complain if we change that.  If the wrist tie was shown on the screen it wouldn't be a problem.

Nina-1474:
Changes made! The default everywhere will be the box-tie and the skill requirements have been updated.

PR #750, opened 2020-04-04
CharacterClass

EveFerox:
Example of converting Character object to class without changing any logic


Ben987:
Thanks, but we don't need classes right now.  Maybe someday.  Cheers!

PR #751, opened 2020-04-04
Bugfix: Allows flat gags to be switched straight to unique gags

elliesec:
**Bug Reproduction Steps**

* Apply a single flat/corset gag (e.g. cloth gag)
* Attempt to swap the gag for a unique gag (e.g. ball gag)
* The gags cannot be swapped directly - the flat gag must be removed first

This PR fixes the above issue. If there are multiple layers of flat gags, the swap will still fail as before, because there is still a flat gag underneath.

I'm afraid the logic is a little more convoluted than before, but I wasn't sure where on the ES spectrum we're aiming, so I went as compatible as possible.


Ben987:
Thanks a lot.

@FeuertinKira, are you fine with this pull request?  Can you think of errors it could create in the gag system? Thanks.

FeuertinKira:
Intention seems fine to me, haven't checked over the code though.

Ben987:
Thanks!  I'll test it soon then.  Cheers :)

PR #754, opened 2020-04-09
Bugfix: Collar restraint removal in kidnap minigame

elliesec:
This is the same issue as #744 - ensures leashes and other collar restraints are removed at the same time as the collar.


Ben987:
Good catch!  Thanks :)

PR #756, opened 2020-04-10
Fixes

Atasly:
- removed prone property on LeatherSlimMaskOpenEyes, DogHood and FoxyMask
- remove ItemNeckAccessories and ItemNeckRestraints when ItemNeck is being removed (untested)
- updated management text and condition to match the actual comparison the server is doing, which is seven days and not three for removing the SlaveCollar
- seam fixes for standard catsuit (missing one pixel line)


Ben987:
Nice fixes, thanks a lot :)

PR #758, opened 2020-04-10
assets adjustements

Atasly:
- Added lock to all piercings and increased their difficulty to not be struggled from (partially tested)
- Allowed glasses to be visible under FoxyMask and LeatherSlimMaskOpenEyes
- Allowed gags to be visible under FoxyMask


Ben987:
This creates a conflict with your previous fix.  Can you review please?  Thanks.

Atasly:
should have expected it since they both modified the same lines, it's okay now

PR #759, opened 2020-04-10
Change the photographer dialogue to standard.

GIT4NICK:
Remove your functions and replace them with standard functions.
Adjustment in the Dialog.js for extended selfbondage


Ben987:
It's great to hear from you :)

These are big changes, have you tested all the possible scenarios?

Thanks!

GIT4NICK:
I guess so. In the photo studio it works as planned.

When testing in the private room and in the chat room, I noticed no errors. So I think it has no side effects.

Ben987:
Thanks!  I will test it soon, maybe I'll change the (CurrentScreen == "Photographic") to something more generic, I'll see.  Cheers :)

PR #760, opened 2020-04-10
Panty Stuffing colour correction

Nina-1474:
Changed the panty stuffing to grayscale and lighter tone so that the final colour more closely matches the user's selected colour.


Ben987:
Very nice!  Thanks :)

PR #755, opened 2020-04-09
New rope tie: Rope cuffs!

Ayesha-678:
Jumped at the chance after seeing the amazing new code for making rope ties, and made rope cuffs!

Edit: Added Chain Cuffs as well!

PR #761, opened 2020-04-10
#755 Fix: Override Priority setting to fix Rope/Chain Cuff clipping

Nina-1474:
Added an OverridePriority setting for appearance item properties that will be checked first before the Asset's DrawingPriority and the Group's DrawingPriority.
In the case of the RopeCuffs and ChainCuffs options, the override priority will be set to 30 similar to MetalCuffs, instead of the default ItemArms' 32, so that the item is hidden behind clothes.

### DIFF
--- a/BondageClub/Screens/Character/Appearance/Appearance.js
+++ b/BondageClub/Screens/Character/Appearance/Appearance.js
@@ -191,7 +191,10 @@ function CharacterAppearanceSort(AP) {
 	var Arr = [];
 	for (var P = 0; P < 50; P++)
 		for (var A = 0; A < AP.length; A++)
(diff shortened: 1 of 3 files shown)


Ben987:
Yes, that solution should work.  Have you tested with all rope/chain positions and a few different outfits?  Thanks.

Nina-1474:
@Ben987 Yes I've tested many different outfits, for both the new positions as well as the previous positions, and for different body sizes. 
The only remaining case of noticeable clipping that I found was over Student Short Skirt when wearing a short top, which is already an existing bug on MetalCuffs for the same priority. I'd like to leave that fix for all 3 for a future PR, just in case lowering the priority value introduces other issues.

Ben987:

PR #762, opened 2020-04-10
Sock Stuffing gag

Nina-1474:
Based closely on Panty Stuffing gag


Ben987:
Thanks!  I will try it soon :)

PR #765, opened 2020-04-12
small fixes for complete tape

fleisch11:
fixes, that the complete ductape was  a bit to high


Ben987:
Thanks!  I will test it soon :)

PR #767, opened 2020-04-13
More fixes

Atasly:
- latex full lockdown now hides all ItemFeet
- When accepting the full collar, subs weren't updated properly, also reverted some unnecessary lines I added in a previous PR
- added color preview on slavecollar
- when choosing last expression of the list then relogging, some double expressions were built to the list, not cycling properly over the length of it, now fixed.

all those fixes have been partially tested


Ben987:
Thanks, but in Login.js, why are you removing these two lines?

		InventoryRemove(Player, "ItemNeckAccessories");
		InventoryRemove(Player, "ItemNeckRestraints");

These lines are used to remove the collar accessories when the collar is removed.

Atasly:
a small adjustement for PR #756, added those two lines too quickly without considering the rest of the code (it checks if the Player is supposed to have an owner but no SlaveCollar, removes the ItemNeck, then adds the SlaveCollar back, I assume we want the Player to keep her ItemNeckAccessories and ItemNeckRestraints there)

Ben987:
Excellent, I'll test it soon.

PR #768, opened 2020-04-13
Stockade

Kitsune-hellfire:
wooden stockade, simple but restrictive

if this is added to the club, please credit it under my club name "Kirsty"


Ben987:
That looks good.  But should it be called "Stocks" instead of "Stockade" ?

I think a stockade is wooden fort to store muskets or something like that.

Kitsune-hellfire:
made a quick name change from "Stockade" to "WoodenStocks", done this as im considering making a metal frame stocks

Ben987:
Awesome!  I will try it soon.  Thanks :)

PR #769, opened 2020-04-13
Inaccessible zones fix - Up button

Nina-1474:
A solution to issues caused by certain poses pushing zones offscreen. For example, using remote vibrators when they're hogtied, or feet arousal options when they're kneeling/all-fours.

An 'Up' button will appear on screen above the character at the top-left under the following conditions only:
- In the character viewing screen
- An item zone has been selected (and the zones are all therefore highlighted)
- The character's height modifier is less than -90. i.e. the zones are being moved downwards, and there's enough room for the button.

On clicking the button, HeightModifier is set to 0 which repositions both the character and item zones up at the default 'standing' location at the top, allowing all zones to be visible and selectable.
This position will be reset to normal as soon as the Exit button is selected, an activity is selected or the character is refreshed for any reason (which covers changing expressions and items).

The zones that would become newly accessible are only: Kneeling allowing ItemLegs and ItemBoots, Hogtied and AllFours allowing ItemPelvis, ItemVulva and ItemVulvaPiercings. ItemMisc for Hogtie/AllFours was the only new zone that seemed to cause problems so this will now be blocked.
I've also tested appearance in general on different poses/screens/rooms and don't think there are any other issues caused.


Ben987:
Since we have a visible zone grid now, would it make it easier to never use the height modifier to show the grid?  We could do a quick survey on Discord to see how people feel about it if you want.

Nina-1474:
Do you mean always have the zone grid filling the screen (as if standing) regardless of the character position?
I did originally try it that way, but the problem is it can be a bit disorienting having to click the 'wrong' location all the time. For example clicking nothing in mid-air above the character to remove a hogtie rope, whereas clicking for the arm zone beside the character at the bottom feels much more natural and immersive.

Ben987:
Very well, I'll test it right now and see how it feels in-game.  Thanks a lot!

Ben987:
@Nina-1474 

I've found two bugs so far.  When using an activity, I now get this error:

Dialog.js:795 Uncaught TypeError: Cannot read property 'HeightModifier' of null
    at DialogClick (Dialog.js:795)
    at CommonClick (Common.js:165)
    at Click (index.html:351)
    at HTMLCanvasElement.onclick (index.html:393)

Also, if the facial expression changes while you're character zones and up was clicked, the character goes back down automatically.  The facial expression will change automatically when the effect of a used item goes off or the arousal meter increases/decreases.

Good luck!

PR #770, opened 2020-04-13
Magic Theatre room gag slot fix

Nina-1474:
Allowing the gag to be applied in any of the three ItemMouth slots instead of only the leftmost one when required for tricks.


Ben987:
Great, thanks a lot!

PR #771, opened 2020-04-13
Beginn Police & Bad Girls Gang

GIT4NICK:
Out of Prison Scene
Can Steel Dices in Gambling Hall, and Hide in Prison
Try to Catch in Gambling Hall, Stable and Mainhall


Ben987:
Very nice, I'm sure that players will be happy to see new content in there.  Thanks!

PR #772, opened 2020-04-14
Torso Rope Menu Rework, And two new rope ties for it!

Ayesha-678:
Copied the code from the new arm rope menu, and reworked it in the torso rope menu, and added two new ties.

Star Harness Tie: Based off the bra item!
Waist Rope: Seems simple, and useless, but could lead to some fun situations!


Ben987:
This looks great, I will try it soon.  Thanks a lot!

PR #773, opened 2020-04-15
Easter Surprise

BunnyMadMan:
Added 4 new items under the ItemButt file; 
- 2 normal butt plugs - bunny tail plug white / brown
- 2 special items with javascripts ; 
  - 1 bunny tail vibe plug, with special text under Dialog_Player.csv 
  - 1 special egg plug, with special text under Dialog_Player.csv + added 3 more speeds for players to 
     choose (mild tease, tease, very high) which fall under the "intensity" option

All items very for now tested, the prize was set at a reasonable limit.


Ben987:
Setting intensity over 3 will probably create problems in chat stuttering and arousal progress.  Everywhere else: "low" is 0 and "maximum" is 3.  Can you set your intensity based on that structure?  Maybe remove a few levels in between.  Thanks :)

PR #774, opened 2020-04-15
Inaccessible zones fix - Up button (#769) version 2

Nina-1474:
Relying on CharacterRefresh resetting HeightModifier was a useful catch-all to ensure the position was always corrected, but it did sometimes unexpectedly snap back the position from outside appearance changes (clothes/pose/items/expression) as you noted in #769 .
Instead the position is now reset whenever the user clicks anywhere outside the current character's item zone boxes. This resolves both issues mentioned.

Ideally the up position would not reset for certain actions, such as clicking to open sexual activities or the next page on the inventory list, but I think this way is still better as prevents the position reset happening unexpectedly and ensures it still always resets when required.


Ben987:
Awesome!  I'll test it right now.  Thanks :)

PR #775, opened 2020-04-15
Inaccessible zones fix - Button visibility

Nina-1474:
Fix to #774 comment. Good catch! The button will now be hidden when the player is selected rather than the current character.


Ben987:
Awesome!  Thanks :)

PR #776, opened 2020-04-15
Body part names in the "Select item/activity" text

Nina-1474:
Including the group/zone name in the text at the top right of the character interaction screen. 
E.g. "Select an item to use on the arms." or "Select an activity to use on the mouth."

This was intended to help #774 when a selected zone disappears offscreen, but could be useful to have in its own right.


Ben987:
Sure!  That's a good idea.

PR #777, opened 2020-04-15
Virgin Killer Sweater

Ayesha-678:
Added a Virgin Killer type Sweater


Ben987:
Good idea!  I'll test it right now :)

PR #778, opened 2020-04-16
Cleaned up the virgin killer sweater images

Ayesha-678:
Title explains it all!


Ben987:
Thanks a lot!  These extra pixels can be a pain to track :)

PR #779, opened 2020-04-16
Chain Wrist Elbow Harness Tie

Ayesha-678:
Same thing as the rope, but for chains!


Ben987:
Awesome!  Thanks a lot, I will try it soon.

PR #780, opened 2020-04-16
LARP - Restrain legs item name fix

Nina-1474:
The item name for the RestrainLegs chat message is showing as "N/A" because it's checking ItemLegs instead of ItemFeet.


Ben987:
Good catch!  I missed that one :)

PR #781, opened 2020-04-16
Minor fixes to recent changes

Nina-1474:
#749 & #772 - Correcting '>' to '>=' for the offset variable reset so that there isn't a blank page when the option list is a multiple of 4.
#749 & #772 - Renaming the offset variables to be distinct from each other.
#776 - Applying a 3 line max limit to the item menu's dialog text so it's readable when there are 7 menu icons (e.g. lockable item on leftmost mouth or extended item on arms).
#768 - Adding a NoItemArms prerequisite for the WoodenStocks device since allowing arm bindings leaves the image in mid-air. Adding LegsOpen prerequisite to ensure legs in the correct position. These have been confirmed with Kirsty.


Ben987:
Thanks for all the fixes :)

PR #785, opened 2020-04-18
simple fix

Kitsune-hellfire:
simple fix to prevent the item (Wooden Stocks) being used while kneeling


Ben987:
Good catch!  Thanks.

PR #791, opened 2020-04-21
Some DE translation

Pinogic:


Ben987:
Danke!  I'm sure that German players will like it :)

PR #793, opened 2020-04-22
Club management returns previous restraints

aldarris:
Hello.  I have just received a report in chat rooms, that after club slave contract a generic collar is returned -- unlocked, if the player is in trial.
Fixed to memorize the player's appearance, which should take care of any locked non-restraints as well.


Ben987:
Thanks!  I can't review it now because I'm preparing the next version but I'll get to it when all the hot fixes will be done for that version.  Cheers :)

Ben987:
Hi.  Does it make sense in all situations?  If your owner locked some items on you in your club slave hour, they will get removed.  Is that what most player would want?  Maybe we should do a survey on that one.

Ben987:
Old Pull Request.  These backgrounds are already in the game.

PR #795, opened 2020-04-23
Missing locks

ace-1331:
3 missing locks for the new chain poses


Ben987:
Thanks!  I will prepare a hotfix :)

PR #796, opened 2020-04-23
Bugfix: Prevents the maid from removing owner-locked bindings after a maid job

elliesec:
#### Bug Description
After completing a job for the maid quarters, all character restraints were being removed, even if owner-locked. This PR changes the behaviour so that any owner-locked restraints are not removed after completing a job. This was mainly any issue for owner-locked heels and leashes (it's generally not possible to start maid jobs with other types of owner-locked restraint).

#### Notes
To avoid duplicating too much code, I've modified the `CharacterRelease` function to take a second optional `Predicate` argument, which is a function used to determine whether or not a given restraint should be removed. This argument defaults to `() => true`, so calling the function without the `Predicate` argument preserves the original functionality. I appreciate that this is a pretty commonly-called function, so if you'd rather not change it, I can revert that and duplicate the original function into the new `CharacterReleaseNonOwner` function.

### DIFF
--- a/BondageClub/Screens/Room/MaidQuarters/MaidQuarters.js
+++ b/BondageClub/Screens/Room/MaidQuarters/MaidQuarters.js
@@ -99,7 +99,7 @@ function MaidQuartersWearMaidUniform() {
 
 // When the player removes the maid uniform and dresses back
 function MaidQuartersRemoveMaidUniform() {
-	CharacterRelease(Player);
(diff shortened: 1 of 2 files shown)


Ben987:
Actually, the Maid Quarters should not remove any locks, since it doesn't apply any.  Check the last commit I've did, CharacterReleaseNoLock() will probably be used elsewhere.  Thanks!

PR #797, opened 2020-04-26
Assets fixes

Atasly:
- reworked priority for some items (aprons on top of itemarms, itemnipples)
- chained nipples piercings now have a override priority to be drawn under the suit when not chained
- reworked suit interactibility (belts can be added on top of a suit now, itemvulva and itembutt are now visible)

haven't tested all combinations of clothing, so there might still be some visual bugs, but it should improve the overall experience


Ben987:

Ben987:
This is huge, I won't be able to test all of it.  Good thing there will be a beta for the next version.  Thanks a lot, cheers :)

PR #798, opened 2020-04-27
Open Back Belts

Nythaleath:
Added Open Back variants of the Metal, polished studded and ornate chastity belts


Ben987:

Ben987:
Very nice.  Maybe someday we could code options on the regular belts to open the butt section.  But for now these will do just fine.  Thanks a lot :)

PR #799, opened 2020-04-27
Id card

Ayesha-678:
An ID card necklace


Ben987:

Ben987:
That's a good idea.  Should we make it a "nurse" specific item?  Is that why you created it?

Ayesha-678:
Not really. I just made it for any kinda roleplay that benefits from it, so I’d prefer if it wasn’t nurse specific.

Ben987:
Alright!  Lets use the ID Card as a regular item then, thanks!

PR #800, opened 2020-04-28
New Character pose fix

Nina-1474:
The default character in the New Character screen was mimicking the pose of the last shown sample character in the welcome screen, despite not wearing the items. Setting it to the default pose.


Ben987:
Good catch!  I will prepare a hotfix.

PR #801, opened 2020-04-28
Stuffing gags in all mouth slots

Nina-1474:
Added the ability to use stuffing gags (cloth/panties/socks) in the 2nd and 3rd mouth slots, provided that all currently worn gags are "open". I.e. there is either no gag, other stuffing gags, or gags with holes (ring/lip/spider gags).


Ben987:
What's the advantage of having stuffing in the other mouth slots?  Isn't stuffing always supposed to go in the first layer so we can use other gags over it?  Thanks.

Nina-1474:
It's to allow more than one object to be stuffed in at the same time, like panties from different people or both panties and socks.
I don't expect the 3rd layer to be used much but it was little effort to add it. Could still have some scenario such as a dom pushing her panties in through a ring gag. I could add only the 2nd layer if that's preferable.
I do hope to create a multiple-stuffing item in future where the item options screen allows the number of them to be selected. But that still wouldn't cover the option to mix stuffing types.

Ben987:
@FeuertinKira , is it ok for you if we have stuffings in multiple gag slots?  Do you see any problem that could be caused by that?  Thanks!

FeuertinKira:
It would make more sense to keep the stuffing to the first slot only as it was, and instead just combine all the stuffings into a single item. Then you can add through an inventory script all the existing versions plus new ones. Including multiple stuffings. (This should then also include new images, so using two panties as stuffing can look distinctly different).

Imho, adding the existing stuffings to all slots is a pointless effort, especially if it just looks the same in every slot using the same image and you would see literally no difference between 1 or 2 panties stuffed in the mouth. 

I was already a bit disappointed whoever added the sock stuffing didn't make any efforts to actually add a more unique visual, it was just panties stuffing with a name change... and wasting a new item on that for whatever reason.

Ben987:
@Nina-1474 , Check Kira's comment.  Since we cannot see them, what are the advantages of having multiple stuffings?  It's only for roleplay?  It will triple the assets each time a new stuffing will be added.  Thanks.

Nina-1474:
Yes the main motivation was for roleplay. I'd envisioned it sitting alongside a single-item-multiple-option rather than replacing it. However the only real benefit, assuming that other change is made, would be mixing stuffing types which probably isn't worth it. 
I'll close this off so. Thanks!

PR #802, opened 2020-04-28
Up Button Fix

Nina-1474:
Applying an item to a kneeling/hogtied/etc player then clicking the Up button left the player hovering in the chat room screen. The position was only being reset when clicking out of the menu, but will now be reset when exiting due to item progress completing as well.


Ben987:
Thanks!  I will prepare a hot fix right now :)

PR #803, opened 2020-04-29
Limited items permission

Atasly:
Requires Server PR [#29](https://github.com/Ben987/Bondage-Club-Server/pull/29)

Added Limited option in item permissions (it cycles between normal (green), block (red) and limited (orange))
it will make the item usable only by you, your owner/lover, and your whitelist if your itempermissions are set lower than 3
This is done to allow some items only with certain people, because uncomfortable with randoms, or only trusting a few. the traditional block items still work well to complete with that.
had to make the whitelist of other characters visible or the whole operation would have been done server side.

known issues : 
- refreshing blockitems/limiteditems for other players is not directly done (requires to exit/enter the room
- blockitems/limiteditems are only blocked through interface : console can bypass it (requires the block/limited feature to be written server side)


Ben987:
Nice!  But this one will be trickier to test.  I'll try to merge the others before I attack this one.  We might also need some kind of survey with players, I'll think about it.  Thanks!

Atasly:
well as for a survey I don't really see why since it only offers more options and subtlety in allowing items, like only allowing certain items to be used by owners/lovers.
already tested it, but the way to do it is pretty easy : two characters, one with her settings open and one test item limited. test. accepting the other character as a lover/owner. test. switching permissions to closed (whitelist only). test. breaking off the relationship. test. 
(might require to exit rooms to update LimitedList because of the first known issue, but that's also a problem for blocked items, didn't try to fix it)

Ben987:
When I've changed the item sorting, I've removed the DialogInventoryBlocked variable and used a SortOrder to sort the array.  Can you review the code from your pull request?  It's trying to push DialogInventoryBlocked back into the code.

Thanks a lot!

Atasly:
fixed

Ben987:
It seems the latest changes is fine with players, I will test it really soon.  Thanks a lot :)

PR #804, opened 2020-04-29
Combination lock

Atasly:
Adds a combination lock with 4 digits
Unlockable by Lovers/Owners
Ability to unlock or change its code if you enter the current one.
Lovers/Owners and the club member's number on the lock can see its current code
Combination number is validated to be only 4 digits long to avoid griefers

Useful for :
- RP (accepting someone as a temporary owner, trying to guess the number, or do some favours to learn a digit at the time)
- selfbondage (to secure some chastity on self)
- many other situations (two people can choose a special code that the wearer will not know and tease her)

Known issues :


Ben987:
This is great, I've been wanting to do such a lock for a while now.  I will test it soon.  Thanks a lot!

PR #805, opened 2020-04-29
Strip single cloth button adjustement

Atasly:
The strip a single cloth button only appears if a piece of clothing is there, not appears otherwise


Ben987:
Good idea!  I'll prepare a hot fix for it.  Thanks!

PR #806, opened 2020-04-29
Vacbed

Kitsune-hellfire:
with Nythaleath's help, we put together a vacbed for the club

if added, please credit it to Kirsty amd Nythaleath

Co-Authored-By: Nythaleath <nythaleath@users.noreply.github.com>


Nythaleath:
also just gonna mention while adding this item we made ItemDevices have a parent group of the upper body, all assets have been duplicated and renamed for all sizes and should work normally

Kitsune-hellfire:
just made a few adjustments, made the vacbed's more central on the characters and added a alpha to stop tails and wings from clipping outside of the item

Ben987:
Wow, this is huge!  Have you tested all the possible combos?  There's lots of stuff in here.

Kitsune-hellfire:
okay, everything i can think of has been checked and works, even added a few more prereqs to the vacbed to make it work with out seemingly being weird, such as being box tied then put inside the vacbed, then bed itself showing the yoked pose

Ben987:
Excellent!  I'll test it really soon.  Thanks a lot.

PR #807, opened 2020-04-30
Maid Soft Lock

ace-1331:
this is an old bug, but now you should not be able to soft lock someone when sending them to by a drink as all gags are removed.


Ben987:
Nice catch!  That's an old one.  I will prepare a hot fix :)

PR #810, opened 2020-04-30
Spreader Dildo Bar

Nythaleath:
added a spreader bar with dildo attachment item, the spreader dildo bar.
Also added a vibrating variant of it


Ben987:
Did you know that you can set the Left and Top property directly on the item?  Check the Gown2Skirt item for an example.  You should only set that property for your item and not change the other items.

Keeping images as small as possible helps use with file size and is especially useful when we need to colorize them, it's faster in-game.

Can you review your pull request?  Thanks.

Nythaleath:
Fixed it now, thanks for letting me know i was not aware i could do that

Ben987:
Awesome!  I will test it pretty soon.  Thanks a lot :)

PR #812, opened 2020-04-30
Management chastity fixes

ace-1331:
Made it so the mistress no longer removes owner locked chastity belts/bras
Forgot to remove head items when fixing the maid soft lock

Fixed chastity release: it will now properly check which slot is chaste and offer to free it (Piercings and butt lock counted as chastity without actually getting removed, causing the infinite 25$ loop.) I assume this happened as items got added without reworking this part. It may now be a little pricey to be completely released. Let me know if you think the price should be lowered. I was thinking 15$ instead of 25$ could be decent


Ben987:
Great!  I'll prepare a hotfix, thanks a lot :)

PR #816, opened 2020-05-04
Merging Open and Closed Back Belts

Nythaleath:
Made the Open and closed back belts one item that is toggleable, not able to change it while locked


Ben987:
Since the scripts are similar, is there a way we could reuse the item code instead of duplicating it?  Like having one function that would handle the clicks of all belts.  Another one that would run all belts, etc.

That way if we change or fix something in the code, it would change it for every belts.

Thanks a lot.

Nythaleath:
I dont see how I could do that, so maybe its possible but is beyond my capabilities sorry

Nythaleath:
There the ornate, polished and studded belts now just call the Metal chastity belt's functions

Ben987:
Awesome!  It will be easier to fix bugs in 1 spot instead of 4 in the future, thanks a lot :)

PR #818, opened 2020-05-06
Image fixes

Atasly:
- Adds missing preview pictures (not sure if can be encountered without the use of console but anyway)
- Corrects some wrong file names


Ben987:
Thanks a lot!  Cheers :)

PR #819, opened 2020-05-06
insertion sort for appearance

Atasly:
rewrote the appearance sort function by using insertion sort
that will allow to have more than 50 layers of priority and finally get a proper layer system without any conflict if someone wants to work on it


Nina-1474:
Hey Jenny, this appears to have dropped the check of Property.OverridePriority. Was that intentional?

Atasly:
fixed it, was an error from me.

Ben987:
Thanks, I've been wanting to do something like that for a while.

Can you use "var" instead of "let"?   We don't use let anywhere in the code right now.  Thanks

Atasly:
done

Ben987:
Awesome!  This will allow us to go beyond 50 for priorities.  Lets hope the numbers won't get too crazy though.  Cheers!

PR #820, opened 2020-05-06
Eslint proof of concept

elliesec:
## Description

This introduces [ESLint](https://eslint.org/) with the [eslint-plugin-sort-keys-custom](https://github.com/elliesec/eslint-plugin-sort-keys-custom) plugin in order to allow keys to be automatically sorted with a custom key order.

I've added a simple `package.json` so that ESLint and other dependencies can be installed via npm.

In order to use the linter:

1. Ensure Node and npm are installed (they can be installed [here](https://nodejs.org/en/))
2. Run `npm install`

You can then either:

- Run `npm run lint:fix` to run the linter and automatically fix any fixable linting issues, or:
- Run `npm run lint` to run the linter, which will list linting issues, but will *not* fix them

## Technical Details

Note that changes to `Female3DCG.js` in this PR are entirely automated and purely for illustration purposes.

They key part of this change that allows customisable object key ordering is the `.eslintrc.js` file. The important parts are:

```
rules: {}, // No global rules configured
overrides: [
    {
        'files': ['Female3DCG.js'],
        'rules': {
            // Sort object keys with "Name", "Value", "Difficulty" and "Time" keys first, in that order
            // 'none' denotes the order in which other keys should be sorted
            'sort-keys-custom/sort-keys-custom': ['warn', ['Name', 'Value', 'Difficulty', 'Time'], 'none'],
        },
    },
],
```
* I have not configured the linter with any global rules, but rules can be configured from the [full list of ESLint rules](https://eslint.org/docs/rules/) if other rules or fixers are wanted in the future.
* I have provided a single rule override for the file `Female3DCG.js`, which sorts object keys in that file. Breaking down the config array (`['warn', ['Name', 'Value', 'Difficulty', 'Time'], 'none']`):
    * `'warn'` tells the linter to warn about keys that aren't in the correct order. This could also be set to 'off' or 'error', as per standard ESLint functionality
    * `['Name', 'Value', 'Difficulty', 'Time']` - this tells the lint plugin that the `Name` property should always come first, if it exists, followed by the `Value`, `Difficulty`, then `Time` properties. This list can be extended to have as many or few property names as desired.
    * `'none'` - this determines the order in which any keys *not* in the above array should be sorted. It can be set to:
        * `'none'` - remaining keys will not be sorted
        * `'asc'` - remaining keys will be sorted in ascending alphabetical order
        * `'desc'` - remaining keys will be sorted in descending alphabetical order

This PR is mainly just to prove the concept, and I expect the key ordering will need to be different - I've just taken the example four keys that came up in the discord discussion.


Ben987:
Is there a way we could build an easy tool from that?  Like an executable file?

Many of our contributors will be lost trying to configure node.js

Thanks

elliesec:
It's not something I've ever done before, but something like [pkg](https://www.npmjs.com/package/pkg) might do the job. Might take some time to figure out, but I'll have a look.

Is this something that could just be used by those who know how in the meantime? I imagine if the file is in a consistent format already, less code-savvy contributors would probably be fairly likely to copy/paste the format of existing assets, and the script could just be run periodically to keep things in line.

Ben987:
Yes, it doesn't need to be run all the time and by everyone.   Could you move that tool to a sub folder in the BondageClub project?  It doesn't need to be in the root directory.

We can check with the gang which properties they want to see first.

Thanks!

elliesec:
I've moved everything into the BondageClub directory now, and checked that it all still works as before.

BondageClub/.eslintrc.js
```diff
@@ -0,0 +1,47 @@
...
+    ],
+    rules: {},
+    overrides: [
+        {
+            files: ['Female3DCG.js'],
+            rules: {
+                'sort-keys-custom/sort-keys-custom': [
+                    'warn', [
+                        'Name',
+                        'Group',
```
elliesec, line 20: Note that I added `Group` into the list here, because the linter applies the sorting to objects at all levels - this keeps the `Group` property at the top of each asset group object.

Ben987:
This is great, I'll try to run as many tests as I can, but the beta version will be the real test for it.  Thanks a lot!

PR #821, opened 2020-05-07
Prison Police fixes

Nina-1474:
Fixing incorrect function references which froze the game when they were hit.


Ben987:
Thanks a lot!  I'm sure git4nick will appreciate :)

PR #824, opened 2020-05-12
Handheld Toys

Nina-1474:
If accepted please credit this to both me and Ayesha-678.

Many new 'non-spanking' items have been added into the Spanking Toys item slot, which can similarly be used all over the target's body. The overall item has been renamed to "Handheld Toys".
These additions are: Feather, Feather Duster, Ice Cube, Wartenberg Wheel, Vibrating Wand, Small Vibrating Wand, Candle Wax, Large Dildo, Pet Toy, Vibrator, Belt, Hairbrush, Small Dildo

This includes general enhancements to the toys item:
- Allowing permissions to be set for individual toys that can be used by other players
- Hand icon in the preview box to distinguish the item from regular applied items
- Individual toys can be restricted to appropriate groups/zones
- Swapping the 2-line chat messages (from general item application and arousal activity) with one combined message
- Displaying 6 options per page in the extended menu
- Removing 199 unnecessary duplicate preview images from non-ItemHand groups

This also includes some fixes for existing bugs for the item:
- Prevent the wrong item appearing when holding one handheld item and adding another into someone else's hand
- Displaying the correct name in the preview box
- Fixing slight misalignment on old images

Previous existing errors that haven't been fixed:
- Adding or removing another's toy while already holding one will still mention the wrong item, but in the chat message only


Ben987:
This is HUGE!  Thanks a lot.  Have you tested all the possible scenarios?

Also, you're changing ActivityArousalItem quite a lot.  The previous code was validating that the effect gets calculated on each player side, to prevent "arousal hi-jacking".  With this new code, will it be possible to make someone orgasm by hacking the target arousal settings in the console?  Thanks.

Nina-1474:
Yes I've tested the all the new individual items on every zone and the overall item in as many different ways as I could think of. Plus checking normal items and normal arousal actions.

I hadn't realised the arousal sync was restricted to the player's arousal only, and it turns out the items weren't actually updating arousal on the target's side. I've redone it so that if a used item has an arousal activity, it's looked up and applied in the Action section of ChatRoomMessage, instead of having to publish both an Action and Activity message. I retested the new items and items in general and it all seems good.

I've undone the addition of the toys to the two piercings zones to avoid confusion in the arousal preferences screen. Idea: The ItemVulvaPiercings zone could possibly be 'rebranded' as Clit since that's effectively how it's being used.

The audio for the toys has a fixup, including not sounding when only equipping a toy.

Ben987:
This is really impressive, and have you also tested it with NPCs in the private room?  Everything works well with them also?  Thanks.

Nina-1474:
The spanking toys have actually never fully worked on NPCs, as the expected "Asset Group + Asset Name" lines in the dialog files were always missing. Trying to use a toy on an NPC does nothing visible, but at least doesn't return any error and still allows another item to be selected instead.
To fix this properly every possible interaction would need to be accounted for, but given the number of different toys, the different zones they apply to and the number of different NPCs, that would mean 6552 new lines of dialog would be needed. Not sure if that would be worth it right now!
The arousal effects and equipping the items to an NPC's hand work as expected.

Ben987:
No need to create dialog for all situations, but do the arousal of NPCs raise when you use items on them?  Have you tested all items on them?  Are they able to get an orgasm from the masturbation items?  Thanks!

Ayesha-678:
Can confirm, Just tested, and orgasms can be achieved with masturbation items, and arousal is affected by all items.

Ben987:
Perfect, I'll need to setup a second beta since this is such a big change.  Thanks you two :)

PR #827, opened 2020-05-12
Up button fix

Nina-1474:
Fix for the Up button which still displayed feet for the all-fours pose, and wasn't blocking devices as intended for the hogtied & all-fours poses.


Ben987:
Good catch!  Thanks a lot.

PR #832, opened 2020-05-13
Item Type Change Fix

Nina-1474:
An error occured whenever a player was wearing an item that has different type options, selected another character to open the character viewing screen, clicked on themselves on the left, and changed the type of their item. The item change would appear on the player's screen but not anyone else's, because the code was referencing the other character instead of the one actually affected.


Ben987:
Awesome catch, that was a tricky one.  Thanks a lot.

PR #833, opened 2020-05-14
Strip Button Layers

Nina-1474:
The 'Strip naked' button in the character appearance screen has been enhanced to strip the character in "layers" with each click. The first click removes outer clothes, the second click removes underwear, the third click removes clothes worn as body parts in cosplay (tails, wings, ears hair-accessories), leaving the character fully naked.
This was mainly motivated by the notable number of players who have strongly-worded insistence in their profiles to never remove their tails and ears, but does also add the benefit of quick stripping down to underwear. 
The existing full-nudity functionality is still preserved as clicking the button three times (if necessary) will remove exactly the same assets as before. This change won't affect nudity conditions anywhere else.


Ben987:
This is a good idea, I've been wanting to do something like that for a while.  I'm not sure about the term "BodyCosplay" though, maybe we can find a better name for that property later.  Thanks a lot!

PR #834, opened 2020-05-14
Multiple Lovers

Atasly:
Requires Server PR #[30](https://github.com/Ben987/Bondage-Club-Server/pull/30)

- Allows dating/engaging and marrying up to 5 players/NPCs
- Breaking with each relationship is done through club management
- Added the ability for lovers to know each other's room name (similar to owners for subs)
- Lovers locks can be unlocked by every lovers
- Character Information screen is now on two parts, one for skills/reputation, one for relationships

Have not tested private lovers except getting in a relationship and breaking out of it.
Tested everything else that I could think of.

Known issues :


Ben987:
Great, I'm sure that players will love that one.  Can you prevent the crash with some validations?  We don't want to leave any scenario that can crash.

Also, can you fully test with NPCs?  Make sure its possible to get multiple NPC lovers and a combo of both players and NPCs.

As we've discussed, I'll have to check with the community to make sure they are fine with having multiple wives, since it's not a situation that can happen in real-life.  I'm not sure how a player would react when they find out that their wife married another girl while still being married to that player.

Thanks a lot!

Atasly:
- fixed many breaking issues with the .csv
- tested with two npc lovers (with the command PrivateStartGirlfriend()) and two players lovers and met no problem
- also must have fixed the crash while doing that cause it didn't happen again
- reduced the calls on the server for building the dialog options (they will only be called if needed, should GREATLY help any potential strain coming from this feature)

Ben987:
I'll review the simpler pull requests first and get back to that one later.  Thanks a lot!

PR #835, opened 2020-05-15
Fixes: spreaders and skirt

ace-1331:
- Fixed the skirt and wedding dresses displaying the spreaders when they shouldn't (as there or no other workarounds for this yet)
- Fixed the new dildo spreader bars so the prerequisite for legs being opened would take them into account (fixes all the odd combos like with the petsuit, straitdress etc.)

Remaining anomalies: the sybian, one bar and saddle can be used with the dildo spreader bars since the spreader bar has always been accepted by those. Could had a requirement or leave it be as it doesn't look too crazy. I left that for the creator of the items


Ben987:
That's a good bug catch, thanks a lot.

PR #836, opened 2020-05-15
2 necklaces

ace-1331:
- Added a lock and a key necklace as often suggested. 

I made it so the pendant item can't be colorized like the ID Card, I felt like it look better when only the chain could be colorized, but it can be easily changed later if people prefer one big blob of color.

On that note, couldn't we remove ParentGroup from necklaces to prevent having to add ignore parent group for each necklace? I assume it was either an oversight, or left there in case a future necklace requires different assets for each body size. Left it as is since it isn't really a problem and there are only a few for now


Ayesha-678:
The parent group was added In case a necklace needed body size assets, so far it's worked out fine, so maybe I'll remove it in the future.

ace-1331:
Yeah, it's not a big deal, might as well leave it

Ben987:
This looks really nice.  Thanks a lot :)

PR #837, opened 2020-05-15
Wired Vibrating Egg

Ayesha-678:
Added a wired vibrating egg.

The remote is strapped to the thigh, causing sizes to be needed, so the ItemVulva section now has a parent group, and all items have been duped with the size images.


Ben987:
Alright!  This adds many new assets but the parent group will be probably be needed elsewhere.

Thanks a lot.

PR #838, opened 2020-05-16
Fixes

Atasly:
- small adjustements to reverse bunny suit to avoid overlap with top
- fixed small oddity with top of suits small, cuffs behind back
- using escape key will use the "exit" function of the item, if it exists, fixes the combination lock input field not being removed if pressing escape key
- overridePriority is now properly recognized when sorting Appearance
- fixed bunnyAccessories not using the correct images


Ben987:
Very nice!  Thanks a lot.

PR #839, opened 2020-05-16
more fixes

Atasly:
- fixed combination lock input fields not being removed when clicking on a zone to interact with


Ben987:
Great!  I'll include it in the second beta.  Thanks.

PR #841, opened 2020-05-17
added variants heart eyes based on player's eyes color

Atasly:
- added variants for Heart and LewdHeart expressions based on player's eyes color
- rename the previous Heart and LewdHeart Expression to HeartPink and LewdHeartPink


Ben987:
Thanks!  This is pretty nice.  Sorry for the delay, I needed to finish all the hot fixes.

PR #844, opened 2020-05-18
small fix

Kitsune-hellfire:
added the all fours pose to the bunny cuffs, making the correct image show


Ben987:
Sweet, thanks a lot.

PR #846, opened 2020-05-18
Locking shoes 

Ayesha-678:
3 Types of locking shoes, may seem useless, but some people are very skilled with their feet~

Added:
Locking Strap Shoes
Locking Full Shoes 
Locking Boots


Ben987:
I'm not sure how useful it will be, but it could be fun :)

I'll make the items free though, since these shoes are free in the main game.  Thanks!

PR #847, opened 2020-05-18
Toe Tie Colour Correction

Nina-1474:
Darkening the toe-tie's colour tone to be closer to all the other hemp ropes, as at the moment it's much brighter when given the same hex code.


Ben987:
True, the color was a little different.  Thanks for the fix.

PR #849, opened 2020-05-18
Missing ItemVulva Images

Ayesha-678:
Added in the missing images for the LegsClosed posed in ItemVulva


Ben987:
This creates a lots of duplicated assets, I'll try to mess with the properties to avoid that.

PR #850, opened 2020-05-18
Leather Armbinder Straps Colouring

Nina-1474:
Colouring the extra straps in ItemHidden the same as the leather armbinder, checking whether it's been left at the default.


Ben987:
Good idea!  I'll add it for the next version.

PR #852, opened 2020-05-19
XL Ballgag Layers

Ayesha-678:
Split the  XL harness ballgag into layers, for better looking coloring on the ball.


Ben987:
Great idea!  Thanks.

PR #853, opened 2020-05-19
Item Vulva Optimization

Ayesha-678:
Added the "ignoreparentgroup: true" function to all of the vulva items that don't need any size images, and removed all of the duped assets, Hopefully this makes it a lot neater.


Ben987:
Absolutely, this will remove lots of duplicated assets.

PR #854, opened 2020-05-19
Fix: inventory screen crash

ace-1331:
Load function was incompatible with locks. Should prevent all the reported menu crash/locks disappearing. (tested and it works fine with locks)


Ben987:
Thanks!  I'll prepare a hotfix.

PR #855, opened 2020-05-19
Fix: Missing assets

ace-1331:
Fixed the missing empty images for the new locking heels in the kneeling position (tested with all 4 sizes and lock)


Ben987:
Thanks!  I will prepare a hotfix.

PR #856, opened 2020-05-20
translucent issue

Kitsune-hellfire:
noted but not caught till after the beta finished and the public version updated
image has been made opaque to match the other 3 sizes now


Ben987:
Great!  I'll prepare a hotfix.  Thanks a lot :)

PR #858, opened 2020-05-20
Fix: Box & Bench

ace-1331:
- removed the yoke pose from the open head box to prevent conflicts
- added CanKneel to both small boxes to prevent odd combinations (like with the latex lockdown suit)
- removed unnecessary assets
- fixed a bug with the display message of the belt strap when they could be used and what I supposed was an inverted color for the button (now it displays the appropriate message at the right time and the icon is grayed out when you cannot apply belts while remaining white when it works to match the other areas of the club)

Edit: slipped in a bug fix for the bench I found while testing


Ben987:
Great!  I will test it and prepare a hotfix for it.  Thanks!

PR #859, opened 2020-05-20
Blacklisting removes items placed by target

ADRCorp:
This change will cause any item which was last interacted by the subject of your blacklisting to be removed. That means "slow reaction" to trolls still works as a means to be saved from them, since the items places before you could blacklist them get removed too.


Ben987:
I fear this could create a few bugs.  There are lots of ways items are updated for a player, not only when a lock is added.  And there could be lots of combos leading to strange results in a chat room.  Lets not use that one for now, maybe in the future if trolls become a huge problem, but I would rather have a "safeword" button then.  Thanks!

PR #860, opened 2020-05-20
Maid Combination Lock Removal

Nina-1474:
Certain assets which are lockable but are not restraints can effectively be permanently locked on right now using the combination lock. Management can remove some of these but not all of them. For example corsets and collars.
The main hall maid will not remove these and if they're locked by a combination lock by someone who won't share the code, they player is effectively stuck locked into the item unless they use the console.
With this change the maid will remove remaining combination locks after untying the player.


Ben987:
That should work!  I'll test it soon and prepare a hotfix if everything works.  Thanks a lot.

PR #861, opened 2020-05-21
New items: 2x Gas mask

ace-1331:
- added a full head gas mask (as a hood)
- added a respirator mask in all three gag slots

The layer for this one is unusual for a gag because it needs to be rendered above all head items and accessories to look good. This is why I pushed up existing items/groups of layer 48 and 49, causing no conflict in the process since those layers are not crowded. I've tested these with collars, hoods, other gags, hats, all hair styles, accessories, glasses, vacbed, emotes, etc. and the results are positive, everything fits surprisingly well. I also re-tested existing items to be sure the top layers being move did not change anything.

This layer will be useful for other gags or eventually masks that need to be above any head item.

Thanks to Jilly for the assets, credit goes to them for the images

Let me know if I missed anything, but I tested it vigorously so I'm hopeful I did not ^^


Ben987:
Thanks a lot!  They look great :)

PR #865, opened 2020-05-22
New Front Hair alts (ALL CREDIT TO AKITA AND GATETREK ON DISCORD)

Ayesha-678:
New front hairs, redrawn from the original hairs.

ALL CREDIT GOES TO AKITA ON THE DISCORD SERVER! THEY DREW THE NEW HAIRS, I JUST ADDED THEM IN WITH PERMISSION! SO PLEASE DON'T CREDIT ME!
Also credit gatetrek too for testing!


ace-1331:
Think they said to credit gatetrek too for helping with testing  ^

Ben987:
Since they look similar to what we already have, should we remove the previous models?  To prevent cycling in 28 styles.  Or maybe we could do a survey on that.  Thanks.

Ayesha-678:
Hmm, I do agree they look similar, Maybe do a survey?

ace-1331:
many already said they wanted to keep the old ones in discord, I think it would only be fair to remove thr few that are almost identical.. some are a bit too different. Overall, the quality is way higher though

Ben987:
Thanks everyone!  This looks great.

PR #866, opened 2020-05-22
Bugfix: Prevents vertical overflow when rendering the suspension pose

elliesec:
When rendering the suspension pose in a chatroom containing more than 5 people, if the character is on the bottom row of the chatroom, their feet will "overflow" and be rendered above portions of the top row.

This change crops the image to prevent this overflow. Visually, it looks like the image below. 

Tested in single player & multiplayer with rope/chain suspension, as well as the water torture cell. Also tested that the cropping only affects the "Suspension" pose - no other poses are rendered differently.

this image


Ben987:
Good catch!  Thanks a lot :)

PR #867, opened 2020-05-23
Fix: Arm chains pose

ace-1331:
fixed a missing pose which prevented one of the alternative chain style from being rendered


Ben987:
Nice!  That's a good bug catch.

PR #868, opened 2020-05-23
More rope dialog

Ayesha-678:
Added more dialog for the other rope ties in the shibari room, private room, and introduction room.


Ben987:
Very nice!  We forgot about this dialog, thanks a lot.

PR #869, opened 2020-05-24
Fix: ItemVulva Assets

ace-1331:
- fixed the ItemVulva asset section (It had unused assets, the wand from the full lockdown suit was missing, and the Kneel folder had not been optimized) 

I tested all items while kneeling/standing and every lock or item rendered properly

Edit: on that note, is it intended that the button to select the color still works but is invisible? Either the icon was removed by accident, or the click handler was kept. Either way, that button is useful to display the color preview when the color picker jumps while inputting a code


ace-1331:
Changed it so this PR only contains the ItemVulva Asset fixes
So all the missing kneel images, wand images because of bad names

Ben987:
Are you sure about that one?  These items were changed recently by @Ayesha-678 .

Please give me an example of an item asset that doesn't load properly so I can test it.

ace-1331:
Full lockdown suit wand belt is invisible because the image is missing.

All items are missing when kneeling (tape, suction cup, dildos, etc.) 

That last change you mentionned is the reason why, because the Kneeling folder was not changed along the base pose.

Ben987:
Thanks, I will test it right now.  Cheers :)

PR #871, opened 2020-05-25
Fix: appearance text menu

ace-1331:
Prevents the text from "sticking" making it so the appearance always keeps the first player's name when you inspect their appearance

### DIFF
--- a/BondageClub/Screens/Character/Appearance/Appearance.js
+++ b/BondageClub/Screens/Character/Appearance/Appearance.js
@@ -444,8 +444,10 @@ function AppearanceRun() {
 	// Draw the background and the character twice
 	var C = CharacterAppearanceSelection;
 	var HideColorPicker = true;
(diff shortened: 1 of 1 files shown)


Ben987:
What if that error triggers?  

				// If we didn't found the group, we warn the user
				if (!Found) {
					CharacterAppearanceHeaderText = TextGet("MustPickItem") + " " + AssetGroup[A].Name;
					return;
				}

It will never show.

Shouldn't we set CharacterAppearanceHeaderText = ""; when we load the screen instead?

ace-1331:
There was a case when it didn't reset, I'll take another look in a bit. 

Does that error show at all, if so, how? I've never seen it.

ace-1331:
My last change makes it so when clicking x or pressing escape to leave the menu, it will reset it, so it will keep the error message as intended

Ben987:
Perfect!  This should work.  Thanks a lot.

PR #872, opened 2020-05-25
Fix: Wand crash

ace-1331:
Fixed a bug where  locks and vibrations on the lockdown suit would collide and make the inventory crash (Like the spreader bar was fixed a couple days ago)


Ben987:
Good catch!  Thanks a lot.

PR #873, opened 2020-05-25
Fix: Locks inventory menu

ace-1331:
- Fixed how the lock menu is displayed so that the unusable locks are clearly labeled in gray like the other items (it used to send the lock at the back, but did not display the color so you could think you can use it but the click never registered)

This was caused because of how the sort order could never reach 3 which is what defines an item as blocked. Test with all types of items, locks, etc. with various permissions and now it all displays ok


Ben987:
Very good, I've never noticed that one before.  Thanks!

PR #874, opened 2020-05-25
Fix for toothbrushes

Ayesha-678:
Fix an error that caused the toothbrush names to not show up,


Ben987:
Great!  Thanks for the quick fix.

PR #875, opened 2020-05-25
Raccoon Tail Blinking Fix

Nina-1474:
The large raccoon tail image was going slightly outside the standard 500 width. This meant that 1) the far right was cut off and 2) when blinking, the edge of the tail appeared out of thin air on the far left due to how GLDraw creates the two canvases side by side. The fix was just to move the image 7 pixels to the left to keep it in bounds.


Ben987:
Great catch!  Thanks a lot.

PR #876, opened 2020-05-25
Kennel Alpha Fix

Nina-1474:
Parts of objects could appear in mid-air at the far right of a character (tails in particular) when inside the kennel (LowCage) as the alpha didn't extend all the way to the right.


Ben987:
Very nice, thanks!

PR #877, opened 2020-05-25
Missing Nipple Weighted Piercings Chain Lock

ace-1331:
Added the blank images for the nipple piercing chain when locked to fix the missing asset errors


Ben987:
Awesome!  Thanks for the fix.

PR #878, opened 2020-05-25
Prison Gagged Fix

Nina-1474:
An issue occured if a player fought the prison guard, was gagged in the fight and then lost. One of the dialog stages had entirely speech-only answers, and since none could be selected the game was softlocked. Now every possible path should have a silent/action option included.


Ben987:
Good catch, that would lead to a dead-end.  Thanks!

PR #879, opened 2020-05-26
Up Button Refresh Fix

Nina-1474:
If viewing one character with the Up button, and other characters refresh during that time, those other characters are also moved up and then left hovering in the chat room (for the viewing player only) until refreshed again.
Now the variable tracks the currently viewed character's number, so that only that character is moved up.


Ben987:
That up button sure is tricky.  Thanks for the fix!

PR #880, opened 2020-05-26
Fix: Wrong csv ids for assets

ace-1331:
After running a script (done by noname) to check which items were missing their description in the CSV, 8 were found and corrected in this PR. All of them had the wrong name/id in the file so they were properly replaced to make sure they were not missing.

### DIFF
--- a/BondageClub/Assets/Female3DCG/Female3DCG.csv
+++ b/BondageClub/Assets/Female3DCG/Female3DCG.csv
@@ -12,7 +12,7 @@ Cloth,ChineseDress2,Chinese Long
 Cloth,TShirt1,T-Shirt
 Cloth,TennisShirt1,Tennis
 Cloth,Sweater1,Sweater
-Cloth,MistressTop1,Mistress Top
+Cloth,MistressTop,Mistress Top
 Cloth,AdultBabyDress1,Puffy Baby Dress
 Cloth,AdultBabyDress2,Bows Baby Dress
 Cloth,AdultBabyDress3,Flower Baby dress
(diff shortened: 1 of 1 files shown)


ace-1331:
For clarity here is the result of the search:
Non-existing items found in the CSV file: [
    'ClothMistressTop1',
    'SuitSeethroughSuitzip',
    'TailStrapsSuccubusStrap2',
    'BodyUpperXlarge',
    'BodyLowerXlarge',
    'ItemButtHorseButtPlug',
    'ItemButtRaccoon Butt Plug',
    'ItemButtVibratingButtPlug',
    'ItemNipplesTapeVibeEggs'
  ],
  Existing items with no CSV counterpart: [
    'ClothMistressTop',
    'SuitSeethroughSuitZip',
    'TailStrapsSuccubusTailStrap',
    'BodyUpperXLarge',
    'BodyLowerXLarge',
    'ItemButtHorsetailPlug',
    'ItemButtRaccoonButtPlug',
    'ItemButtVibratingButtplug',
    'ItemNipplesTapedVibeEggs'
  ]

Ben987:
Are you sure about that one?  I've tested a few of them and the descriptions are showing properly in-game.

ace-1331:
I assume the ones that were just about capital letters showed before, but after fixing them, the ones clearly labeled incorrectly like the spaces, etc. Showed as intended. Just do a quick search and you will see stuff like "MistressTop1" does not exist in Female3DCG.js and "MistressTop" does not exist in the csv, and so on

Ben987:
I've just tested a few items and the description shows properly.  What's the bug in-game?

ace-1331:
The current CSV file requires the exact order to prevent incoherent ids like this. If they are fixed, we can stop ignoring the middle column like it is currently. Which would solve many incompatibilities issues in the future with merges and stuff, only one item would be broken instead of all the ones below it.

This is a quality cleanup more than a "bug fix". With actual consistency the rest can be improved in the future... otherwise this column can just be deleted and it would make no difference.

Ben987:
Alright.  And you've tested the items with that version to make sure it doesn't create any new bugs?

ace-1331:
yes, csv loads and all assets still have the correct description. This will prevent errors the day things are arranged or improved

Ben987:
Excellent, I'll merge it then.

PR #881, opened 2020-05-27
Thrash Spelling Correction

Nina-1474:
Correcting the verb 'trash' to 'thrash'


Ben987:
Very nice, it's an old mistake that nobody saw.  Thanks!

PR #882, opened 2020-05-29
Multiple Lovers - Hotfix

Atasly:
hotfix to be applied to the current client (R57) to support multiple lovers once the server is updated with #[30](https://github.com/Ben987/Bondage-Club-Server/pull/30)
incompatible with the multiple lovers PR #834

Known issues : 
- players with that hotfix and the updated server could ask other players to have multiple lovers and accept (since it's all handled server side) but will not have any impact on their gameplay (except for permissions)


Ben987:
Awesome!  I'll test it soon and prepare a hotfix for the live version if everything works.

PR #883, opened 2020-05-30
Raccoon Tail Blinking Fix V2

Ulith:
Same fix as Nina, but for the forgotten tail plug version,  which has the same overlap.
moved 7 pixels again


Ben987:
Good catch!  Thanks a lot.

PR #885, opened 2020-06-02
Adds a new extended item script for consistency and ease of development

elliesec:
# Description

The motivation behind this PR is to make it simpler to create new extended items (specifically for items and restraints with multiple variations), and to make the extended item selection interface more consistent throughout the game. This has been done by refactoring the common functionality of extended item scripts into a single place, with some capacity for customisation.

This PR:
* Adds a new `ExtendedItem.js` script. This script contains a handful of functions useful for handling drawing and clicks for extended item selection
* Adds a new common function `CommonCallFunctionByName` in `Common.js` - this allows an arbitrary named global function to be called dynamically with any number of parameters. Similar to `CommonDynamicFunctionParams` but this function is intended to be called directly from code, rather than being used to parse a function call string from a CSV file
* Refactors the web arm restraint and web blindfold restraint to use the new extended item functions. 

At the moment, the web restraints are the only items that have been refactored, as I wrote the code for them, but the eventual goal would be to refactor most, if not all extended items to use this framework, which would reduce a lot of code duplication. However, this can be done gradually, on a per-item basis, to reduce PR sizes and ease testing.

# Motivation

While this doesn't change any of the actual gameplay, the benefits of this PR would be:

* Easier and quicker to create extended items in the future - drawing and click handling is all done inside the common functions, so these do not need to be handled. See the size of the newly-refactored web restraint files for an example. This also means less scope for copy/paste errors.
* A more consistent UI for extended items
* Less code (eventually) - due to the removal of the need to duplicate code in each individual item's script. Now the scripts only need to handle item-specific functions (like chatroom message generation)

### DIFF
--- a/BondageClub/Screens/Inventory/ItemHead/WebBlindfold/WebBlindfold.js
+++ b/BondageClub/Screens/Inventory/ItemHead/WebBlindfold/WebBlindfold.js
@@ -27,7 +27,7 @@ function InventoryItemHeadWebBlindfoldDraw() {
 }
 
 function InventoryItemHeadWebBlindfoldClick() {
-	ExtendedItemClick(InventoryItemHeadWebBlindfoldOptions, InventoryItemHeadWebBlindfoldSetType);
+	ExtendedItemClick(InventoryItemHeadWebBlindfoldOptions);
 }
 
 function InventoryItemHeadWebBlindfoldPublishAction(Option) {
--- a/BondageClub/Screens/Inventory/ItemHead/WebBlindfold/WebBlindfold.js
+++ b/BondageClub/Screens/Inventory/ItemHead/WebBlindfold/WebBlindfold.js
@@ -1,6 +1,6 @@
 "use strict";
 
-const InventoryItemHeadWebBlindfoldOptions = [
+var InventoryItemHeadWebBlindfoldOptions = [
 	{
 		Name: "Blindfold",
 		Property: { Type: null, Difficulty: 0 },
@@ -32,8 +32,8 @@ function InventoryItemHeadWebBlindfoldClick() {
 
 function InventoryItemHeadWebBlindfoldPublishAction(Option) {
 	var C = CharacterGetCurrent();
-	const msg = "HeadWebSet" + Option.Name;
-	const Dictionary = [
+	var msg = "HeadWebSet" + Option.Name;
+	var Dictionary = [
 		{ Tag: "SourceCharacter", Text: Player.Name, MemberNumber: Player.MemberNumber },
 		{ Tag: "TargetCharacter", Text: C.Name, MemberNumber: C.MemberNumber },
 	];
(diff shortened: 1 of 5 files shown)


BondageClub/Screens/Inventory/ItemHead/WebBlindfold/WebBlindfold.js
```diff
@@ -17,72 +19,23 @@ const InventoryItemHeadWebBlindfoldOptions = [
...
-
-	for (var I = 0; I < InventoryItemHeadWebBlindfoldOptions.length; I++) {
-		var X = 1175 + I * 425;
-		var Y = 550;
-		var Option = InventoryItemHeadWebBlindfoldOptions[I];
-		if (MouseX >= X && MouseX <= X + 225 && MouseY >= Y && MouseY <= Y + 225 && DialogFocusItem.Property.Type !== Option.Property.Type) {
-			InventoryItemHeadWebBlindfoldSetType(Option);
-		}
-	}
+	ExtendedItemClick(InventoryItemHeadWebBlindfoldOptions, InventoryItemHeadWebBlindfoldSetType);
```
elliesec, line 30: Incorrect variable reference here - need to fix.

BondageClub/Scripts/ExtendedItem.js
```diff
@@ -0,0 +1,230 @@
...
+
+/**
+ * Handles clicks on the extended item type selection screen
+ *
+ * @param {Object[]} Options - An Array of type definitions for each allowed extended type (as defined in ExtendedItemLoad)
+ */
+function ExtendedItemClick(Options) {
+	// Exit button
+	if (MouseX >= 1885 && MouseX <= 1975 && MouseY >= 25 && MouseY <= 110) {
+		DialogFocusItem = null;
```
elliesec, line 76: This branch should return early.

BondageClub/Screens/Inventory/ItemHead/WebBlindfold/WebBlindfold.js
elliesec: Fixed.

BondageClub/Scripts/ExtendedItem.js
elliesec: Fixed.

NoneNoname:
I tried this two times already, never got it accepted, #380 #603

elliesec:
Well, I'll persist with it for now. It's a small enough change, and mostly just a lift-and-shift right now, and the scope of this PR only affects one item that is as-yet unreleased, so there's no danger of breaking any existing items.

Ben987:
Yes, it's a good idea, I've been wanting to do something like for a while now.  I'm not sure how easy it will be to have all items working on a uniform system, but if most items can work it will be great.

Not everything is commented, but we can add the comments later.

Can you change the "const" for "var" ?  We use var almost everywhere right now.

Thanks a lot!

elliesec:
Yes, there are a few items that won't fit with this so well, or it'll need adapting to work with them. But I think we should be able to get to a point where most items are using this. I'll make the `var` changes and add the remaining documentation.

elliesec:
Okay, those changes have been made and I've re-tested, and everything looks to be working as expected.

Ben987:
Awesome, I'm sure I'll be able to use it for some items.  I will test it soon, thanks.

PR #886, opened 2020-06-06
Fix : Owner/Lover Lock on self

Atasly:
when using a lover/owner lock on self, only possible if the player has a lover/owner, it will now use the memberNumber of the player.
to make it clear for owners/lovers when someone unlocked themselves behind their backs and so on


Ben987:
Thanks!  Does it open a new door for hackers to lock players with owner/lover locks even if they are not owner/lover?  Or that door was already opened anyway?

Atasly:
that door was already open, if someone does not have any lover/owner it won't be possible to owner/lover lock them (as before)

Ben987:
Awesome!  It's a good idea.

PR #887, opened 2020-06-08
Bugfix: Allows inventory clicks on the current item to bypass prerequisite checks

elliesec:
# Problem

The web bondage item currently uses the `NotSuspended` prerequisite, but one of the extended variants for it adds the `Suspension` pose. This means that after selecting the web suspension option and clicking on the web item in the inventory menu, the "Remove suspension first" message would be displayed and the extended item screen would not be displayed.

As far as I'm aware, the web item is the first item that has an extended type whose pose configuration conflicts with the prerequisites of the base item, which explains why this issue has not been seen before.

# Solution

This PR adds an additional check which bypasses the prerequisite check if the clicked item is the same as the currently equipped item.

### DIFF
--- a/BondageClub/Scripts/Dialog.js
+++ b/BondageClub/Scripts/Dialog.js
@@ -725,7 +725,7 @@ function DialogItemClick(ClickItem) {
 	// Cannot change item if the previous one is locked or blocked by another group
 	if ((CurrentItem == null) || !InventoryItemHasEffect(CurrentItem, "Lock", true)) {
 		if (!InventoryGroupIsBlocked(C))
-			if (InventoryAllow(C, ClickItem.Asset.Prerequisite))
+			if ((CurrentItem != null && (CurrentItem.Asset.Name == ClickItem.Asset.Name)) || InventoryAllow(C, ClickItem.Asset.Prerequisite))
 				if ((CurrentItem == null) || (CurrentItem.Asset.Name != ClickItem.Asset.Name)) {
 					if (ClickItem.Asset.Wear) {
 


Ben987:
Is that the best solution for this item though?  This will allow us to click on the item but not use the "Remove" button.

Should the item prerequisites be changed or should we change the code to access the "remove" button?

Thanks!

elliesec:
It's a good question, and something I was wondering about. Good point about the remove button. I don't think there's any harm in allowing access to the extended menu click and the remove button like in this (although like you say, it should probably be both at the same time if we went that way). 

The alternative would be to remove the `NotSuspended` prerequisite, but this would mean all of the web variations would need to handle being applied from a potentially already-suspended position, which might be a little fiddly, but probably isn't too difficult.

I think I'm leaning towards this being the right solution (with the addition of allowing the remove button too). I think it's probably a little more flexible in the long run; being able to always click or remove an already-equipped item seems like a reasonable thing to be able to do to me, provided the item isn't blocked. And I feel that prerequisites, as the name suggest, only really need to be checked when equipping the item, not when removing/changing the extended type (and prerequisites for changing types can be handled in the individual item code).

That said, I'm happy to take your lead on this one, so do you have a preference?

Ben987:
Adding some similar code to enable the remove button for that specific situation could work, but it could also cause issues with other items.  Can you try it with many items on your side to make sure we don't create new bugs?  Thanks a lot.

elliesec:
Okay, I'll add that and make sure to give it a thorough testing.

PR #888, opened 2020-06-08
hotfix : lovers stuck

Atasly:
Due to my PR #882, lovers were unable to engage or marry after relogging.
Fixed that bug, but will require two relog (one in the new client, and another once the data has been updated).
The Multiple Lovers Server also fixes the problem.
This PR is intended as a hotfix, it will be useless once [#30](https://github.com/Ben987/Bondage-Club-Server/pull/30) is accepted


Ben987:
Great!  I will prepare another hotfix and warn the players that they might need to use CTRL+F5 and relog twice.  Thanks.

PR #889, opened 2020-06-08
Fix: Infinite refreshes

ace-1331:
The slave collar check would cause infinite refreshes if the collar was invalid (something coming from NPCs) when in chatrooms. This only syncs the slave collar to make sure the whole character does not keep being refreshed back and forth. This way people don't have to break with a NPC if they want to go in a chatroom

Tested with someone who had the "can't kneel or do anything" bug and it fixed it.

### DIFF
--- a/BondageClub/Screens/Character/Login/Login.js
+++ b/BondageClub/Screens/Character/Login/Login.js
@@ -120,7 +120,7 @@ function LoginValidCollar() {
 	}
 	if ((InventoryGet(Player, "ItemNeck") == null) && (Player.Owner != "")) {
 		InventoryWear(Player, "SlaveCollar", "ItemNeck");
-		if (CurrentScreen == "ChatRoom") ChatRoomCharacterUpdate(Player);
+		if (CurrentScreen == "ChatRoom") ChatRoomCharacterItemUpdate(Player, "ItemNeck");
 	}
 }
 


ace-1331:
To add some context, after checking a bit further, I can confirm the infinite loop is not there at all once this fix is applied. (It does not infinitely loop on the neck refresh)

Ben987:
Thanks, but how can I recreate the bug?  I get collared by an NPC and try to enter any chat room?

ace-1331:
I still have no clue how to reproduce it, but about 3-4 people have been reporting this and all had the same pattern (owned by NPC, were locked/tied by the NPC and when trying to go in a chatroom, they can't do anything at all because their client always sends a refresh back infinitely from the slave collar trying to be put on again). I was in a room with one person who had that issue and tested a few thing with them until we figured out that was the only spot this could be coming from. When they tried it with this fixed, the refresh loop was gone. (Regardless of the NPC bug, this infinite loop is dangerous for any case where the slave collar is bugged or messed with)

 I don't use private NPCs so I am not 100% familiar with how they work. I assume this bug has been here since the changes to the refreshes were done since no one mentioned it before the past month. Some thing they do probably removes the collar or does not get synced with the server until they get into a room which causes this when someone else enters

Ben987:
Alright!  I won't be able to test it but the code update makes sense.  We can always apply another fix if that doesn't cover all cases.  Thanks!

PR #890, opened 2020-06-08
Icon Transparency Edit

Nina-1474:
A subtle change that may or may not be wanted. I've changed enclosed spaces in the centre of some icons from transparent to white to make them 'pop' a little when the background is highlighted. Some already did this (e.g. Maid Quarters, LARP battles) but most didn't.


Ben987:
That's alright with me, but can we still see some blue highlight on all of them when you move your mouse over it?  It's still easy to see which button is focused?  Thanks!

Nina-1474:
Yes they all still have noticeable bright blue around the edges. 
Attached is a composite image of what the right-hand hall menu highlighted icons look like. The 'Look For Trouble' near the bottom-left is about the worst-case in terms of reducing the amount that's highlighted.
SelectedNew

Ben987:
Great, it's easy to see which icon is focused.  Thanks a lot!

PR #891, opened 2020-06-09
Owner self lock rule

Atasly:
as requested by some
add possibility for owners to have control over the owner locks of their subs
allowed by default


Ben987:
Sure, it's a funny rule idea, thanks :)

PR #892, opened 2020-06-09
some new ropes works

gatetrek:
added some new ropes works for both legs and feet zones with a little touch up on the mermaid tie to make a little more realistic look in the middle


Ben987:
Very nice!  Should we change the bondage requirement levels to have a progression between ties?  Should it be harder to do the mermaid tie than the new ties?  

Also, should we sort the ties by the required bondage level?

Thanks!

gatetrek:
can do was just going off what was put there before

PR #896, opened 2020-06-10
Changed the behaviour of the strait jackets.

SandrineBC:
- StraitLeotrad now goes under all skirts

- The other StraitJackets now hide all skirts


SandrineBC:
I tested the strait jackets and now the mittens chain things with all clothes I got in my test environment. As to the external link:
If we put the documentation into the main project, we will only add additional traffic to the server, as the documentation is massive. If you don't like it, please remove the comment.

Shiranui-Izayoi:
So I was actually testing this out to see what was best and was going to submit the same changes to the StraitLeotard. I found that 24 works better for it since then it will always go under leg restraints which would be better.

Also damn you for beating me to it. I guess I will just have to do the other items I had planned.

Ben987:
Should we go for priority 24 instead of 25?  Check @Shiranui-Izayoi comment.

Yes, please remove the external link.  Thanks!

Shiranui-Izayoi:
Basically what I found is **sometimes** when you equip two items with the same priority the one put on first will go under the other. So since the only items with a Priority of 24 are the ankle and leg cuffs along with the wand belt which isn't hidden but could be if needed it won't have any major clipping.

SandrineBC:
I see no problem in changing the priority of the Leotard to 23, as there should be no conflicts. But do we wand the wand belt go under the various pants?

Ben987:
From what @Shiranui-Izayoi tested, priority 24 seems ideal.  Can you test it with priority 24 and change your pull request if it works better?  Thanks a lot.  Cheers.

Shiranui-Izayoi:
Admittedly the wand belt should likely have a higher priority but that can be done later. I am currently testing changing some item drawing priorities locally, Mostly I am testing increasing some of the ItemNipple items so that they would be drawn over bras. This means that chains and things won't suddenly disappear under the corsets and under certain bras.

I doubt I will find any issues since most things hide them already when they are covered. I can add the wand belt to the tests. The main thing I can see that might be a issue is skirts since they should be over pants but under skirts. My thought is to have skirts be a higher priority, but again I will test that later.

SandrineBC:
I tested it again. As the leotard blocks the ItemVulva spot, you always have to apply the wand belt before and so the Leotard goes **over** the belt at priority 24.
Thus, I changed it to 23 for the leotard and adjusted the HempRopeWand to 24 as well. So both belts go over panties and the Leotard, stay under all skirts and go over the pants.

Ben987:
Alright!  Thanks you two, I will try it soon.

PR #899, opened 2020-06-11
Change: Remove addon on remove

ace-1331:
Made it so the mask addons do not block the mask itself, allowing you to lock both independently and to remove the mask while wearing an addon to avoid confusion 

We'll see, if there is more confusion during beta I'll think of what else can be done, for now I'm just thinking of a text message on the item menu to say x rectangle contains addons. This change is already much better regardless 

Only reason I did not add a "remove addon" button on the mask itself is because the lock for removing the item would need to be handled there and it's less future-proof and way messier. 

Maybe a button "go to addons" ? It would change the focus to null and display the addon section like if you clicked on it yourself. Let me know what you think.


Ben987:
Great!  Thanks a lot.

PR #900, opened 2020-06-11
Bugfix: Fixes chatroom action text when changing web type

elliesec:
There was a small bug in the way the new extended item script was attempting to retrieve the previously selected type, meaning the web action text would always say "loosens" instead of toggling between "loosens" and "tightens" as appropriate.


Ben987:
Great!  I will test it soon.  Cheers.

PR #901, opened 2020-06-12
Improvement: Forced labor

ace-1331:
Made it so instead of removing items, it will simply not allow the owner to send their sub to the maid quarters.  It will also prevent unusual items with a gagging effect from causing problems in the future. I also think it's an improvement to let the players decide whether or not they should remove the gagging items.

The added check sub-side is to prevent an owner from soft locking their sub through the console by bypassing the check


Ben987:
What was the problem with removing the gag before sending the girl for forced labor?  Did that created a bug with hoods?

Also, there's already a function .CanTalk() on the character to validate if she's gagged on not.

ace-1331:
right, just changed it for ` .CanTalk()`, forgot about it for a second 

Well, there are already the two posture collar gag variants of the slave collar which can gag you. And there were the ItemAddons for the mask until I added ItemOnRemove. 

This is not critical, but I wanted the fix to be both future-proof and transparent about why it does not work. This will prevent  forced labor from being a issue if some day there is a new gagging item in a different slot.

Ben987:
Sure, that's fine for me.  It might prevent some strange combos where the player get stuck.  Thanks!

PR #904, opened 2020-06-13
Bed

Ayesha-678:
Added a bed item along with many addons for said bed

Covers
Bed Ropes
Bed Straps
Bed Tape
Bed Chains

And i made the addons buyable through the store with a dialog option.

Now there is no bed menu to add these addons, they are applicable through the addon catagory at anytime, but only if your on a bed with a new prerequisite (Thanks to Nina!)

This is because I feel it makes more sense to have it like this, cause realistically, the bed wouldn't come with this stuff, and it allows for specific selfbondage levels.

This also opens up the opportunity for more addons like this, and making the addon section a more regular part of the game, and people wouldn't keep forgetting about it.

Hope that's all alright and this all goes by smoothly!!


Ben987:
Thanks a lot, I'm trying to fix all the bugs for the beta version first.  As soon as the beta is released, I review it.  Cheers!

Ayesha-678:
No problem! Thanks!

Ben987:
Very nice!  We have so many crazy items but never had a bed, that was overdue :)

PR #905, opened 2020-06-14
Fix: Layers + AllowTypes

ace-1331:
- Addressed a concern where the AllowTypes property would not behave as intended on Layered items

The check for layers when an item had types was wrong, making it so the base type would never be rendered unless you knew about it and added *""* to types.

This will allow items with various states to also support layers which will make complex items much simpler (Like the gas mask I added a couple days ago... which I might update after this change)

### DIFF
--- a/BondageClub/Screens/Character/Appearance/Appearance.js
+++ b/BondageClub/Screens/Character/Appearance/Appearance.js
@@ -372,7 +372,7 @@ function CharacterAppearanceBuildCanvas(C) {
 				var LayerType = Type;
 				if (CA.Asset.Layer != null) {
 					Layer = "_" + CA.Asset.Layer[L].Name;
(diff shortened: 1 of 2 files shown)


Ben987:
So we'll be able to remove the AllowTypes: [""] from the assets with that change?

ace-1331:
Yes, or, we will be able to make items without wondering why the base pose never renders. If I am not mistaken, there are currently no layered item with types because of this bug. 

I've seen it come up a few times on discord when people tried to make complex items (and it happened to me as well)

ace-1331:
There are 4 items that rely on this (LoveBelt, FullLatexSuit, MermaidSuit, BitchSuit) which work different than any other item, I'm afraid it doesn't help with the confusion though. I will close this PR for now and look at an easy way to make them follow the same pattern. Till then, I recommend adding a comment or note of some sort to help asset makers realize they have to add "" to the allowtypes

PR #906, opened 2020-06-14
Fix: null text

ace-1331:
- Fixed the displayed text for hemp rope feet 

(It said "required level null" instead of "no requirement" like everything else does)


Ben987:
Good catch, I've never noticed that one.  Can you do it for hemp ropes on the lower legs and upper legs?  Both seem to have that same problem.  Thanks.

ace-1331:
Sure, just added the fix for it too

Ben987:
Excellent!  I'll test it soon.  Cheers :)

PR #907, opened 2020-06-14
Ban list size

ace-1331:
Like we spoke It's been a frequently raised concern that the ban list only allows +/- 42 banned players considering the growing list of trolls. This has led many to use console to ban players to bypass this restriction caused by the UI.

I think it is a needed quality of life  change, and it will also be safer for people who are not familiar with the console.


Ben987:
Absolutely.  Thanks :)

PR #908, opened 2020-06-14
Crib and Milk Bottle

mrbakucamp:
Added two items
Crib
Milk Bottle


Ben987:
Thanks a lot, I'm trying to fix all the bugs for the beta version first.  As soon as the beta is released, I review it.  Cheers!

Also, there seems to be a conflict with the latest items that were added in the last week, can you review?  Thanks.

mrbakucamp:
The conflicts look to just be from the rope gag also being added and them wanting to go to the same spot.

PR #910, opened 2020-06-15
New emote behaviour

SandrineBC:
If "/me" or "\*" is followed immediately by a "'" or "," no blank is inserted in front of the emote.
So "*'s tail swishes" will now result in "Sandrine's tail swishes" instead of "Sandrine 's tail swishes"


Ben987:
I'm not sure how well it will work for players that doesn't play in English, but we can accommodate them later.  Thanks :)

PR #912, opened 2020-06-15
Code cleanup and various presentation fixes

SandrineBC:
- Removing duplicate asset properties [#877286](http://www.hostedredmine.com/issues/877286)
- Deleting identical properties in group and asset properties [#877286](http://www.hostedredmine.com/issues/877286)
- Displaying errors with feet items and the wedding dresses [#877236](http://www.hostedredmine.com/issues/877236)
- Breast slot was accessible through vac bed [#877120](http://www.hostedredmine.com/issues/877120)
- Wired vibrating egg could be applied through panties [#877125](http://www.hostedredmine.com/issues/877125)
- Hat bands could be seen on top of polished steel hood
- Usage of hats with the inflating ball hood is blocked, because the presentation looked strange


Ben987:
Are you able to split this in multiple smaller pull requests?  It's much easier for me to try a specific bug first and see if the PR fixes it.  A huge file like that makes it much harder to reject or accept one specific fix.  Thanks a lot.

SandrineBC:
_sighs_
I see your point. I hope I can roll back all my fixes.

PR #914, opened 2020-06-15
Fixes to feet items and wedding dresses

SandrineBC:
- Displaying errors with feet items and the wedding dresses [#877236](http://www.hostedredmine.com/issues/877236)


Ben987:
Awesome!  These are good catches :)

PR #916, opened 2020-06-15
Fixed errors on vac bead and wired egg

SandrineBC:
- Breast slot was accessible through vac bed [#877120](http://www.hostedredmine.com/issues/877120)
- Wired vibrating egg could be applied through panties [#877125](http://www.hostedredmine.com/issues/877125)


Ben987:
Good catches!  Thanks :)

PR #917, opened 2020-06-15
Hat display errors on polished hood and gas mask

SandrineBC:
- Hat bands could be seen on top of polished steel hood
- Usage of hats with the inflating ball hood is blocked, because the presentation looked strange

No error was raised for this, but I found the problems during my code cleanup


Ben987:
Thanks!  I will test it right now :)

PR #920, opened 2020-06-15
Corset Shirt

Ruilov3:
Added Corset Shirt


Ben987:
That looks great!  I'm sure we'll see lots of them in the rooms soon.  Thanks!

PR #923, opened 2020-06-15
Auto-ban on room creation

ace-1331:
Added the ability to auto-ban your blacklist and/or ghostlist when creating a chatroom

This will make it faster for players to create rooms and wont require them to maintain a separate list outside the game.

Ghostlist is on by default but blacklist is optional to make it flexible (many players blacklist players they do not mind having in their room, so each individual can decide how they want it)

This goes with server PR [#34](https://github.com/Ben987/Bondage-Club-Server/pull/34)

Edit: this will allow the ability to maintain a client-side global banlist if it ever comes to that. It will also allow people who use console to send just one request instead of spamming ServerSend with a forEach


ace-1331:
updated to work client-side only

Ben987:
Absolutely, this will save some time when creating rooms.  Thanks!

PR #924, opened 2020-06-16
Lovers fix - clingy Private Lovers

Atasly:
Lover data is now always sent to server, allowing to break properly with Private Lovers
part of the fix for breaking with "clingy" private lovers


Ben987:
Thanks!  I will test it soon.

PR #925, opened 2020-06-16
Solved various problems around neck and butt chain

SandrineBC:
With the introduction of the medium collar chain, several bugs were introduced as well, because the medium chain never made it to prerequisites and effects. This fix does the following:
- Introduction of the "IsChained" effect, when a player is chained to the floor
- collar chains can be applied to hogtied players
- hogties are possible, if players are kneeling
- suspension (normal and hogtied) cannot happen anymore, if a player is chained to the floor
- chains cannot be used, if a player is suspended

This holds for the collar chains, as well as the chains attaching to the locking butt plug and works for
- hemp rope
- chains
- the new webbing


Nina-1474:
Hey Sandrine! Like the changes but there's one issue I see:
The ItemArms HempRope & Chains changes removed the prerequisite checks for everything other than SuspensionHogtied, which reintroduces errors that they were preventing. Use the wooden horse or sensory deprivation hood then a hogtie for examples.

SandrineBC:
Hello Nina,
I'll have a look at this one. Thanks for pointing out.

SandrineBC:
Re-added some checks before being able to hogtie someone, based of @Nina-1474 hint.
Thanks again.
Also removed commented out code.

Nina-1474:
The AllFours tie is the last "ArmsOnly: false" option and that will also need preqrequisite checks reinstated. Wooden horse or legs upside-down suspension then all fours for examples.

SandrineBC:
Good catch, @Nina-1474.
Perhaps we should change the Hemp.js and Chain.js to handle that check the same way, it is done in the Web.js?

elliesec:
If you're suggesting completely moving the hemp rope and chains to use the `ExtendedItem.js` script, I'd say that's probably something for a separate pull request. There are still a couple of kinks (no pun intended) being worked out in that. My intention was to get that and the webs out into the beta to identify any problems with it, and make sure it's completely stable before moving other items over to use it.

SandrineBC:
I'd say so as well. I simply find it ugly to do the same thing in two different ways and think yours is simpler and leads to more maintainable code.
Let's add this to a 'Code Maintenance' task as soon as this PR is merged.

Nina-1474:
For the HempRope and Chains in particular, I'm working on a feature that could make converting those to the extended script much harder. So don't jump the gun on that just yet!

elliesec:
Good to know Nina. Probably worth waiting on those two in that case.

Ben987:
I'm not sure I've followed everything here.  Should we put that one on hold or should I review it?  Thanks everyone.

SandrineBC:
I think you can review it. The changes were tested by Ellie and me and make the overall feeling better as they remove unnecessary restrictions within the game (like standing up before being hogtied).

Ben987:
This PR creates a few conflicts, probably with other PRs that were accepted in the last days.  Can you double-check and test again?  Thanks a lot.

SandrineBC:
I re-tested in on my site after a manual merge and applied the changes to the new nose ring item as well. Should works as intended...

SandrineBC:
I'll redo this request, now that we have a stable version

PR #927, opened 2020-06-16
Beta Bugfix: Fixes to the web and extended item scripts

elliesec:
* Fixes an oversight in the extended item script that didn't deal with NPC targets properly
* Modifies the extended item script to accommodate changing NPC dialog
* Changes the web to always hide shoes (clipping issues)
* Changes the web hogtie and suspension hogtie to hide socks, and to block the `ItemDevices` slot
* Changes web priorities so that collars are now rendered below the web

Tested with NPCs and in chatroom to ensure nothing else has been broken.


Ben987:
That was quick.  Thanks for the big fix.

PR #929, opened 2020-06-17
some more rope designs

gatetrek:
added a few more designs to the legs and feet area and in the pelvis area as well.
in the pelvis there the crotch rope that in the Torso harness still to add some variety


Ben987:
This looks great!  The shibari lovers will certainly like that one.  Thanks!

Ben987:
I love the new rope poses, they are great.  But do you think it would be easy to make it so there's more than 2 options available per screen?  Maybe 3 or 4 options per screen would be great, so there are less clicks to do to find your favorite rope tie.  Thanks!

gatetrek:
i'll have a lookie to see if can fit more on to one page without it being a clogging mess

PR #930, opened 2020-06-17
fix : lovers timer padlock GetLoversNumbers call

Atasly:
fixed the call of GetLoversNumbers call (silly error)
untested


Ben987:
Awesome!  This should take care of that bug.

PR #932, opened 2020-06-17
Info Sheet rework

Atasly:
reworked how owner and lovers are displayed in the information sheet
- longer names will be properly displayed
- owners will be visible at all times
- removed some redundant lines


Ben987:
Nice!  I'll tackles the bugs first and review that one after.  Cheers!

PR #933, opened 2020-06-17
Struggle Audio Fix

Nina-1474:
When the struggle screen is exited, either by escaping or hitting the Exit icon, any audio that's playing ends. However this didn't cover the case where the screen was exited by hitting the Esc key. This currently only occurs for hemp & nylon ropes which have a 37 second long audio file that would play in full. The call to end the audio has been moved and now all 3 exit methods are covered.


Ben987:
Nice!  I've never heard of that one before.  Thanks!

PR #934, opened 2020-06-17
Beta Bugfix: Missing AllowType property in toe tape asset definition

elliesec:
This was causing toe tape type changes not to be propagated to other users in a chatroom.


Ben987:
Nice!  Thanks a lot.

PR #935, opened 2020-06-17
Fix: Block self-use of remotes

ace-1331:
Fixed an oversight where remotes on the vulva and nipples area could still be used while using the blocked remotes rule (on self) because they were still inside the inventory (not confiscated).

So all a sub had to do, is know they can click on the remote itself to control their vibrators (even if the remote icon was gone)


Ben987:
Good catch!  Thanks a lot.

PR #936, opened 2020-06-17
fix owner/lovers lock resetting to player number

Atasly:
as title says, fix owner/lovers lock resetting to player's number after struggling/wand/exiting the room


Ben987:
Great!  Thanks :)

PR #937, opened 2020-06-18
Fix: 2 missing locks

ace-1331:
Renamed two  locks that were incorrectly named which let to missing assets when kneeling with the straight/round clit piercing locked


Ben987:
Nice catch there!  Thanks :)

PR #942, opened 2020-06-18
Crib extend menu stuffed icon fix

mrbakucamp:
Should make the icon for the stuffed crib option more centered


Ben987:
Great!  Thanks :)

PR #947, opened 2020-06-18
Fix: Behavior of burlap sack and baby harness

SandrineBC:
Repeated PR to fix the interaction between burlap sack and adult baby harness.
Fixes:
Wrong drawing order of burlap sack and baby harness [#877793](http://www.hostedredmine.com/issues/877793)


Ben987:
Awesome!  Thanks a lot.

PR #948, opened 2020-06-18
Fix: Butt plugs block the anus not the vulva

SandrineBC:
The locking butt plug chastised the player on the butt, as well as on the vulva region. 
This was solved by adding a new effect "IsPlugged" to all plugs and using it as a prerequisite for the anus access.
So all 'masturbate' activities on the are excluded, as long as a plug is worn, but the vulva of the players is still accessible. All other butt related activities are still possible.

This solves "Locking butt plug blocks sexual activities on vagina [#877111](http://www.hostedredmine.com/issues/877111)"


Ben987:
Sure, the "IsPlugged" is good fix for that situation.  Thanks!

PR #949, opened 2020-06-18
Menu to have more options 

gatetrek:
with the request to have more on single page, have done it for the legs and feet 
spaced it out as best I could so doesn't look like its clogging up the screen


Ben987:
Great, that was fast!  I will test it soon :)

PR #950, opened 2020-06-18
Feature: Gag activity enhancements

elliesec:
## Summary

This pull request changes "open mouth" gags like the ring gag and spider gag to permit the wearer to use the "Lick" and "MasturbateTongue" activities. This is unfortunately necessarily quite a large change, as it touches almost every gag asset in the game.

The gags I've considered to be "open mouth" for this are:

* Ring Gag
* Spider Gag
* Lips Gag
* Dildo Plug Gag (when not plugged)
* Plug Gag (when not plugged)
* Chopstick Gag
* Tongue Strap Gag

## Details

What it does:

* Adds a new `BlockMouth` effect which is applied to all gags that aren't open mouth (gags I've)
* Adds the `BlockMouth` effect to the scripts for gags with multiple types. In the case of the two plug gags, the `BlockMouth` effect is only added to the plugged type
* Replaces the `UseMouth` prerequisite with a new `UseTongue` prerequisite on the `Lick` and `MasturbateTongue` activities.
* Adds handling for the `UseTongue` prerequisite into `Activity.js` and `Character.js`
* Changes the handling for the `UseMouth` prerequisite so that it requires that the character can talk *and* use their tongue (this makes the prerequisite more strict than the `UseTongue` prerequisite), so if the character can use their mouth, they will always be able to use their tongue, but not the necessarily other way around.
* Minor efficiency improvement to the way prerequisite checking for activities is done (changed `if` statements into `else if` statements to short-circuit subsequent `if`s if `Allow` has been set to `false`
* Removes the `Effect` array from a handful of gags where the effects would already be inherited from the parent group by default.

One (intentional) side-effect of this change is that the pump gag now no longer allows players to use mouth activities when the pump is fully deflated.

## Testing

Because this touches so many items, I've been sure to test this very thoroughly (in fact, I spent an order of magnitude more time testing than writing the code). What I've tested:

* Every single gag in the game in each gag slot individually (without any other gag layers) works as intended
* All multi-type gags work as intended for every allowed type
* Layering a gag with the `BlockMouth` effect on top of a gag without it correctly blocks all mouth activities
* Layering a gag without the `BlockMouth` effect on top of a gag with it correctly blocks all mouth activities
* Every head item in the game to check that mouth-blocking head items like hoods do not allow the lick/tongue masturbate activities


Ben987:
This PR creates a few conflicts, probably with other PRs that were accepted in the last days.  Can you double-check and test again?  Thanks a lot.

elliesec:
Thanks Ben - there were a bunch of `HideItem` conflicts with the new nose ring, and a small activity logic conflict with the recent butt plug activity changes.

I've resolved the conflicts and verified that everything still appears to be working as it should.

elliesec:
Just updated this to add the new `BlockMouth` effect to a handful of relevant head items. This fixes some existing issues where players could use mouth actions through certain hoods (like the sack hood), and also prevents players from being able to lick/tongue masturbate through hoods.

Ben987:
This is very nice, I'm sure that players will appreciate.  I probably won't be able to test all of it but we have beta versions for that.  Thanks a lot!  :)

PR #951, opened 2020-06-18
Fix: Seamless straitdresses

ace-1331:
Renamed the incorrectly named seamless straitdresses assets in the kneel folder


Ben987:
Good catch!  Thanks :)

PR #953, opened 2020-06-18
New Nose Ring Item

FeuertinKira:
-Adds new nose ring item with attachable chains or leash to ItemHead


Ben987:
Thanks!  That's a fun idea, I'll test it soon.

PR #954, opened 2020-06-18
Improv: Beep from LARP

ace-1331:
The "beep from x space" was never added for LARP, so here it is.
It will make it clearer someone beeped you from there


Ben987:
Sure!  That's a good idea.

PR #956, opened 2020-06-19
Resets character information sheet to page 1 upon exit

elliesec:
Fixes a slightly unintuitive behaviour of the new character information sheet.

Steps to reproduce:

1. View a player's character information sheet
2. Change to page 2
3. Exit
4. View a different (or the same) player's information sheet
5. Notice that the information sheet is still on page 2
6. Change back to page 1
7. Exit and view a different (or the same) player's information sheet
8. Notice that the information sheet is back on page 1

I'd expect the information sheet to reset to page 1 each time, which is what this PR does.

### DIFF
--- a/BondageClub/Screens/Character/InformationSheet/InformationSheet.js
+++ b/BondageClub/Screens/Character/InformationSheet/InformationSheet.js
@@ -178,6 +178,7 @@ function InformationSheetClick() {
 
 // when the user exit this screen
 function InformationSheetExit() {
+	InformationSheetSecondScreen = false;
 	CommonSetScreen(InformationSheetPreviousModule, InformationSheetPreviousScreen);
 }
 


tetris245:
Or maybe let the user choose the default page? With only one owner and one lover, page 1 is almost empty while there is more info on page 2.

elliesec:
Interestingly enough, @Nina-1474 and I were having this discussion earlier today.

I think we came to the conclusion that letting the user choose would be nice, but also a fair chunk more effort to code. We did also agree that possibly the reputation page should actually be page 1 and relationships should be page 2 for the same reason (reputations will generally be a more fleshed out section, and some people will have an entirely empty relationships section). Obviously that's just the opinion of two people.

I think this PR is a simple fix for one issue (consistency from a user interaction point of view). Potentially it would be nice to make the aforementioned changes too, but I've left it as is for now.

Atasly:
that's the intended behaviour.

Because I didn't want to implement a default page setting for that, I assumed the page with relationships is more important online (skills are not seen other by yourself, and reputation does not matter since the only stat people may look at is dominant which a lot max up just for the clothes)
but if you wishes to view skill/reputation, then it will be saved for the time you're online

but yes, ideally people should be able to choose which page is the default one, just like if they could choose wether to print the /help message when logging in

elliesec:
So I'd argue that from a user experience point of view, this behaviour is confusing, and I'll elaborate a little on my argument for this change (largely taken from a discussion on the topic I had with Nina). I do feel reasonably strongly about this from a UX point of view, but I'm not going to argue it to death, and will close if you strongly feel the current behaviour is the best.

In applications in general, when clicking a button that takes a user somewhere, users tend to assume that the same button will take them to the same place every time, unless they've explicitly expressed a preference about it.

Users have a tendency to autopilot, and regardless of which page they expect to see first, with the current behaviour, there are two potential sources of confusion:

1. If they usually see a particular page first, and for whatever reason happen to view the other page and exit, it's confusing not to see the page they're used to the next time they click
2. If users become used to the application remembering which page they usually view, it then becomes confusing that the application forgets this between browser loads

The optimal solution from a UX point of view would be to allow users to choose using a user setting, but that is perhaps overkill for something this small.

The next best thing, in my experience, is to pander to that tendency to autopilot and consistently show the same page first. Users are much more likely to expect the first page to be the same _every time_ than to expect it to be the first page _most of the time_. If they care about relationships most, they become used to the fact that that is what they see every time they click that button, and that becomes routine. If they care about reputations and skills most, they become used to the fact that the standard flow to view that is to click that button, then click the pagination button, and _that_ becomes routine. If something happens to break that routine, that's something that causes confusion and throws users out their accustomed flow.

As I've said, while I do care about this (ignoring which page should actually be first, because that's only something that can accurately be judged by observing the players), I'm not going to fight it to the death, so will close if you're genuinely not happy with the change.

Atasly:
well the auto pilot thing is a good point, did not think about at all. but the amount of clicks required if you're interested to only see skills/reputation or relationships is way higher with your PR, which I find a lot more tiring than sometimes breaking out of your routine. Especially if you're checking profiles a lot.

anyway do whatever you want

elliesec:
One final comment from me then. Yes, this increases the overall click count for some portion of the user population. However, I'd argue the more important thing is that each click can be made without the user having to stop and think about where they are, and the lack of thinking time decreases the overall time take to get to what they want.

As I've said, click count, in my experience, is a useful metric sometimes, but not the be-all and end-all, and I've seen this exact pattern of behaviour trip users up in real apps I've created or been contracted to do UX work on. A couple of references:

* [Don't Make Me Think](https://books.google.co.uk/books?id=g1QBFJxB_eEC&pg=PT63&lpg=PT63#v=onepage&q&f=false) - a fairly influential book within the UX community generally, but that particular section deals with exactly this case
* [UX Myth #2](https://uxmyths.com/post/654026581/myth-all-pages-should-be-accessible-in-3-clicks) - A collection of links to some related studies on the topic

At this point I'll leave it to Ben's discretion as to whether to merge this, but that's my stance.

tetris245:
Interesting discussion. In the meantime, I have found the console command to be used before login allowing me to always see first the second page when checking a profille.

BufaloAcquatico:
Also consider that this behaviour affects NPCs as well, so if you are currently on the second page, you need to switch to the first one in another profile to see the traits as a button to switch page isn't there (as there is no second page), so this kind of fix would be needed anyway

Ben987:
@Atasly, you've made the 2 pages in the menu so it's your call.  Do you want to include that pull request or not?  I don't mind either way.  Thanks!

Atasly:
as I said, I don't really mind/care, but since it actually fixes a problem, as BufaloAcquatico mentioned above, I think you should accept it

PR #957, opened 2020-06-19
NPC Trait Guarantee

Nina-1474:
Since every trait pair for new NPCs in the private room has an independent chance of being generated, there's a very low but non-zero possibility (0.07%) that no traits are generated at all. This can look like a bug to players and one such instance has been raised as an error.
Perhaps this is intended and a trait-less NPC is an acceptable possibility. But in case not, I've added a check so that the trait loop will repeat until at least one trait is assigned.


Ben987:
Sure :)  I'll try it right now.

PR #958, opened 2020-06-19
Bugfix: Ensures bed cover is removed at the same time as the bed

elliesec:
## Steps to reproduce

1. Equip new bed item
2. Equip bed cover addon
3. Remove bed item
4. Bed is removed, but bed cover stays on

Added a `RemoveItemOnRemove` property to the bed to address this. Not an issue with the bed restraints because they block the `ItemDevices` slot.


Ben987:
Great catch!  Thanks :)

PR #959, opened 2020-06-19
Turn under bed bondage cuffs into arm restraint

Ayesha-678:
Turned the under bed bondage cuffs into an arm restraint to fix some bugs


Ben987:
Alright!  Thanks a lot :)

PR #960, opened 2020-06-19
Fix crib dialog

mrbakucamp:
Fixes the crib writing (SourceCharater closes her crib.) when closing the crib


Ben987:
Good catch!  Thanks.

PR #961, opened 2020-06-19
Tight straightjacket

Ayesha-678:
Just a simple straightjacket that is tight on the body.


Ben987:
Awesome!  The tighter the better, cheers :)

PR #963, opened 2020-06-22
Seperate the glasses and masks

Ayesha-678:
Made a mask category, and put the full face masks, the shinobi and fox masks in that category to allow glasses to be worn with them.


Ben987:
Good idea!  I'll test it right now.  Cheers :)

PR #964, opened 2020-06-22
corset shirt fix

Ruilov3:
added missing priority value to the corset shirt to make it work better with skirts


Ben987:
Good catch!  Thanks a lot.

PR #965, opened 2020-06-22
Feature: New asset ArousalZone for sybian and dildo spreader bar

SandrineBC:
# Summary
The usage of the sybian or a vibrating spreader dildo bar cannot give a character an orgasm via the currently implemented arousal system, because currently the items group is used to check, if it can give an orgasm. Thus, if you would set your legs to do so, you can actually orgasm from the vibrating spreader dildo bar.

# Details
## What it does
- Introduction of the new asset attribute `ArousalZone` 
- this currently defaults to the items group name , so no need for extensive changes in the file. 
- If you add this attribute to an asset, instead of its group, the group name of this attribute is checked for all orgasm purposes.
- Example: `AroursalZone: "ItemVulva"`

## Testing
I tested the new attribute with the vibrating dildo spreader bar and the sybian, to see if I achieve the right effects and used some other vibrating items to see, if they still work without having to specify the new attribute


elliesec:
Just a thought - the Love Chastity Belt might actually be another good candidate for this, as it's a pelvis item, which is a zone players won't often have set to trigger orgasms.

SandrineBC:
Heehee, and I thought, that was the intended behavior...
In my experience, most doms are using the belt this way: to tease and deny their subs. Perhaps we should start a survey, before we change this?

elliesec:
Yes, that did cross my mind, it's one of those unintended "features" that people have gotten used to. Perhaps that's a good call.

SandrineBC:
We could think about the clitoris vibrating egg and the vibrating clit piercing, though.
The clitoris cannot be defined as an 'orgasm hotspot' right now.

elliesec:
That's a good point for sure.

SandrineBC:
According to the result of my [poll](https://discord.com/channels/554377975714414605/554378904308023297/725298995387170910) I added the `ArousalZone` attribute to the love belt and the vibrating heart clitoris piercings. Both items can give an orgasm now, if the player has the vulva configured that way.

Ben987:
I like the idea of "hijacking" the arousal zone, it will be useful elsewhere.  Thanks a lot :)

PR #967, opened 2020-06-23
Feature: Select empty gag slot

SandrineBC:
# Summary
Whenever a player selects an empty gag slot, the focus is transferred to the lowest empty gag slot. This is done for convenience, but also to resolve with 'open gags' like the ring gag in slot two, with no gag underneath (See issue [#878154](http://www.hostedredmine.com/issues/878154))

# Details
## What it does
- Added a functionality to the function `DialogInventoryBuild` in file `Dialog.js` that searches for the leftmost available gag slot, if a free gag slot is selected
- If such a slot exists, the focus is transferred to that free slot
- afterwards, the function commences as usual

## Testing
Tested the feature
- with all three slots, both occupied and free
- by applying gags on myself, as well as on another player
- with all other slots, which are not affected by the change


Ben987:
I'm not sure about this one.  Transferring focus to another zone than the clicked one might be confusing for players or they might think there's a bug.  I'll think about it.

SandrineBC:
Applying this will remove graphic problems from the 'open gags' like ring or spider gag, as they currently don't show a mouth, if they are used on zone 2 or three with no gag underneath. And we think from a usability point of view this should not pose a big problem.

PR #968, opened 2020-06-23
Feature: Levels of deafness add up

SandrineBC:
# Summary
Unlike the various gagging items, multiple layers of deafening items do not add up. This PR changes this and solves issue [#877137](http://www.hostedredmine.com/issues/877137).

# Details
## What it does
This PR does two things

**`Character.js`**
* adds a function `GetDeafLevel` to the Character this function loops through the Appearance of the character and sums up the various deafening level. The function returns the current level of deafness
* The values are defined in constants:
  * `CCharacterDeafLevelFactorLight = 1`
  * `CCharacterDeafLevelFactorNormal = 2`
  * `CCharacterDeafLevelFactorHeavy = 3`
  * `CCharacterDeafLevelFactorTotal = 4`

**`Speech.js`**
* replaces the various calls of `IsDeaf` by `GetDeafLevel` and compares the result with the appropriate level of deafness. If a certain level is reached, the incoming messages are garbled, analog to speech garbling, when gagged
* The thresholds are defined in constants:
  * `CSpeechDeafLevelTotal = 4`
  * `CSpeechDeafLevelHeavy = 3`
  * `CSpeechDeafLevelNormal = 2`
  * `CSpeechDeafLevelLight = 1`

## Testing
The functionality was tested by two characters, one adding dealing items to herself and another repeatedly saying the same phrase. According to the level of deafness, the correctly garbled speech was received, up to "mmmm mmm" at the highest level.

### DIFF
--- a/BondageClub/Scripts/Character.js
+++ b/BondageClub/Scripts/Character.js
@@ -1,4 +1,10 @@
 "use strict";
+
+const CCharacterDeafLevelFactorLight  = 1;
+const CCharacterDeafLevelFactorNormal = 2;
+const CCharacterDeafLevelFactorHeavy  = 3;
+const CCharacterDeafLevelFactorTotal  = 4;
+
 var Character = [];
 
 // Loads a character in the buffer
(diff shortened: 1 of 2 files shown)


Ben987:
Good idea!  I'll probably remove the const though, we don't use these anywhere in the code.  Thanks!

SandrineBC:
Hi @Ben987 ,
thanks for accepting. Would probably be a good idea to let the constants stay in this one. It make fine-tuning the level of deafness easier, if you can do so at one point instead of scanning the various points.
Besides, mos browsers can handle those, thus there should be no problem from this side.

PR #969, opened 2020-06-23
Smooth Sleep mask

Ayesha-678:
Added back the old sleep mask as the Smooth sleep mask, and turned the new one into the 'frilled' sleep mask, after seeing people were upset the old one got removed.


DMCHiggens:
THANK YOU ;-;

TheGnarp:
-_-'

Ben987:
Sure!  Though if we add a third sleep mask or more, we should probably extend these items as 1 item and be able to select the style.  Thanks :)

PR #971, opened 2020-06-23
Fix: 4x missing locks

ace-1331:
Here are some (more) missing locks causing 404's


Ben987:
Alright!  I'll make a hotfix for these also.  You're the locksmith :)

PR #972, opened 2020-06-24
Fix: AllowPose for ball and shackles

ace-1331:
Removed LegsOpen from the AllowPose for both these items as the assets for them are not different (LegsOpen is the same as the base pose) which caused those two items to disappear when used with one bar, saddle stand and several other items that set the pose to LegsOpen due to more 404s in a folder which doesn't (and should not) exist


Ben987:
Good catch!  I will prepare a hotfix.  Thanks :)

PR #973, opened 2020-06-24
Fix: Teddy bear 

ace-1331:
Fixed the TeddyBear when in the AllFours or Hogtied pose. Instead of it clipping through the floor, it behaves like when it is used with other restraints: It is placed next to the character on the floor.

This allows it to be displayed while using ropes/chains now, hence why I removed the hide item property on the rope (it was never hidden for chains, causing more weird clipping like the ones experienced with the pet crawler)

See: [Issue #878737](http://www.hostedredmine.com/issues/878737)


Ben987:
Good catch!  Thanks a lot.

PR #974, opened 2020-06-24
Fix: Crib locks

ace-1331:
Believe it or not, I found more missing locks causing 404s
The creator of the crib confirmed they should not be visually rendered, hence why the locks are invisible


Ben987:
The locksmith strikes again!  Thanks a lot :)

PR #975, opened 2020-06-24
Change: bunny plugs

ace-1331:
Changed it so bunny plugs now behave like all other butt plugs.

These were the only two plugs you could use on someone with a zipped suit because they were missing the correct prerequisite


Ben987:
Good one!  I will hotfix it also.  Cheers :)

PR #976, opened 2020-06-24
Fix: Hogties and various items that chain a player to the ground

SandrineBC:
# Summary
With the introduction of the medium collar chain, the new webbing and the nose ring, several bugs were introduced as well, because the medium chain never made it to prerequisites and effects, as well as the nose ring does not check for hogties or suspended hogties. 
This PR fixes all these problem.
Additionally, a hogtie can now applied to kneeling players, what adds a lot of comfort and realism for the players.
# Details
## What it does
This PR changes the following files:
**`Female3DCG.js`**
- introduces a new Effect `IsChained` that denotes, if a player is chained to the ground. This Effect is used on the following items:
  - `CollarChainLong`
  - `CollarChainMedium`
  - `CollarChainShort`
  - `NoseRing`
  - `ButtPlugLock`

**`Appearance.js`**
- Alters the calculation of the player's position on the screen. In this version you can use a short chain on a hogtied player and the player will stay on the ground

**`Inventory.js`**
- Adds the `SuspensionHogtied` pose to the check, wether a player is hogtied
- Uses the newly introduced `IsChained` effect in the `NotChained` prerequisite, so this has not to be changed again, when new items are added that can also chain a player to the floor.

**`Chains.js`**
- Splits the general prerequisite in more specific ones, depending on the various possible poses

**`HempRope.js`**
- Splits the general prerequisite in more specific ones, depending on the various possible poses

**`Web.js`**
- Removes the `NotKneeling` prerequisite from the various poses, as this is not needed anymore

**`ButtPlugLock.js`**
- Moves the check, if the application of a short chain is possible, to an explicit function, as this check was implemented two times in the code
- Adds the `IsChained` effect, if the player is chained to the floor
- Removes the possibility to use a chain, if the player is suspended

**`NoseRing.js`**
- Moves the check, if the application of a short chain is possible, to an explicit function, as this check was implemented two times in the code
- Adds the `IsChained` effect, if the player is chained to the floor
- Removes the possibility to use a chain, if the player is suspended

**`HempRopeHarness.js`**
- Removes the checks before applying a hogties, as the hogtie was moved to the ItemArms group

## Testing
All changes were tested locally on my machine, as well as by @elliesec. 
@Nina-1474 gave important hints in regards of the hogtie checks. So the prerequisites were added, that hogties are not possible, if the player is mounted or wears a hood with an alpha mask


Ben987:
This one changes a lot of stuff and creates conflicts in the asset file.  I'll get back to it later.  Thanks.

SandrineBC:
Current version merged into this branch

Ben987:
Don't we need new images if we want to be able to leash a hogtied girl to the ground?

Also, please change "let" for "var".  Var are used everywhere, thanks a lot.

SandrineBC:
Hi Ben,
no new images are needed. This just fixes the handling of the poses. I'll change let for var, though let would be the better option imho...

Ben987:
Yep, please change them to "var".  I'll clean the remaining "let" if there are any in the main code when I have time.  Thanks.

SandrineBC:
Merged into current master and changed 'var' to 'let'

Ben987:
Awesome!  Thanks for the changes and code cleanup, I'll test it soon.  Cheers :)

PR #977, opened 2020-06-24
fix : owner locks check for member number

Atasly:
if a sub applies an owner lock on herself, her owner should be able to modify it and not the sub
bug introduced due to the member of the sub on any locks self applied


Ben987:
Awesome!  I will hotfix that one.  Thanks :)

PR #978, opened 2020-06-24
Fix: online dialog builder crash

ace-1331:
- Fixed a bug where the client would crash when trying to click on an online player's dialog options

Due to how functions in the dialog sheet is built, online characters **entering a room you are in for the first time while you are logged in** while you are in a screen other than the chatroom screen (like your friend list) will build the dialog with the wrong function names, causing the dialog options to crash. This fix ensures that any Online character is loaded with the ChatRoom prefix instead of the CurrentScreen.

Crash in action

Tested with self dialog, NPCs and online characters with no apparent side effects.

See [issue 878282](http://www.hostedredmine.com/issues/878282)

### DIFF
--- a/BondageClub/Scripts/Character.js
+++ b/BondageClub/Scripts/Character.js
@@ -113,8 +113,9 @@ function CharacterRandomName(C) {
 // Builds the dialog objects from the CSV files
 function CharacterBuildDialog(C, CSV) {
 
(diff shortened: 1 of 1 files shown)


Ben987:
Great catch there, I'll prepare a hotfix for that one also.

PR #979, opened 2020-06-24
Improvement: Description refresh

ace-1331:
- Made it so the description is refreshed when someone updates their description

This has been a major inconvenience for a while, you cannot see someone's new description until you relog because it was never a flag for a character refresh. This makes it so each time someone changes the description, everyone else will see it without having to relog. No need the desync is also quite confusing when people expect you to see (or not to see) certain information

To test how inconvenient it is right now, you can have two characters in a room, #1 changes their description and #2 will never see it, no matter how many rooms you go in with them, unless you relog.


Ben987:
That's a nice one, I will prepare a hotfix for it.  Thanks.

PR #980, opened 2020-06-24
Bed Straps with Closed Legs fix

Nina-1474:
Adding a prerequisite to have legs open to use the bed cuffs on the arms, as otherwise the leg cuffs connect to nothing in midair.


Ben987:
Nice!  I will prepare a hotfix.

PR #981, opened 2020-06-24
Missing Cattle Prod Pose Files

Nina-1474:
Adding in the missing 1x1 image files for the cattle prod spanking toy in different poses.


Ben987:
Great!  Thanks a lot :)

PR #982, opened 2020-06-24
Fix: plug 404s

ace-1331:
- Fixed 3 butt plugs that were missing `Visible: false`

This caused 404s when in the AllFours position, so I added `Visible: false` and removed the necessary 1x1 empty images from the base pose


Ben987:
Absolutely!  Thanks :)

PR #983, opened 2020-06-24
Improvement: Quickban

ace-1331:
- added the ability to manually ban your ghostlist and blacklist after a room was created.
- improved the logic of autoban/quickban to remove duplicate numbers
- extracted it into a function
- changed the "selection" button so it is now properly grayed out like the rest for non-admins

As pointed out to me by others after adding auto-ban, this is useful when you recently ghosted/blacklisted someone, but it is especially useful for when there are more than 1 admin. It concatenates new numbers to the list to avoid duplicates and provides for an easier way to ban. Truly removing the need to maintain a separate list of member numbers for trolls.


Ben987:
Sure, this should be useful.  Thanks!

PR #984, opened 2020-06-24
Fix: pony training dialog crash

ace-1331:
- Fixed a typo in a stable.js function that caused the dialog to crash when it was called

https://discord.com/channels/554377975714414605/554378725916147722/725455652196974672


Ben987:
Good catch.  I wonder why we never fixed that one before.  Thanks a lot!

PR #985, opened 2020-06-24
Bugfix: Incorrect kneeling asset alignment of normal weight straitdresses

elliesec:
There's a minor alignment issue with the kneeling pose for all four straitdress variants with the normal upper body weight (asset is positioned a few pixels too low). See below image.

I've adjusted the alignment of the assets to line up correctly with the corresponding standing poses, which addresses the issue.

straitdress


Ben987:
Thanks for the alignment fix, I will prepare a new hotfix batch and include it in it.  Cheers!

PR #986, opened 2020-06-25
Fix: more missing locks

ace-1331:
- added missing lock assets causing 404s for the hoof mittens & ponyhood (@Ruilov3 confirmed they shouldn't be visually rendered)

I'm adding missing assets as I come across them so it's been several small PRs in a short time,, but I'm starting to consider going through everything in some way to do a full cleanup. I'm not sure how I'd do it yet. For now, I'll stick to this method


Ben987:
The locksmith strikes again!  Thanks, I will prepare a new hotfix batch.

PR #987, opened 2020-06-25
Bugfix: Missing action text when changing web blindfold type

elliesec:
The web blindfold publish function was missing a parameter, resulting in "((You can use or remove items by selecting specific body regions on yourself.))" messages when its type was changed.


Ben987:
Good catch!  I've never noticed that one, I will prepare a hotfix batch pretty soon and include it.  Thanks!

PR #988, opened 2020-06-26
Improvement: Asset code cleanup

SandrineBC:
# Summary
By adding more and more items and continues tweaking of the various attributes the list of assets got a bit disheveled. This PR tries to tidy up the Assets a bit without adding new functionality to the game, but hopefully makes the code shorter and helps with the further addition of new items.
# Details
## What it does
Changes occur in two files
**`Asset.js`**
* Adding the attributes `Block` and `Hide` on the group level, so those don't need to be defined on every item of a group, like the various gags and the boots
* Making inheritance explicit, so that an undefined item attribute now copies the attribute of their group

**`Femaled3dCG.js`**
* Removed duplicate attributes on item level
* Removed attributes on item level that had the same value as the corresponding attribute on group level (e.g. Priority)
* Made use of the new group level attributes `Block` and `Hide` to shorten the code
* Set the prices in groups `ItemMouth2` and `ItemMouth3` to -1, so they don't show up two or three times in the shop
* Minor reformatting of the code for readability 

## Testing
As we have no automated testing I spend a good two hours to apply various items of clothing and bondage to my test accounts on my local machine. No strange behavior was noted, though I probably did not test any combination.


Ben987:
Cleaning the assets will be a good idea, but we'll need to synchronize on that one.  Since this file is changed all the time, it will create tons of conflicts.

SandrineBC:
Just tell me, when a good opportunity will be. I'll try to keep this one up to date

SandrineBC:
Re-merged into the current master branch.

Ben987:
This changes so much stuff that it makes it impossible for me to test it all of it.  Is there any way you can split it into multiple PR ?  Thanks a lot.

SandrineBC:
Removed the `WearTime` correction from the commit. The rest stems from the corrections mentioned in the original comment to this pull request.

Ben987:
Lets cross our fingers that this doesn't introduce too many bugs.  A big cleanup of the asset file will be a good thing for sure.  Thanks a lot!

PR #989, opened 2020-06-26
Improvement: Removed appearance blank pages

ace-1331:
- made it so the amount of page is tied to the amount of groups you see, not the total amount of groups

Currently, if you go to change someone else's clothes, you will cycle through pages until you end up on a blank page because the body parts are not rendered. This removes any blank page(s) created by the non-rendered groups


Ben987:
Great idea!  That little bug has been in the code for a long time now.  Cheers :)

PR #990, opened 2020-06-26
Fix: Frogtie corrections

ace-1331:
EDIT: I've ended up splitting the changes because all 3 were unrelated

- Fixed the frogtiestraps to correctly apply the forcekneel status (Fixes the problem where you could still click the stand button while wearing them)
- Fixed the frogtiestraps to prevent them from being used while suspended (causing you to look like an ostrich if you used suspension hemprope with them)
- Fixed the frogtie from the leg hemprope in a similar fashion


ace-1331:
I'm adding small item corrections to this PR while I find them since they are all small and have little impact.
Let me know if you prefer  it split the following way (currently): 1- frogtiestraps 2- bed/crib 3- gas mask

Ben987:
Thanks for splitting the fixes, it makes it much easier for me.  I'll try this one right now.

PR #991, opened 2020-06-26
Msl

themeresort:
Pre-alpha.  Details in discord.


themeresort:
There is an uncommitted change to MainHall.js.  It puts a new clickable icon on the view.  On click, a session is precreated for the user on the server, and the user is redirected to the new rooms.

function MainHallGoToMsl(){
	//Perhaps move this server side?
	var sessionId = ServerSocket.id;
	io(ServerURL + "/msl").emit("PreCreateSession", {playerId:Player.MemberNumber, sessionId:sessionId});
	
	//Session storage expires when browser tab is closed
	sessionStorage.setItem("playerId", Player.MemberNumber);
	sessionStorage.setItem("sessionId", sessionId)
	
	//bloaty code because browsers don't understand ../
	var indices = [];
	var url = window.location.href;
	for(var i=0; i<url.length;i++) 	if (url[i] === "/") indices.push(i);
	url=url.substring(0,(indices[(indices.length)-2])+1);
	
	setTimeout(function(){window.location = url + "/Msl";}, 200); 
}

Ben987:
Thanks, this is HUGE!  The big problem with this project are the changes on the server side.  Since we only have one production server, any bug found will crash the game for everyone.  Is there any way this could run on a different server?  So we could protect our current players while we test.  I can't include any pre-alpha in the master branch.  So the regular game would run on the actual server and the Msl would run on it's own server.  Thanks.

themeresort:
But of course!  You start the same server (same codebase) -- on a different port, and change the url in Environment.js (and MainHall.js, where it ).
With the exception of the import process, all functions are wrapped in try-catch.

Ben987:
Is there anyway we can have it running on a different server, without touching the current one at all?  It will be much safer if it can run on a whole different environment and doesn't impact the production server,   You could host that server somewhere else and we could connect to it when the game enters in your new room.  Thanks!

themeresort:
Yes, anything is possible.  The msl code, both client and server, is completely decoupled from either college or club.  Both client and server can be deployed on a different server, on a different domain.  With minimal effort, they can even be made into a separate projects with their own repository. 

The server code still needs access to the database.  Msl server can run its own database -- but character import process will get tricky and error prone.  Now it loads everything it needs from the database, I have two dozen test users in my test environment.

Ben987:
If it can run on it's own database and not impact our current production server that would be great.  Let me know if you can find a solution where we do not need to change the server and little stuff in the current screens of the client.  It needs to be easy to close these new chat rooms temporarily if there are major problems.  Thanks!

themeresort:
There are too many solutions that fit this creteria, including current set up with my dev server running somewhere unconnected to the club.
The current version of the server stores everything in one single database table.  The msl code does not write in the existing table, and will store imported characters in new table.

Ben987:
I'm not sure what's the best way to do it.  But since this is such a huge change in one block, we need to isolate it from the production server until it's fully tested.   That's why doing small changes step by step makes it much easier to incorporate than trying to plug in a huge chunk of code, especially server-side.  Thanks!

themeresort:
The front end client is isolated completely.
The server will run as a separate instance, on a different port.  It reads from the database, but does not write.  When it does write, it will use its own tables.

Ben987:
Great, if you can make a version that doesn't require any server change it would be perfect.  And why do you need to read from the account database?  Could you also work around it?  If all you need is some extra properties, we could load them up at login.  Thanks!

themeresort:
The reading from the database is to import characters.
I will be done by the end of the week with the alfa version.

themeresort:
Server is as separate, only change would be required is inclusion of the lzstring in package.json for the inventory import.
Set the variables in msl/environment.js and run via nodejs ./msl/msl.js

themeresort:
Latest release merged, latest commit pushed.

Ben987:
Hello, is that new version ready to test on my side?  Were you able to make it work stand alone without any changes to the server?  What are the impacts on the client?  Thanks a lot.

themeresort:
Everything is ready to be deployed -- it was last week, when I made announcement in discord.
No effect on the client.
Both server and client have file environment.js, which determines whether the server accepts id login, or requires user name and password to check.
Instead of 'nodejs app.js', run 'nodejs msl/msl.js' to run the server (it might be different at heroku -- I never used app container, VPS every time).
The client can be deployed alongside -- it is a set of files in a separate folder, the page is reloaded completely in the browser.

Ben987:
So we would need to run another server that needs to access the database to make it work?  You could not make it work without accessing the live database?  Heroku works with dynos but I'll be moving out of that environment soon.

themeresort:
Access to the database is required for import process.  The import process 1. Allows to quickly dress up characters in main club, and then test how they look in the new rooms and 2. carries over the friend and other lists.

There are two modes.  I can commit the small change to app.js file, and it will pass the database handle to msl.js -- and code will run as a single application.  At the moment, the settings in environment.js are set up as if the msl.js is started by itself.

There is also another switch in environment.js (server side) -- whether to allow login with member number, or require user name and password.

There are alternatives to import process, such serializing the Player object from the console and cut and pasting it, but this is something that can be done later.

Ben987:
If you need server side changes it makes it much more complicated to merge and approve this.  There's no way for you to code it client side only?

If it can't be done, this will need to wait until I've migrated the server to a new node.js hosting.

themeresort:
In discord, you said 'make a pull request for us to review the code'.
There is the code....
Anyway,, I can make another git project for the server code, or zip and email the server code.  There are many, many ways to deploy it without merge, at least on a regular server.

Ben987:
Yes.  If you can make it so we don't have to modify anything server side and only a few things on the current client side, it will be much easier to merge this into the project.  Thanks a lot.

themeresort:
It has been done.
https://github.com/themeresort/Bc-Msl-Server

But -- while we were doing this back and forth, someone has added a new dependency in the client code.  The asset database module now has to know about the dialog and localization module.

In Female3DCG.js, line 31 has to be commmented out (or value hardcoded), or the whole thing won't start.
line 31:	DialogSortOverride: DialogSortOverride

Nina-1474:
That setting could be changed to just a boolean instead then which will remove the dependency. 
The idea was to give future assets some flexibility on how they appear in inventory lists, but tbh I can't really see it getting much use.
If you want to make such a change here I have no objection to that. Or I can tweak it under a separate PR if you'd prefer.

Ben987:
Lets close this for now until it's updated to the latest version and ready to work without needing any or very few server-side changes.  Thanks everyone.

PR #992, opened 2020-06-27
Feature: Adds a chatroom message and audio cue when timer locks expire

elliesec:
# Summary

Adds chatroom message and audio cues to make timer lock expiry feel a little more natural and a little less sudden/unexpected.

* Timer lock expiry sends a chatroom message if the player is in a chatroom
* Adds a new audio file for unlocking actions, and associates it with the new chatroom message
* Also associates the new audio file with the existing `ActionUnlock` and `ActionUnlockAndRemove` chatroom messages


Ben987:
That's a good idea, it will add to the immersion.  Thanks a lot :)

PR #994, opened 2020-06-28
Fix: Gas mask dialog

ace-1331:
EDIT: I've ended up splitting the changes because they were unrelated.

- Fixed the dialog for the gas mask (Wrong template name)


Ben987:
Awesome!  Thanks for the fix :)

PR #995, opened 2020-06-28
Fix: Bed and Crib prerequisites

ace-1331:
EDIT: I've ended up splitting the changes because they were unrelated

- Fixed the bed and crib with alpha hoods (causing parts of the items to dematerialize ), which also adds a "item cannot be used with alpha hood" prerequisite similar to the one for the hogtied pose for future items.

- Fixed the prerequisites for the crib that allowed for many broken combinations (Fixes safety concerns regarding suspension, hogtie, suspension hogtie, etc.)


Ben987:
Good catch there!  I haven't thought about the alpha hoods.  We might need to find a better for these hoods someday.  Cheers :)

PR #996, opened 2020-06-28
Fix: missing assets

ace-1331:
- renamed incorrectly labeled assets for the LockingBoots1 item inside the Kneel folder (they were all labelled as 3 somehow)
- added the missing lock asset for the LockingBoots1 item in the kneel pose

this fixes all 404s for this item


Ben987:
Great catch!  I've never noticed that one.  Cheers :)

PR #997, opened 2020-06-28
Bugfix: Prevent french kissing through gags

elliesec:
# Summary

A small follow-up to #950 - this changes the `FrenchKiss` activity to only be permitted if the target's mouth is not blocked by a gag (unless it's an open mouth gag like the ring gag). This prevents the previous behaviour where people could be French kissed through any gag as long as it was in the first gag slot.


Ben987:
Nice touch!  I'll test it right now.  Cheers :)

PR #998, opened 2020-06-29
Bugfix: Client-side check for spaces in account name during account creation

elliesec:
# Summary

Currently, the server uses the following regex to validate account names:

```js
/^[a-zA-Z0-9]+$/
```

Whereas the client uses this regex:

```js
/^[a-zA-Z0-9 ]+$/
```

This means that users can attempt to create an account with a space in the account name, receive a "Creating your character..." message, but the server will silently reject the account creation, making the screen appear to hang.

This changes the client-side regex to line up with the server-side one, so users will now receive the "Invalid character, account name, password or email" message instead.


Ben987:
Good idea!  Thanks :)

PR #1002, opened 2020-06-30
Feature: ParentGroup on the asset level

ace-1331:
- Added support for asset-level parent groups.
- Removed parent group property from 5 existing items using it for the wrong reasons
 
Currently, there are some items using this property (which should not), and others that have been created and blew up the amount of properties/assets for a given group which should not happen. This property was also falsely "advertised" as working in the comments at the beginning of the file.

A category might have items needing one or the other (like itemdevices), this way will help to add new items in the future. @Ruilov3 actually has one inbound which will be helped by this.

I've tested with both parent groups on the group level and on the asset level. I've also tested with ignore parent group, everything still follows the same pattern it did before as expected.

**The related PRs following this one all require AND include this PR**


Ben987:
Yes, it should follow the same pattern and help with future items.  I'll test it soon.  Thanks!

PR #1007, opened 2020-07-01
Bugfix: Cleans up visual artefacts in the reverse bunny suit top

elliesec:
# Summary

Some of the reverse bunny suit top assets had some leftover artefacts where parts of the image hadn't been fully erased, resulting in visual errors like the one shown below. I've been through all of the assets for the top, and cleaned up any of these artefacts where I've found them (it also seems that my image editor uses a better PNG compressor than whatever was used for the original assets, so bonus file size reductions too).

image


Ben987:
Very nice!  I've never noticed these before.  And reducing the image size is always good, thanks :)

PR #1008, opened 2020-07-01
Feature: Hand restraints make struggling more difficult

SandrineBC:
# Summary
As of now, it's irrelevant wether the hands of a player are restrained, e.g. by mittens or tape, when they try to struggle out of some bindings. This PR changes this and adds half the difficulty (rounded up) of the player's hand restraints to the struggle challenge.
This resolves issue [#878284](http://www.hostedredmine.com/issues/878284)
# Details
## What it does
Changes occur in two files 
**`Character.js`**
* Adding the functions `GetClumsiness` that returns either 0 (if the player's hands are not restrained) or half (rounded up) of the player's hand restraint base difficulty
**`Dialog.js`**
* Changed the function `DialogProgressStart` so that the clumsiness of the player is taken into account when the challenge level is calculated
## Testing
* Tested that feature on my own server by letting a character struggle out of a restraint and comparing that challenge level with the one after a hand restraint was added
* Tested with the following restraints
    * Paw mittens
    * Padded mittens
    * Leather mittens
    * Padded leather mittens
    * Duct tape
    * Polished steel mittens

### DIFF
--- a/BondageClub/Scripts/Character.js
+++ b/BondageClub/Scripts/Character.js
@@ -85,17 +85,14 @@ function CharacterReset(CharacterID, CharacterAssetFamily) {
 		 */
 		GetClumsiness: function () {
 			var clumsiness = 0;
-			var A = 0;
-			var found = false;
-			while (A < this.Appearance.length && !found) {
+			this.Appearance.some(Item => {
 				// get the difficulty of a hand restraint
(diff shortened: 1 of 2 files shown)


Ben987:
Adding a difficulty to the evasion skill when hands are restrained is a good idea.  Though it should not change the difficulty when removing the hands item itself.  There's probably a much easier way to do that though, I'll add it to my to-do list and create a small change soon.  Thanks for the idea though.

PR #1010, opened 2020-07-01
Bugfix: Minor adjustments to the slim leather masks to ensure full face coverage

elliesec:
# Summary

Something spotted by one of the more attentive-to-detail players, there is a very small sliver of the right side of the face that is not fully covered by the slim leather masks, only noticeable with certain hairstyles (see image). This PR makes some very minor adjustments to the mask assets to ensure more complete face coverage.

image

Below is a before/after comparison (hair and mask coloured to improve contrast).

slim-mask-before-after


Ben987:
Very nice!  Thanks :)

PR #1015, opened 2020-07-02
Bugfix: Prevents combination locks from being opened/updated on blocked zones

elliesec:
# Summary

Changes combination locks so they can't be opened or have their codes changed when on a zone that is blocked by another item. They will now just display the message, "Another item is blocking access to this lock", if the zone is blocked.

I've also added translations to this pull request (something I should do more often, but keep forgetting about). Whereas my knowledge of French and German grammar are just about enough to give them vaguely sensible translations, I'm completely lost on Russian, so I've just left that one as is!


Ben987:
Great idea!  And you've already translated it, awesome :)

PR #1019, opened 2020-07-03
Bugfix: Fix to limited item permissions for bondage bench/gas mask addons

elliesec:
# Summary

There was an issue with the bondage bench straps and old gas mask addons that meant that if the player had the items set to limited permissions, then nobody would be able to use those addons (even if they were owner/lover/on the player's whitelist). This fixes that.

# Steps to reproduce

Requires 2 characters: character A, and character B

1. As character A, set global item permissions to "Everyone", "Everyone, except blacklist" or "Owner, Lover, whitelist & Dominants"
2. As character A, set item permissions for the bondage bench straps in the addon slot to limited
3. As character A, add character B to your whitelist
4. As character B, add the bondage bench to character A
5. As character B, attempt to add the bondage bench straps to character A
6. Notice that character B is unable to add the straps, despite being on character A's whitelist

The same issue exists with the old gas mask addons.

# Changes

## BondageBench.js and OldGasMask.js

* Uses `InventoryCheckLimitedPermission` instead of `InventoryIsPermissionLimited` when checking whether addons should be available
* Changes usages of `var C = (Player.FocusGroup != null) ? Player : CurrentCharacter;` to `var C = CharacterGetCurrent();`, which is functionally equivalent

## Inventory.js

* Adds a new `InventoryItemCreate` function, used to generate `{Name, Group, Asset}` item objects
* Simplifies `InventoryAdd` by changing it to use the new `InventoryItemCreate` function (previously it was duplicating functionality with `AssetGet` anyway) - this change does not change the behaviour of `InventoryAdd`, just simplifies the code. This change was created as a separate commit, so can easily be rolled back if needed.

### DIFF
--- a/BondageClub/Screens/Inventory/ItemDevices/BondageBench/BondageBench.js
+++ b/BondageClub/Screens/Inventory/ItemDevices/BondageBench/BondageBench.js
@@ -15,12 +15,13 @@ function InventoryItemDevicesBondageBenchLoad() {
 
 // Draw the item extension screen
 function InventoryItemDevicesBondageBenchDraw() {
-	
+
 	var C = (Player.FocusGroup != null) ? Player : CurrentCharacter;
 	var strapsBlocked = InventoryGet(C, "Cloth") != null || InventoryGet(C, "ClothLower") != null;
 	var itemBlocked = InventoryGet(C, "ItemAddon") != null;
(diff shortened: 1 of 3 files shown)


Ben987:
Nice one, I've never tried that combo before.  I'll test it right now.  Thanks :)

PR #1022, opened 2020-07-04
Bugfix: Behaviour of previous button in background selection

elliesec:
# Summary

This fixes a small issue with the behaviour of the previous button in the background selection code, where clicking on the button when on the first page would take the user to the penultimate page rather than the last page.

# Steps to reproduce

1. Enter the background selection screen
2. Immediately press the "previous" button
3. Notice that the backgrounds displayed are not the last page of backgrounds, but the second last


Ben987:
Very nice!  I didn't notice that one.

PR #1023, opened 2020-07-04
Fix: missing toe tape

ace-1331:
- added missing blank assets for the full toe tape when kneeling to prevent 404s


Ben987:
Great!  Thanks a lot :)

PR #1024, opened 2020-07-04
Improvement/Fix: Chatroom admin

ace-1331:
- fixed a bug where the chatroom limit would be reset to 10

Currently there is a bad check at line 855 on the server. Because the server sends back the data as an int and the check contains `(typeof data.Room.Limit !== "string")` The already correctly formatted value is reset to 10. I assumed you prefer the fix client side which is contained in this PR (see ChatRoom.js file), but thought I should inform you of that regardless

- improved the chatroom admin screen so that it does not forget the input values when going in background selection (ChatAdmin.js)


Ben987:
Sure, the client side fix works for me.  Thanks a lot :)

PR #1025, opened 2020-07-05
Added neck rope

TheGnarp:
After a very exciting mega rope contortion hogtie in a RP, I decided to add the rope that we had to describe in text form.


Ben987:
Since it's also a hemp rope, can you add it to the same buy group as the other ropes?  With the same value?  Thanks.

TheGnarp:
Done!
Thanks for reminding me! Totally forgot about buygroups!

Ben987:
Great!  Thanks a lot :)

PR #1026, opened 2020-07-05
Fix: WebGL context

ace-1331:
- fixed WebGL draw canvas missing relevant listeners for context lost/restoration which caused the club to be broken if your GPU runs out of resources

Tested and it works as intended. I also verified and the 3D Draw already properly listens for these events.
Note: if you test with the .loseContext() method, you need to use the .restoreContext() method on the same context afterwards as it won't come again by itself when you trigger it manually. (Otherwise you can always run a VM and crash the GPU on it to test in a real scenario.)

On this note, we should probably think about limiting the texture cache for those who stay on the club for a while. The more assets we had, the more often this will occur.

See issue [878403](http://www.hostedredmine.com/issues/878403)

### DIFF
--- a/BondageClub/Scripts/GLDraw.js
+++ b/BondageClub/Scripts/GLDraw.js
@@ -29,6 +29,16 @@ function GLDrawLoad() {
 
     CharacterAppearanceBuildCanvas = GLDrawAppearanceBuild;
 
+    // Attach context listeners
+    GLDrawCanvas.addEventListener("webglcontextlost", function (event) {
+        event.preventDefault();
+        console.log("WebGL Drawing disabled: GPU ran out of resources. If the context does not restore itself, refresh your page.");
+    }, false);
+    GLDrawCanvas.addEventListener("webglcontextrestored", function () {
+        GLDrawLoad();
+        console.log("WebGL: Context restored after it was lost.");
+    }, false);
+    
     console.log("WebGL Drawing enabled: '" + GLVersion + "'");
 }
 
--- a/BondageClub/Scripts/GLDraw.js
+++ b/BondageClub/Scripts/GLDraw.js
@@ -32,11 +32,11 @@ function GLDrawLoad() {
     // Attach context listeners
     GLDrawCanvas.addEventListener("webglcontextlost", function (event) {
         event.preventDefault();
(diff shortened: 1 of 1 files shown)


Ben987:
I don't think I've had this error before.  It happens when too many images are loaded in cache?  Is there a way we could know the remaining cache available so we can handle it?  Thanks.

ace-1331:
That's two separate problems 
1- Handling resources in an optimal way by limiting the cache and getting rid of the oldest items when it's full (items that haven't been seen in the longest time)
this would optimize the club's memory usage

2- No matter what we do, we have no control over a client's current resource usage overall. If someone's GPU runs out of resources, WebGL will lose its context. Once memory is emptied and WebGL regain its context, it can get back to rendering. However in the case of our 2D rendering, we never check to see if we have access to a new context, so the club will stop rendering things like it should.

This PR adresses issue 2, this assures that if such a case happens, the user does not have to reload to start seeing characters again. Of course, it is more likely to happen on slower/older machines.  

That's why testing is slightly difficult. To give it a try and see just how much the UI gets messed up -> you can simulate losing context by doing
`GLDrawCanvas.getContext('webgl2').getExtension('WEBGL_lose_context').loseContext();`
it won't be as bad as the actual thing if test it by forcefully crashing, but you can already see nothing gets rerendered.

If you want to test with this method, the window will not regain context until you explicitly give it back. To do so you need to do the following:
first, keep the extension of the current canvas in a variable to restore it later
`let ext = GLDrawCanvas.getContext('webgl2').getExtension('WEBGL_lose_context');`
then trigger a context loss manually
`ext.loseContext()`
at this step you have no WebGL context, meaning drawing characters won't work. You can test around to see it mess up. When you want to reattach the context, you need to do:
`ext.restoreContext()`
You'll notice characters resume being drawn normally with this fix, whereas the current version does not because we do nothing with the event.

Note that the current 3D engine does this as I've noticed while testing.

Another 2-window view of the graphical bug
Capture (1)

Ben987:
Thanks!  I've tried to recreate this scenario but could only replicate it manually.  I guess it won't hurt to have some code to reload the context if it was lost.  Cheers :)

PR #1028, opened 2020-07-05
Feature: Adds a "Suspension Cocooned" type to the web arms item

elliesec:
# Summary

Adds a new extended type to the web arms item: "Suspension Cocooned".

Credit to Cecilia for the new assets.

Preview of the new type:

kneeling-suspended-5


Ben987:
Very nice!  It looks great, thanks you two.

PR #1029, opened 2020-07-06
Medical Mask

Nina-1474:
Credit to **Z-freak** for the images.
 
A medical face mask in the gag slots. This won't actually apply a gag effect since it wouldn't muffle speech noticeably in reality. Is that alright? If not it could be tweaked to apply the minimum gag effect instead.


Ben987:
Sure, since it's not a random item that could break NPC scripts if it doesn't gag, it's perfectly fine.  Thanks you two!

PR #1030, opened 2020-07-06
Improvement: JSDoc comments added to Dialog.js

SandrineBC:
# Summary
I added JSDoc comments to the functions in `Creation.js` to help with further development.
No functionality was added or removed.

The result of the JSDoc output can be found [here](http://www.hostedredmine.com/projects/bondage-club-issue-tracker/wiki/Dialog)


Ben987:
Thanks!  The file is still easy enough to navigate with all these extra lines?  It doesn't make it too big?

SandrineBC:
I'd say, the gain all contributors get is worth the extra lines. A good IDE reads the comments and helps people that are using the functions. Plus, it makes the code easier to read...

Ben987:
Do we need the @description tag?  Some scripts have it and some don't.  Is it needed in most IDE?

SandrineBC:
The @description tag _can_ be omitted, but then the first lines of the block comment are used as a description for the function. Imho it's good style to use it, but omitting it breaks nothing.

ace-1331:
@desc or @description is optional when the description is above any other tags. 
I personally have no preference, but being more precise isn't a bad thing

Ben987:
I don't mind either.  But if it's optional I guess we can remove the description tag.  The less stuff we need to add to the files the better.

Ben987:
Also, to make everything unified and as short as possible, could we remove the "return" tag if the function doesn't return anything?  And could we remove the empty lines from the comments?  This will eat less space in the code for coders like me that don't use a IDE at all.

SandrineBC:
I can do that, but the empty lines make the comments more in my opinion. I would keep the returns though. Explicitly showing that a function does not return anything has its value...

ace-1331:
let me know what you end up settling for and I'll update my PRs with the same style.

- **Less stuff =/= better** having files with gigantic 1 liners or code blocks with no `{}` is counter productive.  Bad spacing can lead to unreadable code no matter how simple it is and even if it takes less lines in the end. That concept is inherently flawed.

-Not sure what to say about the IDE part. I am still shocked about it. I'd be in favor of those using proper tools to help speed up dev time instead of completely ignoring best practices or simple documentation.

Ben987:
Yep, I still use notepad++ to code the game, so no IDE on my side, one day I'll have to upgrade.  

But I agree that we should add IDE comments to help new programmers.  

Could we go using this standard so you have everything you need and we eat less space for old school coders like me:
- No @description tag, put the description right away
- No @return tag if there's no return value
- No empty lines

What do you think?

Thanks

ace-1331:
That works for me.
Whatever we settle for, we should not change it midway to have to go through all the files again in the future.

SandrineBC:
Ben, I talked to @elliesec and @ace-1331 about this.
We can omit the \@description tag, if you want and compress the empty lines. But all three of us think it's a good idea to stick with the mandatory \@return tag.

Ben987:
Is the @return tag helpful for you if the function doesn't return anything?  If so, we can leave it there.

Can you update your pull requests to go with that new setup?  I'm preparing a beta version and I'll review these changes soon.

Thanks!

elliesec:
The @return tag can be helpful in that it relays the intentions of the person that wrote the code - it's essentially saying "when I wrote this, I intended it not to return anything, and that's by design". It can encourage people to consider the design choices they're making a little more before making drastic changes.

SandrineBC:
Looks good from my side, Ben. Go for it.

ace-1331:
updated my 2 PR as well

Ben987:
Thanks for all the comments and fixing the spacing.  Cheers :)

PR #1033, opened 2020-07-06
Improvement: Server connection feedback on the login page

elliesec:
# Summary

This change adds some additional feedback on the login page for when there are connection issues with the server, which is a little more informative to players, and makes it a little more obvious what's happening when a login attempt isn't working due to a server problem.

This doesn't change the functional behaviour of login submission (you can still enter username/password and submit them at any time, and will be logged in when a connection has successfully been established).

* The login page will now display a "Connecting to server..." message before the websocket connection has been successfully opened.
    * If the connection succeeds, the standard "Enter your name and password" message will be displayed
    * If the connection fails, the "Connecting to server..." message will continue to display while socket.io retries
    * If the connection is still failing after a third retry, the login page will display a "Unable to connect to server. Retrying..." message
    * If at any point during the above process a connection attempt succeeds, the standard "Enter your name and password" message will be displayed, and the usual behaviour will be restored.


Ben987:
Absolutely, giving more server info to the player is a good idea.  I'll include that one for the next beta.  Thanks a lot.

PR #1035, opened 2020-07-07
Allow leg suspension on one's self

Ayesha-678:
Allows the suspension tie on the feet to be used on one's self, since we have the bondage skills, and we can suspend ourselves in a hogtie, this should only make sense.


elliesec:
Just a thought on this one, but might it be sensible to have a self bondage skill requirement for self-suspension rather than just allowing it outright?

Ben987:
Sure!  And adding the self-bondage as a requirement later on would also be a good idea.  Thanks :)

PR #1037, opened 2020-07-07
Fix: Adult baby harness mitten chain in the magic theatre corrected

SandrineBC:
# Summary
When a player uses the adult baby harness with mittens and chains, only the arms restraint is copied to another character in some magic performances, leaving the victim with floating chains as an arm restraint. This fix copies the harness and the mittens as well to the victim, so that no undefined state is reached.
This fixes [#880245](http://www.hostedredmine.com/issues/880245).

# Details
## Reproducing
- Go to the magic theatre
- Wait for the 'bind the assistant' or 'bind the magician' trick pops up
- Bind the other character with harness, mittens and chains
- After the trick, the new character is left with only the chains, without harness or mittens

## What it does
**`Female3DCG.js`**
- Added the `RemoveItemOnRemoval` attribute for the removal of the mitten chains to
  - Adult baby harness
  - Paw mittens
  - Padded mittens

**`Magic.js`**
- Added JSDoc comments to all functions
- Added a new function `MagicRestrainRemove` to remove duplicate code
- Changed the removal of items to the use of the `InventoryRemove` function, so that dependent items are removed automatically
- Changed the behavior of `MagicRestrainCopyTransfer` so that mittens and harness are also copied to the target character, if the mitten chains is detected as an arm restraint

## Testing
Tested the feature
- by reproducing the above mentioned steps
- the victim gets bound with harness, mittens and chains, as expected
- works for all restraint copying tricks


Ben987:
Thanks for all these changes, but since that code belongs to git4nick, we'll wait for his approval.  

@GIT4NICK , are you fine with this pull request?  Should I go ahead and merge it?  Thanks a lot.

Ben987:
@GIT4NICK , do you have a few minutes to review this pull request and give it a thumps up or a thumbs down?  Thanks a lot.

Nina-1474 (approved):
Longer term, I feel the mittens could be implemented better so that it doesn't require specific-item fixes like this. Such as making it an extended item where the added chain is another option which requires the arms to be free and blocks the arms.
For now though this does the trick (pun intended).

elliesec (approved):
I agree with Nina that longer-term a revision of the harness/chains/mittens combo might be a good idea to avoid having to do special casing like this, but this fixes the issue for now.

PR #1038, opened 2020-07-07
Fix: owner rule persisting

ace-1331:
- fixed a bug where the owner rule to block the sub from using remotes on herself would persist after breaking the collar


Ben987:
Great catch!  Thanks a lot :)

PR #1041, opened 2020-07-08
Update Female3DCG.js

Shiranui-Izayoi:
Changed some item DrawingPriorites for the sake of consistency as I mentioned in the Work In Progress channel. Also allowed the Reverse Bunny Suit Cloth version to expose some slots like the Suit Slot version can which I also mentioned.


Ben987:
Very nice!  Thanks a lot, I will try it soon :)

PR #1043, opened 2020-07-08
Adds .vscode directory to .gitignore

elliesec:
Not personally a problem for me, as I don't use VS Code, but I did just have to walk someone through some git to work around the fact that the directory isn't ignored. This should make life a little easier for some VS Code users.


Ben987:
Sure, it could be useful for some coders.

PR #1046, opened 2020-07-09
Bugfix: Adds missing lock assets for Corset2

elliesec:
Adds missing lock assets for Corset2 in several poses.


Ben987:
Nice catch!  Thanks a lot.

PR #1047, opened 2020-07-09
Bugfix: Adjusts the chatroom admin screen to prevent hover overlap with textarea

elliesec:
# Summary

When selecting a chatroom background via the back/forward buttons, the button hover displays underneath the ban list textarea. As it's currently impractical to display the hover text over the textarea (this would probably need another canvas to be temporarily rendered), this change adjusts the layout of the screen slightly to ensure that the hover doesn't overlap by slightly reducing the size of the background preview image. See screenshots below.

*Current layout:*

hover-overlap

*Adjusted layout:*

background-hover-fix


Ben987:
Nice one, I've never noticed that visual glitch.  Cheers!

PR #1049, opened 2020-07-10
Improvements/New: Extended item properties + Shocking Items

ace-1331:
This PRs does a lot at once because everything is related, here is the list in order:

### Improving extended items
- added 2 properties `AlwaysInteract` and `AlwaysExtend` to allow an extended item menu to be available if you are tied and/or if the zone is blocked

I'm hoping this will allow people to be more creative with extended items, and allow for more possibilities than just "types". This will allow us to have interactive items while a sub is tied for example. The fixes and items detailed below all rely on it to be an example.

If people wish to update the items they made to add more functionalities, they'll be able to. Though this mostly gives the opportunity to fix more of the existing items' odd behaviors and bring new kinds of items to the table

### Fixing shock items
- fixed the shock collar/unit to allow it to be accessible while it is blocked
Currently, even if a shock collar is remote controlled, you cannot change its intensity or trigger it if you tied your sub in a way that blocks the zone. With the new item properties, I made it so you can always interact with it, and also added the "trigger" button to the menu. 
- revamped shock items to use the same menu. (Changed the Draw/Click/Set function to be more dynamic)
currently, both items used a different menu builder which caused undesired differences like one of them being less intuitive to pick intensities. These 2 and the items I've added all use the same draw and click function to make it more maintainable. This will also make adding any other shock item way easier.

### Adding new items
- added a few items to leverage the new extended item properties: cupholder gag, bluetooth earbuds, shock clamps, shock vibrator, shock dildo

Most of these items have often been suggested and the shock collar issues have a frequently reported annoyance. I decided to make a new pair of earbuds to keep the current one as the version where you cannot change the volume while bound 

Thanks to Ayesha for the vibrator assets, clamps assets and plug preview. 
Thanks to Rui for the cupholder gag assets and earbuds preview


ace-1331:
just merged with the other plugs + tested to make sure everything was ok

Ben987:
Can you resolve the conflicts and test?  Lots of new stuff was merged today.  Thanks a lot :)

ace-1331:
ah, the bunny plug  fix conflicted once more, fixed ^

Ben987:
This is huge!  I probably won't be able to test all of it but since I want to do a second beta it should be just fine.  Thanks :)

PR #1050, opened 2020-07-10
Fix: OrnateCuffs hidden

ace-1331:
- fixed ornate cuffs with straitdresses

The strait dresses did not properly hide the ornate cuffs like other cuffs
https://discord.com/channels/554377975714414605/554378725916147722/730437559179804824

- fixed an oversight with the legbinders causing them to render the ornate cuffs when they should not

the leg binders also did not reference the ornate cuffs properly, causing them to be rendered when they were not intended to


Ben987:
Awesome!  I will test it right now.  Cheers.

PR #1052, opened 2020-07-10
Fixes for the Strait Dress and Removal of redundent ItemHides.

Shiranui-Izayoi:
Removed the HideItems from the MermaidSuit since all of them were ItemFeet which is already hidden via the normal Hide.

Added "LegsClosed" to any item that has "StraitDressOpen". Since there doesn't seem to be a reason not to and it will better respect functions looking for "LegsClosed" which "StraitDressOpen" is very similar to. If there is a reason not to then feel free to revert this the effect items are as follow. StraitDressOpen, SeamlessStraitDressOpen, Bitchsuit, BitchSuitExposed & MermaidSuit.

Removed "ItemFeetMermaidRopeTie" since the item no longer exists in game due to being merged with HempRope thus the hide is no longer needed.


Ben987:
This pull request creates a conflict with another fix to the strait dress.  Can you review please?  Thanks a lot.

Shiranui-Izayoi:
I assume you mean the fix for the ornate cuffs sure I can fix that. If I remember correctly I will only need to add in ItemFeetOrnateAnkleCuffs to the dress. I was thinking of including that but Ace said he already was. I will update the request soon.

Shiranui-Izayoi:
There I think that should fix the conflict if not I will update it again.

Ben987:
There are still 6 conflicts left in your pull request.  Can you review them?  Thanks a lot.

Shiranui-Izayoi:
Ok hopefully that should fix it. I really should have just made those changes myself in the commit instead of removing them when I heard Ace was working on it.

Ben987:
Awesome!  Everything seems to work now.  I will add it to the next beta.  Cheers :)

PR #1053, opened 2020-07-10
Fixes for the Short Chain to allow use with Bitchsuits & Strait Dresses 

Shiranui-Izayoi:
Removed the seemingly uneeded check for "StraitDressOpen" and replaced it with "BlockKneel" as it made more sense and still prevents the Mermaid Suit from having issues.


Ben987:
Thanks!  I've never noticed that one.  Cheers :)

PR #1054, opened 2020-07-10
Improvement: JSDoc comments added to Element.js

ace-1331:
- added JSDoc comments to Element.js

I also did a quick refactor of `ElementIsScrolledToEnd` to simplify it, we can return the value implicitly instead of a conditional return. It doesn't change the behavior as you can see in the diff.


Ben987:
Absolutely, it cleans the code a little.  Thanks!

PR #1055, opened 2020-07-10
Improvement: JSDoc comments added to Reputation.js + Refactor

ace-1331:
- added JSDoc comments to Reputation.js

- cleaned up the `ReputationSave` method

I also did a quick refactor to remove the `ReputationSave` method. It dates back from a long time ago as seen in the git history. It's essentially useless, you can see the parameter was not even taken into account. So the two spots it was used for were changed in favor of a direct call to `ServerPlayerReputationSync`


Ben987:
Sure, we can cut that middle function.  Thanks!

PR #1056, opened 2020-07-10
Bugfix: Hides boots (and some other items) with the new web cocoon

elliesec:
# Summary

The new web cocoon type doesn't currently hide the `ItemBoots` slot. This fixes that, and also hides a few other slots that shouldn't be visible.


Ben987:
Good one.  Thanks a lot :)

PR #1059, opened 2020-07-10
Allow feet suspension with chains

Ayesha-678:


Ben987:
Absolutely!  Thanks :)

PR #1060, opened 2020-07-10
Kneeling Hogtie

Nina-1474:
An option for the arms hemp rope and chains to tie a character on their knees with arms back.

I've also moved the prequisites checks into the option settings arrays and removed the no longer used ArmsOnly properties.


Ben987:
Sure!  The more poses the better :)

PR #1061, opened 2020-07-10
Improvement: JSDoc comments added to Skill.js

ace-1331:
- added JSDoc comments to Skill.js

I also removed the `SkillModifierReset` because it is actually never called anywhere in the whole project, it's also not very useful at the moment.

Allso, I knew about this but never quite realized how frequent these occured: `for-if-return loops` are less optimal than using `.find`, on top of `.find` taking less space for smaller operations. There could be a meaningful performance gain for core functions handling very large arrays like Dialogs and Assets

### DIFF
--- a/BondageClub/Scripts/Skill.js
+++ b/BondageClub/Scripts/Skill.js
@@ -7,7 +7,14 @@ var SkillLevelMinimum = 0;
 var SkillBondageRatio = 1;
 var SkillBondageRatio = 1;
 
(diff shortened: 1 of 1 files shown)


Ben987:
That's what I also thought, but strangely enough, the for loops I've tested are often faster than .find()   Maybe it's faster on some other browser though, I've only tested on Chrome.

Nice catch, I didn't know we had a dead function in there.  Thanks!

PR #1062, opened 2020-07-11
Handheld Toy Effect fix

Nina-1474:
A recent change to have assets inherit the Effect property from their asset group caused an issue with handheld toys. They're getting effects such as Prone on legs, which is preventing the message/activity from being sent to the chat.
The effect check will now look at the length so that the asset-group-level setting can be overridden at asset-level with an empty array.


Ben987:
Nice catch there, thanks a lot.

PR #1063, opened 2020-07-11
Lock Permissions Blocking

Nina-1474:
Unlike regular items, locks don't show as 'worn' in the inventory permissions screen at the moment, so they can always be changed. Since the lock validation removes locks the user has blocked, players can remove locks they're currently wearing by changing its permissions.
Now if a player is wearing a given lock anywhere, that lock will be greyed out in the permissions screen.


Ben987:
Yes, it makes a lot of sense, thanks a lot, I will try it right now.

PR #1064, opened 2020-07-11
Major Zip-Tie update

Ayesha-678:
Majorly updated the zipties, giving the existing images a much better asset, and many variations.


Ben987:
This is huge!  Thanks a lot, I'll try to test all of it soon.

PR #1065, opened 2020-07-11
Improvement: JSDoc comments added to NPC.js

ace-1331:
- added JSDoc comments to NPC.js

I never really looked at them before, so I learned quite a bit about NPCs which was interesting.


Ben987:
Thanks a lot, you're on fire.

PR #1066, opened 2020-07-11
Fix: Active pose

ace-1331:
- fixed an issue where item prerequisites were not taken into account when using the save state

The save state function did not reset the active pose. This means that many items were broken when they required the character to be kneeling/standing. Instead of bloating the asset files and creating a bunch of conflicts, and to be more futureproof, I've added the pose to the list of things it. This means all items will work retroactively, and new items won't need to worry about this feature.

To reproduce the current bug. You can wear a vacbed/stock/bed and many more in the main hall, go into a room, then remove the item, then kneel, then use your safeword. You'll be stuck

https://discord.com/channels/554377975714414605/731199817111568384/731478171349811230


Ben987:
Good catch!  Yes, we need to reset the pose also.  Thanks!

PR #1067, opened 2020-07-11
Translation: Tools fix

ace-1331:
- fixed the translation tools reference

Since the translation tools have been moved up a folder, they were broken because the relative paths pointed to nowhere
I fixed them and used them to update most of the existing french files that had missing/changed lines


Ben987:
Absolutely!  We need to focus more on our foreign language users.  Cheers :)

PR #1068, opened 2020-07-11
New_items_Hat_Masks_ItemButt

BunnyMadMan:
A set of new items ready to be added to the game;

Under group Hat : - SmallHat, Veil1, Veil2 
 Comment : 
- all three items were tested and shouldnt cause any problems with the existing items

Under group Masks: -BunnyMask1, BunnyMask2, BunnyMask3, KittyMask1, KittyMask2, LaceMask1, LaceMask2 
Comment : 
- as before all masks were tested; in the current situation the masks have a special priority added to them, so that they appear 
  over the hair to make them more visible (tested for them to be under the hair, but looked very weird). Still the hair is not 
  hidden, only the group Glasses and certain items (look bellow). This can still be adjusted later on, according to the opinion of 
  the player base, 
- in some cases there were problems with the interaction of already existing items with the masks; because of that certain items 
  were made to be hidden in the situation that a player uses that mask,  
- this is mostly in case of the Bunny Masks, where the mask ears could cover HairAccessory items or it would be illogical; in cases 
   of this happening, the troubled item was made hidden (this can still be adressed later on, depending on the player reactions).

Under group ItemButt: - BunnyTailVibePlug, EggVibePlugXXL 
Comment: 
- both items have a script added to the game coding, aswell as their own chat lines,
- they were both tested to see if they work with the current arousal meter and after testing there werent any troubles (in case of 
  any other problems, they can be still fixed later on).


Ben987:
This is great!  Lots of new items.  I'll try them right now.  Cheers :)

PR #1070, opened 2020-07-11
Removes Next button for ziptie items with 4 or less options.

Ayesha-678:


Ben987:
That was quick!  Thanks a lot, cheers :)

PR #1071, opened 2020-07-11
Fix: new bunny plugs

ace-1331:
while testing, I found the bunny plugs could be used through zipped suits, so I added the extra prerequisite like the other plugs


Ben987:
Very nice!  Thanks.

PR #1072, opened 2020-07-12
Ziptie Kneeling Hogtie

Nina-1474:
Adding the new Kneeling Hogtie option to the zipties as well, along with the prequisite & property cleanup.
Also made minor corrections to the image files and preview icon sizes.


Ben987:
Great!  Thanks for that extra option with zipties :)

PR #1074, opened 2020-07-12
Improvement: Proof of concept for the usage of button event handlers

SandrineBC:
# Summary
One of the main issues, we currently face, is the increased usage of gpu time, by constantly redrawing the whole canvas of the game, even if nothing spectacular is happening, like in the Information Sheet screen. To tackle this issue, we should try to move more and more implementations that were done 'by hand' to the existing JavaScript standard mechanics.
This proof of concept shows how a first step can be done for the numerous buttons in the game.
# Details
## What it does
Changes are done to three files
- `Styles.css`
- `Element.js`
- `InformationSheet.js`

The details to these changes are explained below

### `Styles.css`
- Added styles for the standard buttons and the behavior of tool tips (hovering text, when the mouse pointer is placed over a button).
### `Element.js`
- Added the function `ElementCreateButton` that works like the other `ElementCreateXXX` functions. It draws a button, adds a label or image to it and - most importantly - takes a function as an argument that is attached to the click event handler for that button
- Changed the function `ElementPositionFix` to a more gently setting of element styles. This makes the usage of CSS possible
### `InformationScreen.js`
- Added the function `InformationScreenLoad`. That's a standard function that is called dynamically, as soon, as a screen is loaded/ called for the first time. Within this function, the buttons are created.
- For sake of argument, all possible positions for the tool tips are used, knowing full well, that they will not fit on the screen fully, when the position "right" is chosen
- Added the function `InformationSheetButtonsRemove` that handles the removal of the buttons. This function is called from `InformationScreenExit` but also, every time a sub screen is called.
- Changed the function `InformationScreenRun` so that the already existing buttons are just repositioned and not recreated, every time the function is called.
- Changed the function `InformationScreenExit` so that it removes the buttons, when the screen is left
## Testing
- I tested that on my local version of the club and it runs smoothly. 
  - iOS 10.15.5
  - Chrome 83.0.4103.116
  - Safari 13.1.1
- All used technologies are standard and supported by all common browsers
- I did not take measurements on GPU or memory usage, but as the buttons are now just repositioned and not drawn new, every frame, the change should reduce GPU time
### Pros
- By using CSS for buttons and tool tips we get a central point to change the look and behavior of a GUI element that is used at various places. Adding a lot to the maintainability of the code
- The usage of event handlers makes the function `InformationScreenClick`obsolete and reduces the possibility of errors, because you don't have to calculate where your button is right now, when the user clicks on the screen
- Tool tips are easily added and handled by the browser instead of having them drawn manually, when `<Screen>Run` is called
- As the buttons a drawn by the use of a separate function we can apply this change gradually and don't have to change huge chunks of code as a whole to transition to a - code wise - better version
### Cons
- You have to manually remove the buttons, else they will stick around
## Further development
If this change is accepted, the next goals should be 
- dynamically calculate the best position for a tool tip
- change `ElementPosition` to set the element's style more gently, so we can use CSS attributes 
- remove the constant redrawing of the buttons to a function that's called, when a resize event occurs, thus reducing more and more functionality from the `InformationScreenRun` function that is called for every frame


elliesec:
This looks nice. Reduces the amount of drawing required of the canvas, and allows us to tie button functionality to the buttons themselves, which would help avoid silly mistakes like #1076.

Ben987:
I've tried using regular HTML buttons back then but it didn't feel as integrated as drawing them on the canvas.  Do you also get that feeling when you use the interface with regular buttons?  Thanks.

SandrineBC:
At the current state of development, you see no difference at all between the HTML buttons and those you draw directly on the canvas. Even the resizing works fluently with the current functions. Only difference you'll see are the tool tips, which have rounded corners, just for show ;-)

elliesec:
There's very little difference visually between the new HTML buttons and the existing ones (as Sandrine said, the tooltip corners are rounded) - if we wanted to make them visually indistinguishable, that's also absolutely possible with a bit of appropriate CSS.

One thing to note is that currently the buttons have different tooltip positions (I presume for demonstration purposes) - it'd be worth making those all consistent before merging the change.

This is what it looks like visually:

html-buttons

SandrineBC:
Yes, as @elliesec said, I just wanted to demonstrate the different positioning options. The final look of the tool tips can be altered later, as we only have to change the CSS for this.
Perhaps you can merge the change to your test environment and give me a thumbs up/ down decision? Then I would correct the tool tip positioning and commit the changes to this PR.

Ben987:
It's too late in the cycle to test it right now, but once the full version is release I'll give it a go to see how it works.   There's tons of buttons in the game though and reworking & testing all the screens will take many, many hours.  Maybe our time could be invested in features that will impact the players.

SandrineBC:
In the long run, this will impact all players, because we're decreasing the GPU/CPU usage on the client side, making the game more stable for all players and make the development of new features easier.
So I'm happy to follow along, as I'm not so good at implementing new stuff...

elliesec:
I agree that it's too late for this release cycle, as we're already in beta. But as Sandrine said, this has an impact on the long-term health of the game in my opinion. Overhauling every button in the game, as you say, would take a whole load of work and testing, but it's the sort of thing that can be done slowly, a little at a time, I think keeping a healthy balance of code/performance improvements like this and new content/features for users is an important balance to strike.

Ben987:
I've ran a few performances test and drawing the buttons we have right now doesn't impact the frame rate of the game.  Since using dom buttons could create problems on some browsers or when using some extensions lets not go that route.  Especially if we want to merge the game to another non-browser engine someday.  Thanks for trying that possibility though :)

dgrp:
Using DOM buttons would've been nice to make the game more accessible for people who prefer using their keyboards for whatever reason. Right now, all buttons need mouse or touch input as there's no way to use the tab or arrow buttons to navigate between them. It would've also been a nice separation of concerns between gameplay and its UI, paving the way for easier to understand and maintain code.

Regarding extensions... as someone who wrote a small extension for the game; the game does not offer official support for extensions at this moment, so I wouldn't hold back on new features because of that. If you want to officially support extensions, you could consider creating (or letting someone create) an API of some sort which extensions can use.

Personally, I saw the potential of this improvement and I hope that it's reconsidered someday.

MrBondt:
Using DOM buttons would've been nice to make the game more accessible for people who prefer using their keyboards for whatever reason. Right now, all buttons need mouse or touch input as there's no way to use the tab or arrow buttons to navigate between them. It would've also been a nice separation of concerns between gameplay and its UI, paving the way for easier to understand and maintain code.

Regarding extensions... as someone who wrote a small extension for the game; the game does not offer official support for extensions at this moment, so I wouldn't hold back on new features because of that. If you want to officially support extensions, you could consider creating (or letting someone create) an API of some sort which extensions can use.

Personally, I saw the potential of this improvement and I hope that it's reconsidered someday.

SandrineBC:
Hi @Ben987 
Personally, I see this as @MrBondt 
The pros outweigh the cons by far. And of course, just the removal of the buttons from the redraw functions would not yield such a big performance gain, but the game will win on the stability side and you reduce possible sources for errors, as mentioned above with the thing that happened to @elliesec 
And if you plan on someday porting this game to a non browser engine, you probably will use click event handlers anyway, so slowly porting the game to this concept will help you on this point as well.
I don't want to bug you, but perhaps you really should reconsider incorporating this in the game.

PR #1075, opened 2020-07-12
Bugfix: Login screen fix + login screen JSDoc

elliesec:
# Summary

This fixes a bug in the login and relog screens where players could not reattempt a login after a failed login without reloading the page. It also slightly simplifies the login message handling logic, and adds JSDoc documentation to the login, relog and password reset files.

# Steps to reproduce

1. At the log in page, enter an incorrect username and/or password
2. Press enter or click the login button
3. The "Invalid username or password" message is shown as expected
4. Now enter the correct username/password and press enter or click the login button
5. Note that nothing happens

The same issue is present on the relog screen.


Ben987:
Sure, I will try to soon.  Thanks!

PR #1076, opened 2020-07-12
Bugfix: Corrects the coordinates for click handlers in the chatroom admin screen

elliesec:
# Summary

Follow-on fix to #1047 -  I moved the buttons, but forgot to move the click handlers for them! This fixes that.

This also removes a declaration of the `ChatAdminPrivate` variable, as it was being declared in two places (lines 4 and 7).


Ben987:
Alright!

PR #1078, opened 2020-07-12
Improvement: JSDoc comments added to Timer.js

ace-1331:
- added JSDoc comments to Timer.js

I've found several unoptimal functions I was semi-aware of before, I've listed them in [#880890](http://www.hostedredmine.com/issues/880890?issue_count=57&issue_position=10&next_issue_id=880188&prev_issue_id=878339)

We'll have a look at ways to optimize these in order to improve performances (hopefully).
I was thinking we could merge the Time to String functions into one, then use an optional parameter for the least used formats

On that note, I noticed there was still a function that didn't use the newer `ItemUpdate` function. I tested and it seems to have no side effect other than being faster. (We don't need to sync the whole appearance of a character just to add a removal timer for a lock/expression)

There is also a huge technical debt involving the character array. Each time someone relogs, their online ID changes and an extra character is added to it without removing the old one (or just changing the old one), this makes it so your Character array can easily get in the 100+. Having to iterate through this constantly while in a room can make the club slow down with time. Ideally we'd need to just keep the characters we need in there, and while in a chatroom we should only iterate through the characters we see

Let me know what you think

### DIFF
--- a/BondageClub/Scripts/Timer.js
+++ b/BondageClub/Scripts/Timer.js
@@ -7,7 +7,11 @@ var TimerLastArousalProgress = 0;
 var TimerLastArousalProgressCount = 0;
 var TimerLastArousalDecay = 0;
 
-// Returns a string of the current remaining timer
+/**
+ * Returns a string of the time remaining on a given timer
+ * @param {number} T - Time to convert to a string in ms 
+ * @returns {string} - The time string in the DD:HH:MM:SS format (Days and hours not displayed if it contains none)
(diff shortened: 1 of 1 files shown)


Ben987:
It's too late to merge & test it now since I want to release a new version soon.  But using the optimized call should be a good idea, we'll do in the next version.  Thanks :)

Ben987:
It should save some network, traffic.  Thanks a lot!

PR #1080, opened 2020-07-12
Moves the two new latex leotards into the same BuyGroup

elliesec:
# Summary

As suggested, I've moved the two new latex leotards into the same BuyGroup - speaking with Cecilia, they were intended to effectively be two views of a "reversible" leotard - one worn backwards, and one forwards, so I think this makes sense (and will help players out financially).


Ben987:
Yes!  It should help players with $$$

PR #1082, opened 2020-07-12
Improvement: JSDoc comments added to Text.js/Import.js

ace-1331:
- JSDoc comments added to Text.js/Import.js

These two files are small, happily I found nothing odd to point out in these!

Hoping to have gotten through most of core script files soon, there are a few long ones left before that part will be done


Ben987:
Awesome!  Thanks :)

PR #1085, opened 2020-07-13
Improvement: JSDoc comments added to Minigame.js/Audio.js

ace-1331:
- added JSDoc comments to Minigame.js/Audio.js

Another 2 small core files commented. Slowly but surely getting to something quite useful to help people get into the project.

My only comment on this one is:
All the sounds are hardcoded in Audio.js... I think it would be a lot nicer to have Asset properties and simply be able to name the sounds there so `AudioPlayContent` only has to dynamically trigger a given sound. This would also make it easier for item makers to add a sound to their items if they already don't know how to

Unless there are any objections to this, we'll probably get to it at some point, created a task for this on the board a few of us have been using

### DIFF
--- a/BondageClub/Scripts/Audio.js
+++ b/BondageClub/Scripts/Audio.js
@@ -1,7 +1,12 @@
 "use strict";
 var AudioDialog = new Audio();
 
-// Players a sound with a set volume
+/**
+ * Plays a sound at a given volume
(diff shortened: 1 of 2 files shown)


Ben987:
Yes, removing the hardcoded stuff would be nice someday.  Thanks!

PR #1087, opened 2020-07-13
Improvement: JSDoc comments added to Common.js

ace-1331:
- added JSDoc to Common.js

other than CommonIsClickAt being severely under used, a few things look a bit odd in Common.js, biggest one is how `CommonDynamicFunctionParams` seems like it could break easily if we need more than 3 parameters or if a parameter contains a comma. 

I keep the PRs separate so they are smaller chunks, easier to keep track of/discuss and to make sure two people don't work on the same file. Also, sometimes there are small refactors so it's better to keep them separate in those cases. Let me know if you prefer bigger PRs for JSDoc instead of several small ones like this

And as always, let me know if any comment can be improved.


Ben987:
That looks fine, thanks a lot!

PR #1090, opened 2020-07-13
Improvement: Removal of Ben-Wa balls from the anus trigger a sexual activity

SandrineBC:
# Summary
Very simple. If a ball is removed from the player's anus, a sexual activity is triggered.
# Detail
## What it does
Small change to `InventoryItemButtAnalBeads2SetBeads`in `AnalBeads.js`. 
- The number of removed beads is calculated and for every removed bead a `MasturbateItem` activity is performed on the player's anus. For good or for worse.
- No additional message is sent to the chat window, except the one that's already sent.
## Testing
Tested it on my personal version of the club. It works with one or two players.


SandrineBC:
There seems to be a problem to propagate the effect of the arousal to all other players in an easy way.

PR #1095, opened 2020-07-14
Improvement: Character array optimizations

ace-1331:
- optimized how online characters are saved to removed instances of characters that are no longer present within your room. 

### The reason
A few functions require to loop through the character array, and it would quickly grow over several hundreds easily the longer you stayed online and the more rooms you went in.

There was also an oversight where if you log back in, a new entry would be created for you in everyone's character array since your online ID changed.

### Potential solutions
There are a few ways to go on about it, I thought about just fixing the duplication problem, but then I thought about saving only X amount max and get rid of the old ones, but in the end, I found no good reason as to why we would need to keep online characters that are not within the room so I went with trimming as much as possible to optimize things.

Nowhere in the code it seemed to be useful to keep track of all online players met. We already receive the data, so this is not like we would query the server more frequently

### Explanations
This way, it means there will ever only be NPCs you saw and the current chatroom's character inside the array, so it will never be in the 100-1000s and will reduce the time it takes to iterate through the array since there wont be 100s of useless iterations

I've tested with several accounts by leaving/entering/dcing, joining rooms and moving around. There was no visible side effects from this change

If you have a preferred way of handling this, or if you have any ideas as to why we would need to save hundreds of characters, let me know. I'll alter this PR if we identify something better to do with this.

### DIFF
--- a/BondageClub/Screens/Online/ChatRoom/ChatRoom.js
+++ b/BondageClub/Screens/Online/ChatRoom/ChatRoom.js
@@ -826,6 +826,10 @@ function ChatRoomSync(data) {
 			if (ChatRoomCharacter.length == data.Character.length + 1) {
 				ChatRoomCharacter = ChatRoomCharacter.filter(A => data.Character.some(B => A.MemberNumber == B.MemberNumber));
 				ChatRoomData = data;
+				// Remove non-present chatroom characters from the characters array as they are no longer needed
+				for (var C = 0; C < Character.length; C++)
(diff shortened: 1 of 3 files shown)


Ben987:
Thanks, but we'll need to wait for the next version to be fully "hotfixed" before I can check that one.

ace-1331:
sure, no worry. Ill merge the upstream when needed so it'll be ready to test whenever its time

Ben987:
It won't be easy to measure the performance increase after a while but I'm sure it could help.  Thanks a lot!

PR #1096, opened 2020-07-14
Lighter latex

Codefreak64:
This turns the latex a lighter shade of black, making it easier to re-color

It keeps the shine of the latex while enhancing the coloring abilities


Ben987:
Thanks!  But it will need to wait for the next version, I'm too close to release date.

Ben987:
Alright!  I hope we will get even better colors with that change.  Thanks a lot :)

PR #1097, opened 2020-07-14
Fix: Player get deaf level

ace-1331:
- fixed an issue with `Player.GetDeafLevel` that caused extended item effects to be overlooked

The get deafness method never took extended properties into account which prevented headphones from working properly (both old and new ones) This will make it more futureproof


Ben987:
Absolutely!  I'm uploading the release version right now but I'll add this as a "pre-hotfix".  Thanks :)

PR #1101, opened 2020-07-14
Fix: Priorities for hair accessories

SandrineBC:
I'm terribly sorry, but I overlooked the hair accessories in the gag fix.
Thus I had to re-assign most of the priorities again, to keep the 'old distances' between the various groups.


Ben987:
Good catch!  I'll prepare a pre-hotfix-hotfix :P

PR #1105, opened 2020-07-15
Fix: Maid responds wrongly if player cannot be freed

SandrineBC:
# Summary
If a player, restrained with a locked item, asks the maid in the maid quarters to free her, the maid responds, that she freed the player, but in fact did nothing (or removed all safe the locked restraints from the player).
With this fix, the maid correctly shakes her head and does nothing.
This fixes [Bug #881335]( http://www.hostedredmine.com/issues/881335)
# Details
## What it does
This fix changes two functions in the script
* added a check for locked restraints in `MaidQuartersCanReleasePlayer`
* changed MaidQuartersCannotReleasePlayer` to return the inverted result of `MaidQuartersCanReleasePlayer` so that code maintenance gets easier
## Testing
Tested that on my local instance:
* a player with any locked item is sent away
* a player with unlocked items get freed

### DIFF
--- a/BondageClub/Screens/Room/MaidQuarters/MaidQuarters.js
+++ b/BondageClub/Screens/Room/MaidQuarters/MaidQuarters.js
@@ -2,7 +2,7 @@
 var MaidQuartersBackground = "MaidQuarters";
 var MaidQuartersMaid = null;
 var MaidQuartersMaidInitiation = null;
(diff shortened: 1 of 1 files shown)


Ben987:
Nice catch!  But since there's a few extra edits in the file, I won't push it as a hotfix, it can wait until the next version.  Thanks :)

Ben987:
No, MaidQuartersCanReleasePlayer() isn't the reversed version of "CannotRelease" since both function needs to check if the player is restrained.  Check the fix I've just did to correct that situation.  Thanks for reporting, cheers :)

PR #1106, opened 2020-07-15
Bugfix: Changes the latex leotards to hide clit items

elliesec:
# Summary

I've noticed some small clipping issues between the new (and existing) leotards and clit items (notably the heart piercing). As there is no clit item that should reasonably clip out of the leotards, I've added `ItemVulvaPiercing` to the `Hide` array for the leotards.


Ben987:
Nice little catch.  I will prepare a hotfix, thanks :)

PR #1108, opened 2020-07-16
Improvement: Pulling of Ben Wa balls trigger arousal

SandrineBC:
# Summary
Pulling one or more beads of the Anal Bead XL item now trigger a sexual activity. This is done repeatedly, for every bead that is pulled, up to four times max. The inserting does not trigger a sexual activity.
These changes could be used as a basis for further changes along that line, like the Ben Wa balls on the vulva.
# Details
## What it does
Four files were changed for this effect:
### `Female3DCG.js`
* Added the `Activity` property to "AnalBeads2" so that the right sexual activity is triggered, when a bead is pulled
### `AnalBeads2.js`
* Enhanced the directory that is pushed to the server. This is needed to propagate the action correctly to other players in the room, mainly, if player B pulls the beads from player A
* Added a loop that calls `ActivityArousalItem` for every bead, if the player pulls beads from her own butt
### `Activity.js`
* Added a new optional parameter to `ActivityEffect` that multiplies the calculated factor, if an activity is done repeatedly
### `ChatRoom.js`
* Reading the enhanced dictionary in function `ChatRoomMessage` and
* Passing the count for the activity to `ActivityEffect`
## Testing
Tested that on my local instance of the club with a player pulling beads out of her own butt, as well as having another player do that:
* Inserting of beads does not trigger anything
* Pulling of beads triggers the sexual activity according to preferences
* Pulling of several beads multiplies the effect
* If the activity is blocked, nothing happens
# Credits
Thanks to @Nina-1474 for giving hints and @elliesec for a review of my first approach and her patience


Ben987:
This one causes conflict in ChatRoom.js, probably from some other changes we added recently.  Can you double-check please?  Thanks.

SandrineBC:
Hi @Ben987 
I merged the branch and it still works on my side.

Ben987:
Thanks for that one!  Multiplying the factor by the count might be a little too much though, it will often cap at 25 after the first one that way.  I'll see if I can reduce that multiplication a little.  Cheers :)

PR #1109, opened 2020-07-16
Improve: Coloring of unusable restraints

SandrineBC:
# Summary
This change applies a color code on restraints that cannot be used on a character, at the moment, either because of worn clothes or other restraints that are currently in use. The items in the inventory are additionally sorted in the following order

1. The worn item
2. Items that can be used (white)
3. Items that cannot be used (gray)
4. Items that are blocked (red)

This solves the issue [#881151](http://www.hostedredmine.com/issues/881151) and is based on a proposal by [Senken](https://discord.com/channels/554377975714414605/554378904308023297/732294616182816831)

# Details
## What it does
All changes were done in file `Dialog.js`
* changed the hard coded sort order of different groups to constants so that additional groups can be added later more easily
* added a distinction of cases in function `DialogInventoryBuild` so that items that are not blocked, but not applicable because of their prerequisites are sorted into different groups
* changed the coloring of the different item groups, so that blocked items are shown in red and unusable items in grey
## Testing
Tested this on my local machine with one and two players. In both cases the change works as it should.

### DIFF
--- a/BondageClub/Scripts/Dialog.js
+++ b/BondageClub/Scripts/Dialog.js
@@ -590,10 +590,10 @@ function DialogInventoryBuild(C) {
 			if (C.ID != 0)
 				for (var A = 0; A < Player.Inventory.length; A++)
 					if ((Player.Inventory[A].Asset != null) && (Player.Inventory[A].Asset.Group.Name == C.FocusGroup.Name) && Player.Inventory[A].Asset.DynamicAllowInventoryAdd(C))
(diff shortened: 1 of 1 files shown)


SandrineBC:
Created "PR #1109":https://github.com/Ben987/Bondage-College/pull/1109

Ben987:
That's a great idea, it will be convenient for players to know if the item can be used in advance.  I see you have a function that returns a value that's not used anywhere though, but I can clean that up.  Thanks a lot!

PR #1110, opened 2020-07-16
Bugfix: Missing text and extended item menu fixes for the leather armbinder

elliesec:
# Summary

This fixes a couple of issues with the leather armbinder:
* There is no available dialog text for the "wrapping straps" option
* The text for both the ring & wrapping straps were being displayed in the same location (on top of each other)
* Both options would always appear greyed out when viewing the armbinder's extended item menu.

Image of the bug here (it doesn't always manifest in exactly this manner because of the missing dialog line):

Armbinder Bug


Ben987:
Nice fix.  This one has been there for a while now.  Cheers!

PR #1111, opened 2020-07-18
Fix: zipped suits behavior

ace-1331:
- fixed an issue where the suits would not load with their intended effects

The normal petsuit, the latex lockdown suit and the mermaid suit currently render the zip state first but do not block the zones until we unzip/zip them again. This fixes the issue


Ben987:
Thanks a lot!  I didn't know about that one.

PR #1116, opened 2020-07-19
Improvement: JSDoc comments added to Drawing.js 

ace-1331:
- added JSDoc to Drawing.js

Unless I forgot one, all main modules in /Scripts should now be covered. Since they are the baseline for the whole club, it should help people understand things better if they have a look. I'm not too preoccupied by the rooms and such, so I'll probably just do them as I change or add things in those files.

I removed some unused parameters from certain functions.. However I believe the following functions are no longer used. I left them in just because, though we could probably get rid of them (most likely due to how suspension drawing was changed a little while back)
- `DrawImageZoomMirror`
- `DrawImageMirror`


Ben987:
Of course!  Thanks :)

PR #1118, opened 2020-07-19
Improvement: Auto focus text area in chat

SandrineBC:
# Summary
When a player changes clothes, applies a restraint or activates the whisper function (for example), the text input field looses its focus and the player has to actively click into it, before entering their chat message.
This change catches the keyDown event and activates the text area before sending the key there, thus helping the game flow.

# What it does
* added an if statement to `ChatRoomKeyDown` in the `ChatRoom` script that transfers focus to the text area

# Testing
* tested it locally with one player, as this only changes the client behavior


Ben987:
Thanks!  That should be handy.  And thanks for cleaning the white spaces.

PR #1121, opened 2020-07-19
Fix: Asset cleanup

ace-1331:
- removed duplicate prerequisites on a few items
- fixed hideitems for several pieces of clothing (duplicate and/or not labelled properly)
- changed the hobbleskirt prereq to allow using it while kneeling (right now you need forcekneel to kneel with it)

This fixes a lot of redundancies and incorrectly labelled hideitems for a few cloths. 

The only behavioral change is the hobbleskirt, it can already be used while kneeling but only if you use forcekneel (short chain, etc.) This will make it easier/more intuitive to use


Ben987:
Yes!  It seems to make sense, I'll try to test it soon.  Thanks :)

PR #1123, opened 2020-07-19
Bugfix: Changes gag level calculation to be more consistent

elliesec:
# Summary

([Testing Document for this change](https://gist.github.com/elliesec/e97c8165de2fcfb309d66efa41197d81))

This PR fixes some issues with several extended gag items by reworking the `SpeechGetGagLevel` function. Previously, gag effects defined in `Property.Effect` would stack with gag effects in `Asset.Effect`. This lead to some slightly confusing behaviour where a gagging item with `AllowTypes` would behave differently upon initial equip than when selecting the default (null) type. This affects the following items:
* Dildo Plug Gag
* Duct Tape (all 3 slots)
* Plug Gag

This pull request changes `SpeechGetGagLevel` so that gag effects in `Property.Effect` _override_ gag effects in `Asset.Effect`, which is more in line with the way other effects work and should prove less confusing for asset creators in the future.

Extended items have been re-adjusted to match their previous values where possible, but I've taken a couple of liberties with some items:

* Slightly reduced the level of the unplugged version of the Dildo Plug Gag and the standard Plug Gag
* Increased the gag level of the plugged version of the Dildo Plug Gag to give it a higher gag level than the plugged version of the standard Plug Gag
* Decreased the gag level of the Small version of the Duct Tape gag in all three slots to make it mechanically distinct from the Crossed version

The pull request also makes some efficiency improvements to the `SpeechGetGagLevel` function:
* Creating a new internal function every time the `SpeechGetGagLevel` function is called is unnecessarily expensive - the new function uses a static lookup to get gag levels from effect text
* Using a lookup is faster than the previous `if`/`else if` construct
* The new function uses a `some` loop to end execution early once the correct asset group has been found, rather than iterating through every appearance item

## Reproducing the bug

The issue is present on all of the gags mentioned above, but is most noticeable on the Dildo Plug Gag.

1. Equip the dildo plug gag (do not select a type)
2. In the console, notice that `SpeechGetGagLevel(Player, "ItemMouth")` evaluates to 3
3. Change the type of the dildo plug gag to "Plugged"
4. In the console, notice that `SpeechGetGagLevel(Player, "ItemMouth")` now evaluates to 10
5. Change the type of the dildo plug gag back to "Open"
6. In the console, notice that `SpeechGetGagLevel(Player, "ItemMouth")` now evaluates to 6, rather than the 3 from earlier

The following demonstrates the above - in between each change to the gag, the character named "Test" is saying "Testing":

> Test: Testing
> (Test Two uses a dildo plug gag on Test's mouth.)
> Test: eeseinh
> (Test Two plugs up Test's mouth with the dildo plug.)
> Test: mmmmmmm
> (Test Two pulls out Test's dildo plug from her gag.)
> Test: eeheemm

Note that the second and fourth times come out differently (contrary to what you would expect).

## Testing

Obviously, this change will affect every gag (and gagging item) in the game, so I've been very thorough about testing it. I've tested the result of `SpeechGetGagLevel` for every gagging item in the game via a script, and recorded the results. Extended items have all been tested manually, and every extended type has been tested.

The following document contains a full list of the before/after values of running `SpeechGetGagLevel` on every asset in a potentially gagging slot:
* [Gag Level Testing Document](https://gist.github.com/elliesec/e97c8165de2fcfb309d66efa41197d81)

The key results are:
* The gag levels for every basic item have remained the same
* The gag levels for extended item types have been adjusted to keep the behaviour similar to their previous behaviours


Ben987:
Thanks!  Is there any simple way we could rename the properties or it would be a huge chore?

	GagTotal4: 20,
	GagTotal3: 16,
	GagTotal2: 12,
	GagTotal: 8,
	GagVeryHeavy: 7,
	GagHeavy: 6,
	GagMedium: 5,
	GagNormal: 4,
	GagEasy: 3,
	GagLight: 2,
	GagVeryLight: 1,

This is not very intuitive.  There must be a better system to label the gag effectiveness.

Also, could you keep the for loop instead of using .some, I don't think we're using .some anywhere in the code right now.

Thanks a lot!

elliesec:
Hmm, I agree they're not particularly intuitive. Did you have a naming convention in mind. I was thinking it might be preferable to just name them something like:
```
GagLevel1
GagLevel2
GagLevel3
...
GagLevelX
```
This would mean renaming them in `Female3DCG.js` as well, and checking for other usages in the code. It would be a bigger task for the lower values (GagVeryLight - GagMedium) I think, because they get used a lot. A lot of the higher values don't get used frequently (I actually don't think `GagTotal3` or `GagTotal4` get used at all). If you'd like me to look at doing it, I can certainly take a look, although I could see it causing merge conflicts with other things - let me know.

I'll also amend the pull request to remove the `.some`.

elliesec:
Hi Ben, I've made the `.some` loop change. I think if you'd like me to look at changing the gag effect names, it's probably best done in a separate pull request, otherwise this one will probably get unnecessarily large - let me know.

Ben987:
Yes!  And I like the way you cleaned the code.  We can rename the different gag levels later so it makes more sense.  Thanks :)

PR #1128, opened 2020-07-20
NEW ITEM & POSE(S): X-Cross, OverTheHead pose & Spread pose

Nythaleath:
Added in a St. Andrews Cross along with 2 poses that it uses, one for the upper body (OverTheHead) one for the lower body (Spread)

Note: a LOT of the assets for the new poses are sub-par and are there as place holders until someone else can make a better looking version

Credits:
 Pose(s): Kirsty & Nythaleath
 Clothing Assets: Kirsty, Ruilove, & Nythaleath
 X-Cross Asset: Sybil
 Item Assets: Nythaleath
 Adding Items into the code: Nythaleath


Ben987:
Wow!  This is INCREDIBLE!  Thanks a lot everyone.  I won't be able to test all the possibilities so please test it again with the latest version and create new pull requests with fixes if we need some.  Cheers :)

PR #1133, opened 2020-07-21
Improvement: Newer collars for the slave collar

ace-1331:
- added the 2 newer collars to the slave collar


Ben987:

PR #1134, opened 2020-07-21
Renamed ItemHead zone

ace-1331:
Renamed the zone to "Eyes" and adjusted some dialogs as discussed


ace-1331:
also did some more testing with newer stuff and found the long back hair that goes in front didnt work with the new vacbed priorities so I also slipped a fix in for that

PR #1135, opened 2020-07-21
Fix: Framerate in the tennis mini game

ace-1331:
- fixed an issue with 144hz monitor and the tennis minigame

The current tennis minigame moves the ball forward with every frame, causing it to move WAY too fast on 144hz monitors. I dont have one myself but I've tested with a 30FPS  cap on 60FPS, and the ratio method makes it so the ball moves smoothly. This makes the minigame playable when the current  FPS is higher than ~60.

At first I wanted to skip a frame, but that actually didn't look smooth. Capping the club to 60FPS would have been an option, but I don't think it would have been needed.

If there are other games affected by this, a similar fix should work. I'm not knowledgeable in singleplayer stuff, so if anyone mentions them, I can have a look.

### DIFF
--- a/BondageClub/Screens/MiniGame/Tennis/Tennis.js
+++ b/BondageClub/Screens/MiniGame/Tennis/Tennis.js
@@ -87,7 +87,7 @@ function TennisRun() {
 			DrawText(TextGet("StartsIn") + " " + (5 - Math.floor(MiniGameTimer / 1000)).toString(), 1000, 600, "black");
 		} else {
 						
-			// Moves the ball the way a 60FPS monitor would if we are above 60FPS
+			// Checks for the frame ratio to keep the ball from moving too fast if the game is running above 60 FPS
 			var TennisCurrentFrame = performance.now();
 			var FrameRatio = 1;
 			if (TennisLastFrame != null && TennisCurrentFrame - TennisLastFrame <= 16.666667) {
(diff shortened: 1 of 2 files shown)


MrBondt:
I just tested this on a 144 Hz screen, but an easy tennis still gets very hard after only a few hits as the ball speeds up. When I try it on 60 Hz, the ball does not speed up as much.

I'm not terribly familiar with the code, but maybe I can help you with debugging.

ace-1331:
Mm it does get quite fast over time even on 60FPS for "easy", maybe there's even more to it though.

Can you send me a screen capture of what it looks like R59 vs this PR?

Ben987:
@MrBondt , can you pull the latest version uploaded by Ace and let me know if the game is beatable at 144hz ?

I can't really test it on my side.  Thanks a lot!

ace-1331:
After some more thinking and testing, I found the problem with the ratio, when testing with a fake framerate it seems to speed up/down like it should. Just waiting for confirmation. I looked and couldn't identify that problem elsewhere in the code.

There was actually 1 more reported "bug" which I added to this PR and it should be addressed whether or not the tennis fix ends up working. Since we check for timed events every 100 frames in TimerProcess, someone on 100FPS+ will get beeped by their NPC owner twice as much compared to someone on 60FPS, and someone at 30FPS would get beeped twice as less.. this should be independent of the framerate, this is why I changed it from 100 frames to "100 frames if it were at 60FPS". That will make owner beeps consistant across varying framerates without changing the intended behavior. 

If we find more on the topic of FPS, I'll have a look

Ben987:
Alright!  I'll test it soon on my side, but since it was already working, we'll have to wait for other players.

MrBondt:
I just got around to testing it again, and it looks like the ball is speeding up exactly the same on 144 Hz and 60 Hz now. So I consider this fixed!

I haven't encountered other places where a higher refreshrate is an issue right now as I mostly use the online chatrooms, but I'll keep an eye out for it if strange things happen.

PR #1136, opened 2020-07-21
Fix: Cross

ace-1331:
- added the alpha hood prerequisite
- added missing allow pose for mittens that caused them to float
- added missing assets for a few spanking toys in the missing pose

The almighty black holes strike again, this time taking your cross and your arms all at once. Like items and poses that go near the head, alpha hoods don't work with this item/pose

Also, since the first missing prerequisite is returned, I moved `"CuffedArms", "CuffedFeet"` at the beginning of the array to get these messages first, which will increase the odds of filling out the other prerequisites by the time you are in cuffs, which is just a small UX thing (To avoid having things like "must not be hogtied, must be able to open legs, must have cuffs, etc."


Ben987:
Thanks!  That was quick, I haven't had time to test it myself.  Cheers :)

PR #1115, opened 2020-07-19
Feature: Lovers Vibrator (new item)

elliesec:
# Summary

Adds new "Lovers Vibrator" and "Lovers Remote" items to the game.

## How they work

* The Lover's Vibrator can only be added to a character by one of their lovers
* When added, the vibrator is given a member number belonging to the player that added it
* A player may only change the settings on the vibrator if they own the lovers remote, are the lover of the character it is equipped on _and_ their member number matches the member number on the vibrator
* _Anyone_ may remove the vibrator (as it is not a lockable item), provided it is not blocked by another item
* The remote is subject to the usual owner rules:
    * Confiscating remotes will confiscate the player's lovers remote
    * Blocking the player from buying remotes will prevent the shop from selling the lovers remote to the player

## Implementation

The changes that were made in order to implement this are summarised below:

* Adds two new properties to assets:
	* `CharacterRestricted` - causes an `ItemMemberNumber` property to be added to the item when it is equipped, which records the member number of the player that added the item, or -1 in the case of an NPC
	* `AllowRemoveExclusive` - can be added to a `LoverOnly` or `OwnerOnly` item to allow any player to remove it, regardless of whether or not they are the character's lover/owner
* Modifies the `CharacterAppearanceSetItem` function to take an optional `ItemMemberNumber` parameter, used to set an `ItemMemberNumber` property for items with the `CharacterRestricted` flag
* Modifies all usages of `CharacterAppearanceSetItem` to handle the modified function signature
* Modifies the `InventoryWear` function to take an optional `MemberNumber` parameter, to be passed to the `CharacterAppearanceSetItem` function
* Modifies all usages of `InventoryWear` to handle the modified function signature
* Refactors remote usage checks in `Dialog.js` into a common `DialogCanUseRemote` function

## Testing

I believe that I've tested the new items and functionality quite thoroughly.

* Players that are not lovers of a character cannot control their lovers vibrator
* Players that _are_ lovers of a character cannot control their lovers vibrator _unless their member number is on the vibrator_
* Players that are not lovers of a character cannot add the lovers vibrator to them
* Anyone, including the player themselves, can remove the lovers vibrator, provided it is not blocked by another item
* The lovers remote is confiscated when a player's owner activates remote confiscation
* The lovers remote cannot be bought from the shop if the player's owner has activated the rule blocking them from buying remotes
* The vibrator can only be added to or controlled on NPCs if they are the lover of the player
* The vibrator cannot be controlled with a standard vibrator remote
[...]

PR #1137, opened 2020-07-21
Improvement: Prevents null fields from being sent in chatroom item updates

elliesec:
# Summary

This change was mostly prompted by the comments on #1115.

This makes some minor optimisations to the `ChatRoomCharacterItemUpdate` function by setting fields to `undefined` rather than `null` if they should not be set. This will prevent the field from being sent in the websocket message. I've tested this in as many ways as I can think of and not seen any unexpected behaviour. I've included before/after examples of the websocket message changes below - these are all recorded from the actual websocket messages being sent by my browser. As you can see, this reduces the message size on item additions, removals and updates, and as far as I've seen, doesn't cause any issues with gameplay.

Note that the changes in #1115 will only send the `Property` field when the `CharacterRestricted` is present on an added asset - `Property` will not be sent (even as `null`) for other items.

### Before

```json
[
    "ChatRoomCharacterItemUpdate",
    {
        "Target": 1,
        "Group": "ItemArms",
        "Name": "LeatherArmbinder",
        "Color": "#404040",
        "Difficulty": 10,
        "Property": null
    }
]
```

### After

```json
[
    "ChatRoomCharacterItemUpdate",
    {
        "Target": 1,
        "Group": "ItemArms",
        "Name": "LeatherArmbinder",
        "Color": "#404040",
        "Difficulty": 10
    }
]
```

### Before

```json
[
    "ChatRoomCharacterItemUpdate",
    {
        "Target": 1,
        "Group": "ItemArms",
        "Name": null,
        "Color": "Default",
        "Difficulty": 10,
        "Property": null
    }
]
```

### After

```json
[
    "ChatRoomCharacterItemUpdate",
    {
        "Target": 1,
        "Group": "ItemArms",
        "Color": "Default",
        "Difficulty": 10
    }
]
```

### Before

```json
[
    "ChatRoomCharacterItemUpdate",
    {
        "Target": 1,
        "Group": "ItemArms",
        "Name": "LeatherArmbinder",
        "Color": "#404040",
        "Difficulty": 10,
        "Property": null
    }
]
```

### After

```json
[
    "ChatRoomCharacterItemUpdate",
    {
        "Target": 1,
        "Group": "ItemArms",
        "Name": "LeatherArmbinder",
        "Color": "#404040",
        "Difficulty": 10
    }
]
```

### DIFF
--- a/BondageClub/Screens/Online/ChatRoom/ChatRoom.js
+++ b/BondageClub/Screens/Online/ChatRoom/ChatRoom.js
@@ -589,10 +589,10 @@ function ChatRoomCharacterItemUpdate(C, Group) {
 		var P = {};
 		P.Target = C.MemberNumber;
 		P.Group = Group;
-		P.Name = (Item != null) ? Item.Asset.Name : null;
+		P.Name = (Item != null) ? Item.Asset.Name : undefined;
 		P.Color = ((Item != null) && (Item.Color != null)) ? Item.Color : "Default";
 		P.Difficulty = SkillGetWithRatio("Bondage");
-		P.Property = ((Item != null) && (Item.Property != null)) ? Item.Property : null;
+		P.Property = ((Item != null) && (Item.Property != null)) ? Item.Property : undefined;
 		ServerSend("ChatRoomCharacterItemUpdate", P);
 	}
 }


Ben987:
Yes, I love these kinds of changes to shorten the sent packets.  

Setting the property to undefined is enough?  We don't need to delete the property from the object?  Thanks!

elliesec:
Correct. Setting the property to `undefined` is sufficient. This is because socket.io converts JS objects to JSON via `JSON.stringify()` before sending them, and `undefined` is not a valid JSON type. For example, running the following:
```js
JSON.stringify({a: "test", b: undefined});
```
Returns the following:
```json
{"a":"test"}
```
Notice that the `undefined` value has been skipped.

Ben987:
Excellent!  The smaller packets the better.  Cheers :)

PR #1140, opened 2020-07-22
Improvement: Corrected some typos in the JSDoc comments

SandrineBC:
Corrected small typos in
* `Drawing.js`
* `FriendList.js`
* `Server.js`


ace-1331:
Now that I think of it, without defining the object, the compiler probably doesn't like the optional object properties being defined inline.

To represent the same thing inline, it might be better to change the type for a union type with undefined `...| undefined` instead of just removing the `?` so  `Property?: object` becomes  `Property: object | undefined`

Ben987:
Sure!  Thanks a lot :)

PR #1141, opened 2020-07-22
Inspect Lock Fix

Nina-1474:
The step to remove items that are no longer worn is an issue for locks since they don't actually occupy the ItemMisc slot. It prevents the extended screen from opening on hitting the Inspect Lock menu button.


Ben987:
I didn't know that one.  Thanks!

PR #1142, opened 2020-07-22
Self Lover Locking Fix

Nina-1474:
Lover-only locks will now only appear as options for a player on herself if she has at least one non-NPC lover.
This is in response to a player reporting in the discord server that they'd self-locked themselves, and both the NPC and maid could not remove it.
I suspect there are still potential cases for self-bondage irreversible-locking where a player has only NPC lovers or owners, but this would require investigation.


Ben987:
Sure!  Thanks a for that one.  Cheers :)

PR #1143, opened 2020-07-23
Improvement: Background tagging

SandrineBC:
# Summary
With the constant growth of backgrounds to select from, we thought it would be nice to be able to apply a filtering by certain tags, added to the backgrounds. So that you can filter by 'Indoor', 'Outdoor' or special themes. 
This PR adds the tagging mechanism to the various background screens and adds a filter drop down to then background selection screen.
image
# Details
This PR adds two new files to the repository and changes eight others. The details are listed here
### New graphics
* `Icon/Dropdown.png` The down arrow for the drop down menu 
### `Backgrounds.js`
This file holds the definitions for the tags, the tag list and the backgrounds list. All definitions are done as constants, because if someone messes with them within the code, this could be game breaking.
### `index.html`
Added the import of the 'Backgrounds.js' script on top of the list, to make sure, the background definitions are there, before anything else happens
### `Common.js`
Replaced the list of background names by a line referring to the new array of backgrounds, defined in `Backgrounds.js`
### `Styles.css`
Added the css definitions of the styled needed for the new drop down list items
### `BackgroundSelection.js`
* Added the dropdown item for tag filtering in the top row
* Cleaned the dropdown item upon leaving of the screen
### `Element.js`
* Added a function for the creation of dropdown items. This one got quite lengthy, because I had to overwrite the player's OS behavior on the HTML `<select>` and `<option>` tags to match the BC look and feel. What exactly happens, is documented in the JSDoc comment
* Added a function that closes all open dropdown items
* modified `ElementPosition` and `ElementPostionFix` to set the css attributes of an element in a more graceful way (and not simply overwrite all definitions
### `ChatCreate.js`
* Added a parameter to the call of `BackgroundSelectionMake` to hide the dropdown box, if we come from the Asylum
### `AsylumEntrance.js`
* Changed the call to `ChatStart` so that the list of Asylum backgrounds is added dynamically by using the appropriate tag
### `LARP.js`
* Added the list of possible LARP backgrounds to the list of public backgrounds by using the appropriate tag.
## Testing
Tested this manually on my local instance: 
* tested browsers
  * Safari on iOS -- Version 13.1.1 (15609.2.9.1.2)
  * Chrome on iOS -- Version 84.0.4147.89
* the general game is not affected by this change
* the drop down menu works as intended
* if a tag is applied, the text filter still applies, so that you can look for backgrounds tagged with 'Outdoor' and then look for 'Old' in addition
# Credits
Thanks to @elliesec for helping out with some JS magic and a review of my code, as well as @ace-1331 @Nina-1474 and @Ruilov3 for the discussion leading to this change


Ben987:
That's a good idea.  We should do a survey on what categories players want though.  I would love a "Dungeon" category or "Regular House" category myself.  I'll try to setup a little survey soon.

You're touching way too much stuff with that change though, don't mess with the full index.html, please use "var" instead of "const" and use for loops instead of .filter / .map / .some, like in the other modules.

Thanks!

SandrineBC:
I changed the `const` to `var` though I still think this is a bad idea and undid the formatting of `index.html`
When it comes to the usage of the array prototype functions, though...
* `.map` is used 179 times in 17 files
* `.filter` is used 80 times in 17 files
* `.some` is used once in `ChatRoom.js`
Do you really want me to clean up all these occurrences?

Ben987:
Most of them are in external modules, if you can do a regular for loop it goes well with the rest of the code.  If it's too tricky we can clean the code later.

SandrineBC:
This is the initial tag list, we start with:
* Indoor
* Outdoor
* Aquatic
* Special Events
* SciFi & Fantasy
* Club rooms
* Regular house
* Dungeon

I also fixed a minor issue to the element drawing, where the display of hovering text hid the filter input box and did not bring it back and added sorting to the list of backgrounds.

Ben987:
That's a great list to start with, thanks.  

And what about the Asylum?  Do the special background selection still works in there?  Do you hide the drop down list in the Asylum?

SandrineBC:
Good catch. I added the Asylum and LARP handling, while I was at it. There are two hidden tags now for Asylum or LARP specific backgrounds. If you add them to `Backgrounds,js` they will be added automatically and you don't have to touch the other files. I'll update the general PR description accordingly.

Ben987:
Excellent!  I'll test it soon and we can clean the code structure later.  Cheers!

PR #1144, opened 2020-07-23
Difficulty Consistency Fix

Nina-1474:
When an applied item is updated through ChatRoomCharacterItemUpdate, the difficulty is refreshed using the player's bondage skill. However this player may not have applied the item originally and will often be just adding a lock or changing the colour, making the base escape difficulty quietly increase or decrease when we'd expect it to stay the same. (Note: difficulty increases from locks are not stored in this number).

You can recreate this easily by having A tie B, then having C with a different bondage skill to A make any trivial change to the item. The escape difficulty changes.

I couldn't see any cases where we actually want this difficulty change to happen. I've changed the formula to be the difference between the asset difficulty and applied difficulty (original skill + asset difficulty) to ensure the skill-based difficulty is not changed by item updates. I've tested cases such as the Bondage bench which calls the function directly and the resulting numbers all looked consistent.

### DIFF
--- a/BondageClub/Screens/Online/ChatRoom/ChatRoom.js
+++ b/BondageClub/Screens/Online/ChatRoom/ChatRoom.js
@@ -596,7 +596,7 @@ function ChatRoomCharacterItemUpdate(C, Group) {
 		P.Group = Group;
 		P.Name = (Item != null) ? Item.Asset.Name : undefined;
 		P.Color = ((Item != null) && (Item.Color != null)) ? Item.Color : "Default";
-		P.Difficulty = SkillGetWithRatio("Bondage");
+		P.Difficulty = Item != null ? (Item.Difficulty - Item.Asset.Difficulty) : SkillGetWithRatio("Bondage");
 		P.Property = ((Item != null) && (Item.Property != null)) ? Item.Property : undefined;
 		ServerSend("ChatRoomCharacterItemUpdate", P);
 	}


Ben987:
If someone changes the position (ex: from regular tie to hogtie), will it use the skill from that new person?

Nina-1474:
The added difficulty for that kind of option is stored in Appearace[A].Property.Difficulty which has not been touched. Appearance[A].Difficulty is the only thing changed and that only covers the base item difficulty plus the Bondage skill.

Nina-1474:
That's a good point. With this change, no it won't. But while it makes sense to reapply the skill with something like changing the rope tie, it doesn't as much when changing other item variants like zipping/unzipping, linking cuffs, opening/closing belt panels, etc.
To resolve that, we'd have to add a property that indicates when changing an extended item option should reapply the skill. Would that be worthwhile?
Otherwise, I feel like this fix is the lesser of two evils. Not reapplying the skill when a second player changes a rope tie seems like a smaller issue to me than a second player reapplying the skill by changing the colour or adding a lock.

Ben987:
I'm not sure.  If you change the rope color, it logically would mean that you're untying the person and tying her again, so the new bondage skill level would make sense also.  I'm not sure what's the best solution here.

Nina-1474:
While it could be interpreted that way, that fact that you don't go through the "Alternate A and S" screen makes it not really feel like removing and reapplying the item.
The same could be said for changing rope types actually. If you really wanted it to represent untying and retying your target, that screen should appear again (but that would probably annoy people!). 
So in that sense, I feel like it's appropriate that the skill is only applied when you go to the effort of pressing a few buttons for it!

Ben987:
Yeah, I guess there's no perfect system there.  Maybe we'll find a better solution later, we can use that one in the meantime.  Thanks!

PR #1145, opened 2020-07-23
Bugfix: Fixes for the OverridePriority of hemp rope and chains cuff variants

elliesec:
## Summary

This fixes a minor issue in the cuff variants of the hemp rope and chains arm items to ensure they are drawn below the `Cloth` slot.

## Reproducing the bug

1. Equip any piece of clothing with the default priority of 30 which would clip with the aforementioned cuff variants (the teacher outfit works well)
2. Equip the hemp rope/chains arm item
3. Select the rope cuffs/chain cuffs type
4. Note that the rope/chains render above the clothes rather than below


Ben987:
Sure!  Thanks :)

PR #1146, opened 2020-07-23
Self Target Menu Fix

Nina-1474:
The step to rebuild the item/activity menu whenever the target character's appearance changes has been corrected to account for when the player focuses on themselves on the left instead. The menu wasn't appearing before.


Ben987:
Yes!  Thanks a lot :)

PR #1147, opened 2020-07-23
Feature: Eyes split

ace-1331:
- split the eyes into Right/Left for more flexibility

The ability to have different eye colors or various winking expressions has been requested a few times. With this, you can independently change your eye style to mix colors.

I tried my best not to bloat existing code while maintaining somewhat of a seamless transition. Here are the key aspects:

- expression menu now has close left/right eye buttons to allow winking with any expression
- any expression set for "Eyes" will be applied to eyes 2, this allows for backward compatibility and removes the need to double the expression triggers everywhere in the code.
- the existing NPCs and the arousal meter will properly recognize the new right eye group
- If your character is missing the right eye, it will take the left eye by default. This will prevent us from seeing cyclops in the beta, having to redo all the wardrobe slots, or logging in with different eyes.
- Generating a full character always keeps the same eye style (for NPCs and random character in the appearance screen). They will also keep the same eye color, heterochromia will actually spawn in at a realistic rate which makes heterochromic NPCs extremely rare
- removed the "Eyes" from the hidden properties for items that did not need it, also added "Eyes" to the lenses so they'd look as intended

These changes were made with the idea that this should not be the norm, and that we shouldn't have to deal with two eye slots everywhere in the code/club. The transition should be smooth and the need to change your two eyes should only occur if you start a new outfit and change your eye style (similar to how body size/color is right now)

After these tweaks and discussing with a few people, I'm quite satisfied with the output. If you have any comments or think certain things should be tweaked further, let me know. This isn't the biggest change but it gives a few new possibilities and will allow for more flexibility in the future

Note: I used a script to split all the assets in half which took only a few seconds instead of hours. If you think such a tool can be interesting in the future, I can share the script in any way deemed useful


Ben987:
That's a good idea.  So if someone changes "Eyes" it also changes "Eyes2", but if "Eyes2" is changed, it doesn't change "Eyes" ?

ace-1331:
No, like the body size and any items with parent items, youll need to manually set eyes 1 and 2 how you want them. However, existing characters and wardrobe slots will have the same eyes.

Also, since the generating character function will always keep the same eyes, your wardrobe and the random button will also always have the same eye style. That means you really just have to change both eyes if youre actively going to change your eyes. (Like youd have to change your skin color for hands/upper/lower body if you were going to make a totally different character)

ace-1331:
To be clear, I tried what you mentioned at first, the behavior is horribly unintuitive. If you want to change just your eyes1, you'd need to change Eyes1 then rechange Eyes2 because it was swapped in the process, etc. Same for expressions.

That's part of why I kept expressions for eyes as 1 block, except for closing eyes (they also didn't really match anyway)

Ben987:
So there's no quick way for the player to synchronize both eyes?  They both need to be selected and colored manually?

ace-1331:
Indeed. We could add a sync button for eyes and body for ease of use, though I am not sure where these would fit in the current UI

Ben987:
It's fine, I guess we don't change the eyes very often anyway, it won't be a big thing.  Was there any way to do it without hardcoding the "Eyes2" in our main functions?  It's not very clean and it complicates the code a little.

ace-1331:
well there are two different things to this. 

For full random, I wanted to keep a rare chance to get a heterochromic npc. If I were to remove that, I could simply use ParentColor and all eyes would always be the same color for NPCs without complicating that part of the code

For appearance validation, I could remove it, but then all the online characters would have a second basic eye on their first login/for each wardrobe (which would be utterly annoying for the week during/after the beta. This is really the only group who needs to be "mirrored".

I agree, its not optimal. Actually, to make it worth it, I could refactor it into a "MirrorGroup" Property, so if in the future we want the same effect for other groups, it wouldnt have to be hardcoded. (If we ever need left arm/right arm for example, not entirely sure where it would be useful)

ace-1331:
I just pushed a change. I left the random heterochromic NPC, but I turned the new code into something we can re-use at a later date if needed. You can set mirrorgroups to have the validation try to take the other item if its missing.

This also enhances the split to also match the eyes if for some reason Eyes1 goes missing.

Again, it is not required, but it will make this and future transitions a lot easier to deal with for the players. Other wise we would have cyclops for the whole beta and have to redo a lot of outfits (which would also end up taking more space due to the added slot). If we ever want mirrored groups for other body parts, this will help keep things synchronized while validating the groups

ace-1331:
I also have another change I just pushed

 Iused Object.Assign to create the groups from a common "AssetEyes" object.

This will remove a lot of duplicate stuff, and would also make it clear that both groups need some common properties to work as intended.

I'm tweaking a few more things and Ill update you again in a hour or two

ace-1331:
I also have another change I just pushed

 I used Object.Assign to create the groups from a common "AssetEyesProperty" object.

This will remove a lot of duplicate stuff, and would also make it clear that both groups need some common properties to work as intended.

Everything should be in order, let me know if there is anything else

Ben987:
The MirrorGroup is a good concept.  Why are you using Object.assign({}, AssetEyesProperties, { in Female3DCG.js ?

ace-1331:
I just merged and removed object assign. Originally it was to make sure both groups shared the same properties, but I dont expect those to change anyway.

Ben987:
Excellent!  Thanks a lot.  I will test it soon.

PR #1148, opened 2020-07-23
Feature: Override blinking

ace-1331:
**This PR contains #1147 inside to avoid conflicts. #1147 should be merged first, but if it gets rejected, I can remove it from this PR as they are completely independent**

- Added a new property to have items override their blinking status

Especially now that ItemHead will be "eyes" item, if any item is affected by blinking, this will allow the item to carry the blink property and have a close state to match the blinking eye. Anything from contact lenses or things around/on the eyelids, etc. will be feasible

Concretely, this fixes one existing item: the blackout lenses. They do not blink since the ItemHead group does not.. but since it is within the eye, it should. This makes the lenses look much cleaner (Only technical limitations led it to staying this way till now)


Ben987:
Can you make a separate pull request for that one?  It will be easier to review without all the extra files.  Thanks :)

ace-1331:
I just merged, you should see the proper difference now

Ben987:
Awesome!  That should work :)

PR #1149, opened 2020-07-23
Bugfix: Ensures spanking toys are always drawn on initial equip

elliesec:
## Summary

This fixes a bug where equipping a spanking toy on another player would not initially draw the toy (the crop in the case of initial equip).

## Steps to reproduce

1. Enter a chatroom with another character
2. Equip a spanking toy on the other character, but do not select another toy type
3. Note that the toy does not get drawn


Ben987:
Yes!  I've seen that one before.  Thanks.

PR #1150, opened 2020-07-23
Improvement: CommonWait

ace-1331:
- changed how the wait time was handled to avoid the club from freezing

currently, changing your lists freezes your club for a second on purpose which was a quick fix for item permissions. This increases the time to avoid conflicts and will remove the freeze to make it a lot less annoying. I still believe we should have some kind of response sent back to the server to do a proper async callback, but this can do for now.


Ben987:
Absolutely.  That's what I was talking when we chatted on Discord.  Thanks :)

PR #1153, opened 2020-07-23
Handheld Item Order

Nina-1474:
This moves equipped handheld items to the front of the inventory list, since the players will likely want to use it more than the other options and currently have to often trawl through multiple pages to find it.


Ben987:
Yes, that should be handy.  Thanks for that one.

PR #1154, opened 2020-07-23
Improvement: JSDoc comments added to Shop.js

ace-1331:
- added JSDoc to Shop.js

We're getting somewhere! Most of the vital parts and many screens are now properly documented. We're still hunting for bugs, oddities and possible refactors as we go, so far it's been quite helpful.

We keep the output at http://www.hostedredmine.com/projects/bondage-club-issue-tracker/wiki but we could keep the generated HTML output in a folder if people want to view it straight from the repo. Let me know what you think.


Ben987:
Thanks!  I don't use an IDE myself so I can't tell if it's working as you want.  But if it's useful for you guys I'm good with it.

PR #1155, opened 2020-07-23
Feature: Layer Priorities

elliesec:
## Summary & Motivation

This enables the use of a `Priority` property inside asset layers, so an asset can have multiple layers with different priorities.

This is something that's been requested by multiple asset creators (I'm aware of at least 3 creators that would like to take advantage of the functionality), and would allow for the creation of more sophisticated assets without the need for workarounds like using cutouts for body parts in assets (which can lead to clipping issues), or the use of the `ItemHidden` slot (for example, the leather armbinder can now be reworked to remove the need for the straps to be in the `ItemHidden` slot - they can just be layers with different priorities in the armbinder asset).

Below is a demonstration asset that I created (not included in this pull request), which demonstrates what is now achievable.

Do let me know if you have any thoughts/concerns, or want me to explain anything about the change in a bit more depth.

This demonstrates the new functionality in practice (note that this is all one item):

layers-demo

And this is the code for the above item:

layers-3

## Technical Details

This is a summary of what has been done as part of this change:

### Login.js

* Replaces the calls to `JSON.stringify` on the character appearance with a custom inventory serialisation function to prevent issues with the new asset/layer structure (as discussed with Ace following the merge of #1217)

### Inventory.js

* Adds a new `InventoryStringify` function which serialises a character's inventory into a string representation to allow easy before/after inventory comparisons.

### Asset.js

* Changes `AssetBuildLayer` so that every asset now has at least one layer upon construction (if the asset has no layers defined in `Female3DCG.js`, a default layer will be created)
* Relevant layer properties are inherited from the parent asset and/or group as previously
* Removes the `IgnoreParentGroup` property from assets - with the new drawing logic, this functionality can now be achieved by setting `ParentGroup: null` or `ParentGroup: ""`
* Renames `Layer.NewParentGroupName` to `Layer.ParentGroupName` and allows this to be defined as `ParentGroup` in `Female3DCG.js` for consistency with assets and groups

### Female3DCG.js

* Changes all instances of `IgnoreParentGroup: true` to `ParentGroup: null` as described above

### Character.js

* Replaces the call to `CharacterAppearanceSort` in `CharacterLoadCanvas` with a call to a new function: `CharacterAppearanceSortLayers`, which sorts the layers rather than the assets

### Appearance.js, GLDraw.js & CommonDraw.js

These files are where the bulk of the changes have been made.

* `CharacterAppearanceSort` has now been replaced with `CharacterAppearanceSortLayers`, which sorts all drawable layers in the character's appearance according to their drawing order
* Note that the sorted layer array is attached to the character, but it exists on the client only - we never try to sync it to the server as it exists purely for drawing purposes
* Most of the logic in `CharacterAppearanceBuildCanvas` and `GLDrawAppearanceBuild` have been extracted into common functions in the new `CommonDraw.js` script: `CommonDrawCanvasPrepare` and `CommonDrawAppearanceBuild`. This was done as most of the logic was the same between the two scripts, so having it in one place should make it easier to maintain in the future
* The asset drawing logic from `CharacterAppearanceBuildCanvas` and `GLDrawAppearanceBuild` has been translated into layer drawing logic: instead of looping through the character's appearance array, the code loops through the character's sorted layer array
* Some loops and unnecessary complexity have been simplified or removed from the drawing logic in order to improve the code's efficiency
* The only notable change to the actual drawing logic is that an item's parent group can now overridden to ignore the parent group at the asset or layer level, removing the need for the `IgnoreParentGroup` property, as explained above
* Lock-drawing logic was the only slightly awkward part of this, as they are drawn per-asset rather than per-layer. Locks for an asset are drawn after the last drawable layer of the asset has been drawn (this should mirror the previous behaviour)

## Testing 

As this is a reasonably extensive rework of the drawing code, it will inevitably affect the way in which every item in the game is drawn. To this end, it's impractical to test all of the possible combinations, but I have tried as many combinations as I can think of including:

* Standard items
* Items with multiple layers
* Locked items
* Extended items with types
* Coloured items
* Items with different parent groups

I've tried the above in as many combinations as possible, and I've also done my best to test out items with some of the rarer drawing properties found in `Female3DCG.js`, and have so far not seen any issues.

I've also been sure to test the drawing of assets through both `Appearance.js` and `GLDraw.js`, and have noticed no differences between the two.

I've also done some very light performance monitoring, and as far as I can see, there is no noticeable hit to performance between the old code and the new code.


elliesec:
I've now merged master to include Ace's blink override changes from #1148 to incorporate them into `CommonDraw.js`, so blackout lenses are working as expected.

elliesec:
Quick note on this: because of the `JSON.stringify` calls that were introduced as part of #1217, this will now break on login because of the new item structure. I've discussed with Ace, and am working on a solution for that. Will comment again here once that's been resolved.

elliesec:
The conflicts with #1217 in`Login.js` have now been addressed, and this all looks to be working correctly again.

ace-1331:
Besides the various PRs waiting for this + creators waiting for this to make other items, I also have a branch ready to be submitted when this is merged. It will allow the gas mask to use any combination of addons together

PR #1157, opened 2020-07-24
Improvement: JSDoc comments added to Cell.js & LARP.js

ace-1331:
- added JSDoc to Cell.js & LARP.js

I also removed a button that is drawn twice on the Cell screen as well as simplifying the click checks to use MouseIn.

I still have the script to reformat all the checks automatically, but for now Ill do them manually while doing JSDoc while I come across them. It makes the code a lot easier to read and a lot shorter, it also gives me the opportunity to test the buttons file by file


Ben987:
Thanks!  I'm not sure why that button was drawn twice.

PR #1159, opened 2020-07-24
Fix: Remove collar attachments with the collar

SandrineBC:
# Summary
This fixes the bug of 'floating collar' attachments that were not removed, when the collar was.

# Details
## What it does
Three files were changed to accomplish this:
### `Asset.js`
* Added the `RemoveItemOnRemoval` attribute on the AssetGroup level
* Modified the Asset object so that it inherits the `RemoveItemOnRemoval` correctly from the appropriate AssetGroup
### `Female3DCG.js`
* Added `RemoveItemOnRemoval` to the `ItemNeck` group, so that all collar attachments will be removed, if the collar goes
### `Dialog.js`
* Removed unnecessary checks for associated items, if a collar is removed
## Testing
Tested the following cases on my local instance
* Apply a collar and an attachment, the remove the collar --> Collar attachment disappears
* Apply another item with the `RemoveItemOnRemoval` attribute, for instance a leather arm binder, including the straps. Upon removal of the restraint, the leather straps vanish as well.


Ben987:
That's a good idea.  Though these collar accessories are removed manually through code most of the time, we'll need to get rid of that code if we use that method.  Thanks!

ace-1331:
we should probably also add neck restraints to this. Though the collar is usually blocked by things to make sure you remove it last, this is a nice failsafe in case of console use, bad scripts or simply places in the NPC areas that dont take those into account like they should

SandrineBC:
I think, you are wrong here. If there's no attachment present, when the collar is removed, nothing happens at all. So I don't think you need to change legacy code for this to work.

Ben987:
I'm talking about these kinds of lines, do we still need them?

			if (InventoryGet(C, "ItemNeck") == null) InventoryRemove(C, "ItemNeckAccessories");
			if (InventoryGet(C, "ItemNeck") == null) InventoryRemove(C, "ItemNeckRestraints");

SandrineBC:
We don't really 'need' them anymore, but they don't hurt as well. I think, we can delete them, when the file is touched the next time. I'll have a look and do that in the next days.

Ben987:
Sure, we can remove the lines that won't be needed anymore later.  Can you check the conflict?  It seems another pull request clashed with that one.  Thanks.

SandrineBC:
I can see no conflict there...

SandrineBC:
All set now.

Ben987:
Thanks!  There are probably other places where we could clean the code lines for ItemNeckAccessories and ItemNeckRestraints but I guess it doesn't hurt to clean them later.  Cheers :)

PR #1160, opened 2020-07-24
Bugfix: Chatroom text for chastity belt shield addition/removal

elliesec:
## Summary

The chatroom text when adding/removing a chastity belt shield had an extra `'s` attached - this PR removes it.

Example:
> (Selene removes the shield from the back of Not Ellie's's belt.)


Ben987:
Good catch!  I've never noticed that one.

PR #1161, opened 2020-07-24
Fix: vacbed priorities

ace-1331:
- improved the vacbed priorities, hidden items and blocked zone

talked with kirsty who made the vacbed to get this where it should be now with all the priority changes and fixes. It properly displays the front hair only and hides neck items. On top of that, blindfolds, hoods and ears are accessible while all gags are drawn above the vacbed like they should be


Ben987:
Thanks for that one!

PR #1178, opened 2020-07-26
Improvement: OnlineProfile.js

ace-1331:
- added JSDoc to OnlineProfile.js
- added the option to leave without saving

When viewing your own profile, you have to logout if you need to revert your changes, this change allows you to exit without saving.

JSDoc wise, it seems like we have finished /Scripts , /Character, /Cutscene. On top of that, /Online, /Room and /MiniGame are fairly advanced to so we're getting closer to be done with this. 

We'll just have to comment properly from now on to avoid having to play catch up like we're doing right now. We're probably also going to do some guides on how to add basic things like items and such. Hopefully this will be enough to get people started and understand everything better


Ben987:
Yes, the extra option should be handy, thanks :)

PR #1180, opened 2020-07-27
Add Diaper Harness

mrbakucamp:
Adds one item: the diaper harness
Also, for some reason the rope harness was not properly added to Female3DCG.csv, so this fixes that as well.


Ben987:
Great!  And I didn't knew that this entry was missing.  Cheers :)

PR #1182, opened 2020-07-27
Improvement: Thoughtful NPCs

ace-1331:
- made NPCs listen to the player's permissions

Currently, NPCs bypass blocked items which some players have complained about in the past. Any specific case can't necessarily be fixed, but at least all calls to the randomwear function will properly check and remove blocked items from the random pool which solves nearly all cases. 

Due to how the limited system works, I can't factor in limited items for NPCs.

This makes it so any blocked item will not be used UNLESS, there is nothing BUT a blocked item that can be used. This is not necessary, but I made it this way to prevent someone from cheating by blocking all restraints. If you think it is not needed, I can remove that concept and always remove blocked items from the pool. The odds that someone blocked all the basic items that could be used instead of blocked items is almost null, so I don't expect this to cause issue, it will only block cheating.

Let me know what you think

### DIFF
--- a/BondageClub/Scripts/Inventory.js
+++ b/BondageClub/Scripts/Inventory.js
@@ -291,13 +291,16 @@ function InventoryWearRandom(C, GroupName, Difficulty) {
 				break;
 			}
 
-		// Builds a list of all possible items and use one
+		// Builds a list of all possible items and uses one of them, prevents the user from blocking everything to cheat
 		var List = [];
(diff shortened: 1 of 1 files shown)


Ben987:
It should be fine like that, we actually need that fix because some categories have lots of items that are Random: false.  So we need to make sure we can still apply a random item.  Thanks a lot!

PR #1185, opened 2020-07-27
Improvement: Unblocked breast slot for the bitch suit exposed

SandrineBC:
# Summary
Unblocked the `ItemBreast` slot for the `BitchSuitExposed` so that the breasts of the wearer can be used for sexual abilities


Ben987:
Sure, that makes perfect sense.

PR #1187, opened 2020-07-27
Improvement: Relog screen

ace-1331:
- preloaded the CSV for the reconnection screen in the main hall
- removed the possibility to show duplicate messages
- replaced click conditions to use MouseIn

Currently, the "Enter your password..." message displays twice if this is what the server sent back, so I added a check to avoid this from happening. 

Also, if you actually disconnect, the screen is just 3 white boxes because you fail to fetch the CSV, this fix makes it so when you run the main hall, it will fetch the CSV and cache it so you will have the text available if you DC later which will make sure the page displays properly in all cases of disconnections.

### DIFF
--- a/BondageClub/Screens/Character/Relog/Relog.js
+++ b/BondageClub/Screens/Character/Relog/Relog.js
@@ -45,7 +45,7 @@ function RelogRun() {
 	
 	// Draw the relog controls
 	if (!LoginMessage) LoginUpdateMessage();
-	DrawText(LoginMessage, 1000, 150, "White", "Black");
+	if (LoginMessage != TextGet("EnterPassword")) DrawText(LoginMessage, 1000, 150, "White", "Black");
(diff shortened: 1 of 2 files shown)


Ben987:
It's strange that I've never noticed that one, it's a new bug from another change we've did?   The reconnect screen seemed to worked well last time I've seen it.

ace-1331:
It's contextual. 

If your socket disconnects, but you still have internet: you will still be able to fetch the .CSV from the server since that is not related to your login state.

If you disconnect because you have no internet: the odds that you see the reconnect screen and had time to fetch the CSV is almost null. You see that screen because you actually disconnected. There is also a case where if you put your computer to sleep, when you turn it back on, you will see the empty relog screen. This is because the connection was suspended so it never managed to fetch the file.

You can test this by disabling internet in the Devtools > network then disconnect the socket to see the relog screen.

Ben987:
Now I get it.  Thanks for that one :)

PR #1188, opened 2020-07-27
Change: Nursery Milk

ace-1331:
- fixed the nursery milk to allow it to be seen by others

This one has been requested a fair bit recently so I tried to get something decent working with our current systems. 

Since the speech changes a few months back, other players in chatroom havent been able to see the nursery milk speech effect since it is client side only (stored on a global variable). There did not seem to be a lot of easy ways to do this simply because we do not have a static effect array. I also didn't want to add a value to the Player object and sync it just for this.

Having it as an item will allow everyone to see the effect, and it still requires to spend time in the nursery since that's where it is earned. Also, players will be able to block this item if they do not want it which also makes it user friendly in general and does not introduce new concept. It's basically a non blocking "gag". If there is ever a better way to do this, or a new speech system, we might revisit this one at a later date, at least now it works as intended.

I discussed with several people on the matter and it seems to be the most sensible thing to do for the code complexity and users seem to like the idea of being able to toggle it like this.
To be clear, I also spoke with Wild about this to discuss the matter since they were the one who made the nursery in the first place and firefly was kind enough to land me the milk asset for the item preview.


Ben987:
@wildsj , can you review that pull request and let us know if everything is ok for you?  It changes the baby milk a little.  Thanks a lot :)

wildsj:
Seems okay to me.

Ben987:
I'm sure that ABDL players will like it :)

PR #1190, opened 2020-07-27
Fix: text

ace-1331:
- added a missing line for the canvas hood name
- renamed the close eye buttons to "Toggle eye" instead to reflect the change


Ben987:
Yes.  Thanks a lot.

PR #1191, opened 2020-07-27
Bugfix: Changes deafness and sensory dep definitions to use GetDeafLevel

elliesec:
## Summary

This changes all checks against the `DeafLight`, `DeafNormal` and `DeafHeavy` effects in the code to use`Character.GetDeafLevel` instead for consistency. This also fixes a small bug that has come about with the recent deafness changes where a player would not be counted as being in sensory deprivation if deafened with three items with the `DeafLight` effect rather than a single item with the `DeafHeavy` effect (which sum to the same deafness level of 3).

## Reproducing the bug

The easiest way to see the bug is to set sensory deprivation levels to "Total", then equip three `DeafLight` items, and notice that total sensory deprivation effects aren't applied. For example, the following three items reproduce the effect:
* Light earplugs
* Polished Steel Hood
* Display Case

### DIFF
--- a/BondageClub/Screens/Character/Preference/Preference.js
+++ b/BondageClub/Screens/Character/Preference/Preference.js
@@ -760,5 +760,5 @@ function PreferenceMainScreenExit() {
  * @returns {boolean} - Return true if sensory deprivation is active, false otherwise
  */
 function PreferenceIsPlayerInSensDep() {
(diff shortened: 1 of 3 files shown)


Ben987:
Yes, this will normalize the deaf level a little.  Thanks!

PR #1193, opened 2020-07-28
Fix: Maid soft lock

ace-1331:
- fixed a soft lock at the end of the maid drinks mini game when done online

Currently, it didn't properly check if the player was gagged so if you wore no restraint but were still gagged when claiming your salary, you had no dialog options shown to you. (The two dialog options had improper prerequisites)

( as reported here: [discord message link](https://discord.com/channels/554377975714414605/554378725916147722/737513222781009930 ))


Ben987:
Good catch!  I've never seen that one.  Thanks :)

PR #1194, opened 2020-07-28
Fix: winking matching eyes

ace-1331:
- made it so the eye toggle actually maintains the same eye expression

I purposefully made it so you can't mix and match expressions because they don't work well together in the first place, but forgot the toggle would re-open this possibility. This fix makes it so instead of opening the eye, it just sets it to the same expression.

If both eyes are closed, it will just open them as normal to be a lot more user friendly and keep the intended design.


Ben987:
Very good!  Thanks :)

PR #1195, opened 2020-07-28
Fix for /safeword command

Lucritius12:
This fixes the /safeword command working even if safewords are disabled in the preference menu.
If you try to use the /safeword command anyways while it's disabled, you get a client-side only message in chat saying "Your safeword is disabled, it cannot be used at the moment."


Ben987:
Thanks for the quick fix :)

PR #1197, opened 2020-07-28
Fix: changing table

ace-1331:
- change the negative priorities to 1 to properly render the item behind everything while maintaining consistency 
- fixed the prerequisites to forbid using the table while hogtied/suspended and with alpha hoods (black hole effect!)
- removed unused values (Extended IgnoreParentGroup AllowType)


Ben987:
Yes, I didn't knew we were using negative priorities.  Thanks!

PR #1198, opened 2020-07-28
Fix: Wink change

ace-1331:
- made it so it will always use the last used expression instead of worn (does not reset when blinking both eyes)


Ben987:
Great!  I will test it right now.  Thanks :)

PR #1201, opened 2020-07-29
Feature: Hide Items

Nina-1474:
This feature allows players to configure assets (both items and clothes) to not appear on their screen. It was mainly motivated by discussions when the webs were added and mentions of arachnophobia. It could potentially help smooth concerns for other controversial items in future too.

After an asset is set as hidden, it won't be added when building a character's canvas and will not hide other items/zones that it usually does. Poses are still applied. It is still worn invisibly and all gameplay effects (blocking/effects/prerequisites/etc) will still be intact.
If any character has any hidden items on, an icon will appear at their bottom-left to indicate this and help avoid confusion. I tried to toe the line between it being easy to notice without being too visually offputting/distracting.
The preview icons in the inventory menu and shop will also be hidden.
The wear random item function will try to apply visible items first and only use a hidden one if no other options are available.

There's a new preferences screen where users can use back/next buttons to cycle through asset groups and then assets within them. This is admittedly a little clunky but I'd argue that's not the worst thing, as it's better if users don't use this feature excessively.
The preview icon of the current asset is displayed, if any. The changes can be saved or cancelled. There's also a reset button to set everything to visible again, requiring two clicks.

It's possible there may be some visual oddities depending what items players hide that are hard to predict. But since the change only happens on the player's side there's a 'use at your own risk' aspect with this. If it turns out there are significant issues, we could consider limiting the options to specific assets only (or all except specific assets), but then we'd need to work out that list.

~~This has a corresponding server change [PR # 40](https://github.com/Ben987/Bondage-Club-Server/pull/40).~~

Visibility1
Visibility2
Visibility3


ace-1331:
I also see the potential to use this screen for opt-in item or to make the permission menu more intuitive, this is great 👍

Ben987:
That's a good idea, some players don't like to see diapers or other items.  To help players select the items that they want to hide, should we only present items that are on the blocked list?  I guess all items to hide will be on the player blocked list.  Thanks!

ace-1331:
I think we should keep everything, that way we might be able to use this for the limited and blicked permissions too in the future (it would be a lot more intuitive.) 

Instead, hiding an item could make it default to blacklist (like ghosting auto adds to blacklist)

Nina-1474:
Showing blocked items only would actually mean you couldn't hide clothes since they can't be blocked at the moment.
I like Ace's idea so I've added that in now. Hiding an item will add it to the player's blocked list if it's not already set to blocked or limited permissions, and is something that can be blocked in the usual inventory permission screen.

Ben987:
True, it wouldn't work with clothes in that case.  Yes, adding to the block list is a a good idea.  I've changed the preference screen in the Fetish update, can you check the conflicts?

Also, does the HiddenItems property needs to be kept in the server memory?  Since it's not sent to other players, we could save the data and purge it from the server character object, so we don't keep it in the server memory for nothing.

Thanks!

Nina-1474:
Conflicts resolved. I've added the new array into Ace's cleanup too.
I misunderstood the purpose of Account in the server code so yes that PR is not needed. It has been closed.

Shiranui-Izayoi:
I just want to quickily pop in and offer a alternative use for this, that is for easier outfit creation. Personally I don't think hidden items should auto blacklist or similar. I have shown off on the discord a few times that using the console I can have the game hide items. 

I mostly do in order to make things look nice e.g. say I wanted to make a futuristic prison suit. I could use the seamless carsuit along with either leather cuffs or ornate cuffs, but then I want the restraints in my sci-fi suit to be intergrated I then use the console to tell the game to hide the cuffs, the moment someone changes the cuffs the hide is removed but until then the cuffs work like normal it is just that they are invisible.

Basically what I am getting at is if it was possible to adapt this/modify this so you can also hide items on yourself only like with the example of the sci-fi prison suit, or say hide a chastity belt and RP it as say a spell or rune preventing the removal of items etc.

I am not sure if that is possible since for a while I was trying to do it, but couldn't figure out a way to implement it that would work well.

Nina-1474:
This is Aeren from discord I take it?
While hiding an item does automatically add it to your blocked-list, you can still go to your item permissions and unblock it after, leaving it only at hidden. I'm not opposed to leaving hidden items at the default permissions though.
However do consider the primary purpose of the feature: To give people who find certain images genuinely offputting a way to avoid them. The ability to hide items to make certain combos look nicer is something I'd be inclined to leave as a bonus effect for "advanced" players rather than something to be advertised. Mainly because if it gets used excessively it could leave to confusion.

Shiranui-Izayoi:
Yer that is me and sure all sounds good. Though from what I have seen of the request currently my idea wouldn't really work. As it only hides it for you everyone else can see it so it would need to be added to later I really should get to trying to do my request if I could just figure out a good way to do it.

Currently the best I can think of is to make like a certain colour code that like how "Default" is used to not have any colouring applied this code would when put in make the item render as invisible/hidden.

elliesec:
I think a feature like that (aside from the fact that it would need to be a completely separate feature) would need to be thought about very carefully. As is, this can lead visual inconsistencies, which we can consider the price to be paid for wanting to hide an item, which is fine when the item is hidden on your end only. If we were to then give players the freedom to hide any item _and_ have it presented to other players - it means people can impose their visual inconsistencies on other players, which has the potential to lead to a lot of confusion.

Shiranui-Izayoi:
That is true and is sadly nearly unavoidable the best way I thought to get around this would be to disallow certain items from being able to be hidden like that such as all full body restraints like bitchsuits, hobble dresses etc, basically things that hide the body itself. That said it should be saved for another time as I don't think it is a good idea to discuss this here, better to use the discord since I kinda got this off topic... Whoops.

Ben987:
Would it be easier to change the "Visible" property on the item asset itself when we set the HiddenItems?   So we would not need to change the character drawing functions?  Thanks.

ace-1331:
That would mean the settings need to be taken into consideration and assets have to be rebuilt after logging in.

Nina-1474:
In addition to Ace's point, I don't think that would help avoid the drawing functions anyway.
CharacterAppearanceBuildCanvas and GLDrawAppearanceBuild only have a single line "C.HasHiddenItems = false" which is still needed to initialiase that property so we can determine whether to draw the "Hidden items are present" icon.
CharacterAppearanceVisible then still needs two changes: one to check if HasHiddenItems should be set to "true", and one to ensure hidden items don't hide other items that they usually do.

Nina-1474:
Resolved conflicts so that this works with the layer priorities change.
The preferences screen will show the previews for clothes too now that we have them!

Ben987:
Alright!  Time to test that one.  I'll probably purge the HiddenItem from the account server side since it won't be needed in the server memory.  Thanks!

PR #1202, opened 2020-07-30
Fix: Character Release Total

ace-1331:
- fixed a crash with character release total when the slave collar had no property object

This caused the safeword to break when your collar had no property. It's an edge case, but it is possible. Usually this happens when the collar is replaced by the client through validation.


Ben987:
Good catch!  Thanks a for that fix.  Cheers :)

PR #1203, opened 2020-07-31
Improvement: Clean up permissions

ace-1331:
- added login validation to remove non-valid data from the permission arrays to maintain data integrity

currently you can push anything to those arrays and it will be saved permanently, moreover, with items changing slots and the likes, it's important to make sure we don't keep a lot of data we do not need in the db. 

For example, all the hoods being moved will require people to re-blacklist them, but the old entries would stay in their array forever unless this change is approved. The server sync will only be triggered to remove bad data someone may have pushed through console, or when changing club version to clean up from changes.

I extracted it into a separate function in case we ever have more things we wish to cleanup that follow the same pattern

### DIFF
--- a/BondageClub/Screens/Character/Login/Login.js
+++ b/BondageClub/Screens/Character/Login/Login.js
@@ -237,6 +237,27 @@ function LoginValideBuyGroups() {
 }
 
 /**
+ * Checks if the player arrays contains any item that does not exists and saves them.
+ * @returns {void} Nothing
+ */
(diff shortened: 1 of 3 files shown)


Ben987:
That's a good idea, we could also add the HiddenItems in there when the pull request is ready.

Can you use a for loop instead of a .find ?

Thanks!

ace-1331:
Done

Ben987:
Awesome!  The less junk and legacy data the better :)

PR #1205, opened 2020-07-31
Improvement/Bugfix: Moves sturdy leather belts to use the extended item script

elliesec:
## Summary

This PR refactors all 3 sturdy leather belt items to use the extended item script. It also fixes a bug where the exit button on the extended item screen for the sturdy leather belts would not work if the belts were locked.

This also slightly changes the behaviour of the belts when locked. Rather than disallowing the player from changing the type of the belts at all when locked, the player can now change the type of the belts _if they are able to unlock them_ (as defined by the `DialogCanUnlock` function).

## Steps to reproduce the bug

1. Equip sturdy leather belts on a character (arms, legs or feet)
2. Lock the sturdy leather belts
3. Enter the extended item menu for the sturdy leather belts
4. Note that the exit button cannot be clicked


Ben987:
Absolutely!  Thanks a lot for that one, it will clean the code a little.

PR #1206, opened 2020-07-31
Feature: Help kneel/stand

ace-1331:
- moved options to their own dialogs
- added the ability to help someone stand/kneel when you are not tied.
- fixed "help struggle" to no longer bypass player permissions 

First iteration of something that has been asked for a long time. For now I made it so you can only help stand/kneel while untied, and you can also help someone kneel when their legs are tied which is not possible right now, which gives more possibilities.


Ben987:
That's a good idea.  But are you checking for permission?  If the person doesn't have permission to use an item, she should not be able to make you kneel or stand.  Thanks :)

ace-1331:
yes, Help struggle, help stand and help kneel has the .AllowItem check

Ben987:
Great!  Maybe someday we could code a server transaction to only change the active pose and send less data through the network. 
 But this will do for now, thanks a lot :)

PR #1207, opened 2020-07-31
Feature: New emoticons

ace-1331:
- added 5 new emoticons
- added a chat command to quickly use the afk emote (Faster than scrolling through when in a hurry)
- added a new option to use the wardrobe icon automatically when going in the wardrobe

The wardrobe emoticon wont overwrite your current emoticon nor show for yourself to avoid taking space in the appearance menu. That setting is off by default. I did this to spare the server because anyone who activates this would be doing it manually anyway so this is more convenient. If you think it is fine to have it on by default,  I can change that.

Thanks to Gnarp for making the icons


Ben987:
Sure!  These are good ideas.  Thanks :)

PR #1209, opened 2020-08-01
Feature: activities prerequisites rework and related fixes

ace-1331:
### EDITS AFTER THE MERGES:
- new double ended strapon contains the fillvulva effect and penetrate activity
- dildo plug gag properly merged
___
So it has been often suggested some activities should be available when the zone is not blocked. This PR changes how the button is built to allow this and tidies up a few things. I dislike activities, but I dislike hearing the same complaint over and over again more than I hate activities so I thought it would be nice to make them a little more intuitive now rather than later.

This fixes so many things, everything was dependent on the other so I could not split them. Every time I found a hidden problem, I found another under it... now overall everything is very smooth with this. I'll list everything this fixes and adds:

## Activities

- changed it so the activity button is hidden when there is no activity available for the focus group
- changed it so the activity button shows regardless of if the zone is blocked. The button is hidden if the player is enclosed
- added a new "Zone accessible" prerequisite to denote an action that needs the zone to be unblocked
- added a new "use arms" prerequisite to allow an activity while mittened. (Pet)
- made it so spank, grope, polite kiss, pet, hand gag and cuddle work regardless of if the zone is blocked or not
- added penetration to the mouth. One of the prerequisite checks if the mouth is blocked, this means the strapon will work when the sub is wearing a lip/ring/spider gag. This new property will be usable for other things in the future
- added a new "fill vulva" effect and a corresponding "IsVulvaFull" function on the character object for future use.  It will make it easier to know whether the zone is occupied by an accessory or if the pussy is full. 
- added a "vulva empty" prerequisite check on penetration to make sure penetration cannot be used when the target's vulva is filled by something while still allowing it with things like suction cups, piercings, wand belts, etc.
- added a "vulva empty" prerequisite check on fisting to prevent fisting while wearing a dildo
- added bite to nose and used nibble for nibbling (was forgotten after bite was added)

### Related bug fixes
- added a missing line in the CSV for the pet fetish
- fixed a problem with extended items not loading with the proper effects when the basic option's effect are not the same as the item's base effect which forced you to swap that option on/off to have the intended effect. This applied to block properties as well and will apply to any other array in the future. I applied this fix to several items as well as the extended script to help with future development. Basically we were not refreshing the character, and we were not publishing the change. This caused multiple desyncs in prerequisites across characters in the room. I went through every single extended item and did not find any that were still bugged after these fixes.
- fixed a potential desync when characters left the chat and the inventory menu was still up. If the character comes back, the person will still see the item. (Character array optimization revealed a few quirks in how items are handled on online characters) I modified a previous fix to prevent this from happening and added back the optimization related to duplication since it was needed again.
- fixed two butt plugs that were missing ""IsPlugged" which caused some weirdness with a few prerequisites
- fixed a prerequisite on the vacbed to avoid blackhole hoods (Again!)


Ben987:
This one is huge, I'll review it later.  Are you sure you want to remove the "CharacterDelete" for the chat room script?

ace-1331:
There are two options for it:
The problem is that if you remove an item and the character leaves/rejoins during that moment, you will see the item on screen while everyone else won't because the NPC changed in your characters array.

I removed CharacterDelete to retain the same behavior as before, but we could do the following:
- When someone leaves, we quit that character's dialog if we are in it

That would allow to keep CharacterDelete for maximum optimization

ace-1331:
Merged and ok with the other PRs

ace-1331:
The longer this PR sits, the more likely we are to miss a few items with new effects.

Note:
#1234 's new dildo will need the "FillVulva" effect.
#1236 's change to the base plug gag type will need the "OpenMouth" effect

ace-1331:
handled the change when merging #1234

ace-1331:
Handled conflicts with #1236 as mentionned

ace-1331:
feel free to close this one, I won't waste my time merging and fixing everyone's bug if it's not going to be looking at.

elliesec:
Just noting that this PR fixes quite a few issues that will likely get reported over the course of the curent beta, and is in my opinion definitely worth getting merged before the next beta/release.

ace-1331:
this PR is becoming a meme with how its being ignored lol
merged the changes

Ben987:
Finally could catch up to it.  There will be lots of stuff to test in that one.  But since we'll need to do a second beta, the timing is perfect.  Thanks!

PR #1213, opened 2020-08-01
Fix: Thermometer + Suspension

ace-1331:
- fixed the arousal meter being drawn before the suspension offset is applied which caused its hitbox to be in the wrong place

Currently, if you are suspended, the thermometer is drawn in the wrong spot so your clicks don't register like they should.

I also moved the online dynamic draw function in case there is ever a click handler on those as it would cause the same issues. For now, that one line being moved does not change LARP

As reported here -> https://discord.com/channels/554377975714414605/554378725916147722/739203452449587220


Ben987:
Good one.  I will test it right now.  Thanks.

PR #1215, opened 2020-08-02
Improvement: Sound and fetish adjustments

ace-1331:
- added some assets to the rope sound trigger
- added some fetish to a few asset and changed mislabeled ones

Had a deeper look and fixed a few inconsistencies.

My other question, the chopstick clamps/gag and bamboo gag are labelled as rope which I find odd. I assume its because its Shibari dojo themed?

In case you did not read discord, that was also my comment after your last post in #programming : [discord message link](https://discord.com/channels/554377975714414605/557728347468070912/738945545493807154)


Ben987:
Yes, since the chopsticks are tied with ropes and it fits in the whole "Shibari" mood I've used "Rope" for it.  Maybe someday we could add a "Shibari" or a "Wood" fetish but it might be too much.  Thanks for the fixes :)

PR #1217, opened 2020-08-02
Improvement: Reduced inventory sync calls

ace-1331:
- reduced the amount of inventory sync calls on login to 1 max.

Currently, every login triggers 2 inventory sync, and has the odds of triggering even more, This removes all of them in favor of only one sync. This singular sync only occurs if the player actually had something changed in their inventory.

I also have two questions:
1- The following line seems useless to me, but I saw you added it so I thought I would ask. I understand this was added when a new maid outfit was available, but with the buy group validation, this item will be earned automatically for anyone with the base maid outfit since they share the same buygroup. I tested and had no problem without this line.
 `if (LogQuery("JoinedSorority", "Maid") && !InventoryAvailable(Player, "MaidOutfit2", "Cloth")) InventoryAdd(Player, "MaidOutfit2", "Cloth");`

2- The other topic is, any groups of InventoryAdd can trigger several sync calls when it happens for the first time. I was thinking of refactoring those into a InventoryAddItems or something like that to add several at once which would trigger less server calls overall. Let me know what you think. If you say you'll accept it, I'll do that optimization.

For example, this would trigger 6 calls when earned for the first time:
```
InventoryAdd(Player, "NylonRope", "ItemFeet");
InventoryAdd(Player, "NylonRope", "ItemLegs");
InventoryAdd(Player, "NylonRope", "ItemArms");
InventoryAdd(Player, "ClothGag", "ItemMouth");
InventoryAdd(Player, "ClothGag", "ItemMouth2");
InventoryAdd(Player, "ClothGag", "ItemMouth3");
```

### DIFF
--- a/BondageClub/Screens/Character/Login/Login.js
+++ b/BondageClub/Screens/Character/Login/Login.js
@@ -360,6 +360,7 @@ function LoginResponse(C) {
 			SarahSetStatus();
 
 			// Fixes a few items
+			var InventoryBeforeFixes = JSON.stringify(Player.Inventory);
 			InventoryRemove(Player, "ItemMisc");
 			if (LogQuery("JoinedSorority", "Maid") && !InventoryAvailable(Player, "MaidOutfit2", "Cloth")) InventoryAdd(Player, "MaidOutfit2", "Cloth", false);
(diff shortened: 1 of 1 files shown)


Ben987:
Absolutely, the less server calls the better.

True, the maid outfit is probably from a time when we didn't validate the BuyGroup at login, it could be removed.

Yes, you could optimize the multiple InventoryAdd call, it would not hurt.

PR #1220, opened 2020-08-02
Fix: Odd prerequisites

ace-1331:
- Removed notkneeling from the sybian
- Removed duplicate hide items from the x-cross (It hid ItemBoots but also listed many boots individually which is not needed)

Kneeling before applying the item does not cause any visual problems and is a lot less annoying, especially for the sybian which is more intuitive.

Unless I am missing something here, I believe this notkneeling is 100% unnecessary and probably leftovers from copying prerequisites or old technical restrictions.


Ben987:
It could be a leftover from something else yes.  If that change causes an issue we can always revert it.  Thanks!

PR #1222, opened 2020-08-02
Bugfix: Resets character appearance text when accepting appearance changes

elliesec:
## Summary

This PR fixes a minor bug where the appearance screen text would not be reset after accepting changes to a character's appearance.

## Steps to reproduce

1. Change character A's appearance and accept the changes - note that while in the appearance screen the text should say "Select Character A's appearance"
2. Enter the appearance screen for character B - note that the text still says "Select Character A's appearance", rather than "Select Character B's appearance" (or "Select your appearance" if character B is yourself)


Ben987:
Good catch, I've never noticed that one.

PR #1223, opened 2020-08-02
Fix: chatroom visual stutter

ace-1331:
- fixed a visual stutter when re-entering a chatroom caused by a desync in the activepose

It took me a while to pin point what was going on here, but here goes:

Currently, going back to the mainhall changes your active pose without syncing with the server, this means that if you were kneeling before going in the main hall from a chatroom, it was not saved on the server.

This caused a visual stutter when entering a room -> the character is refreshed with the old pose (kneeling) then refreshed again with the right pose (standing). causing unnecessary redraws.

 This fix makes sure to sync with the server only if necessary, so it will not trigger a bunch of active pose syncs for no reason.


Ben987:
Alright!  Thanks for that one, I'll see if I can recreate it :)

PR #1225, opened 2020-08-03
Feature: Shock item arousal

ace-1331:
- added shock item support to the arousal system
- fixed the shock item action text to properly use DestinationCharacterName 
- added missing fetish to the shock plug

Self explanatory. Similar to the anal beads, triggering a shock can lower/raise arousal based on the player's settings

Edit: removed the dialog leave from shock items when clicking the checkbox as requested. I agree it was not very intuitive and is probably an old leftover


Ben987:
Yes!  Thanks a lot :)

PR #1227, opened 2020-08-03
Improvement: JSDoc comments added to RhythmGame.js

ace-1331:
- added JSDoc to RhythmGame.js


Ben987:
Absolutely!  I'm sure that Sonic7 will like it.

PR #1228, opened 2020-08-04
Improvement: JSDoc comments added to KidnapLeague.js

ace-1331:
- added JSDoc to KidnapLeague.js
- changed the click handlers to use MouseIn


Ben987:
Thanks for that one also :)

PR #1229, opened 2020-08-04
Improvement: Inventory screen refresh

ace-1331:
- optimized the interruption message to not send any if the player never progressed through the action (this prevents a message from being sent if a character was sitting on it.)
- improved how changes are handled while refreshing the screen to retain the current page and to take every possible changes into consideration.

With the new inventory colors change, I wanted to improve the new refresh system ahead of the beta so that is is more consistant. This means that it takes into account changes in the blocklist, limited list and active pose to always reflect the correct status.

This prompted me to change the inventory build function to take a possible offset. This means you will stay on the given page when rebuilding the screen to avoid having to scroll back constantly. This greatly improves the experience since it won't kick you back to the first page, it's especially good when nothing changed for the zone you are focused on.

With the amount of things in the next beta, I have a feeling some will complain about items moving or just changes in general... This should help alleviate those complaints, Regardless, this only happens if the player's appearance change in a way that affects the current zone, paired with interruptions, this assures we do not allow conflicting items and get characters soft locked.

### DIFF
--- a/BondageClub/Scripts/Dialog.js
+++ b/BondageClub/Scripts/Dialog.js
@@ -5,7 +5,7 @@ var DialogTextDefaultTimer = -1;
 var DialogProgress = -1;
 var DialogColor = null;
 var DialogColorSelect = null;
-var DialogPreviousCharacterAppearance = null;
+var DialogPreviousCharacterData = {};
 var DialogProgressStruggleCount = 0;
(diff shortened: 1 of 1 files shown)


Ben987:
Sure, going back to the previous page is a good idea.

Yes, the new beta will change lots of stuff :)

PR #1230, opened 2020-08-04
Fix: LARP option screen

ace-1331:
- Fixed the team selection action text to only show when the player actually changed their team/class
- Fixed the team selection screen to update live while inside of it to prevent having to go in and out to start when the admin is the last one to select a team. (Bad checks in `GameLARPCanLaunchGame`)
- Removed the `CommonWait` call for the host which made the game freeze
- Removed `CommonWait` now that it is unused. (And should remain as such.)

I tested many times and removing the commonwait call had 0 side effect. The order and turns were the same for all 4 test players and you can actually see the characters being shuffled on the host's screen since shuffling triggers ChatRoomSync. I'm not sure exactly why it was needed, but I couldn't figure out any bug this hid.

When it comes to leave/dc's the characters in the larp player lists are completely out of sync and break the game. Sometimes you could even get stuck in there forever and had to F5 out.

 I'm still investigating the matter. I will likely go for the following -> host dc/leave will forfeit the match, anyone leaving/dc'ing is flagged as out of the game and it is stopped for them. It will at least make it playable, then we can go for host migration and more complex stuff later. Though with the very low amount of people in LARP rooms, I don't think it's worth the time.


Ben987:
The CommonWait here is to make sure the room shuffle happened before the game is started, since everything is async you never know.  We could use a SetInterval call there also though.

ace-1331:
that leads to the same issue as the friend list where the server doesn't send back an acknowledgement, but yes I set that on a delay now with a slightly bigger delay to be safe

Ben987:
Yes!  That should work just fine.  Thanks a lot :)

PR #1231, opened 2020-08-04
Fix: Lock verification

ace-1331:
- fixed a scope issue when validating combo locks

We were testing the for let change and we already identified several code issues where it was already not working or broke because the code relied on bad practices. After fixing 2-3 of them, all the code runs fine after switching the 900 for-var loops as far as I can see for now, and each frame is drastically faster. I can publish my test branch if you want to see.

This bug was found while doing tests, the regex used the same var as the loop which caused scoping issues. After verifying the combination, it breaks out of the loop because `regex++ = NaN` and `NaN < anything = false`. This causes combo locked items to be "unsafe" and is a clear reason why `let` and `const` should  be used over var in 99% of cases.


Ben987:
Yep, I would be curious to see if changing the "for (var" to "for (let" do give better performance.  I'm checking with Sandrine on that one to see where I could notice an impact.  Thanks for that little one.  Cheers :)

PR #1232, opened 2020-08-04
Improvement: More Inventory Sync Optimisation

ace-1331:
- added a new `InventoryAddMany` function to add several items at once
- refactored existing calls to `InventoryAdd` to use the new `InventoryAddMany`. 
- reduced the amount of inventory syncs in the shop

This new function allows to add many items at once without spamming the server with the inventory array which is a large one. This function replicates the existing behavior of `InventoryAdd`. Meaning singleplayer areas that had the potential of triggering 0-6 syncs now only sync up to 1 time. 

Also, the shop spammed inventory syncs when an item has a buy group, this makes it so the inventory is synced after everything was added. Reducing any item bought from 1+ syncs to a singular one.

There is also a huge drop in performance in shopping mode, I'll have a look at that eventually.

### DIFF
--- a/BondageClub/Screens/Room/AsylumEntrance/AsylumEntrance.js
+++ b/BondageClub/Screens/Room/AsylumEntrance/AsylumEntrance.js
@@ -472,6 +472,8 @@ function AsylumEntranceEscapedPatientLeave() {
  * @returns {void} - Nothing
  */
 function AsylumEntranceGiveNurseUniform() {
-	InventoryAdd(Player, "NurseUniform", "Cloth");
-	InventoryAdd(Player, "NurseCap", "Hat");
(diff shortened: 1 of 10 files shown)


Ben987:
Sure, less server queries are always a good thing.  Have you tested all of these paths that you've changed?

ace-1331:
yes, I made sure all items were added and synced properly after using the new function, I also tested in the store to make sure items with buygroups also synced properly (by adding then reloading to make sure it was saved)

ace-1331:
This PR will conflict with #1239, but the changes are simple so I'll resolve them when one of them is merged

Ben987:
Thanks!  I will test it right now.  Cheers :)

PR #1233, opened 2020-08-05
Concealing Cloak

Codefreak64:
As you will see it is required to use Ellie's item priority integration

the purpose of this item is to hide some lower items while still having a realistic touch.

credit goes to ZFreak for the item
and Ellie for her change on the way items are loaded in


Ben987:
Thanks for that, but since you're also pushing some code from Ellie it creates conflict.  Can you only push the new item?

Codefreak64:
forgot to remove ellies item priority when merging, there should be no conflicts now

Ben987:
That looks great, thanks.  I will test it right now.

PR #1234, opened 2020-08-05
New Items to play around with 

gatetrek:
new items included in the pull are:
heart necklace
crotchless pantyhose
Leather Arm Splints
a dildo
Pelvis Chain Leash
neck accessories for the collar pup bone
a gag and pantyhose that goes on your face like a robber in the old movies


Ben987:
New items are always welcomed!  Thanks a lot :)

PR #1235, opened 2020-08-05
Fix: Dialog extended with NPCs

ace-1331:
- fixed an edge case where items could not be extended while focused on a NPC

Nina's previous fix was correct, but did not apply to extended items, so they were broken when targeting yourself while in the middle of a dialog with a NPC.

This fix should cover all cases.


Ben987:
Alright!  Thanks for that one.

PR #1236, opened 2020-08-05
Bugfix: Gag level tweak for the dildo plug gag

elliesec:
## Summary

This PR is just a small modification to the gag level of the dildo plug gag following #1123 (the gag level consistency fix). Upon examining the previous code, I realised that the intention was actually probably to make the open version of the gag have the `GagEasy` effect for a gag level of 3 (rather than 6 which is how it behaved before that fix), which would line up with the gag level on the ring gag. This PR makes that change, bringing the open version of the dildo gag in line with the ring gag.


Ben987:
Sure!  That makes sense.  Thanks :)

PR #1237, opened 2020-08-05
Fix: missing assets

ace-1331:
- added missing previews for the new wardrobe. (Could not find anything else missing)
- made the X button quit for UX 

Everywhere else in the club, X leaves the current menu, this was super unintuitive at first. Maybe the checkmark should be changed to the exit icon instead and the X becomes "revert"? whichever you prefer

I am not sure whether or not you finished your work on this screen, but now that we can, we can enable permission mode in there as well as extended clothes, prerequisites and more! that's a lot of work but it will enable a lot


Ben987:
For the "X", I've actually used the same logic as the color picker.  Should we also change the color picker?  Thanks.

ace-1331:
The current color picker leaves the current mode and restores the previous color. It does not just revert and stay on the screen. I'm not sure what you mean by that

Ben987:
I better check the color picker then, thanks for that!

PR #1238, opened 2020-08-05
Fix: scope issue line 255

ace-1331:
This is the fix you need for the let tests and it is a good example of how var+hoisting can cause silent scope issues in other places. This was not an issue with var but it could have been in another context


Ben987:
Yep, this is a good example.  Thanks!

PR #1239, opened 2020-08-05
Optimization: Shop mode

ace-1331:
- optimized Shop.js shopping mode

Currently, shopmode loops through `Asset` x1 and `Player.Inventory` x12 every single frame causing drastic frame drops compared to other screens.

This PR does a few things to optimize drawing in a similar way than Inventory.js. This cuts down a frame's render time by more than half.
The shopping cart is re-built when a change happens, and drawing/click handling is made based on the cart rather than being re-calculated every time.

I tested the normal mode and the "buy all mode", I tested normal items and buygroups, and made sure all clicks/pages were all behaving as intended. After buying every single item from a fresh inventory, I found no problem at all.

https://cdn.discordapp.com/attachments/727085309082206209/740607428499996833/Capture.PNG
(left side is with this PR, right side is current release.)

### DIFF
--- a/BondageClub/Screens/Room/Shop/Shop.js
+++ b/BondageClub/Screens/Room/Shop/Shop.js
@@ -9,7 +9,6 @@ var ShopText = "";
 var ShopRescueScenario = "";
 var ShopRescueScenarioList = ["BoughtEverything", "CatBurglar", "BoredVendor", "SleepingAtWork"];
 var ShopItemOffset = 0;
-var ShopItemCount = 0;
 var ShopDemoItemPayment = 0;
 var ShopDemoItemGroup = "";
 var ShopDemoItemGroupList = ["ItemHead", "ItemMouth", "ItemArms", "ItemLegs", "ItemFeet"];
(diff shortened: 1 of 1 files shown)


ace-1331:
This PR will conflict with #1232, but the changes are simple so I'll resolve them when one of them is merged

Ben987:
Are players having issues in the shopping screen?  I've never had any reports on that one.  How can I measure that?

ace-1331:
It's not really noticeable until you look at how unoptimized the page is. Take a look at the time per frame before and after the change, you will see it's a lot slower before. 

You can use console.time or performance.now (note that performance.now can be highly unprecise on certain browsers if you have some security flag enabled that prevent it from being accurate, so if you use firefox I recommend console.time for this)

Before this PR, the menu is rebuilt every single frame, that's like if we would rebuild the  inventory every frame when it is not necessary.

ace-1331:
Merge the optimisations from #1232

ace-1331:
merged and applied the proper changes to the things changed by the new hide item feature

PR #1240, opened 2020-08-05
Improvement: Gas mask

ace-1331:
**This PR depends on the new layer priority changes in** #1155 

- moved gas mask addons to their own group to be compatible with item devices addons. Now that we are getting more addons in that slot (bench, bed, etc.) it felt wrong to keep them there.
- added versions of the addons as pairs to allow wearing both the lenses and another add on at the same time as often requested.
- added a ItemHoodAddon category. This will now clean up groups: ItemHidden is for arms, ItemHoodAddon for hoods and ItemAddon for devices. We should make it clear that these should not be mix-matched in the future as we start to get more and more complex items.

The new layer change will be great to make 3D-ish (a bit like the private room cage is faking to be an item, but as a real item!) 

This is yet another example of what can be done with the layer priorities. Since the addons need various priorities, I could not make a combo before. Now that the technical limitation is lifted, it actually behaves like it was desired when I coded it a little while back.

These have often been requested, now that technical limitations are lifted, I could finally do it.


Ben987:
Thanks!  I hope the new zone won't be too small for mobile players, I'll test it right now.

PR #1241, opened 2020-08-05
School lockers

Ayesha-678:
Added 2 lockers, a standing one, and a small one, both with opaque and seethrough variants.


Ben987:

PR #1242, opened 2020-08-05
Feature: Chat search additions

ace-1331:
**This PR requires the related server PR for some of its functionalities.** [#41](https://github.com/Ben987/Bondage-Club-Server/pull/41)
No hotfix is needed for the current version once the production server is updated, and no effect should be felt from this change

- added the option to toggle on/off showing full rooms in the chat search
- added the option to order the chatroom list by number of friends
- added a friend counter next to the room name (good for mobile as they cant see the friend overlay)
- added support for pages to support more than 24 rooms during busy moments
- added a throttle to block search query spam. (forces 2s between queries)
- added a throttle to block join query spam. (forces 1s between queries)
- extracted the page per room into a variable to easily change the number of rooms displayed per page in the future if we want to display more information or have different modes (aka detailed vs concise mode)
- changed click handlers to use MouseIn
- added the ability to ignore rooms

## Note on sorting, pages, throttling
This PR is a follow up to the poll we made as this was one of the most popular QoL request. 
It aims to reduce confusion in the chat search area by allowing full rooms to be shown (less confusion of private vs public, and why a room cannot be seen when you search for it).

I tried many different things, I tried to balance functionalities, current behavior and server load as best as I could.

This helps with busy days when all rooms are not shown and you might miss out on some. This is also more future proof as the club getting busier won't become a problem in the near future.

The delays to prevent click spam should greatly reduce the amount of redundant  queries when a user clicks repeatedly on search/join while still remaining fast enough to be intuitive and not "miss your chance" because of a delay. I was able to trigger a search/join query about 20 times within 2 seconds which all got blocked by the delay. We can tweak it later on if we want it higher/lower.

On the server side PR, I changed the check for "if a room is full" and the 24 rooms limit. I raised it to 60 for safety, but later if that ever becomes a problem again, we can think of a complete revamp to the search functionalities. (That's not likely to happen any time soon) and the full rooms are still taken if it was specified in the query. With eventual room filters or more spaces, I dont think it would be common to have more than 60 rooms in the same space any way.  The specifics about the lines changed on the server are detailed in the server PR.

Tested the new sorting options with various different rooms as well as about ~60 rooms. The new default behavior will show full rooms, but it can be turned off.

There is no issue with merging this PR before the server is updated and vice versa.

This also means, if you want to keep the server the same way or change it in your way, this PR will still work without it. The only features that won't work are room ignoring, pages and viewing full rooms since these are limited by the server.

Feel free to comment any suggestions for improvements or added features while I'm at it so chat search can be perfected more.

## Note on ignoring

Following the rise of trolls and the efficiency of ghosting + autoban, they have started to just create rooms as this is the last thing we cannot block. With this PR and the server side PR, I have added a "ignored room" list. The list is cleared on relog. It is not saved on the server and the user can use the existing screen to toggle them on/off.

This new permission mode replicates the existing behavior of permission mode for items. When turned on, you see all rooms you chose to hide and you can toggle them on/off. This mode avoids misclicks and helps to keep the UI concise.

The goal here is to hide offensive rooms or any room spam to avoid having the whole list filled with irrelevant rooms. The server has a new check to pass ignored rooms to allow users to display as many relevant room has the server allows which takes down the last option trolls have to ruin everyone's experience in the club. It is vital for the ignored room list to be passed to the server in order to filter them out, otherwise a troll can still make 24+ rooms and prevent you from seeing legit rooms that way.

This will also be useful to simply hide rooms you are not interested in.  I extracted the normal and the permission mode draw/click handlers so eventually we could add more modes to the search screen. Like detailed vs normal, private vs public (if we want to display owner + lover rooms) or even filters, etc.

If you have any questions, let me know.


ace-1331:
After trying to find a way to do the filter without keyboard, someone gave a really good suggestion for filter mode. I will change the filter mode behavior later today for that more intuitive option

ace-1331:
Added the new filter mode to the screen. It makes it a lot more intuitive and easy to just click a room without being inconvenient

Ben987:
This will be the first change test on the new Azure server.  Thanks a lot!

PR #1243, opened 2020-08-05
Fix: new items fixes

ace-1331:
- repaired the CSV file (typo)
- added a missing file for the new double ended dildo in the kneel pose

#1209 contains the relevant fixes for the activities as that PR is what touches all of that


Ben987:
Good catch!  Thanks :)

PR #1244, opened 2020-08-05
Chat Room Calculation Fixes

Nina-1474:
This changes the sizing and positioning of characters in chat rooms to use formulas that make it slightly more accurate and consistent.
In the text to follow the format (2/3/4/5) represents the values for rooms with (2 people / 3 people / 4 people / 5 or more people) correspondingly.

- Zoom changes from (1/0.9/0.7/0.5) to (1/0.8325/0.625/0.5). This is the most substantial change but with the biggest benefit. The amount that the player canvases overlap each other is currently inconsistent, which means that asset makers making large rectangular assets have to limit themselves to the worst case (4 person rooms) to ensure there's never overlap. This zoom adjustment means every room size to has the same best-case overlap. The number of horizontal pixels available for wide items changes from (500/370/357.14/400) to (500/400/400/400).
- Space, the horizontal distance between character starting points, changes very slightly from (500/333/250/200) to (500/333.333/250/200) which can make the 3 person positions a tiny bit more even.
- The X offset goes from (0/66.667/50/40) to (0/83.33/62.5/50) which ensures the leftmost and rightmost characters have the same amount of space at their left and right respectively.
- The Y offset goes from (0/50/150/250) to (0/83.33/187.5/250) to centre the new zooms.
- The background drawing inputs are unaffected, other than to account for new Y offset and zoom values.
- Each characters' starting X and Y positions have been extracted into a variable to make the click & drawing inputs much more readable, but nothing changed here numerically.
- Replaced MouseX/Y checks with MouseIn.
- A redundant DoClick was removed as it will always be true at that point.


Ben987:
That looks good, it's not a huge change, I'm sure players won't mind.  I'll test it right now, thanks :)

PR #1245, opened 2020-08-06
Handheld Item Free Fix

Nina-1474:
The new free-item source for the item inventory didn't have the DynamicAllowInventoryAdd check that the other ones do, meaning the handheld toy would appear as an option everywhere even if the player wasn't holding one.


Ben987:
Alright!  Thanks :)

PR #1246, opened 2020-08-06
Tightjacket upgrade

Ayesha-678:
Upgraded the tight jacket with new assets, and tons of variants.  (extra padding, pulled out straps, lining straps etc...)

thanks to Clarity Velox Higgens for most of the ideas!

added a leotard version of the tightjacket with the same variants.


Ben987:
You're moving the BedChains asset out of the asset array, was that intended?  Thanks.

Ayesha-678:
Fixed, i think.

Ben987:
Very good!  I will test it right now.  Thanks :)

PR #1247, opened 2020-08-06
Fix: new appearance mode

ace-1331:
- fixed a bug where leaving the new wardrobe mode caused the focus group to stick

Right now, if you go to the wardrobe mode and cancel the focus group is kept so the next character to inspect will be in the inventory with the cloth category selected. (Since that button leaves the dialog now, it needs FocusGroup = null like the save changes checkmark)


Ben987:
Absolutely!  Thanks :)

PR #1250, opened 2020-08-07
Fix: Slow effect mass stopping

Kira22703:
As tittle mentions, this is a fix to having everyone with anything slow on to be stopped. Only target character is stopped now.

Also the text for stopping another player wasn't displaying fully.

Also also, stopped players are prevented from leaving for 30seconds instead of 45.


Ben987:
Thanks!  But can you make sure you merge the latest master branch before you apply your fix?  Some stuff was changed and some code was cleaned in the meanwhile.  Thanks a lot :)

PR #1253, opened 2020-08-08
Feature: Advanced alpha masks

elliesec:
## Summary

This PR modifies the behaviour of alpha masks by making them more flexible, and allows them to be applied selectively to specific item groups and specific character poses. This can remove the need to hide boots in certain devices (e.g. bed, crib, changing table), as well as the need for the `CannotUseWithAlphaHood` prerequisite, as alpha masks can now be applied selectively to the `BodyUpper` or `BodyLower` groups without affecting other layers.

This change also modifies boots (in both the `Shoes` and `ItemBoots` slots) with alpha masks, and the bed, changing table and crib to allow these items to be used together (I've held off on modifying the hoods for now to limit the scope of the PR, and also because I'm aware that @Nina-1474 has a separate, preferable, solution for alpha masks on head items on the way - #1271).

As an example, here is the bed, in combination with a pair of boots which define a group-specific alpha mask (note that the mask clips out the feet & socks, but leaves the bed intact):

group-masks

## Technical Details

Alpha masks are now defined in `Female3DCG.js` like this:

```js
{
    Name: "MyAsset",
    Alpha: [
        // Equivalent to the old alpha definitions - mask will be applied to every layer underneath the current one
        { Masks: [[200, 100, 100, 50], [50, 800, 400, 30]] },
        // Masks defined like this will only be applied to the named groups
        { Group: ["BodyLower", "Socks"], Masks: [[200, 100, 100, 50], [50, 800, 400, 30]] },
        // Masks defined like this will be applied in the same way as previously, but only when the character is in
        // one of the named poses
        { Pose: ["LegsClosed"], Masks: [[200, 100, 100, 50], [50, 800, 400, 30]] },
        // Masks defined like this will only be applied to the named groups when the character is in one of the
        // named poses
        { Group: ["BodyLower", "Socks"], ["LegsClosed"], Masks: [[200, 100, 100, 50], [50, 800, 400, 30]] }
    ]
}
```

The `Alpha` property is an array that can contain as many group/pose-specific mask definitions as desired. Masks can also be defined as above on a per-layer basis (not something that is currently used, but it's easy to enable without changing the logic, and could potentially have uses in the future).

To enable this, the following changes have been made:

* Modifies `Asset.js` to enable the new alpha definitions and allow `Alpha` properties to be defined on layers
* Modifies `CharacterAppearanceSortLayers` to keep track of group-specific alpha masks
* Modifies `CommonDrawAppearanceBuild` to handle group & pose specific alpha masks
* Modifies the signatures of the `drawImage`, `drawImageBlink`, `drawImageColorize`, and `drawImageColorizeBlink` callbacks in `CommonDraw.js` to permit an array of alpha masks to be passed in
* Modifies `DrawImageCanvas` and `DrawImageCanvasColorize` to handle the new `AlphaMasks` parameter
* Modifies the shaders in `GLDraw.js` to allow clipping masks to be passed in
* Modifies `GLDrawImage` to handle the new `alphaMasks` parameter
* Updates all assets with an existing `Alpha` definition to fit the new alpha definition format
* Modifies the Mistress Boots, Thigh High Latex Heels, Pony Boots, Ballet Heels, and Ballet Wedges (in both the `Shoes` and `ItemBoots` slots as appropriate) to use group/pose-specific alpha masks
* Modifies the open hair latex hood to use a group-specific alpha mask
* Modifies the Bed, Crib and Changing Table to no longer hide the aforementioned boots
* Removes a leftover `CannotUseWithAlphaHood` prerequisite on the vacbed

## Testing

As these changes are only applied when an `Alpha` definition exists on an asset, they should be relatively safe for the majority of assets. I have tested the boots as mentioned above, and have also tested group-specific masks in conjunction with hoods that use alpha masks (but have not modified hoods as part of this change), and have seen no problems using either `Drawing.js` or `GLDraw.js`.


Ben987:
This one is huge!  Thanks a lot.  Good thing there's an early beta coming up, players will be able to test with us.  Cheers :)

PR #1255, opened 2020-08-09
Ventless lockers

Ayesha-678:
Ventless variants of the school lockers.


Ben987:
Thanks!  This one creates a conflict with other pull request, can you double-check?  

Also, should we create a buy group for all the lockers and maybe only sell one model?  It could be priced a little higher but you would get all models at once.

Ayesha-678:
Fixed! also, i don't think so, since realistically you couldn't morph the locker into a a smaller one/ make or remove vents

PR #1256, opened 2020-08-11
Fix: speech.js

ace-1331:
- fixed hood addons non garbling properly

Forgot to change this line when splitting items. Since there are no more gagging items in ItemAddon, I just swapped it for ItemHoodAddon

Also updated the texts for consistency. 

_reminder that #1209 has multiple fixes too_


Ben987:
Yes!  Absolutely, thanks :)

PR #1257, opened 2020-08-11
Fix: item combos

ace-1331:
that item hide the cuffs in a few poses when it shouldnt, after trying all  poses I couldnt find any strange clipping


Ben987:
Sure, lets show our cuffs :)

PR #1258, opened 2020-08-11
Fix: TightJacket

ace-1331:
the folder was not named correctly so the previews were not showing for the new tightjacket.


Ben987:
Yes!  I've had a few reports on that one.  Thanks :)

PR #1259, opened 2020-08-11
Beta Bugfix: Latex Sleeveless Leotard fixes

elliesec:
I thought I'd already fixed some of these, but something must have slipped through the cracks or gotten lost in the merges. Fixes the following for the latex leotards:

* `Extended` property in the sleeveless leotard
* `Hide` property (ensures `ItemVulvaPiercings` is hidden) in the sleeveless leotard
* `AllowType` property in the sleeveless leotard
* Sets `AllowTypes` and `HasType` properties appropriately on the individual layers in the sleeveless leotard
* Adds missing lock assets for the polished versions of all three leotards
* Fixes the `ExtendedItemLoad` function to prevent object assignment issues (as per #1269 )


Ben987:
It's fine, this is why we do beta versions.  Thanks!

PR #1260, opened 2020-08-11
Beta Bugfix: Fixes for the plug gag and dildo plug gag

elliesec:
## Summary

This fixes some issues with the plug gag and the dildo plug gag. These have likely come about as a result of the layer priority changes, but I'd actually consider them fixes for a bug in the old appearance build functions (and haven't looked too hard into why this actually worked before, because it looks like the intention was for it to work as it now does). 

Note that this bug _might_ also apply to a couple of other items (any with both layers and extended types, where `HasType` was omitted on some layers) - I'll do a more thorough search and correct these later today, but these are the ones that came to mind initially.

* Adds `HasType: false` to relevant layers in the dildo plug gag and plug gag
* Corrects the open gag effect in the `AllowEffect` array for the dildo plug gag (as the effect was changed to `GagEasy` but I forgot to update the `AllowEffect` array)
* Removes the redundant `Open` type from the dildo plug gag (not needed as this is now using the extended item script)
* Removes redundant assets for the plug gag under the `ItemMouth2` and `ItemMouth3` directories, as the plug gag is no in those mouth slots.


ace-1331:
Just went through every item with layers and these seem to be the only two with that edge case

elliesec:
Awesome, thanks @ace-1331!

Ben987:
Yes!  Thanks for cleaning that up :)

PR #1261, opened 2020-08-11
Fix: InventoryRemove crash

ace-1331:
- fixed a crash with items containing a RemoveItemOnRemove item.

ways to reproduce -> Char1 and Char2 are in the same room, Char2 uses a leather armbinder with straps or a suspension hogtie , Char2 changes the color of it, Char2 tries to remove it -> Char1 has an error and the item stays

https://discord.com/channels/554377975714414605/554378725916147722/742716102185844796


Ben987:
That's quite a combo, yes, it could happen with a few items.  Thanks!

PR #1262, opened 2020-08-11
Beta Bugfix: Fixes for the high security straitjacket

elliesec:
## Summary

This PR makes some minor modifications to the high security straitjacket.

* Allows `HideItem` to be used in item properties
* Slightly ups the jacket's priority
* Reworks the groups/items that the various jacket combinations hide (mostly this prevents tail plugs from being hidden by the jacket)


Ben987:

PR #1263, opened 2020-08-11
Fix: SelfUnlock extended items

ace-1331:
- fixed several extended items that did not work as intended or had left over properties

All these items had irrelevant property checks or was trying to set SelfUnlock on the property level which did not work anymore when the item did not restrain the player following the check in Server.js 
(line 355) `["SetPose", "Difficulty", "SelfUnlock", "Hide"].forEach(P => delete Item.Property[P]);`

This caused an item to behave differently when you put it on yourself vs when someone else did.
I added the property on the asset level for the items that clearly intended to have it always set to false. Some others deleted that property even if it never existed.. and most of them already had SelfUnlock: false on the asset itself so it was useless to have that misleading line in the script. The other newer items never had that property work for them so I only removed that line to prevent others from copy pasting that mistake. This will keep the same behavior across the board without causing weird behaviors. 

let's see how much longer these bugs will be left ignored, similar to the desyncs in 1209


Ben987:
True, these are probably leftovers or copy pasted properties we were carrying for nothing.  Thanks!

PR #1264, opened 2020-08-11
Fix: Locked items locking locks with more locks trying to be normal locks

ace-1331:
- fixed a bug where copy pasted extended items kept the "re-apply lock" code when it never actually sets or resets Property.Effect which caused an item to have multiple locks..

Step to reproduce -> lock a straitjacket, change the tightness and you will have an extra lock effect

Several of these are bugged and the other half has unreachable code. For example you cannot change a chain pose while locked so I also removed the unreachable instances to prevent more bugs in the future or to prevent people from repeating that mistake

https://cdn.discordapp.com/attachments/727085309082206209/742765707351294023/unknown.png


Ben987:
Yes, I've seen a few items with multiple lock effects before.  Can you double-check the Tight Jacket scripts?  There's a conflict.  Thanks!

ace-1331:
merged

PR #1265, opened 2020-08-11
Fix: Missing asset

ace-1331:
So on this one, instead of having too many locks, we were missing one so I took one of the extras from the other PR and slammed this one in.


Ben987:
The locksmith strikes again!

PR #1266, opened 2020-08-11
Fix: Nose layer

ace-1331:
- adjusted the nose layer to render behind gags

Edit: left the nose ring on layer 43 as gags already hide it when they should, this prevents the attachments from clipping with all gags


FeuertinKira:
This will cause issues with things like the nose ring attachments being displayed under gags too. Probably needs a more nuanced solution to fix layering issues?

ace-1331:
the nose ring is already hidden by gags that should hide it, I didnt find a combo that caused problems

ace-1331:
Actually there are a few edge cases. Since the nose ring is already hidden by what should hide it, I left it on its own layer and everything else (nose hook, basically) is properly rendered under gags as intended. If some specific combos are found besides these we can look at them precisely

Ben987:
Sure!  Thanks.

PR #1267, opened 2020-08-11
Fix: exit icon in chatroom

ace-1331:
- fixed the door phasing out of existence while slow and stuck

The exit icon was not being drawn while wearing an item that ties you to a room and a slowing item (ball and chain and long chain for example)


Ben987:
Good catch!  I'm sure Kira will like it.

PR #1268, opened 2020-08-11
Fix: Gask mask remove

ace-1331:
- moved the item remove on remove to the asset level to prevent the addons from sticking when swapping from one hood to the other


Ben987:
Good one, swapping would be a problem.  Thanks!

PR #1269, opened 2020-08-11
Fix: Magic locks

ace-1331:
- fixed how default properties are assigned to pass the value **NOT** the reference which caused several issues
- fixed the locker locks disappearing

There was a bug where the lock was not reapplied, it was actually getting stored in the option objects due to how that object was assign when an item was loaded. That means, reapplying an item you previously locked the default version of would re-apply that same lock to the target character.

This is a major flaw in the logic which can also cause more strange behaviors, that's why I applied the fix to deep copy objects to every item that follows this pattern.

For the locker specifically, we are only interested in the type so instead of blocking the perspective when changing, we just assign the type (otherwise the lock disappears!)


ace-1331:
added the same fix for ventless lockers

Ben987:
Sure!  Since there's no function in the property object, we can clone it like that to prevent the object copy bug.  Thanks!

PR #1271, opened 2020-08-11
Improvement: BodyUpper and Head Split

Nina-1474:
**NOTE:** This change is built on and includes the inherit colour feature from PR #1270.

A new "Head" group has been created and the head parts of all the BodyUpper images have been removed. The head colour is automatically copied when the body colour changes without requiring a separate change in the wardrobe.
This allows the hoods which use alpha masks to hide parts of the head poking out to instead simply hide the Head group, which allows them to be used with the items they couldn't before.
It'll also give us a lot more flexibility in future if ever we want to hide either the head or the body but not both.

This didn't cover Hogtie and AllFours which still include keep the head part in BodyUpper, since I don't have the artistic skills to fill in the gaps! If someone else is able to do this, it will be a simple change to use the Head and allow hoods with them too.
I also didn't include the Leather Hood with open eyes as the head is showing here. It would be possible to cover this using a 'face' layer that inherits the body colour, but @elliesec's advanced alpha masks in PR #1253 would probably be a neater fix.

I tested every body size, colour option and pose, and a healthy cross-section, and everything looked as it always does.


Ben987:
Yes!  Absolutely.  I've been wanting to do something like that for a while now.  I hope this will take care of all the alpha hoods bugs.  Thanks a lot :)

PR #1273, opened 2020-08-11
Fix: yes, a lock, again

ace-1331:
**just in case we werent sure locks were utterly broken**

- fixed the click handler for owner timer locks

The click handler was different than the draw check, so you could click the button even if you added time once. This follows the pattern of the lover lock and the pattern of the buttons to work as intended


Ben987:
Alright!  Thanks again, this is great.

PR #1274, opened 2020-08-11
Beta Bugfix: Fix for gag level returning 0 on locked items

elliesec:
## Summary

This PR fixes an oversight in the new `SpeechGetGagLevel` function where the `Property.Effect` array would be used to calculate a non-extended item's gag level when it was locked (and hence it's `Effect` array would be `["Lock"]`, causing a gag level of 0 to be returned).

The function now checks for `Property.Effect` arrays containing only the `Lock` effect, and falls back to `Asset.Property` if the item isn't an extended item with allowed types.

## Steps to reproduce

* Equip any non-extended gag
* Talk - note that speech is garbled as expected
* Lock the gag (with any lock)
* Talk - note that speech is no longer garbled


Ben987:
Yes, I've had a few bug reports on gags, thanks for that fix!

PR #1277, opened 2020-08-12
Fix: Hidden Items Autoblock

Nina-1474:
There are two issues with the last-ish minute change to automatically block hidden items. 1) Players could hide items that they are currently wearing and the block would result in them immediately being removed. 2) If a player hides one item, unblocks it, then goes back and hides a second item, the first item becomes blocked again along with the second.
Changes could certainly be made to handle both of these issues, but it gets a bit more complicated. I'm second guessing the whole idea since it's not very clear for the player that hiding an asset will have that effect. It seems neater to just have the hide screen handle hiding and the permissions screen handle permissions. Plus I'd need to know if we're including PR #1249 or not. What do you think? For now I'm removing the BlockItems step.
I've also added two previews for ItemHidden that were missing.


ace-1331:
Would a text saving it is blacklist along being hidden help for clarity?

Nina-1474:
It would indeed. I did plan to do so if I'm adding the feature back in.

ace-1331:
Mm now that I think of it, I think the "on the fly" permission system we've known is useful, but adding a box for blacklisted on that screen which get checked automatically when clicking hide would actually be enough. The person would see it and could just uncheck it (though that would surprise me, unless they want effects of items with no visual for rp)

Nina-1474:
We think alike. I have in fact already started on that very idea!

Ben987:
It's perfectly fine, later on when blocking clothes becomes available it won't make a big difference anyway.  I'll see if I have time to review that one before the full release.  Thanks!

PR #1278, opened 2020-08-12
Fix: Orgasm Group Check

Nina-1474:
The group name was not being passed into the arousal activity call successfully, meaning the check for whether it allows an orgasm never returned true and an orgasm could never be reached.


Ben987:
Good one, I've had a few reports on orgasms that couldn't be reached.  Thanks!

PR #1279, opened 2020-08-12
Fix: vacbed typo + mask

ace-1331:
Spoke with Kirsty, the vacbed didn't block breasts as intended, and she wanted the fox mask back on the layer it was before. The cloth version of it was intended to show under front hair (similar to the domino mask) so I actually just used the layer masks use to be under front hair and not 28 like before. We can give that a try for now to respect nearly the same behavior those had before.


Ben987:
Nice one!  I'll do a quick look to see if we have other "ItemBreasts" items.  Thanks :)

PR #1280, opened 2020-08-12
Minor Fix: Unusable Fetishes

Nina-1474:
The AssetSpankingToys (the non-ItemHands version), LeatherCrop and LeatherWhip assets have been given fetishes that call ActivityFetishItemFactor, but these have no effect. This is because since they are "Wear: false", they'll never occupy a spot in Appearance and therefore can never affect arousal for these fetishes.
Leaving the fetishes there wouldn't break anything but it is a bit misleading, so this removes them.


Ben987:
Sure, we can remove it for now and add them later if we change the fetish code.

PR #1281, opened 2020-08-13
Hidden Items - Block Option

Nina-1474:
This is the change that was discussed in PR #1277.
The item visibility preference screen now includes a Block checkbox under the Hide one, which allows the player to set whether they want the item blocked. This will not appear as an option if the item is currently being worn or is already set to Limited.


Ben987:
That was quick!  I'll test it right now, thanks :)

PR #1284, opened 2020-08-13
Hidden Items - Block Auto-Tick

Nina-1474:
A small QoL change: clicking the Hide checkbox will now automatically change the Block checkbox to match it.
Sorry for the continuous trickle of PRs. Should be the last thing!


Ben987:
It's fine, it's a good feature that's getting even better :)

PR #1288, opened 2020-08-13
Beta Bugfix: Fixes chatroom messages for equipping wearable items with a defined Activity

elliesec:
## Summary

This PR fixes an issue where equipping wearable items with an `Activity` defined would cause incorrect chatroom action messages to be shown. This was happening because the equip action would be caught by the checks on line 631 of `ChatRoom.js` rather than the intended check on line 632. I've modified the check on line 631 to only include non-wearable items (handheld toys, etc.).

## Steps to reproduce

1. Character A equips a wearable item with an `Activity` on character B (e.g. Shock Dildo, Anal Beads XL) - note that this doesn't work for swapping, only equipping when the slot is already empty
2. Note that the chatroom message has no correct translation (see example image from discord)

Example


Ben987:

PR #1291, opened 2020-08-14
Improvement: wardrobe selection

ace-1331:
- made it so the red box of what is currently worn properly updates while browsing in cloth mode

The red box to determine what is currently worn does not get updated, so selecting through several items can become unintuitive. Especially now that we might have more buttons based on what's selected (Extended items, locks, etc.). It struck me as quite annoying when working on updating the necklaces which I should submit soon to showcase extended clothes and begin testing the system more.

Might conflict with #1287 but it will be easy to resolve
This might be worth merging first to polish the current update as 1287 is new tech for asset makers


Ben987:
Thanks!  But could you use a for loop instead of a map?  Cheers :)

ace-1331:
No. 
We took the step to add jsdoc, we finally declare variables properly. I think it's about time we use array functions. 
Feel free to fix this yourself with a for loop if you prefer, or if someone else wants to pick it up. Or just close this PR and leave the current behavior in

Ben987:
It's ok, we can leave it there until I have time to change it myself.  Cheers!

elliesec (approved):
LGTM!

Nina-1474 (approved):
Tested and works well.

PR #1292, opened 2020-08-14
Improvement: wardrobe restore properties

ace-1331:
- made it so canceling out reverts the extended properties 

Cancelling out a change did not save extended properties, this applies them properly when applicable.


Ben987:
Yes, it will be useful in the future.  Thanks!

PR #1293, opened 2020-08-14
Improvement: Extended script support for clothes

ace-1331:
- added explicit support for clothes in the extended script

This will allow an extended item to prevent being refreshed/sending messages so that changing an item will not change the appearance for others in the room. This will also allow us to tweak things based on restraints vs clothes in the future

Talked with Ellie to make this change as best as it could be.


Ben987:
Yes!  Something like that will be needed for advanced clothes.

PR #1294, opened 2020-08-14
Feature: Extended Necklaces

ace-1331:
**This PR works as-is, and is already functional, however it should be merged after all the other wardrobe improvements to behave smoothly:**
#1292 
#1293
___
- fixed a merge problem that occurred in the past update that duplicated the necklace lines
- added extended options to tuck them in/out

These items were actually intended to behave that way, but I didn't want to duplicate them back then, they are the best candidates to test the clothes extended system. This will also show that asset makers don't have to restrain their clothes to one layer if it makes sense + we no longer need separate assets for everything when unnecessary.

I made this as clean as I could and as re-usable as I could. This way, asset makers can have an example of what you can do with extended clothes without copy pasting any mistakes. I always prefer having a tech demo so this is kind of it, at least now people will be aware they can make more complex clothes and it gives us a way to flesh out the system more with only a few items which is easier to handle (already found a few improvements as per the other PRs). It's also another good example of what can be done with the new layer change.


elliesec (approved):
Looks good. Tested and working correctly!

Nina-1474 (approved):
Have you considered using OverridePriority for the Tucked types? The advantage is that it avoids image file duplication (not that there's much to worry about in this case), while the disadvantage is the new priority isn't located in Female3DCG.js and so is easier to miss when re-ordering them.
Looks good as it is though.

ace-1331:
Good call, just changed it for that. Although, we should probably have a check to prevent console abuse on that property. I never realized it was a thing

PR #1296, opened 2020-08-14
Fix: X-Cross Torso Items

Nina-1474:
The ItemTorso items were all using the standard pose versions, some of which include cut-outs for the hands. These gaps in the item appeared when the hands were overhead in the St Andrew's Cross.
The affected items are the Leather Harness, Hemp Rope and Nylon Rope which have been given images.
There is a small mis-alignment for them with the body but this seems the case for a lot of items in that pose, so that should be looked at separately.


Ben987:
Great catch there!  I did not notice that one.  Thanks a lot :)

PR #1297, opened 2020-08-15
Beta Fix: XCross priorities

ace-1331:
- moved the cross to layer 1 like any other background item (bench/bed/etc) which prevents conflicts with rear items as well as a few head items
- adjusted the butt/pussy area as it was slightly off (see [discord msg](https://discord.com/channels/554377975714414605/743491048352907355/744269196048990288))

Currently working with Sybil to allow going up to the XCross and being attached being separate for more flexibility which exposed the fact the priority was off compared to existing items


Ben987:

PR #1298, opened 2020-08-16
Beta Bugfix: Incorrect chatroom messages for the lovers vibrator

elliesec:
## Summary

The chatroom messages for the lovers vibrator referred to it as an egg. This adds new chatroom messages (that other vibrating items could use in the future to reduce the need for specific chatroom messages for every vibrating item) that address this, and hook the new messages up with the audio system.


Ben987:
Yes, very good.  Thanks for that one.

PR #1299, opened 2020-08-16
Beta Bugfix: Chatroom messages for the ventless lockers

elliesec:
## Summary

This PR changes the chatroom messages for the new ventless lockers so that they don't mention vents (change checked with @Ayesha-678).


Ben987:
Yes, that makes a lot of sense.  Thanks a lot :)

PR #1300, opened 2020-08-16
Beta Bugfix: Minor issue with layer sorting for layers belonging to the same asset

elliesec:
## Summary

This PR fixes a small issue with layer sorting between layers that belong to the same asset. If two layers belonged to the same asset with different priorities, the layer sorting algorithm would respect the position of the layers inside their parent asset's `Layer` array, rather than respecting their priorities. This meant that the sort order could become inconsistent, causing the layers to be rendered incorrectly (see image below).

This PR changes the sorting algorithm so that priority is most important, and position within the asset is used as a fallback in the case of equal priorities.

Incorrect layer order


Ben987:
Excellent!  Thanks for that fix :)

PR #1301, opened 2020-08-17
Beta Fix: extended leave

ace-1331:
- fixed a case where going in the wardrobe revealed restraints extended menus

some items do not reset DialogFocusItem under certain circumstances, this makes sure to reset any of them if they were still active.


Ben987:
Good catch there, I've never seen that case but it makes sense that it could happen.  Cheers!

PR #1303, opened 2020-08-17
Fix: Added ArousalZone attribute to TapedClitEgg

SandrineBC:
Added `ArousalZone = "ItemVulva"` to the vibrating clit egg, so that this item can trigger an orgasm as well.


Ben987:
Absolutely!  Thanks a lot :)

PR #1302, opened 2020-08-17
Cloak Fix

Codefreak64:
Some other assets were poking through the cloak, this is a fix for it

PR #1305, opened 2020-08-18
Fix: Conflict Cleanup

Nina-1474:
Part of the PR #1302 fix for the duplicate Necklace rows seemed to be lost in merge conflict resolution. This fixes that up.


Ben987:
Good catch!  Thanks a lot.  I'll prepare a hotfix.

PR #1309, opened 2020-08-18
Bugfix: Prevents NPCs from equipping the high security straitjacket

elliesec:
## Summary

This PR removes the high security straitjacket from the NPC random item pool, as its default type isn't `null`, and therefore having it equipped by an NPC would cause it not to render, as the extended item load function is never called. Because of the way the types are coded, changing the default type to `null` would be awkward, so I've opted to remove it from the random item pool.


Ben987:
Absolutely, it would not work great with NPCs.

PR #1310, opened 2020-08-19
Cape fix (again)

Codefreak64:
collar tags go over the cloak now, and refined the hidden items


Ben987:
Changing the ItemArms group priority might have side effect for other items, are you sure you've tested all the possible combos?  Also, why is that change targeting the NecklaceLock item?  Thanks!

Codefreak64:
sorry about how late this is. i made sure to test every item arm and they are fine. as for the necklace lock im not sure, chances are it got moved and deleted by mistake and I put it back the way it was. why its targeting that I don't know, but its unchanged

Codefreak64:
fixed the issue with the necklace and its good to go

Nina-1474:
Is the removal of the ArmbinderJacket priority necessary? It was presumably set that way for a reason so this could reintroduce some other priority issue.

ace-1331:

When you resolve the conflicts, please make sure the items retain the proper SetPose/AllowPose, etc. set in 1336 as it appears this PR might change some of the same lines.

Ben987:
Lot of assets were changed since that pull request was done.  @Codefreak64, if that fix is still needed, please recreate it with the latest asset file and make sure to test all the possible combos.  Thanks a lot.

PR #1312, opened 2020-08-21
Improvement: nursery milk

ace-1331:
- changed it so the milk wouldn't be taken as a restraint

To better replicate the previous behavior of the milk, and to keep it as a modifier more than an item, I changed it so it doesn't get removed when a NPC releases the character. This way people don't have to reapply it every time and it doesn't create soft locks since that was never counted as "gagging"


Ben987:
Yes, absolutely :)

PR #1314, opened 2020-08-21
Inherit Colours - Hands Fix

Nina-1474:
Moving the Hands group down below the customizable groups so that it doesn't leave a gap and disabling AllowColorize again.


Ben987:
Sure, that should fix it.  Thanks!

PR #1319, opened 2020-08-23
Fix: Removed unused property

ace-1331:
- removed `PropertyLocked` from collar tags as these are not in use

Noticed this leftover property as I was working on the dynamic assets framework


Ben987:
Good catch, I don't remember why we added that property back then.  Cheers :)

PR #1320, opened 2020-08-23
Fix: Background select text error

ace-1331:
- fixed the text behind the images displaying [object Object]

This one has been bugging me for a while, but never got around to fixing it.
You can see the text behind the image (selected background) being incorrect while the image loads because we passed an object instead of the text before. (`DrawButton(X - 4, Y - 4, 450 + 8, 225 + 8, BackgroundSelectionView[i], "Blue")`)


Ben987:
Yes, I've noticed that one also.  Thanks!

PR #1322, opened 2020-08-23
Fix: enclosed activities

ace-1331:
- fixed players being able to perform an activity on others while enclosed

Forgot to add this to the check when building the activity button


Ben987:
Nice!  This should fix that situation.  Cheers :)

PR #1323, opened 2020-08-23
Lock for cross

Uwnodi:


Nina-1474:
The item itself is still in the image rather than having the locks only.

Ben987:
Thanks, but it now creates a conflict with other pull requests, can you validate?  Also, have you made sure that the cross creator intended the item to have locks?  @Nythaleath , do you want to add a lock to the cross?  Thanks!

Nythaleath:
Yeah this is fine I did intend on making it lockable and as far as im aware Sybil was working on making custom lock assets as she is the one that made the cross asset in the frist place, so this should be fine for now and then when she finishes she can replace the lock asset with her custom one

Ben987:
Awesome!  Thanks you two :)

PR #1324, opened 2020-08-23
Advanced feature: Dynamic assets

ace-1331:
- built a framework from scratch to allow for asset drawing hooks (Before, After, Script)
- added a new custom tag to give an example of the new system
- added the option to disable scripts on assets on other online characters (for slow devices)

We've had a lot of new tech to make even more complex and fun items, but there was still something missing for a lot of suggestions to be feasible. This framework was born from dozens of ideas and was built with them in mind. This opens up sooooooo many possibilities, it's actually insane. Of course, it will need decent quality control to prevent asset makers from releasing unoptimized items. The built-in features of this should be used as they handle storage and refreshes in the most optimized way possible.

In short, this enables an item to carry a script, to alter itself and to draw things like text dynamically. This will allow things to be animated as well such as an animated shine effect on latex. It will also allow certain projects to take a lot less space than they currently do. For example, you wouldnt need 200+ assets just to simulate dynamic text

I added an option to disable the script part of the hooks in case you are using a slow device.

Also, ghosted players' assets wont be using their hooks as a safeguard. That way, someone wont be able to use dynamic assets to grief if it could ever be an option. This made me think, should we just not render the character at all when they are ghosted? it would be simpler and cover any case without asset makers having to think about that? Let me know what you think about that

We have a lot of ideas in mind, and this tag should give a good starting example of the AfterDraw method. Ellie also has a shock collar change and new vibrator modes that use the other functions ready to be PR'ed. That way we will have an example for each available feature already.

Of course, I expect this system to be improved and expanded over time as we discover more uses for it.

For full details on what's available and how it works:
[Dynamic Assets explanation](http://www.hostedredmine.com/projects/bondage-club-issue-tracker/wiki/(Advanced)_Dynamic_Assets)

Let me know if you have any question.


BondageClub/Scripts/CommonDraw.js
```diff
@@ -136,6 +147,7 @@ function CommonDrawAppearanceBuild(C, {
 		if (Layer.Name) L = "_" + Layer.Name;
 		if (!Layer.HasType) LayerType = "";
 		var BlinkExpression = (A.OverrideBlinking ? !AG.DrawingBlink : AG.DrawingBlink) ? "Closed/" : Expression;
+		var Color = CA.Color;
```
elliesec, line 150: Just a tiny one that probably came out of the merge with the colour inheriting changes - this variable is now declared twice. I've already removed this declaration in my shock collar branch.

ace-1331: fixed, thanks

ace-1331:
Edit: I have a concept for dynamic body writing and tattoo applier, just needs this to be accepted first as these would be complex, I dont want to waste even more time with something thats not in yet

ace-1331:
Merged #1253

ace-1331:
Quick note: I added a shared data system for items to be able to communicate. Explanations for this option is inside the wiki page as well.

ace-1331:
I'm sorry, but we're at a point yet again where this is non-sense. If anything major is merged beforehand and conflicts with this PR, Im closing it. I won't be wasting more of my time dealing with improper project management

BondageClub/Screens/Inventory/ItemNeckAccessories/CustomCollarTag.js/CustomCollarTag.js
```diff
@@ -0,0 +1,122 @@
...
+	}
+}
+
+// Leaves the extended screen
+function InventoryItemNeckAccessoriesCustomCollarTagExit() {
+	ElementRemove("Color");
+	ElementRemove("TagText");
+	PreferenceMessage = "";
+	DialogFocusItem = null;
+	if (DialogInventory != null) DialogMenuButtonBuild((Player.FocusGroup != null) ? Player : CurrentCharacter);
```
elliesec, line 86: Might as well use a `CharacterGetCurrent()` call here.

elliesec (approved):
Just the one tiny comment, but all looks good to me! This opens up a load of really nice possibilities for assets in the future.

Ayesha-678 (approved):
Looks good!

PR #1325, opened 2020-08-23
Feature: Adds a blinking light to the shock collar

elliesec:
## Summary

---

**Important**: This PR depends on and contains #1324 and should only be merged after that

---

This PR demonstrates the `BeforeDraw` and item refresh capabilities of the dynamic assets system by adding a periodically blinking light to the shock collar and the collar shock unit.

Below is a gif of the feature in action.

shock-collar


ace-1331 (approved):
This is a good example of how to use the features in Animation.js script for animations. Optimized as it should be!

Ayesha-678 (approved):

PR #1327, opened 2020-08-23
Feature: Advanced Vibrator Modes

elliesec:
## Summary

---

**Important**: This PR depends on and contains #1324 and should only be merged after that

---

This PR uses the dynamic asset system to add 5 new advanced vibrator modes to several vibrators. These new modes are unique in that they will automatically adjust the vibrator intensity from time to time. The PR also adds the concept of "automatic" chatroom messages, and a player preference to prevent automated messages that don't involve them from showing in chatrooms, to avoid potential chat spam (although I've tried to make the updates infrequent enough that spam shouldn't be a major problem most of the time).

The new modes have been added to the following items:
* Vibrating Nipple Clamps (nipples)
* Vibrating Heart Piercings (nipple piercings)
* Lovers Vibrator (vulva)
* Vibrating Dildo (vulva)
* Vibrating Heart Clit Piercing (clit)
* Vibrating Dildo (butt)

The new modes are as follows:
* Random - the vibrator will automatically change intensity randomly every 30 seconds to 2 minutes
* Escalate - the vibrator starts at minimum, then periodically increases. When it reaches maximum, it will then reset back to minimum intensity and repeat
* Tease - Behaves like "Random" mode, except when the player's arousal becomes sufficiently high, the vibrator will either drive them to orgasm, or it will edge the player for a period of time, before turning off and restarting
* Deny - Behaves like "Tease" mode, except the vibrator will never bring the player to orgasm
* Ramps the vibrator up to maximum and applies the `Edged` effect, preventing the player from reaching orgasm

Below is a preview of what the extended item screen looks like for items with the new modes.

vibrator-modes

## Technical Details

The following changes have been made:

* Adds a new user chatroom preference to prevent automated messages (from automatic vibrator updates) from being displayed if they do not involve the player - by default the preference is set to _not_ show automated messages
* Adds a new `VibratorMode.js` file. This contains:
    * All the logic for load/draw/click handling for both basic and advanced vibrator modes (other vibrators could easily be refactored to use this logic to show one or both sets of vibrator modes in the future)
    * Automatic vibrator update logic and chatroom message publishing
* Adds a new `Edged` effect. When the player is wearing an item with the `Edged` effect, they will be unable to orgasm. This effect is used by the "Edge" and "Deny" vibrator modes
* Splits several chat settings out into a separate online settings screen.


elliesec:
I need to rework the chat preferences screen as part of this change as we now have more checkboxes on there than can fit. Going to avoid pushing my merge conflict resolutions until I've gotten around to that.

elliesec:
Conflicts have now been resolved, and there is a new Online Preferences screen.

ace-1331 (commented):
Good example of how to use script draw from Animation.js. Nice to have an example for all among the current PRs.

Thanks for splitting the page, we've been needing it for a while now,

For consistency, I think some of the main page options should be moved there -> Show AFK bubble & Show wardrobe icon 

Let me know what you think about that, otherwise everything looks good

elliesec:
Seems reasonable to me. I've moved the two icon preferences into the online section.

ace-1331 (approved):

elliesec:
Updated to increase the time between vibrator updates (while hopefully keeping escalate mode roughly in line with the thermometer).

Nina-1474 (approved):
This all looks really good. I just have one observation:
If one vibrating item has an Edge effect but there are other vibrating items which do not, I would expect an orgasm to be possible, whereas at the moment it's blocked entirely.
It's a small enough drawback that it could be addressed at a later date.

elliesec:
> This all looks really good. I just have one observation:
> If one vibrating item has an Edge effect but there are other vibrating items which do not, I would expect an orgasm to be possible, whereas at the moment it's blocked entirely.
> It's a small enough drawback that it could be addressed at a later date.

Yes, my original implementation of `CharacterIsEdged` actually handled it in the way you've said you'd expect, but I changed it to the current implementation for simplicity. I was pretty on the fence as to which way to go with it. I can revert back to the original implementation if you think it makes more sense?

elliesec:
Urgh. The preferences got messed up, probably in the merge. I need to fix them before this can go in.

elliesec:
Done! Preference screen is fixed, and you can the "Edged" effect now only applies to the character if every vibrating item on a zone that can trigger orgasm has the "Edged" effect.

PR #1328, opened 2020-08-24
Fix: Reset to old emoticon after afk

SandrineBC:
# Summary
If a user has the "Auto AFK Setting" enabled and does nothing for 5 minutes, the "AFK" emoticon is shown over her head. As soon as she returns to the game, the Emoticon disappears, instead of resetting to the Emoticon, she had defined before.
This fix saves the old emoticon, before the "AFK" is displayed and reverts back to the old emoticon, when the player returns to the game.
# Details
## Step to reproduce
* Select an emoticon, e.g. "Read"
* Wait until the "AFK" emoticon is shown
* Click with your mouse in the game screen ==> The emoticon disappears
## What it does
* Added lines to save the old emoticon to a global variable in `AFKTimer.js`
## Testing
Did the above described steps on my local instance of the game and the behavior is as expected.


Ben987:
Yes, this is a good idea.  Cheers!

PR #1329, opened 2020-08-24
Fix: Missing asset

ace-1331:
- renamed the image as it was not named properly


Ben987:
Classic XLarge / Xlarge bug.  Thanks!

PR #1330, opened 2020-08-24
Fix: Web Gag Name

Nina-1474:
Changing the web gag name in mouth slot 1 to match the other two mouth slots.


Ben987:
Yep!  Thanks :)

PR #1331, opened 2020-08-24
Random Asset Selection

Nina-1474:
Now that permissions and prerequisites are possible for clothes, they should have the same checks that items do when picking a random option.

### Random Selection
A new InventoryGetRandom function now handles the random selection of an asset from a group or a defined list or an intersection of both. It expands the blocked/hidden checks from before into a more flexible system.
This now uses a binary ranking to check 3 conditions (at the moment) and find the most preferred combination that has some assets available to pick from. For example if the only options are assets that are blocked and visible or assets that are unblocked and hidden, we prefer the unblocked hidden ones.

The current conditions are, in order of preference: not blocked, not hidden, not limited permissions.
Or to put it another way, we want to pick our random asset from the first non-empty one of these lists:
- 000 - Not Blocked & Not Hidden & Not Limited
- 001 - Not Blocked & Not Hidden & Limited
- 010 - Not Blocked & Hidden & Not Limited
- 011 - Not Blocked & Hidden & Limited
- 100 - Blocked & Not Hidden & Not Limited
- 101 - Blocked & Not Hidden & Limited
- 110 - Blocked & Hidden & Not Limited
- 111 - Blocked & Hidden & Limited

(I know blocked and limited are mutually exclusive but it's still better/easier this way!)
The advantage of this ranking method is we avoid a long list of 'if' statements and if there's ever a fourth condition we want to check, it's a small change instead of needing to double the number of cases.

### Random Items
The InventoryWearRandom function will use the new GetRandom function. The only effective difference is the improved selection method. 
This function now also selects and uses a random colour from the group's pre-defined colour schema. Since this function is currently only used for item groups, and these groups all have a colour schema of ["Default"] only, the colour will not be affected for existing calls.

### Random Clothes - Full
The appearance screen now has a Randomize Clothes button next to the Randomize Character one. This will call the usual function to generate a random set of clothes without changing body/hair/size, as I expect most players will have settled on these details after a while and won't want them to change.
The FullRandom function now calls the new GetRandom method. There are also a few small fixes: an error in the parameters meant the last option in both clothes and colours could never be used, and poses are now applied if relevant.

### Random Clothes - Group
This button appears in the new clothes selection screen. It randomly selects and colours an owned piece of clothing in that group.

### Random Locks
This now uses the GetRandom function to avoid blocked/hidden/limited locks if possible. The checks around owner and lover locks are intact.


elliesec (approved):
Looks good to me! Tested and all working as expected. I feel like there should be a more elegant way to implement the permissions handling in `InventoryGetRandom`, but I can't think of one right now, and it looks about as efficient as it gets, so I'm happy for that to go in.

ace-1331 (approved):
Looks good to me.

Its likely to conflict with the other wardrobe PRs, depending on the order. 
Just need to make sure conflicts are resolved as intended. So whoever goes to merge needs to re-verify real quick

Nina-1474:
Merge completed to resolve the conflict from #1385 which was also adding an input parameter to InventoryWearRandom. This also resolves an error from that PR where it was passing the Refresh setting into Difficulty.

Nina-1474:
InventoryWearRandom used to only be called for item groups, which never had predefined colour schemas, so the new step to randomly colour items in that function wouldn't have recoloured existing random-item selections.
Since #1368 adds colour schemas to those groups, the function now explicitly checks if the group is clothing to applying a colour to it.

PR #1332, opened 2020-08-24
Feature: Keep NPCs from removing Cosplay items

SandrineBC:
# Summary
It was [requested](https://discord.com/channels/554377975714414605/721493158625804298/742824830679908422) that NPCs do not remove cosplay items like ears, tails or wings. This PR does exactly this. 
NPCs, like the nurse in the Asylum or the Club Mistress don't remove those items anymore, when they strip the player. Accordingly, wearing only cosplay items now counts as being naked.
# Details
## What it does
* Introduced an optional parameter `keepCosplay` to `CharacterAppearanceNaked` and `CharacterNaked`. Setting this parameter to `true` makes the Function keep all cosplay items on the player. The default for this parameter is `false`
* Changed the way the nurse in the Asylum strips the player, making her keep the cosplay items as well
* Added the parameter to all calls of one of the above mentioned functions within the main game. I did not touch the 'left hand side' parts of the game (theatre, stables, etc.)
* If wearing only cosplay items now the player counts as naked
* If wearing only underwear and cosplay items, the player is only wearing underwear for game purposes
## Small changes on the side
* Removed two lines from `Management.js that are not needed anymore due to changes in the removal behavior of collars
* Removed all appearances of the `KeepNaked` attribute as that was never used in the Asset definition to shorten several `if` statements
## Testing
* I tested the function with the Asylum nurse and the club Mistress
* After being a club slave, the maid in the main hall did not appear, if only cosplay items were worn


elliesec:
Hmm, question for both @ace-1331 and Sandrine on this one - with the ability to block slots from being touched on the way, would it be worth holding off on this one temporarily and then updating it to have NPCs respect the character's block settings, rather than flat out not touch cosplay items?

SandrineBC:
Well... I don't want to complain about all the dust on this PR... but when I did this, there was nothing 'on its way'...
If we will have a more generic approach in the near future, then please close this PR.

ace-1331:
Well, I'm waiting for the backlog to be cleared because right now it would be quite messy to do the intended change as I need the new preference screens and merge wardrobe stuff. 

Although, yes with the new settings it could easily just be done so NPC respects the player's preference which leaves the option for everyone to either keep the current behavior, or the one where ears and tails are respected. I already have a draft tested and all, but can't submit/finalize with the current backlog. Let's put this one on old and when the backlog is cleared we'll have a better idea of what works better

SandrineBC:
To be honest, it **is** old already (Sorry, couldn't resist)

ace-1331 (changes requested):
This will be a nice change for sure, check #1399 for the cosplay toggle, with it, you could check for it in functions where those groups could potentially be removed (but shouldn't)

Sadly, it requires a server-side pr so it might take a while, but you can still test on yourself with the right settings

ace-1331:
It appears you aren't using the latest changes brought by 1399, everything looks okay, but you'll have to change which accessory group is protected and which one is always removed (I added a comment to differentiate 1-2-3)

BondageClub/Screens/Character/Appearance/Appearance.js
```diff
@@ -212,18 +212,19 @@ function CharacterAppearanceFullRandom(C, ClothOnly) {
...
  * @param {Character} C - The character to undress
  * @returns {void} - Nothing
  */
 function CharacterAppearanceNaked(C) {
 
 	// For each item group (non default items only show at a 20% rate)
-	for (let A = 0; A < C.Appearance.length; A++)
-		if (C.Appearance[A].Asset.Group.AllowNone && !C.Appearance[A].Asset.Group.KeepNaked && (C.Appearance[A].Asset.Group.Category == "Appearance")) {
+	for (let A = 0; A < C.Appearance.length; A++) {
+		if (C.Appearance[A].Asset.Group.AllowNone && (C.Appearance[A].Asset.Group.Category == "Appearance") && !(C.OnlineSharedSettings.BlockBodyCosplay && C.Appearance[A].Asset.Group.BodyCosplay)) {
```
elliesec, line 223: I'm getting the following errors when `C` is an NPC (presumably because NPCs don't have settings):
```
Uncaught TypeError: Cannot read property 'BlockBodyCosplay' of undefined
    at CharacterAppearanceNaked (Appearance.js:223)
    at CharacterNaked (Character.js:705)
    at ManagementLoad (Management.js:334)
    at ManagementRun (Management.js:351)
    at CommonDynamicFunction (Common.js:196)
    at DrawProcess (Drawing.js:953)
    at MainRun ((index):309)
```
It's probably safest to make an NPC check here (and probably also in the two modified `Character` functions) to handle that case.

BondageClub/Screens/Character/Preference/Preference.js
```diff
@@ -203,6 +203,10 @@ function PreferenceInit(C) {
 	if (C.ChatSettings.ShowAutomaticMessages == null) C.ChatSettings.ShowAutomaticMessages = false;
 	if (!C.VisualSettings) C.VisualSettings = { ForceFullHeight: false };
 
+	if (!C.OnlineSharedSettings) C.OnlineSharedSettings = {};
```
elliesec, line 206: These lines look like they're duplicating the existing stuff down on line 238. Probably a merge problem?

elliesec (changes requested):
Mostly looks good - just a couple of comments, but should be fine after they've been addressed.

BondageClub/Screens/Character/Preference/Preference.js
SandrineBC: Jupp, looks like it...

BondageClub/Screens/Character/Appearance/Appearance.js
SandrineBC: I think I fixed it. How did you provoke this?

elliesec: Hehe, don't we all?
In my case I ran into it in the club management (as the stack trace says, `Management.js:334`), when it gets run on the NPC sub in the management room.

SandrineBC: I'll try it first. If I don't run into issues, I'll push it anew

SandrineBC:
Looking good now. I added a new function IsNpc that should help us to get rid of all those string comparisons...

elliesec (approved):
Looks good, all tested and working as far as I can see!

ace-1331 (approved):

PR #1333, opened 2020-08-24
Improvement: Optional slave collar (a.k.a The one)

ace-1331:
Here goes _clears throat_
____
**I know this change is likely to be rejected or require a painful discussion, but since this is very often requested, was brought up before and was voted high on Patreon, I did this as an example of how it could be done, and I personally think it's the right way to make everyone happy. It didn't take long to do it, and wouldn't take long to tweak the texts or tweak how it works based on feedback**
____

- added a club owner rule to allow removal of the slave collar

The default behavior stays the same, NPCs are left unchanged, existing checks to see if there is an owner locked collar accommodate for owner locked collars, the rule is deleted when losing an owner to maintain the default effect, etc.

Essentially, this gives the option to remove the collar and disable the slave collar validation if the owner wishes. This will essentially give access to the neck area in a normal matter which is perfectly fine now that we have owner locks. It also keeps all power in the hands of owners who should definitely be able to decide how they deal with their subs' collars

This also means, for anyone who still wishes to have the forced collar just like it is right now, they can ignore this new rule and it will not change anything for them whatsoever.


elliesec:
Looks good to me, although I'll hold off on an approval until we've put it to the community and let them have their say.

ace-1331:
I've added the slave collar status to the new rules menu while merging. This PR is ready

Nina-1474 (approved):
The only small suggestion I have is should the Release/Give dialog options for the collar have conditions so that only the relevant one of the two is displayed? It's a minor detail though and looks good otherwise!

ace-1331:
> The only small suggestion I have is should the Release/Give dialog options for the collar have conditions so that only the relevant one of the two is displayed? It's a minor detail though and looks good otherwise!

That's not something you can do currently, since you can't access the sub's log. That's how all the rules are, and it's a separate improvement that we definitely need. 

I avoided just checking for the slave collar in case someone does stuff with console/custom scripts, what they are wearing might not reflect their actual rule state.

gatetrek (approved):

elliesec (approved):

PR #1334, opened 2020-08-24
Bugfix: Fix for timer locks removing gag effects upon expiry

elliesec:
## Summary

There was a small bug with the way timer locks interact with the new gag level calculation code which meant that when the timer expired, it would leave the gag's `Property` with an empty `Effect` array, causing the calculated gag level to be 0.

## Steps to reproduce

1. Equip a (non-extended) gag
2. Attach a timer padlock (any kind) - don't set the padlock to remove the item upon expiry
3. Wait for the timer padlock to expire
4. Note that when the padlock expires, the player can talk normally (no speech garbling)


Ben987:
Nice catch, I've never noticed that one in-game.  Thanks!

PR #1336, opened 2020-08-26
Feature: Active pose system rework + Enhanced character menu

ace-1331:
**This PR requires the server side PR to enable active poses to pass as an array [#43](https://github.com/Ben987/Bondage-Club-Server/pull/43) for chatroom synchronisation to work, but everything works and can be tested on its own without the server change. The server change and this is compatible with R60 and wont cause any crashes**
___
- Improved the character menu to allow for many pages
- Added a character page to view currently active owner rules when fully collared
- Added a page to change active poses
- Added support for multiple active poses
- Changed how active poses work to ignore them if a pose in the same group is used 
- Added helpers for prerequisites to discern active poses and item poses

This PR contains several enhancements to current item behavior due to how the knew behavior handles active poses. Everything should behave the same as it used to. However some items will work better now that active poses are left out if needed.  Also now asset makers can go back and improve certain items to enhance their prerequisites and make them compatible with more items by adding a prerequisite check for a given pose group rather than a whole group of items. An example of this is the "must stand up" message, it will no longer be needed if you are only kneeling due to an active pose. 

I added a base pose to denote items that force the common standing pose (for example, the bouquet forces you in the normal arms pose) among other things. This will make it easier to add or change existing items by adding a "Free arms" prerequisite. Same for legs.

This means now we need to always define a pose, we always need to add a category and if the pose is completely finished for all items, we can add it to the menu. I made the menu dynamic so adding to it is as simple as adding a property on your pose. I'm hoping this will also give a really good incentive for asset makers to finish integrating their pose 100% rather than only doing it for their item

[Menu in action](https://cdn.discordapp.com/attachments/721493158625804298/748020597065384006/ewee.gif)

The following poses could realistically be added if their makers finish off the few items that dont work with them:
* backcuffs (needs arm cuffs to work in that pose
* spread (need shackles, ball chains and all itemboots)
* allfours (needs nipple zones finished)

I've provided helper methods to check if the character has a pose due to an item, which is what should be used for checks as this means an active pose will not prevent you from using an item (and will be ignored if necessary).

It's good to know that an active pose is never added when a pose in that category is set, this means if you want to allow active poses to contain something like kneeling, you need the new property to whitelist it (This is an edge case, the hobble skirt, straitdresses and legbinders had this behavior as a side effect of active poses, but I decided to make it a feature instead as I fixed the system)

It's definitely something that asset makers will need to adapt to, but it will make it WAY easier to handle poses when making assets.

Note that the screen is only available in chatrooms, this will prevent messing up anything related to NPCs as those will stay the same.

Let me know if you have any questions.

Credit goes to Ayesha for the new pose icons.


ace-1331:
I'm sorry, but we're at a point yet again where this is non-sense. If anything major is merged beforehand and conflicts with this PR, Im closing it. I won't be wasting more of my time dealing with improper project management.

elliesec (approved):
Tested along with the accompanying server change, and all looks good to me.

ace-1331:
Just added the missing icon for OverTheHead so everything is ready and it now includes the OverTheHead pose

Nina-1474 (approved):
Everything that I checked looked pretty good!
One trivial change I made was to trim the bottom of the OverTheHead icon as it was sticking out of the bottom of its box.

ace-1331:
Waiting for server to be restarted with the changes, but before we merge, there are a few updates to the cuffs we'll add to include the BackCuffs pose. I'll keep you updated and let you know when this can be merged.

ace-1331:
Now includes BackCuffs, thanks to ayesha for the updated cuffs + icon. This PR is fully ready to get merged once the server is restarted.

PR #1337, opened 2020-08-26
Fix: afk timer

ace-1331:
fixed the exception when the emoticon asset was in its initial state (often when loading from wardrobe or randomized character)

I also simplified the check to use InventoryGet


Ben987:
Awesome!  Thanks for that quick fix :)

PR #1339, opened 2020-08-26
Fix: text alignment for the duct tape

ace-1331:
- fixed the text alignment for the duct tape in the arms slot

The text wasn't aligned properly unlike the other slots


Ben987:
Sure!  Thanks

PR #1340, opened 2020-08-28
Fix: Remove background colour from activity messages for sensory deprivation

SandrineBC:
# Summary
When a player is in sensory deprivation, all actions and activities are anonymized, but activities are still displayed with a colored background, so the target is still able to guess, who did what to her.
This fix removes that background color from the message before it is displayed.


Ben987:
Good one, I've seen it before and it should make it more immersive.

PR #1341, opened 2020-08-28
Fix: Remove collar cuffs with the collar

SandrineBC:
# Summary
When a club slave has been fitted with collar cuffs and then goes to the club management to have her contract fulfilled, the management removes only the collar, leaving the player with the cuffs that cannot be removed.
This PR fixes that by adding the collar cuffs to the `RemoveItemOnRemoval` attribute of `ItemNeck` assets


Ben987:
Yes!  Good catch :)

PR #1342, opened 2020-08-29
Infatable leotard

Codefreak64:
Like the Inflatable Body Bag, but portable and as an "ItemArms"

Please merge when the time is right, I know about the server migration so no rush on this


ace-1331 (approved):
Appears to be working well!

BondageClub/Assets/Female3DCG/Female3DCG.js
```diff
@@ -1620,6 +1620,7 @@ var AssetFemale3DCG = [
...
 			  Prerequisite: ["NotHogtied", "NotMounted", "NoFeetSpreader"], Hide: ["Cloth", "ClothLower", "Suit", "SuitLower"], Block: ["ItemHands",  "ItemNipplesPiercings"] },
+			{Name: "InflatableStraightLeotard", ParentGroup: null, Fetish: ["Latex"], Value: 150, Top: 137 ,Left: 3, Difficulty: 1, SelfBondage: 6, Time: 30, RemoveTime: 50, AllowLock: true, Prerequisite: ["NotSuspended", "AllFours", "NotHogtied", "NotYoked", "NotMounted", "NotKneelingSpread"], Hide: ["Cloth", "Suit", "ClothLower", "ClothAccessory", "ItemButt", "TailStraps", "Wings", "ItemNipplesPiercings", "ItemVulva", "ItemPelvis", "ItemTorso"], HideItem: ["ItemNipplesLactationPump"], SetPose: ["BackElbowTouch"], Effect: ["Block", "Prone"], AllowType: ["Light", "Inflated", "Bloated", "Max"], Block: ["ItemBreast", "ItemButt", "ItemHands", "ItemNipples", "ItemNipplesPiercings", "ItemPelvis", "ItemTorso", "ItemVulva", "ItemVulvaPiercings"], SelfUnlock: false, Extended:true},
```
elliesec, line 1623: You probably want to add `ItemVulvaPiercings` to the `Hide` array - they clip out the bottom of the leotard.

elliesec (changes requested):
Looks good! Just the one comment that needs addressing. 

The only other thing is that it might be worth considering changing the spelling to "strait" instead of "straight". I won't insist on that because I don't really mind and I think "straight" _is_ actually a valid spelling, but I know we've had people complain about "gh" in previous straitjackets. As I said though - I leave that decision up to you.

PR #1344, opened 2020-08-29
Fix: Lock Permissions

ace-1331:
- fixed locks still being able to be blacklisted while worn as a check was missing in the click handler

Steps to reproduce: wear a lock, go in lock permissions and click the gray icon.


Ben987:
Nice!  I didn't knew that one.  Thanks :)

PR #1345, opened 2020-08-29
Fix: timer lock

ace-1331:
- fixed a rare edge case where the CurrentTime is altered while wearing a timed lock which makes them think they should be unlocked in the future.

This uses commontime to make sure it is checked every 1.7seconds consistantly regardless of how we manipulate CurrentTime


Ben987:
This one will be hard to recreate and test but thanks :)

PR #1346, opened 2020-08-29
Refactor: Audio.js

ace-1331:
- Refactored audio.js to be more reusable

I've overhauled the audio checks to be re-usable instead of hardcoded.
An asset can determine an audio track to play through Audio or DynamicAudio.

Sounds can be listed to map properties to them (for now, just file name), Actions that will play sounds can be listed based on their full name (prioritized as such) or a condition, etc.

That way, an asset maker just needs to add their audio file and/or set the name of an existing one for their item. Ive also added more recent items and extended items that did not use the existing sounds that matched them. Having them in the asset file should also remind people of being able to add sounds as right now they are often left behind.

Listing the actions separately decouples them from the information sent, helping cross-version compatibility and removing the need to increase data passed between clients which is why I did not take the route of attaching a sound name to an action. It will also give more control to the receiving in through future settings if that ever becomes a thing


ace-1331:
I'm sorry, but we're at a point yet again where this is non-sense. If anything major is merged beforehand and conflicts with this PR, Im closing it. I won't be wasting more of my time dealing with improper project management

elliesec (approved):

Ayesha-678 (approved):

elliesec:
For the record, I've now re-tested this against the new vibrator changes and it all seems to be working as expected.

PR #1347, opened 2020-08-30
Fix: safewording in spaces

ace-1331:
- Fixed a problem where safeword in the asylum was enough to be freed

Instead of safewording into the main hall, I made it so it goes to chatsearch. This keeps the exact same behavior and ensures we do not skip the chatroom space. In the case of LARP, it does not matter, but for the asylum, it makes it so a patient does not have a free pass.


Ben987:
True!  I'm sure Lucritus will be fine with that.

PR #1348, opened 2020-08-30
Bugfix: Fix for character appearance backup

elliesec:
## Summary

This PR fixes character appearance backups not working fully when modifying a character's appearance via the appearance screen. This is only an issue for the player changing the appearance (of herself or another player).

The PR adds custom character appearance backup/restore functions that can be used to serialise a character's current appearance to a string, and restore a character's appearance from a serialised string. These functions are currently only used in the appearance screen, but could be useful elsewhere in the future.

## Steps to reproduce

1. Enter the character appearance screen (for yourself or another player)
2. change the colour of an appearance item (and accept the new colour if doing it through the colour picker)
3. Exit the appearance screen via the cancel button
4. Note that the modified item's colour does not reset to what it was before entering the appearance screen

Note that if modifying another player's appearance, the cancelled changes are not sent to the chatroom, which causes the character's local appearance to be out of sync with their actual appearance.


Ben987:
Nice one there, thanks a lot :)

PR #1326, opened 2020-08-23
Extended Items Conversion

Nina-1474:
A bug was raised recently where reverse-suspension and ziptie frogtie could both be applied causing the player to be kneeling upside-down. This turned out to be because the item's script wasn't checking the Frogtie's prerequisites properly. Examination of other similar items showed similar incorrect copy-pasted versions, along with other inconsistencies such as where to display the requirement message.
Rather than correct these details, I've changed the items to use the standed ExtendedItem scripts which achieves the same thing. The extended script has a small change to support the Prerequisites settings without needing custom Validate functions that all do the same thing.
I tested each item and didn't find any issues or changes in behaviour other than the intended fixes.

PR #1355, opened 2020-09-02
HiddenItem rework

NoneNoname:
Since layers now have priority, ItemHidden no longer has a purpose, they do not allow color update and they require extra workarounds while they work like layers, we should only have one way to express one functionality so it is more simple to keep bugs away.

Added functionality to Layers:
HideAs: layers now can be hidden just like items or groups this allow the exact behavior as the previous
Top, Left:  allow other image size and position for layers


elliesec:
Might be worth checking with @Nina-1474 that the suspension ropes/chains in `ItemHidden` are okay to remove or whether that's going to cause pain for her upcoming variable suspension height changes.

Nina-1474:
I've moved the ItemHidden rope/chains to layers in my local branch in very much the same way as here, so I'm good to work with this.
Thanks Ellie!

Also a heads up, most of SetPose cleanups will conflict with #1326 which replaces them with ExtendedItem calls.

elliesec:
I figured that might be the case, but just wanted to double-check!

Ben987:
I'm glad to see a new change from you, it's been a while, cheers!  I don't know if you still check Discord, but developers will be able to approve pull request in a near future, so yours will probably be reviewed by Ace, Nina, Ellie, Kira or Ayesha, check the Discord channel for more detail.  Welcome back!

ace-1331 (approved):
Looks good, nice to finally get rid of this group.

When multi color is added, we can add the straps to the colorable stuff (#1368)

elliesec (approved):

PR #1358, opened 2020-09-03
Bugfix: Makes struggle progress per keypress independent of framerate

elliesec:
## Summary

This fixes a bug that would make it much harder to struggle out of items at higher framerates. `TimerRunInterval` was being used to calculate the amount of struggle progress on each keypress. This meant that at higher framerates it would take many more keypresses to struggle out of the same item. Note that the framerate is still used (correctly as far as I can see) to calculate the amount of progress decay/growth. Note that the 42 multiplier was calculated (and rounded up) using the expected value of `TimerRunInterval` at 60Hz, so this change should keep the struggling experience consistent with the current behaviour at 60Hz regardless of framerate.


SandrineBC:
Of course it had to be 42...

PR #1362, opened 2020-09-04
Improvements: More wardrobe changes

ace-1331:
- added the ability to view locked groups (useful to see what someone else's colors are)
- save properties to the wardrobe
- added a property to the DrawButton functions to "disable" them

This is boiler plate code for an upcoming PR, WardrobeGroupAccessible will be used to enable things like lovers being able to change the hair color, having an option for anyone to allow touching your character as well as adding an option to block touching tails/ears as well as enabling locks. This function can be used to tweak what's changeable by online characters without having to redo all the wardrobe and appearance screen code

I have a lock PR in the works, my only problem is dealing with NPCs right now, but this function will also greatly help. Seeing where extended clothes are going is also why I was able to determine that the best option is to save properties to the wardrobe as of right now. Most extended clothes will only have a type and a few minimal properties.

Adding the disabled button property will help simplify future and existing code as it wont require having to do if/else statements to draw a button or a rectangle only to hide the hover color. If anyone wants to, they can already simplify existing screens.


elliesec (approved):

Nina-1474 (approved):
Regression testing hit no issues. All good!

PR #1364, opened 2020-09-05
Fix: slave collar change

ace-1331:
- added the newer collar type in the mistress options
- fixed an issue where the collar properties were not actually kept which caused discrepancies when the type was changed by the mistress (Empty effect and block array)
- fixed some grammar errors

I had to add the "Leave" dialog to the list of that stage since you can select a slave collar type that gags you which would soft lock you otherwise.


elliesec (approved):

Nina-1474 (approved):
Looks good. I made a few additional grammar corrections around the same lines.

PR #1367, opened 2020-09-05
Improvement: JSDoc comments added to Private.js

ace-1331:
- added jsdoc to private.js
- removed an unused variable
- corrected a few typos in function names and calls
- changed some click handlers to use MouseIn

One more big file to go before we are done! So from now on we can have JSDoc added to all the new stuff and not have to worry about backtracking,


elliesec (approved):

Nina-1474 (approved):
Adding one missing param. Fine elsewhere.

PR #1368, opened 2020-09-05
Feature: Multi Colouring

elliesec:
## Summary

This PR adds support for multiple colours to be applied to items on a per-layer basis, along with a reworked item colour UI. This PR doesn't add explicit support for multi-colouring to any items except for the High Security Straitjacket, although any item with multiple layers with `AllowColorize: true` will automatically be multi-colourable by default. There is a separate PR in the works that will add multi-colour support to many items in the game, but I will raise that separately, as this is already a very large PR. The UI (and underlying system) works with both restraints and clothes, and is fully backward-compatible with the old string colour format. 

Below is a little preview of the new colour selection UI. Items that only have one colourable layer go straight to the colour picker as per the current behaviour.

colour-picking-preview

## Changes in this PR

* Allows item colors to be either a string (as per the current behaviour) or an array of strings
* Modifies asset layers to assign them a "colour index" - if an item's colour is a string array, and it contains an entry at a layer's colour index, the layer will be coloured according to that item in the colour array
* Modifies several existing functions across the codebase to ensure they are compatible with the color arrays
* Adds a new `TextCache` class, which can be used as a simple key/value text lookup (with translation) for arbitrary key/value type CSV files
* Adds a `CommonDebounce` function to allow a function to be debounced
* Adds a new multi-colouring UI component that can be used to apply multiple colours to an item
* Modifies the `Appearance` and `Dialog` UIs to include the new multi-colouring UI
* Splits the `MouseIn` function into two `MouseXIn` and `MouseYIn` functions for additional utility
* Adds colour schemas to most restraint asset groups

## A multi-colouring example

To give a more concrete example of how multi-colouring works from an asset creator's point of view, consider the following simplified example of a jacket with a crotch panel and straps:
```js
[
    { Name: "Jacket" },
    { Name: "CrotchPanel", CopyLayerColor: "Jacket" },
    { Name: "BodyStraps", ColorGroup: "Straps" },
    { Name: "CrotchStrap", ColorGroup: "Straps" },
]
```
* The `CopyLayerColor` property specifies that the `CrotchPanel` layer must always have the same colour as the `Jacket` layer.
* The `ColorGroup` property specifies that both the `BodyStraps` layer and the `CrotchStrap` layer belong to the same colour group. This means that they can be coloured together in the UI (they can also be coloured individually if the player desires)

In this example, the item would be automatically assigned 3 colour indices:
1. The `Jacket` layer _and_ the `CrotchPanel` layer
2. The `BodyStraps` layer
3. The `CrotchStrap` layer

Assigning the item a string colour (e.g. `#ccc`) would cause all layers to be rendered in that colour. However, suppose we were to assign the item the following string array as its colour:
```js
["#111", "#222", "#333"]
```
Now, the layers would be coloured as follows:
* The `Jacket` layer would be drawn with the colour `#111`
* The `CrotchPanel` layer would be drawn with the colour `#111` (as it copies the jacket's colour)
* The `BodyStraps` layer would be drawn with the colour `#222`
* The `CrotchStrap` layer would be drawn with the colour `#333`
Finally, because the `BodyStraps` and `CrotchStrap` layers share a `ColorGroup`, they can be coloured either together or individually through the new colouring UI.


ace-1331:

When you resolve the conflicts, please make sure the items retain the proper SetPose/AllowPose/AllowActivePose, etc. set in 1336 as it appears this PR might change some of the same lines.

Nina-1474:
The addition of colour schemas to item groups means that #1331's random colouring will now apply colours to randomly chosen restraints, which it previously didn't.
I'm assuming this is something we don't want? If not, I can change that PR to only colour clothes.

elliesec:
> The addition of colour schemas to item groups means that #1331's random colouring will now apply colours to randomly chosen restraints, which it previously didn't.
> I'm assuming this is something we don't want? If not, I can change that PR to only colour clothes.

Hmm, good point, and I'm not sure. I think there are some items that it makes sense for, and some where it makes less sense. Perhaps making it only colour clothes would be the most sensible thing.

Nina-1474 (approved):
No issues encountered and the new interface is very user friendly. Great feature!

ace-1331 (changes requested):
I couldn't find any compatibility issues which is great, everything seems to be working as intended, this will be great! I feel like with all the new features, we can now do anything we've ever wanted haha time to make and upgrade a bunch of items!

Small notes
1- all clicks are eaten by the color picking mode, I assume that's a side effect of how the click is currently forwarded to the color picker zone. Before, clicking on yourself made it quit the color picker. But that also means other menus just seize to work (such as the expression menu when looking at yourself)

2- After merging, it looks like the wardrobe buttons should probably use the new disabled property when no item is selected to retain a uniform behavior

elliesec:
@ace-1331 - those two points have now been addressed.

ace-1331 (approved):

PR #1371, opened 2020-09-06
Fix: Extended Items Next Page

Nina-1474:
The extended item script's click function had a reference to InventoryItemArmsWebOptions instead of the options passed in. This hasn't matter so far since the Web is the only item currently using the extended scripts with enough options to use a second page.


ace-1331 (approved):
Looks good 👍

elliesec (approved):
Good spot! All looks good to me.

PR #1373, opened 2020-09-06
New Item: Leather Bolero

Ruilov3:
Added new Leather Bolero top

As the item comes with a built-in "collar" piece - it covers most regular collars, having the "special" gag-colalrs visible underneath.
Attachments display properly on the item's own collar attachment points.

Duplicated the Leather Corset over to bras so it can combine with the new Bolero


ace-1331 (approved):
Looks lovely 💜

elliesec (approved):
And colours beautifully~

PR #1374, opened 2020-09-06
Fix: Missing locks

ace-1331:
- added some missing locks for the tight jacket types


elliesec (approved):

Nina-1474 (dismissed):
I think the two PulledPadding options should have an empty image rather than visible locks, since there are no buckles at the front.

ace-1331:
> I think the two PulledPadding options should have an empty image rather than visible locks, since there are no buckles at the front.

There was already one of the padding option with those locks on one of the two jacket (yet had no visible straps in front either) I only re-used the image that was used for the similar -ish styles. I don't mind having it either way

Ayesha-678 (approved):
Looks good!

PR #1376, opened 2020-09-08
Added new slots as a temporary thing till a better system can be made

Shiranui-Izayoi:
As I said these slots are designed as a temporary thing. I have already checked everything for issues such as clipping etc. And all items that needed it have had their priority changed, These shouldn't be added to as I hear that their are systems comming down the line that will make this obsolete but with no known due date it could be many months before that so I whipped this up.

In addition this makes a few smaller changes to bring a few bits of old code inline. Such as changing the mistress and pony item code to the newer and seemingly better InventoryAddMany function. I couldn't find a similar version for deleting so I left it at that.

Another change made was making it so the seamless versions of the normal and see-through catsuits now expose the same zones as their zippered equivelents. While I am aware it isn't "realistic" for them to do so. In this case I think it is better and justified to allow it. Mainly since as it is, them not exposing the zones makes it annoying to add or remove items from people wearing them and since clothing can't be locked it doesn't prevent people getting access simply requires more time. Which since they need to either be changed or stripped then ruins peoples outfits.

The final change made was due to adding the extra categories. I reorganized the categories so that all the suits and the gloves are on the same page. In addition I added a hidden dummy category in order to keep the hair accessories on the same page. The idea of this is to make it easier to customize outfits. Since things like hair accessory 1 and 2 tend to be customized together. Thus if they are on different pages you constantly have to keep scrolling all the way through each time. So the wardrobe now looks like the image provided below.
Menu Reorganization 2


Shiranui-Izayoi:
Hey um don't suppose I can get a reason why this is on hold?

ace-1331:
Hi,

After discussing it between a few of us, we've come to a similar conclusion we previously had on discord which I'll try to summarize here.

Duplicating groups several times -even temporarily- consists of quite the bloat and complexity for the resulting gain considering what can be done now and soon. This is some of our observations when it comes to it, I'll try my best to summarize the core points. I think if implemented that way, these extra options will be great in-game, and much more cleaner overall :)

Often temporary changes end up being permanent, and since most of the things this PR tries to do can be achieved in a permanent way right now, it would save everyone time, reduce risks and be a lot cleaner to do it the permanent way first.

1- To overlay gloves, the catsuit(s) can have the gloves added as an extension in the extended menu to avoid having to duplicate the gloves group

2- With multicoloring (which is currently ready in the backlog, so available to work from right now), the catsuit can be split in several layers to allow for different/optional color patterns.

3- Some fixes and other changes would need to be implemented/be adapted with this as several things will clash. Most notably, the "dummy" group should be removed altogether. To allow something like group categories, Noname has done something in their wardrobe that could achieve this result in a clean way (#747) if you're interested to have a look at it. Basically, clicking the top/bottom/head show the categories of that group, it could already re-utilize the fact some categories use ParentGroup and some don't. This makes it futureproof and won't require future changes to accommodate (or more than 1 dummy to make it fit) 

On a side note, the inventory add many change has no weight in this case, since it doesn't actually have an effect, It was added to prevent pushing to the server several times when adding more than 1 item, but during login, they all do not push to the server so either way is fine. (Can still be included in the final version of the PR though)

With the delay brought with last beta, the backlog is currently a little out of control, we're trying our best to approve/merge everything in a productive order.

Shiranui-Izayoi:
Ok that is understandable. And since it seems that all of this is coming down the line soon then this won't be needed and I will likely remove the pull request in the future once I see those implemented.

Also yer the dummy category was really only added for the sake of organization. But if what you said is true then that will not be needed with the wardrobe change. While I still think it might be good to reorgonize the general list I think that is something that can be done in a seperate pull request. I will look into that later since I still think it would be good to have groups that are regulaly editted together be on the same page.

Once again thank you for writing out why this is on hold I appriciate it. Though I will also admit that I am wondering how the multi colour stuff would deal with something like say combining the See-Through Suit with the normal one. I will likely take the parts of this that aren't related to the multi slot items and rework them as a seperate request such as the expose on the suits etc.

Shiranui-Izayoi:
I am going to close this request since with the new extended system we should be able to mix and match stuff with a little work. Maybe if I can be bothered I will merge the suits together and make various parts different layers so you could say have the See-Through version as a base then add the bunny suit over the top. I would likely make it so that it would go Bunny Suit at the highest, See-Through then normal version. But that will be a bit.

I still might look into reorganzing the slots in order to keep slots that are commonly editted at the same time together but that is a different matter.

PR #1378, opened 2020-09-08
New Item: Mermaid outfit

SandrineBC:
# Summary
The PR introduces a new set of items - the mermaid outfit
image
It consists of:
* The snorkel mask,
* a clamshell bra and
* the mermaid tail with a built in vibrator

This is a cooperation between @codefreak64 and me, where @codefreak64 had the bigger part in doing all the artwork.

# Details
## What it does
This PR 
* includes the images for the above named items, as well as the appropriate entries in `Female3DCG.js` and `Female3DCG.csv`
* adds the logic for the vibrator of the mermaid's tail in a new file `ItemLegs/MermaidTail.js`
* adds the file for the tail to the `script` section of `index.html`
* introduces a new prerequisite `NoOuterClothes`. This prerequisite ensures that the targeted character does not wear anything in the cloth or clothLower slot to avoid awkward situations with the new tail
## Prerequisites
This PR relies on the PRs #1324 and #1327 and should be merged *after* those two
## Testing
I tested this on my local BC instance and it works with a single player, two players and a player and an NPC as well.

# Credits
go to @codefreak64 for the great artwork, as well as to @elliesec and @ace-1331 for the basic work, I could rely upon (and the support of annoying me... ;-)


SandrineBC:
I tried to resolve the merge conflicts, but had to give up when it came to `Preferences.js`

ace-1331:

There does not appear to be conflicts with this PR, however I see you added AllowPose for the tail with many poses, Just need to double check if it will work well with all the poses included in the pose menu (OverTheHead, etc.)

elliesec (approved):
Tested, no issues found. Looks good!

ace-1331 (changes requested):
I'm leaving it at requesting changes until Sandrine replies. 

It looks like the OverTheHead pose does not have the asset like the others, but that leaves me wondering, wouldn't 1 base pose be enough with the right priorities? Either way, you might want to check overthehead in case you need to allow that pose with the right style

SandrineBC:
I added the 'OverTheHead' pose.

ace-1331 (approved):

PR #1379, opened 2020-09-08
Improvement: Unique Gag Prerequisites

Nina-1474:
Gags with the GagUnique prerequisite cannot be applied when another GagUnique gag is already worn. While this makes sense if the previous unique gag is on a lower ItemMouth layer, it shouldn't block the change if they're on the same layer, as the previous unique one will be removed.
The blocked gag also isn't being given a grey background because this check is in DialogItemClick instead of in InventoryPrerequisiteMessage.
GagCorset gags also can't be applied more than once and have these same two problems.

There is an existing prerequisite check which blocks GagUnique gags when a lower layer has either a GagFlat or GagCorset gag. This now checks for GagUnique gags. 
The check has also been extracted into a function so that GagCorset gags can use it to only be blocked by lower layer GagCorsets.
This change means we no longer need to use SetPose for those three gag types since the Prerequisites alone cover it.

Since this hits a lot of lines in Female3DCG.js, I'm keeping it as a draft for now until other PRs are merged to avoid conflicts elsewhere.


elliesec (approved):
Tested with a bunch of gags, and everything seems to be working as expected. I'm positive I fixed exactly this problem several months back, but a lot's changed since then, and building the checks into the prerequisite is definitely an improvement.

ace-1331 (commented):
Tested and everything seems ok. I didn't approve yet just to wait and make sure conflicts were resolved properly and do a final test after

ace-1331 (approved):

PR #1382, opened 2020-09-09
More disabled buttons

ace-1331:
**Depends on #1362 and should be merged afterwards**

- added the new disabled option to more buttons where appropriate (full rooms, chat admin when not an admin, extended menu for extended script items when a type cannot be selected, etc.)

This is more intuitive UI-wise as it brings a more common type of UI behavior. I'm hoping this will be used instead of hiding buttons or to remove over when buttons are disabled to be more clean.

Let me know if there are more buttons I didn't find/think of and I'll adapt them as well.


elliesec (approved):

ace-1331:
#1362 was merged, so this should be easier to test and review, as well as merge

Nina-1474 (approved):

PR #1384, opened 2020-09-09
fix eslint so it is more simple to run

NoneNoname:
npm scripts now can be run from `BondageClub/Tools/ESLint/` directly no need to copy or anything extra

.gitignore updated so npm modules can be installed to that path

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,5 @@
 .DS_Store
 
 /BondageClub/node_modules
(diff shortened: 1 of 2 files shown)


ace-1331 (approved):

elliesec (approved):
Eventually, I'd like to move the `package.json` higher up in the directory tree (which is where it was originally) and set up some proper tooling, but this is fine for now.

PR #1385, opened 2020-09-09
Reduce CharacterRefresh calls as #639

NoneNoname:


elliesec (approved):

SandrineBC:
Could we use a default parameter in the function declaration, like 
`function InventoryRemove(C, AssetGroup, Refresh = true)`
and this way simplify the refresh logic, as we will always have a value for `Refresh` provided?

ace-1331 (approved):

ace-1331:
> Could we use a default parameter in the function declaration, like
> `function InventoryRemove(C, AssetGroup, Refresh = true)`
> and this way simplify the refresh logic, as we will always have a value for `Refresh` provided?

currently the project uses the "true or null" pattern so for consistency, I personally prefer to level it like that, but I agree it would be nice to change them all to a default param. Though, it should be done to everything at once which is out of this PR's scope

NoneNoname:
we can create an issue here and once we get to it a pr can cover every change about default parameters, for now, I would too keep it consistent too

PR #1386, opened 2020-09-09
remove expensive JSON.stringify call from Dialog draw

NoneNoname:
CharacterRefresh now refreshes the dialogs if the user selected that character to view, this is much less heavy then calling a check on every draw.

DialogFocusItem = null; moved to DialogLeaveFocusItem() so no need to call after

Locks are now refreshed or unloaded depending on the state of the character.


ace-1331 (approved):
Perfect ^^

Nina-1474 (approved):
Tested and works well. 
I'd make a small suggestion to put the CharacterRefresh changes into their own function similar to the other steps (or at least add a comment) to clarify what it's doing. That's just preference though.

PR #1387, opened 2020-09-09
Allow owners to lock lovers locks

NoneNoname:
The MemberNumber should be the one who locks the lock, either the Player herself, or the Owner, or one of the Lovers, this changes will make it visible what currently left hidden.

Validate OwnerOnly and LoverOnly items properly allowing all combinations: 
OwnerOnly: applied by Player or Owner
LoversOnly: applied by Player or Owner or one of the Lovers

Add rules where the lovers can prevent the owner from using the lovers' locks, though the owner will still be able to remove it, this could allow the current behavior (except the bugs that the validation commit is fixing), this may be not necessary and may be unfeasible with the current rules that are only visible to the player

This change should prepare the way for a Lover's Belt that can truly be a lover's belt or an owner only belt in a single item:
The current plan is to add a Property 'OwnerOnly' & 'LoverOnly' that will limit the item usage just like if it would be on an asset


elliesec (commented):
I made a couple of little changes to the dialog file to fix the grammar (and also changed lovers' to lovers to keep it the same as the actual item name). I'm still working my way through this, but need to go to bed, so I'll leave the comments I have so far here - all just little things. I'll resume looking at it tomorrow.

elliesec (changes requested):

elliesec:
Not going to test this right now, but once everything has been addressed, I'll pull the latest down for testing.

BondageClub/Screens/Online/ChatRoom/Dialog_Online.csv
```diff
@@ -92,37 +93,42 @@ PlayerGagged,,,(You don't have access to use or remove items on her.),,!DialogDo
...
 170,100,(Release her from the slave collar.),(You remove her slave collar.),"SendRule(""CollarRelease"")",
 170,100,(Give her the slave collar.),(You lock the slave collar back on her.),"SendRule(""CollarWear"")",
-170,0,(Back to main menu.),(Main menu.),,
\ No newline at end of file
+170,0,(Back to main menu.),(Main menu.),,
+210,200,(Allow her to use lovers locks on herself.),(She will be able to use lovers locks on herself.),"SendLoverRule(""SelfLoverLockAllow"")",
+210,200,(Block her from using lovers locks on herself.),(She will not be able to use lovers locks on herself anymore.),"SendLoverRule(""SelfLoverLockBlock"")",
+210,200,(Allow her owner to use lovers' locks on her.),(Her owner will be able to use lovers locks on her.),"SendLoverRule(""OwnerLoverLockAllow"")",
+210,200,(Block her owner from using lovers' locks on her.),(Her owner will not be able to use lovers locks on her anymore.),"SendLoverRule(""OwnerLoverLockBlock"")",
+210,0,(Back to main menu.),(Main menu.),,
```
elliesec, line 134: The lover rules dialogue options don't chain together properly. The 210s in these 5 options should probably be 200s.

BondageClub/Scripts/Server.js
```diff
@@ -544,6 +509,30 @@ function ServerAppearanceLoadFromBundle(C, AssetFamily, Bundle, SourceMemberNumb
...
+	if (Item.Property.RemoveItem != null) NewProperty.RemoveItem = Item.Property.RemoveItem; else delete NewProperty.RemoveItem;
+	if (Item.Property.ShowTimer != null) NNewProperty.ShowTimer = Item.Property.ShowTimer; else delete NewProperty.ShowTimer;
+	if (Item.Property.EnableRandomInput != null) NewProperty.EnableRandomInput = Item.Property.EnableRandomInput; else delete NewProperty.EnableRandomInput;
+	if (!NewProperty.EnableRandomInput || NewProperty.LockedBy != "LoversTimerPadlock") {
+		if (Item.Property.MemberNumberList != null) NewProperty.MemberNumberList = Item.Property.MemberNumberList; else delete NewProperty.MemberNumberList;
+		if (Item.Property.RemoveTimer != null) NewProperty.RemoveTimer = Math.round(Item.Property.RemoveTimer); else delete NewProperty.RemoveTimer;
+	}
+	Item.Property = NewProperty;
+	ServerValidateProperties(C, Item);
+	if (Item.Property.LockedBy == "OwnerPadlock") InventoryLock(C, Item, { Asset: AssetGet(AssetFamily, "ItemMisc", "OwnerPadlock") }, NewProperty.LockMemberNumber);
```
elliesec, line 532: `AssetFamily` here and on the next line aren't defined.

elliesec (changes requested):
I've started testing this locally, but have been unable to get very far with it so far. Aside from the 2 comments made in this review, I'm also running into the following error when entering a chatroom whilst wearing a lover locked item:
```
Inventory.js:635 Uncaught TypeError: Cannot read property 'IsLock' of null
    at InventoryLock (Inventory.js:635)
    at ServerItemCopyProperty (Server.js:533)
    at ServerAppearanceLoadFromBundle (Server.js:405)
    at CharacterOnlineRefresh (Character.js:308)
    at CharacterLoadOnline (Character.js:396)
    at ChatRoomSync (ChatRoom.js:1089)
    at r.<anonymous> (Server.js:41)
    at r.emit (index.js:83)
    at r.onevent (index.js:83)
    at r.onpacket (index.js:83)
```
If you could test and address any issues, then I'm happy to pull down the changes and test again.

NoneNoname:
I fixed a bit more and also added way so the rules now kind of visible for the lover/owner as in only the reverse option is visible.

ace-1331 (approved):
That's definitely a nice improvement, that menu has been rather annoying for a while. I'm sure we can add more simple lover rules in the future to make them more interesting.
Although, I do have some small comments you might want to think about:
- The rules are not displayed in the "current rules" menu when viewing yourself (they're simple to add)
- The dialog when clicking on rules sends you back to the beginning instead of staying in the "Lover rules" stage so its a little unintuitive

NoneNoname:
I added the return to the lover rules, it may worth considering to do the same with owner rules and add a back to selection option above the back to the main menu

elliesec (changes requested):
Hmmm, I can't make the rule work for preventing owners from using lovers locks. Could you take a look?

What I'm trying:

Characters 1, 2 and 3
1 lover of 2
3 owner of 2

- Character 1 clicks character 2, selects "Block her owner from using lovers locks on her"
- Character 3 is then still able to use lovers locks on character 2

elliesec:
I'm still seeing the same issue with the rule for owners.

I'm also seeing this issue:

- Character 1 clicks on character 2 and selects either of the new rules
- Both rules get toggled

This is only a problem once - after that the rules seem to toggle as intended. Upon leaving the chatroom and entering again, I see the same issue. Kind of looks like the rules aren't being initialised properly the first time, or something like that?

elliesec:
I've resolved the merge conflicts on this, and will be testing it thoroughly over the next couple of days, with the aim of getting it merged after the R63 release. @NoneNoname - sorry for the very long delay. If you'd like to take over, just let me know. Otherwise I'll push any fixes if I find bugs, and we can hopefully get this merged soon.

elliesec:
As far as I can see this is all now working as expected. I'll continue testing over the next day or so, and will merge once I'm confident there's nothing I've missed.

elliesec (approved):
Okay, I've tested this as exhaustively as I think I'm able, and everything now seems to be working correctly.

PR #1388, opened 2020-09-09
Extended Item Option Layout Enhancement

Nina-1474:
This makes some notable changes to the common ExtendedItem functions around how extended item options are displayed in the selection screen. Namely:
- The name of the option will now be displayed within and at the bottom of the white rectangle, to be more consistent with other selection menus and leave more space for more options.
- Items will show 8 options per page by default. A developer can configure a specific item to show fewer options per page if they wish by passing the number into the Draw & Click functions. (If enough time passes and no-one uses this option we could probably remove it for simplicity).
- The options will be laid out based on pre-configured positions in order to fill the space evenly.
- There are now Previous and Next buttons, instead of just a cyclic Next button.

### DIFF
--- a/BondageClub/Scripts/ExtendedItem.js
+++ b/BondageClub/Scripts/ExtendedItem.js
@@ -28,6 +28,19 @@
  */
 var ExtendedItemOffsets = {};
 
+/** The X & Y co-ordinates of each option's button, based on the number to be displayed per page. */
+const ExtendedXY = [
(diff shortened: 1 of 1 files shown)


ace-1331 (approved):

Ayesha-678 (approved):
Looks good asset menu wise!

elliesec (approved):

PR #1391, opened 2020-09-10
Refactor: Item arms leather cuffs to extended items

SandrineBC:
# Summary
This PR refactors the leather cuffs for the arms to use the extended item script.
To keep this item consistent with the others, several files had to be renamed (and added)...


elliesec (approved):
As Nina mentioned, it might be worth looking at making `HasType` available at the asset level. Combining that with `AllowPose` would probably let us achieve the same thing with fewer assets. But this is an improvement for now!

ace-1331 (changes requested):
#1396 Adds HasType on the asset layer to allow phasing out `Restrain` , if you merge and start from there, you can keep all images as they were before, looks good otherwise ^^

ace-1331 (changes requested):
All good when it comes to the first comment.

But before we merge, something from the singleplayer also needs to be adjusted ->
`SarahSophiePreparePunishCharacter` and `SarahSlaveLockedCuffs` both need to use type instead of Restrain now.

This PR is also dependent on the fix in #1414

SandrineBC:
Problems in `Sahrah.js` resolved. How did you spot that one?

ace-1331:
Was already aware of it, but remembered when doing a projectwide search for .Restrain when I realized we were about to get rid of it

ace-1331 (approved):

ace-1331:

When you resolve the conflicts, please make sure the items retain the proper SetPose/AllowPose/AllowActivePose, etc. set in 1336 as it appears this PR might change some of the same lines. It looks like you just need to add` HasType: false` to the incoming change, however please double check the type array to make sure they all set the proper pose and block the pose menu

SandrineBC:
Everything works fine.

PR #1399, opened 2020-09-10
Wardrobe permissions

ace-1331:
- added an option to allow everyone to touch your character
- added an option to block tails/wings/accessories from being removed by others (could be used for NPCs)
- added a new online shared settings object to characters for settings that need to be synced
- changed the strip button to support the new accessibility function. This will make sure nothing blocked can be changed (will be especially useful when locks are introduced!)
- re-organized the hair groups to have ears&accessories, ears, accessories as 3 groups


ace-1331:
#1362 was merged, this should now be easier to test

Nina-1474 (approved):
I hit an error on loadup and put in a quick fix for it. Otherwise seems to work as intended!

ace-1331:
ah yeah possible merge conflict gone wrong

elliesec (approved):
Looks good, and all seems to be working as intended. 

The only thing that could perhaps be added would be to enforce those permissions on the receiving player's end in `ServerValidateProperties` - although maybe that's overkill? Not sure, let me know what you think. Otherwise, happy for this to go in though.

ace-1331:
It currently is not validated already, and I don't want to mess with existing custom code because it would disable a lot of cool stuff.  This is really just a way to not have to use console if you want the same effect.

elliesec:
Eh, that's fine. It was just a thought because the code already validates and blocks certain changes that are similar, and it feels like something people might want some protection on, but I'm not particularly fussed either way personally.

PR #1405, opened 2020-09-11
Hotfix for audio.js

ace-1331:
- removed duplicate properties
- fixed dialog sound

So, I was totally unaware ropes were the only item with sounds in singleplayer, so I adapted it with the new system.
This also means now that items will have sound in singleplayer as well, so that's a nice suprise upgrade!


Ayesha-678 (approved):

Nina-1474 (changes requested):
The idea behind RopeLong was that it only plays when struggling to get free from ropes. And it would cut off whenever the struggle screen was exited.
RopeShort covered the application of ropes. By using RopeLong instead, the sound is going on for much longer than it should.

Nina-1474 (approved):
Sounds good now!

PR #1411, opened 2020-09-11
Fix inventory refresh with locks

ace-1331:
Depends on #1386 

This fixes cases where a refresh is called while browsing through locks. It is especially annoying with animated assets (you can test by trying to lock the shock collar before/after the recent changes.

I simply check if we are in lock mode, then if we are, we rebuild locks instead of items.


Nina-1474 (approved):
Checked before and after and much better now!

ace-1331:
I've added 2 lines to Dialog.js to find more weirdness I've found ->
- Going in permission mode leaves lock mode to prevent stuttering when refreshing (among other things)
- Going in lock mode will properly rebuild the buttons (like the next button should disappear)

elliesec (approved):

PR #1414, opened 2020-09-11
Asset has type: lock

ace-1331:
- this makes it so an asset where we marked the type as irrelevant won't try to look for extra locks.

This will prevent having to rename locks with the items being moved over to the extended script and will make them more uniform

This has to be merged before more refactors are introduced.

(This currently fixes some that already have been merged such as the chastity belts)


elliesec (approved):
Yep, good spot.

Nina-1474 (approved):

PR #1417, opened 2020-09-12
Refactor: Chains (ItemFeet)

ace-1331:
- refactored the itemfeet chains to use the extended script

### DIFF
--- a/BondageClub/Screens/Inventory/ItemFeet/Chains/Chains.js
+++ b/BondageClub/Screens/Inventory/ItemFeet/Chains/Chains.js
@@ -1,95 +1,59 @@
 "use strict";
+var InventoryItemFeetChainsOptions = [
+	{
+		Name: "Basic",
+		BondageLevel: 0,
+		Property: { Type: null, Difficulty: 0, SetPose: null },
+	},
+	{
+		Name: "Strict",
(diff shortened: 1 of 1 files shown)


elliesec (approved):

Nina-1474 (changes requested):
Since this item has a Validate function, the Prerequisite aren't being checked within ExtendedItem.
The reasoning is an item may want to handle prereqs in its own way, for example Web which needs to ensure the prereq isn't caused by itself. The usual preqreq checks will need to be covered in this item's Validate.

ace-1331:
Used the same process as the web, should be ok

elliesec:
Just fixed a couple of little things - I'll pull it down to test soon.

Nina-1474:
Since the options don't have two types that block each other (unlike the web which has seperate types with SetPose: Suspension and Prerequisite: NotSuspended), I don't think the removal and reapplication of the item in Validate is actually necessary.

ace-1331:
sure, as long as no one forgets about it if/when more types are added (though arguably, that should be handled in inventory allow

Nina-1474 (approved):
Made a small fix to add the character line back in. All good!

PR #1419, opened 2020-09-12
Fix: lower suit

ace-1331:
- fixes the ghost problem when wearing a suit hiding item + a lower suit

The suit is layered on top anyway, I couldn't find any reason why it would hide the body.

See -> https://discord.com/channels/554377975714414605/554378725916147722/754133691864711238


gatetrek (approved):
was hidden due to some the body was clipping, but if not being clipping now then it's all good

ace-1331:
After double checking, I can confirm I don't see any clipping with all sizes/colors in the standing pose and kneeling pose 👍

elliesec (approved):

PR #1429, opened 2020-09-13
Fix for leg and active poses

ace-1331:
Just realized these changes got lost somewhere along the line of merges and stuff, but leg items almost all need the new AllowActivePose, this will make them work as they should.


Nina-1474 (approved):
Tested and seems good.

elliesec (approved):
Tested, looks good.

PR #1431, opened 2020-09-14
Important fixes

ace-1331:
- fixes the submodule linking that were incorrect which caused issues when checking out. (I just removed it, since no one other than reviewers have push access anyway. If we want, we could add a wiki folder we can copy paste)
- moved the preference migration under the gameplay settings as it caused crashes when creating accounts (because gameplaysettings was undefined)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "clubwiki"]
+	path = BondageClub/Wiki
+	url = https://github.com/Ben987/Bondage-College.wiki.git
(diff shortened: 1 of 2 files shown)


ace-1331:
Note that it's possible you might have to add the .gitmodules file to your working tree when switching branches if you get the mapping error.

There currently is no .gitmodules files which was problematic

elliesec (approved):
Ah, good call - didn't think about that!

PR #1432, opened 2020-09-14
SpamFilter

NoneNoname:
Add a simple way to filter out fast and consecutive action like kneel/stand or enter/leave.

The current UI image could be improved so as the wording for it


Nina-1474:
I'm getting this error:

Preference.js:1191 Uncaught ReferenceError: R is not defined
    at Object.CheckAction (Preference.js:1191)
    at Object.Check (Preference.js:1179)
    at ChatRoomMessage (ChatRoom.js:850)

Nina-1474:
The preference descriptions are a little unclear to me at the moment. Mainly the terms 'block' and 'test' and a general indication of when these rules apply. How about something along these lines:
"
When a character spams messages in chat:
◻ Ghost the character
◻ Block the character
◻ Kick the character from your room
◻ Ban the character from your room
◻ Exclude friends from the check
◻ Exclude whitelisted characters from the check
"

NoneNoname:
Thanks,

Should I allow a way to disable the hole filter, even the soft block (It has low false positive with this current setup)?

And I did extend the /help but it may really be unclear how to actually use it (/unblock), should, and how could we add a way for the user to get information about the usage


I have an [ipementation](https://github.com/NoneNoname/ace12401-cursedscript/blob/fa065e7bd664befea3d544973f54cd0aa46f33d1/Curse/Room/CommandsV2.js) for a better help which maybe be useful for this.

Nina-1474:
Some feedback & thoughts on this (numbered for ease of response):
1. There are ServerSend calls misspelling the type as 'AccoountUpdate'.
2. I'm seeing just the word "SpamFilterBlockedAction" appearing in chat instead of the intended message.
3. I'd say there's no harm to letting players disable the whole thing if they want. Although does the "soft block" actually do anything if the other settings are off?
4. If a player triggers the filter, they can continue causing problems depending on the settings. For example if they're kicked from the room, they can re-enter and spam again without triggering the filter and getting kicked anymore, since they're now on the blocked list.
5. Could you add that line in preferences above the checkboxes saying "When a character spams messages in chat:" as well? It will help explain what the settings are for.
6. Maybe you could add another line below the checkboxes on the preferences screen saying "Use '/unblock MemberNumber' to unblock a player who triggered the spam filter" to help explain it?
7. I have a change in progress for the preferences screens to give more space for new menus. I could make the Spam Filter page separate instead of it being inside the Online page if that's ok with you? I could also replace the icon, as we should avoid putting English words in images.
8. When trolls can't spam kneel anymore, they'll probably just spam text instead. Should we check for rapid chat messages as well? We'd need to be much more careful of false positives there.
9. This is a bit of an aside, but do we even need the kneel button anymore now that we have it in the active pose menu?

NoneNoname:
1, 2, 5, 6 - fixed/added

3, 4 - soft block removes the action messages like kneel stand enter leave, so you no longer see even if further spam happens, yes kick may allow them to continue

7 - sure go ahead

8 - the filter is made so it can be extended to any other type of filtering, like a chat message filter triggered by in a similar way or a word filter

9 - this may be outside of the scope of this pr, but it may worth considering, as well as adding the rest of the pose change messages if they are easily spammable

ace-1331:
I purposefully avoided making messages with the pose menu as they'd be even easier to spam (we also do not have chat message with expressions so it is not entirely different than current behavior), but I agree with the points above that normal messages should be filterable too.

The kneel button is a quick way to reset your pose/kneel and provide a message for it, but the way the menu works is enough on its own, so yes, technically we could remove the kneel button. 

I wont have a lot of time till later this weekend, but I'll do some more testing with this as soon as I can to see if I have more comments. As long as its optional + you get a visual cue when someone is auto-ghosted, I think itll be fine

Nina-1474 (approved):
I think the Enable checkbox would be better above the "When a character" line with something a little more explicit like "Enable the spam filter". Since that line is explaining the conditions for the _other_ checkboxes and not the Enable one, so the Enable checkbox following it immediately is a bit confusing.
Otherwise this all seems to be working fine.

Nina-1474:
I've merged in the master and integrated this PR as a standalone page in the new preferences menu structure from #1599.
I swapped the icon out for one that doesn't have English text.
I also applied the line re-ordering suggestion from my comment above. Hope that's ok!

PR #1435, opened 2020-09-15
AssetTypeInfo V-Final

NoneNoname:
Turns ExtendedItems to unified single object settings like #603 and #380

`WIP` --- `WIP` --- `WIP`


elliesec:
Please do be aware that the larger the scope of this pull request gets, the longer it's going to take to fully review and test. I like the premise and where this is going, but do try to keep the scope focused - multiple small PRs with limited, targeted scopes are much easier to review and test (even if there is no immediate end-user benefit to a PR - that's fine).

NoneNoname:
Well, there is an option to not include as many items in the first pr, that would reduce the amount of test necessary for this pr, maybe even down to one single item, if that helps, for example to all assets in ItemMouth.

I also didn't want to make a partial ExtendedItem script that would have seemed just like a copy of what we already have, I also left out any file removing changes (all Inventory<Group><Asset><Load|Draw|...> can be removed)

elliesec:
It's fine as is right now, I just wanted to point that out before it becomes to unwieldy to review effectively - something to bear in mind. Smaller, incremental changes are always easier to review than everything being shoved into a single monolithic PR. Even if there's no immediate tangible benefit, as long as the _intent_ can be understood and is sound, that's fine.

NoneNoname:
I added all items from the ItemMouth, so now there is no more extended item in that group, as this gets accepted I can start converting more items, that should mostly be code reduction from this point only a few functionalities left unimplemented, I leave them until I actually add items that need them

Yet to be implemented:
- Vibration control
- Lock exit to AssetTypeInfo screen
- Expression
- Full Items, ex: Spanking Toys
- Owner/Lover only item support
- Server load check for TypeLocking
- Edit Dialog_Player.csv to remove moved lines
- Move pictures from Inventory/ to Assets/.../Preview/Type
- Translation fixes and finishes, tool to move old ones to the new place
- Better format for Dynamic functions in Tools_AssetTypeSaveJs ???

Nina-1474 (changes requested):
When I apply a gag and select the first option, nothing happens. When I select an option after the first one in chatroom, a "Cannot read property 'FocusGroup'" error occurs.

PR #1436, opened 2020-09-15
Refactor: ClitSuctionCup to extended item

SandrineBC:
# Summary
Refactors the clit suction cup to use the extended item script for better maintainability.
As the cup does not use images for the different suction levels, the extended item script was extended for this special case.

# Reviewer
Perhaps it's a good idea, if the review would be done by @Nina-1474 and @elliesec because both of them worked much on the extended item script.


elliesec (approved):
Looks fine to me. Tested and working.

Nina-1474 (approved):
This looks good!
When archetypes come around we could possibly take another look at this, since at the moment it's limited to 8 options per page even though there's plenty of space for more. And since the button arrangement is more regular/predictable than the image ones, it might be better to use a formula rather than a big predefined array. Think of how many options are in the collar tags for example!

ace-1331 (approved):

PR #1443, opened 2020-09-16
Improvement: Remove loop variable modification

SandrineBC:
# Summary
The code uses the following pattern at several places:
```JavaScript
for (A = 0; A < array.length; A++)
   if (something) {
      doSomething();
      array.splice(A,1);
      A--;
   }
```
As it is a bad idea in general to modify a loop variable within the loop, this PR changes this and transfers the above mentioned pattern to:
```JavaScript
for (A = array.length - 1; A >= 0; A--)
   if (something) {
      doSomething();
      array.splice(A,1);
   }
```

### DIFF
--- a/BondageClub/Screens/Character/Appearance/Appearance.js
+++ b/BondageClub/Screens/Character/Appearance/Appearance.js
@@ -47,20 +47,18 @@ function CharacterAppearanceValidate(C) {
 
 	// Remove any appearance item that's not in inventory
 	var Refresh = false;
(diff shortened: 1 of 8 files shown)


elliesec (approved):
I've not tested every instance, but they all look fine. I think ideally it'd be cleaner and easier to read if we converted these to use `Array.prototype.filter`, but looping in reverse is definitely safer than the current method.

ace-1331 (approved):

PR #1392, opened 2020-09-10
Feature: Multi-colouring asset migration

elliesec:
---
**Important**: This draft PR depends on #1368 and should be merged after it. Additional assets may be added to this PR while it is still a work-in-progress.

---

## Summary

This PR converts a lot of items in the game to enable their layers to be coloured separately. It also splits several assets out into multiple layers to enable them to support multi-colouring. Several items have also been refactored to remove parent groups where they were unnecessary.

Huge huge thanks to @Ayesha-678 and @Nina-1474 for all the help converting assets.

A (roughly up-to-date) list of all of the assets that will support multi-colouring after this change can be found [here](https://gist.github.com/elliesec/76eabcb4c79f937a7ca182a35f4394b9).

PR #1444, opened 2020-09-16
more cow printed clothing

gatetrek:
adds printed socks, gloves, tail
added some horns, tail plug


Nina-1474:
CowHorns has the Fetish tag twice.
And a heads up: some of the tail strap image changes are going to conflict with the split into layers under #1392.

elliesec (approved):
Looks good! There are some notable clipping issues with the gloves in the `OverTheHead` pose, although I imagine that's because they're an edit of the long gloves, which have the same issues anyway.

gatetrek:
yeah it was but managed to remove most the clipping so both glove2 and the printed ones should be better now

elliesec:
Nice, looks good!

ace-1331 (approved):

PR #1445, opened 2020-09-17
New Item: Studded Harness

Ruilov3:
new harness in both Bra and lockable variants


elliesec (approved):
Looks lovely! Not sure if you wanted to make the metal/studs colourable separately now that multi-colouring is in, or if you were intending on coming back to this later. If you want to leave it as is now though, this all looks good to go in.

Ruilov3:
I tried colouring the metal but it ended up looking flat outside of very few specific colour options and considering some parts of it were 1-2px in size i don't think i'm up to splitting the metal parts into base and shine layers and trying to align those afterwards.

we keep the metal as it is for now

ace-1331 (approved):

PR #1446, opened 2020-09-17
New handheld: Lotion bottle

Ayesha-678:
another handheld, a lotion bottle.


elliesec (approved):
I made a very small fix to the asset name in `Female3DCG.csv`, but other than that, looks good to me!

ace-1331 (approved):

PR #1447, opened 2020-09-17
Improvement: Multi Colour Button

Nina-1474:
This replaces the standard colour picker button image with a "multicolour" version whenever the item to be coloured supports multi-colouring.


BondageClub/Screens/Character/ItemColor/ItemColor.js
```diff
@@ -526,6 +526,27 @@ function ItemColorStateBuild(c, item, x, y, width, height) {
 	};
 }
 
+/**
+ * Returns layers of the asset which can be given distinct colors
+ * @param {Item} item - The item to be colored
+ * @returns {array} - The colourable layers
```
elliesec, line 532: Could be a bit more specific about the return type here - we have a `Layer` type definition in `Asset.js`, so this could be `Layer[]`.

elliesec (approved):
Nice idea! Just the one comment, but otherwise looks good!

Ayesha-678 (approved):

PR #1449, opened 2020-09-19
Refactor: Bondage bench to extended items

SandrineBC:
# Summary
The PR refactors the bondage bench to use the extended items script. This has the following implications
* Getting rid of `BondageBenchStraps.js`
* Using the new feature of layer priorities for the bench and the straps
* Introducing the new feature to allow the usage of locks on type level, including all checks in the dialog screen and the receivers end. If you want the whole item to be lockable, just add `AllowLock` on the asset level, as usual. Else add `AllowLock` within the properties of the type. 
* The receiving end was modified to accept those locks as well

# Points to discuss
As the prerequisite for a certain type is not defined in the type's properties, we could think about moving the `AllowLock` attribute from the properties to the main entry as well...


Nina-1474 (approved):
Looks good. One possible suggestion: the "None" option's image used to represent the addition of straps but now it represents the removal of them, so the straps in the image are a bit misdirecting. What about using an edited version of the item's main preview image without straps instead?

BondageClub/Assets/Female3DCG/Female3DCG.js
```diff
@@ -2921,14 +2933,7 @@ var AssetFemale3DCG = [
 		Top: -250,
 		Zone: [[400, 0, 90, 100]],
 		Asset: [
-			{ Name: "BondageBenchStraps", Value: -1, Difficulty: 12, SelfBondage: 5, Time: 5, IsRestraint: true, AllowLock: true, Hide: ["HairBack", "Wings", "TailStraps", "ItemButt"], SetPose: ["LegsClosed"], Effect: ["Block", "Prone"], AllowType: ["Light", "Normal", "Heavy", "Full"], Block: ["ItemDevices"], SelfUnlock: false, Extended: true, RemoveAtLogin: true },
-			{
-				Name: "Covers", Value: -1, Difficulty: 1, SelfBondage: 0, Prerequisite: "OnBed", BuyGroup: "Bed",
```
ace-1331, line 2926: Im not sure if its with the recent merge, but Im not sure if the cover layers being changed here is intended?

ace-1331 (dismissed):
Seems to be working as intended and its a good option to have until we figure out layer locks.

It was intended for the crib to be only lockable when closed so it could be possible now.

Just double check my 1 comment when you can

BondageClub/Assets/Female3DCG/Female3DCG.js
SandrineBC: Perhaps a merging problem and I changed that back.

Nina-1474:
Since the preview images were originally 200 pixel squares rather than the typical 221 pixels, they weren't quite centred. I've adjusted them to correct this, along with making the None image suggestion from above.

Nina-1474 (changes requested):
Since there are no colourable layers, clicking the colour button hits an error. Both the bench and straps were colourable before so I think they should remain so.

SandrineBC:
I added the colorization of the layers. It seems, I got your images included, @Nina-1474 , but you better double check.
When I tested, the colorization seemed a bit unstable, Sometimes the picked color was not shown in the icon, sometimes I was not able to colorize the item as a whole. Perhaps there's still a glitch hidden somewhere...

Nina-1474 (approved):

elliesec (approved):

PR #1450, opened 2020-09-19
Fix when given mittens in the nursery

wultir:
A small bug when joining the nursery after the first time while already wearing a diaper.

The following dialog path doesn't actually cause you to wear mittens due to attempting to equip it on arms instead of hands

* I'm ready to join the nursery again
* I am already diapered and ready to enter
* Can I have some restraints please?
* I've been a good girl honest. My hands are cold.


elliesec (approved):
Good catch! Tested and working.

ace-1331 (approved):

PR #1451, opened 2020-09-19
Fix tray check when talking to cafe maid

wultir:
Clicking the maid in the cafe while not having anything equipped in the ItemMisc slot causes an error and no dialog.

Steps to reproduce:
* Start maid online drink job
* Remove the tray
* Talk to the maid in the cafe


elliesec (approved):
Thanks - looks good! Tested and working.

ace-1331 (approved):

PR #1453, opened 2020-09-21
Refactor: Nose ring to use extended items

SandrineBC:
# Summary
This PR refactors the nose ring to use the extended item script.

### DIFF
--- a/BondageClub/Assets/Female3DCG/Female3DCG.js
+++ b/BondageClub/Assets/Female3DCG/Female3DCG.js
@@ -2667,7 +2667,7 @@ var AssetFemale3DCG = [
 					{ Name: "Hook" },
 				]
 			},
-			{ Name: "NoseRing", Priority: 43, Fetish: ["Metal"], Value: 25, Difficulty: 50, Time: 15, Random: false, AllowLock: true, Left: 50, AllowEffect: ["Tethered", "Freeze", "ForceKneel", "IsChained"], AllowType: ["Base", "ChainShort", "ChainLong", "Leash"], AllowPose: ["Kneel"], Extended: true },
+			{ Name: "NoseRing", Priority: 43, Fetish: ["Metal"], Value: 25, Difficulty: 1, Time: 15, Random: false, AllowLock: true, Left: 50, AllowEffect: ["Tethered", "Freeze", "ForceKneel", "IsChained"], AllowType: ["Base", "ChainShort", "ChainLong", "Leash"], AllowPose: ["Kneel"], Extended: true },
 			{ Name: "DuctTape", Fetish: ["Tape"], Value: 50, BuyGroup: "DuctTape", Audio: "DuctTape", Difficulty: 2, Time: 10, RemoveTime: 5},
 			{ Name: "NosePlugs", Value: 20, Difficulty: 3, Time: 5, RemoveTime: 5},
 			AssetSpankingToys
--- a/BondageClub/Assets/Female3DCG/Female3DCG.js
+++ b/BondageClub/Assets/Female3DCG/Female3DCG.js
@@ -2667,7 +2667,7 @@ var AssetFemale3DCG = [
 					{ Name: "Hook" },
 				]
 			},
-			{ Name: "NoseRing", Priority: 43, Fetish: ["Metal"], Value: 25, Difficulty: 1, Time: 15, Random: false, AllowLock: true, Left: 50, AllowEffect: ["Tethered", "Freeze", "ForceKneel", "IsChained"], AllowType: ["Base", "ChainShort", "ChainLong", "Leash"], AllowPose: ["Kneel"], Extended: true },
+			{ Name: "NoseRing", Priority: 43, Fetish: ["Metal"], Value: 25, Difficulty: 10, Time: 15, Random: false, AllowLock: true, Left: 50, AllowEffect: ["Tethered", "Freeze", "ForceKneel", "IsChained"], AllowType: ["Base", "ChainShort", "ChainLong", "Leash"], AllowPose: ["Kneel"], Extended: true },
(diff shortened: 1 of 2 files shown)


BondageClub/Screens/Inventory/ItemNose/NoseRing/NoseRing.js
```diff
@@ -1,116 +1,135 @@
...
-	}
 
-}
\ No newline at end of file
+/**
+ * Validates, if the chosen option is possible. Sets the global variable 'DialogExtendedMessage' to the appropriate error message, if not.
+ * @param {Option} Option - The next option to use on the character
+ * @returns {boolean} - Returns false and sets DialogExtendedMessage, if the chosen option is not possible.
+ */
+function InventoryItemNoseNoseRingValidate(Option) {
```
elliesec, line 97: Can this not be done by just setting `Prerequisite`s on each of the options and not having a `Validate` function, or is there something in here that can't be done with prerequisites?

BondageClub/Assets/Female3DCG/Female3DCG.js
```diff
@@ -2667,7 +2667,7 @@ var AssetFemale3DCG = [
 					{ Name: "Hook" },
 				]
 			},
-			{ Name: "NoseRing", Priority: 43, Fetish: ["Metal"], Value: 25, Difficulty: 50, Time: 15, Random: false, AllowLock: true, Left: 50, AllowEffect: ["Tethered", "Freeze", "ForceKneel", "IsChained"], AllowType: ["Base", "ChainShort", "ChainLong", "Leash"], AllowPose: ["Kneel"], Extended: true },
+			{ Name: "NoseRing", Priority: 43, Fetish: ["Metal"], Value: 25, Difficulty: 1, Time: 15, Random: false, AllowLock: true, Left: 50, AllowEffect: ["Tethered", "Freeze", "ForceKneel", "IsChained"], AllowType: ["Base", "ChainShort", "ChainLong", "Leash"], AllowPose: ["Kneel"], Extended: true },
```
elliesec, line 2670: Is there a strong reason for reducing the difficulty here? I personally feel like this should probably be a reasonably high difficulty item, because the idea of struggling out of one without hands seems kind of... ouch... to me.

SandrineBC: Perhaps I'm wrong here, but a difficulty of 50 is neigh impossible to remove (like the ring is welded shut).
Why do you think that the hands could not be used? All other none chastity piercings are like 10 (which is still to high for me, because removing a piercing is not soo difficult, but perhaps we can go for 10, then?

BondageClub/Screens/Inventory/ItemNose/NoseRing/NoseRing.js
SandrineBC: There currently is no prerequisite to check the character to not be suspended, but allow a hogtied suspension. That is used for the application of the long chain or the leash.

BondageClub/Assets/Female3DCG/Female3DCG.js
elliesec: My point was more that difficulty only really has any meaningful effect when either the hands are not free, or the ring is locked. If neither of those apply, even a difficulty 50 item is trivial to remove. I've no idea how you actually lock a nose ring in reality, but I'd imagine in the hands tied case, it would be pretty difficult to remove. Granted 50 seems arbitrarily high, but 1 would be trivial to remove for most people, even if locked/hands tied.

BondageClub/Screens/Inventory/ItemNose/NoseRing/NoseRing.js
elliesec: Ah, I see. Could most of these blocks be simplified using `InventoryAllow` in that case? For example:
```js
case "ChainShort":
    if (!InventoryAllow(C, ["NotSuspended", "StraitDressOpen"], true)) {
        DialogExtendedMessage = DialogText;
        ChainShortPrerequisites = false;
    }
```

BondageClub/Assets/Female3DCG/Female3DCG.js
elliesec: In any case, 10 seems reasonable enough to me.

SandrineBC:
I reworked (and condensed) the validate function

BondageClub/Screens/Inventory/ItemNose/NoseRing/NoseRing.js
```diff
@@ -1,116 +1,118 @@
...
+	switch (Option.Name) {
+		case "Base":
+			break;
+		case "ChainShort":
+			if (!InventoryAllow(C, ["NotSuspended", "StraitDressOpen", "NotMounted"], true)){
+				DialogExtendedMessage = DialogText;
+				ChainShortPrerequisites = false;
+			} // if
+			break;
+		case "ChainLong":
```
Nina-1474, line 109: ChainLong is missing prereqs now!

Nina-1474 (dismissed):

BondageClub/Screens/Inventory/ItemNose/NoseRing/NoseRing.js
SandrineBC: Shouldn't the missing break give "ChainLong" the same prereqs as "Leash"?

Nina-1474: I see! My mistake.

SandrineBC: Oh, given my JScript skills, it's better to double check...

ace-1331 (approved):

elliesec (approved):

PR #1454, opened 2020-09-21
Halloween assets

aldarris:
Deploy to prod on or before October 26th


ace-1331 (dismissed):
fixed a typo, otherwise it seems to be working well

Nina-1474 (changes requested):
Mostly looks good. A few remarks though:
The Pumpkin Hood's colour button has text like ItemHoodPumpkinHeadFront. It needs entries in LayerNames.csv.
Things like hair & accessories can poke out of the Hood. Using advanced alphas could help here.
The Hood's priority is putting it in front of Emoticons.
The monsters & familiars aren't visible when Hogtied and in AllFours because they're low in those images.
Is there a reason the Familiar set is so much more expensive than the other two sets?

aldarris:
1.  Could not figure out the layer names mapping yesterday for the coffin -- and help in discord did not help.
2.  There is one mask that removes everything above the hood, such as hair and accessories.  Clearing the sides would look force than hair peeking on the sides.
3. Changed the priority to 56 (same as cages) (the number 70 came from copy and paste of other hoods).
4. Fixed.  The crawler-cuffs are sill in front of the monsters, but they are in front of the teddy bear as well, as we don't have pose-level based priority.
5. Live pets cost orders of magnitude more than fluffy toys.  Familiars are not expensive -- they are cheap, it is Halloween sale.

elliesec:
The layer name mapping for the pumpkin hood should be straightforward. You need the following entries in `LayerNames.csv`:
```
ItemHoodPumpkinHeadFront,Front
ItemHoodPumpkinHeadBack,Back
```
For the coffin (if you want to include it), look at something like the `Cage` asset in `ItemDevices` to see how both colour groups and layer names are handled.

aldarris:
Ok, it worked (for the pumpkin).
Also, the rectangular alfa masks can't cut off the hair naturally, so hiding the HairBack entirely

ace-1331 (commented):
Retracted my review with the added item, I'll test the coffin soon and re-approve after conflicts are merged if all is ok.

ace-1331 (commented):
All seems to be working so I have nothing to add on top, will double check with merges but it should be good.

Nina-1474 (commented):
Happy that the issues have been resolved. The coffin looks good!
The one potential issue is that the lid can still be opened and closed after locking it. Is that intended? If not, the Validate function might need to be changed to something more like Chains for ItemLegs.

Nina-1474 (approved):

aldarris:
The coffin does not allow variant change if lock is applied,  It is on line 43 in the Validate function, copied from some other item validation .
	if (DialogFocusItem.Property.LockedBy && !DialogCanUnlock(C, DialogFocusItem)) {
		DialogExtendedMessage = DialogFind(Player, "CantChangeWhileLocked");

ace-1331:
That line means you can still change the type if you can unlock the lock.
If you want a behavior where the type can't be changed at all unless the lock is removed (which I think does make sense), you need something like the chains ->
```
if (InventoryItemHasEffect(DialogFocusItem, "Lock", true)) {
	DialogExtendedMessage = DialogFind(Player, "CantChangeWhileLocked");
	return false;
}
```

aldarris:
Yes, the belts can't be changed while locked, the gags can be changed while locked.
In the end, I went with the latter behaviour, as there are boxes out there that one shut and auto lock.

ace-1331 (approved):

ace-1331:
Just a heads up, even if the PR was approved, we cannot merge due to conflicts. Since the beta is by Saturday, we have till Friday to merge new stuff.

aldarris:
Conflict in layer names -- trivial to resolve.
Conflict at the wooden horse line -- the leggings and monsters were added to the hideItem array, trivial to resolve.
Conflict at the teddy bear line -- looks like there were poses removed from AllowPose?  Should I keep all the poses, or follow what is in master branch?

Nina-1474:
The Bolero and StraitDressOpen poses have been removed entirely so yes you can remove them for your asset too.

aldarris:
All done.

PR #1457, opened 2020-09-22
Change: SelfUnlock

ace-1331:
SelfUnlock wasn't actually preventing you from unlocking your own item if you were not blocked from interacting. Currently, all selfunlock: false item seem to block interactions which hid this problem.


ace-1331:
I've added a new `InventoryGetItemProperty` method to return the value of an item. This will allow us to refactor and squash a lot of awkwardness related to null checks, Property checks, etc.

elliesec (approved):

Nina-1474 (approved):

PR #1462, opened 2020-09-22
Enchanted help for commands

NoneNoname:
Has conflict with #1459, I resolve it as soon as one is merged


BondageClub/Screens/Online/ChatRoom/Commands.js
```diff
@@ -0,0 +1,328 @@
...
+        Action: () => {
+            const Heads = (Math.random() >= 0.5);
+            const Dictionary = [
+                { Tag: "SourceCharacter", Text: Player.Name },
+                { Tag: "CoinResult", TextToLookUp: Heads ? "Heads" : "Tails" }];
+            ServerSend("ChatRoomChat", { Content: "ActionCoin", Type: "Action", Dictionary: Dictionary });
+        }
+    },
+    {
+        Tag: 'safeword',
```
ace-1331, line 184: The safeword command does not seem to work as intended (Does not kick you out of the room as it should, and it doesn't ask between the two, so I'd say it might be safer to leave it up to the dialog menu for now)

BondageClub/Screens/Online/ChatRoom/Commands.js
```diff
@@ -0,0 +1,328 @@
...
+        Description: ' [Message], send emote: "<i>*Message*</i>", alternative start message with **',
+        Action: (_, msg) => ChatRoomSendEmote(msg)
+    },
+    {
+        Tag: 'invalid',
+        Description: ', do nothing',
+        Action: () => { }
+    },
+    {
+        Tag: "CommandsKey",
```
ace-1331, line 295: Is this command working as intended? It does not seem to work (I think it doesnt work on the curse either, but that might be me not understanding it)

ace-1331 (changes requested):
Just double check the 2 comments, other than that, everything works like it should

ace-1331 (approved):

Nina-1474:
Overall it does look good, but I do have some findings and thoughts:

### Errors

- “/action Message” is including the space at the start of the message.
- "/beep" doesn't seem to work. I don't see what "T < ChatRoomCharacter.length" is trying to do given T is the MemberNumber.
- "/showfriends" says "Blacklist" in the description.

### Descriptions
The initial help message doesn’t make it clear that “/help” lists everything. A suggestion for the text:
`(Type “/help” for a list of chat commands or “/help [Keyword]” to get help for a specific command)`

I think the four "view" commands should include 'list' in the keyword (e.g. "/showblacklist" instead of "/showblack") as that's how they're referred to everywhere else and it's more consistent with the other commands.

Some of the chat descriptions I personally find a little unclear. And in particular the dice description isn’t fully correct. Below is a suggestion of what I would have: 
```
Help: 
/action [Message]: Display the message as an action, without including your name. **[Message] does the same
/afk: Set AFK emote
/ban [MemberNumber]: Ban the player from the room
/beep [MemberNumber]: Send a beep to the player
/blacklistadd [MemberNumber]: Add the player to your blacklist
/blacklistremove [MemberNumber]: Remove the player from your blacklist
/coin: Toss a coin
/demote [MemberNumber]: Demote the player from room administrator
/dice [Number]: Roll a 6-sided die. Adding “[Number]” after the command for any number between 2 and 100 will roll a die of that many sides instead. Adding “[Number]d[Number]” will roll multiple dice and sum the total.
/friendlistadd [MemberNumber]: Add the player to your friendlist
/friendlistremove [MemberNumber]: Remove the player from your friendlist
/ghostadd [MemberNumber]: Add the player to your ghostlist
/ghostremove [MemberNumber]: Remove the player from your ghostlist
/help [Keyword]: Display help for the command with prefix Keyword. Exclude [Keyword] to view all valid commands
/kick [MemberNumber]: Kick the player out of the room
/me [Message]: Display “[PlayerName] Message” as an action. *[Message] does the same
/promote [MemberNumber]: Promote the user to room administrator
/showblacklist: View your blacklist
/showfriendlist: View you friendlist
/showghostlist: View your ghostlist
/showwhitelist: View your whitelist
/unban [MemberNumber]: Remove a ban on a player
/whitelistadd [MemberNumber]: Add a player to your whitelist
/whitelistremove [MemberNumber]: Remove a player from your whitelist
```

### Translations
Does this support translations? I see there's a link to a Commands.csv but the file doesn't exist and all the descriptions & messages are hardcoded in English. Also not using the standard Text methods might make this harder to change (and not forget) if/when we move to a better translation system.

### Invalid Commands
Is there a purpose to having the "/invalid" command?

Returning the ‘invalid’ message when the command doesn’t exist means that people running script rooms can’t define and use their own custom commands as easily. I don’t see the benefit of this change, other than hiding a mistyped command attempt. Allowing the text to just post to chat normally would mean the current double forward-slash “//” behaviour could be maintained as well.

NoneNoname:
Yes translation should be improved, this was made to be a snap-in replacement for personal use originally, so that's why I didn't rewrite every part, and sadly I was never great at those descriptions

generally, longer commands are more hassle to type as I would just use /ghost /friend but I also do not want to change what we already have. so I guess I add the list to them (`/ghostadd` currently do not have it either), on another note maybe I should add Aliases for short ones

for the last part: 
- adding custom commands never been easier after this change just call `CommandCombine(YourCustomCommands)` and you can use them afterward
- /invalid is triggered when the command is invalid or the prerequisites didn't meet and in this case you probably not want to send it out as a public chat message like `/ghostaddd 123` would make it public that you wanted to ghost someone, I never liked this current behavior personally

(actually, this /invalid Message need to be translated I shell fixed that when doing the rest)

(since we are in beta, I will get this ready somewhere after the release)

Nina-1474:
Cool if aliases are possible then I'd suggest the longer keywords like "/showghostlist" as well as the shortened versions to cover all keywords that users are likely to try.
For custom commands I wasn't thinking about the person running the script, but other users sending commands to them. Maybe using backslash in keywords is not the usual custom in such cases? But yeah if we should be hiding mistaken commands that's fine. In that case, I think '[Message], send "/Message"' also needs to be reworded as it took me a long time to figure out what it meant!
Returning a message for non-existant commands and the "/invalid" keyword itself were two separate points. Does the "/invalid" command that returns "do nothing" need to exist?

NoneNoname:
I didn't think about custom rooms but yeah probably a good idea to use different keywords for that, though changing the / key is somewhat possible but I have yet to test it

I fixed up the translation so it works normally now (I didn't add the language txts)
I have yet to add aliases but I added autocomplete that should help too, maybe I add the aliases in a later pr

/invalid clears the chat that's all it did and that's why it is necessary, maybe a `ChatRoomSendLocal` with the new timeout would work better

Nina-1474:
Looks pretty good. Just two last things I think could be improved:

_(Use : /help [keyword] to get help for commands)_
This should also make it clear that just "/help" will list all commands.

_// [Message]: Send message starting with /_
I found this line very confusing. Being the first line doesn't help matters! I can't think of anything much better though.

kimei-nishimura:
Hi @NoneNoname ! Would you be willing to bring this PR up to date with the current club master, and see what's left for it to be usable? You've added some features that could be really beneficial to script and extension makers, and I'd like to see it brought to completion.  Thanks!

NoneNoname:
@kimei-nishimura I don't intend to work on this until I am sure it would get reviewed and merged

PR #1465, opened 2020-09-22
Feature: Extended item permissions

ace-1331:
- added permission mode to extended items
- refactored the permission toggle setting into a function to regroup the piece of code which is now used in dialog, appearance and in the extended item script
- fixed a bug where toggling perms while focused on another character would mess up your limited items (this was hidden since you couldn't go in permission mode while focused on others before)

Hoping this will help regroup several items like the petsuits, etc.

Some points to discuss about this ->

1- this enables blocking stuff like the min/normal/max type of extended items like suction cups, straitjackets, etc. I personally don't mind since it adds depth to permissions, but it would be possible to add a new property to disable perm mode on certain items or types. What do you think? (It's frequent to see people who list vibes on max as being a dislike, for example, but overall it doesn't hurt to keep it)

2- I did not do any checks like "if base type is blocked, remove all other blocked types since the item is completely blocked" simply because this made it so toggling through the menu would reset your type-based permissions. I think the current behavior is more intuitive.

3- The mode is always toggleable, even while inspecting others. This is because most items would not allow you to interact with them once you've used them on yourself (After the initial time you've put it on) therefore, it might be easier to just toggle things the way you want them when you see someone else wearing an item. However, this leads me to say -> What if we always allowed viewing the extended menu, but disabled it when you can't interact? (Then implement AlwaysExtend/AlwaysInteract within the script itself) Of course, that would only be possible once everything is done being migrated as we don't have a way to check for if an item is using the extended script (Would've been useful to have something like Extended: "V1", "V2" instead of true/false)


Nina-1474:
When we exit the inventory permissions mode, ChatRoomCharacterUpdate is called which ensures everyone in the room gets the new permissions immediately. This doesn't occur for the type permissions. Should it?

elliesec (approved):
Looks good to me, and seems to be working well. Regarding your points:
1. I think that's fine - it might not make the most sense on some items, but in my opinion, it's better to give people the flexibility.
2. I think that's fine too
3. I'm inclined to say that eventually we might want to move the functionality into its own menu somewhere (not exactly sure where yet). With #1435 on the horizon, that would give us enough unification of extended items to be able to have a single place from which we could manage permissions for all types rather than having to use the item or see it on another player.

ace-1331:
Good point, I just added the sync.

And yes, for #3, it can be in its own menu at some point. It'll potentially be easier to handle that way once everything is migrated

Nina-1474:
Could the sync be put in for the Exit button too (if in permissions mode)?
The permissions also won't sync if exiting the screen by picking a new focus group on the character, though it seems this desync issue actually already occurs for the inventory permissions screen too.

ace-1331:
I looked everywhere the mode is reset, this should help prevent desyncs in general

Nina-1474:
I still see desync potential but rather than let this PR keep expanding, and since it covers most cases, I'm happy to merge this if you are.

Nina-1474 (approved):
Works well! The one outstanding sync case I'll either take a look at or raise an issue for.

PR #1468, opened 2020-09-23
Login Random Character Fix

Nina-1474:
The recent change to reduce CharacterRefresh calls passed in False to the ClothOnly parameter of the Full Random Appearance function. This revealed a long-standing bug where True and False were both being treated as True, and only undefined was treated as False.
This should fix the appearance of both the Login NPCs and the Shopkeeper after doing a job, to randomise the body as well as the clothes.


elliesec (approved):

Ben987:
Yes, that one was annoying, thanks a lot.

PR #1472, opened 2020-09-23
Halloween assets -- Coffin

aldarris:
The layer names mapping worked after another merge with the main.


aldarris:
Actually, it no longer has dependencies -- the two fixes have been merged yesterday.
Would it be more optimal to cancel this PR and put it with the other halloween assets?

elliesec:
I added the label because this is built on top of the other halloween assets PR, so needs to be merged after that one. If you want to close this and include it in the other PR, that's fine too.

aldarris:
Combining two PRs into one

PR #1475, opened 2020-09-24
Seamless suits expose same slots as the zipper versions

Shiranui-Izayoi:
Made it so the seamless versions of the normal and see-through catsuits now expose the same zones as their zipper having equivelents. 

While I am aware it isn't "realistic" for them to allow access to these slots. In this case I think it is justified in having it. Mainly since all them not exposing these zones does is make it annoying to add or remove items from people wearing them, And since clothing can't be locked it doesn't prevent people getting access simply requires more time and messages. 

Which also has the side effect of either potentially ruining peoples outfits, since some people don't take note of what they are changing.

I also changed the Fetish of the suits from Nylon to Latex. It might just be me and if that is the case feel free to change it back, but the suits do not look like any kind of nylon, they look much more like latex. As shown by how closely they can match the Seamless Hobble Dress and Leg Binder along with the original Strait Leotard. Which are all Latex.

### DIFF
--- a/BondageClub/Assets/Female3DCG/Female3DCG.js
+++ b/BondageClub/Assets/Female3DCG/Female3DCG.js
@@ -197,22 +197,22 @@ var AssetFemale3DCG = [
 		AllowPose: ["TapedHands", "BackBoxTie", "BackCuffs", "BackElbowTouch", "Bolero", "Yoked", "Hogtied", "OverTheHead"],
 		Asset: [
 			{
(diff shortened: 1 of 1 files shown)


ace-1331 (commented):
@gatetrek I requested a review from you to see if its ok with you. I personally dont mind, but I was also wondering if you already planned to add unzip vs zip version.

If you're ok with it, then Ill approve too and we can merge already since its a simple change.

gatetrek (approved):
I do agree with fetish change, have been meaning to add an unzipped version for them but just haven't gotten to it yet.

The side effect of either potentially ruining people's outfits, not something I get as can say the same with other clothing that people can remove and not put back the way was before.
but if people want it changed that way so can access them till an unzipped one made I don't mind it

Shiranui-Izayoi:
Honestly it is mostly just for the sake of convenience. Since the seamless versions wouldn't get a unzipped version. Unless you are planning to do something crazy with it. Still nice to know that this is most likely going through.

ace-1331 (approved):

PR #1477, opened 2020-09-25
New item: wooden sign

ace-1331:
- added a custom sign
- added a few extra bits to increase what can be done with After/Before draw

It fits best in ItemMisc as it has similar properties than the wooden tray.
I can just revert the last commit if you prefer to have it in neck accessories.. however that makes it a lot less usable and slightly awkward with collars.

I changed the login fix to ignore this item.

I added a custom font folder and the font license that came with it, but the license can be removed. Now we can add custom fonts used with items in that folder (fonts are only loaded when first used)

Thanks to Gnarp for the assets!

PR #1478, opened 2020-09-25
Change: Custom tag to use multi color system

ace-1331:
- moved the text color to its own layer for simplicity

Since multicolor wasnt a thing when I made it, decided to improve it now rather than later.

requires #1477 for a small change to commondraw


Nina-1474 (changes requested):
The default gold colour is initially applied to both the tag and text, making it very hard to read. I'd say it would be better to have the text default to black.

ace-1331:
good point, just updated it

Nina-1474 (approved):

elliesec (approved):

PR #1497, opened 2020-09-26
Handheld Toys Refactor + Ruler item

Nina-1474:
Refactoring SpankingToys to use extended item scripts. The only notable difference to other extended items is the options screen will only show the owned items. All other irregular behaviour for the item luckily happened outside the extended screens functions.
Adding a ruler as a spanking item.


ace-1331 (changes requested):
Everything related to the refactor itself is working as expected, but there are 2 problems I wanted to point out which we should probably address first before merging

- Extended item pages are not working, when I reach the end, I have to go back but can't go forward + I have an empty page (I'm not sure whats the exact intended behavior, but the empty page is definitely a mistake)
- Refreshing the menu while in a different page of the extended screen resets the page number (it should keep the offset, try with the shock collar while testing)

Nina-1474:
The empty page should be fixed by #1463.
The last page not having a next button to cycle back to the first page was intended since it has a back button and quite few pages. Though since most other menus cycle (albeit out of necessity) this one could too. I'm going to raise another PR soon with more miscellaneous ExtendedItem tweaks so I can make that change there if we want it.

ace-1331 (approved):
Alright, #1498 #1463 will be dependencies for this one then, alls good with the actual refactor

And yeah, I would keep both the back and next button at all times to cycle through (And out of habit since its a common behavior)

elliesec (approved):

PR #1499, opened 2020-09-27
Fix multicolor picker crash

ace-1331:
- fixed a crash when the character is refreshed while in the multicolor picker (if the item is removed or changed, it crashes because the menu is not closed as expected)


elliesec (approved):
Good catch!

Nina-1474 (approved):
The Confirm/Cancel buttons are still left behind, but they don't seem to cause any issue so looks good.

PR #1503, opened 2020-09-28
Preference screen crash

ace-1331:
- fixed a crash in the preference screen when switching account too fast.

If you went to email, then changed screen faster then the query, you crashed with 
```
Server.js:615 Uncaught TypeError: Cannot set property 'placeholder' of null
    at ServerAccountQueryResult (Server.js:615)
```
because the input no longer exists. This prevents it


elliesec (approved):

Ben987:
Yes, I've seen that one recently.  Thanks for the fix!

PR #1505, opened 2020-09-29
Bunny Ears Multicolouring

Nina-1474:
Split the Floppy Bunny Ears into an outer and inner layer to allow different colours.


ace-1331 (approved):
One word:

YEEEESSSSS!

elliesec (approved):

PR #1507, opened 2020-09-29
Optimize Drawing.js

NoneNoname:


BondageClub/Scripts/Drawing.js
```diff
@@ -1010,18 +1039,52 @@ function DrawProcess() {
 			MainCanvas.height = MainCanvas.width / 2;
 			MainCanvas.canvas.style.width = "100%";
 			MainCanvas.canvas.style.height = "";
+			const MainCanvasRect = document.getElementById("MainCanvas").getBoundingClientRect();
```
elliesec, line 1042: Might be a little neater if you shifted this and the chunk on line 1058 out after the if/else, and just define a scale value (`2000 / W` or `1000 / H`) inside here.

BondageClub/Scripts/Drawing.js
```diff
@@ -66,6 +81,10 @@ function DrawLoad() {
 	MainCanvas.textAlign = "center";
 	MainCanvas.textBaseline = "middle";
 
+	DrawWindowResize();
+	setTimeout(DrawWindowResize, 1000);
```
elliesec, line 85: Why are the deferred calls to `DrawWindowResize` needed?

elliesec (commented):
One suggestion, one question. But otherwise, looks like a nice set of optimisations.

BondageClub/Scripts/Drawing.js
NoneNoname: previously this was called every DrawProcess, and now only on resize, when the window loads the size can be incorrect, actually on mobile I did experience this effect even more and few extra deferred calls were necessary:
```
	DrawWindowResize();
	setTimeout(DrawWindowResize, 1000);
	setTimeout(DrawWindowResize, 3000);
	setTimeout(DrawWindowResize, 5000);
	if (CommonIsMobile) setTimeout(DrawWindowResize, 10000);
```

elliesec: It seems odd that it's not getting called as a result of the resize event listener of it's not accurate the first time - presumably there's some underlying cause to that? In any case, it might be worth adding a comment over these to explain what they're there for.

BondageClub/Scripts/Drawing.js
```diff
@@ -187,13 +208,13 @@ function DrawArousalMeter(C, X, Y, Zoom) {
...
 				if (C.ArousalSettings != null && Player.ArousalSettings.VFX != "VFXInactive" && C.ArousalSettings.Progress > 0 && ((C.ArousalSettings.Active == "Automatic") || (C.ArousalSettings.Active == "Hybrid"))) {
-					var Progress = 0
+					let Progress = 0
 					if (!((C.ArousalSettings.VibratorLevel == null) || (typeof C.ArousalSettings.VibratorLevel !== "number") || isNaN(C.ArousalSettings.VibratorLevel))) {
 						Progress = C.ArousalSettings.VibratorLevel
 					}
 
 					if (Progress > 0) // -1 is disabled
-						var max_time = 5000 // 5 seconds
+						const max_time = 5000 // 5 seconds
```
elliesec, line 217: Changing this to a `const` has exposed an issue in the existing code: `max_time` only gets defined in the `if` - the drawing then proceeds regardless. This `if` needs braces.

BondageClub/Scripts/Drawing.js
```diff
@@ -374,7 +410,7 @@ function DrawAssetGroupZone(C, Zone, Zoom, X, Y, HeightRatio, Color, Thickness =
  * @returns {boolean} - whether the image was complete or not
  */
 function DrawImageZoomCanvas(Source, Canvas, SX, SY, SWidth, SHeight, X, Y, Width, Height, Invert) {
-	var Img = DrawGetImage(Source);
+	const Img = DrawGetImage(Source);
 	if (!Img.complete) return false;
 	if (!Img.naturalWidth) return true;
 	if (Invert) Img = DrawInvertImage(Img);
```
elliesec, line 416: This line is now re-assigning a `const` - should be a `let` now.

BondageClub/Scripts/Drawing.js
```diff
@@ -497,7 +529,7 @@ function DrawImageZoomMirror(Source, X, Y, Width, Height) {
  * @returns {boolean} - whether the image was complete or not
  */
 function DrawImage(Source, X, Y, Invert) {
-	var Img = DrawGetImage(Source);
+	const Img = DrawGetImage(Source);
 	if (!Img.complete) return false;
 	if (!Img.naturalWidth) return true;
 	if (Invert) Img = DrawInvertImage(Img);
```
elliesec, line 535: This line is now re-assigning a `const` - should be a `let` now.

BondageClub/Scripts/Drawing.js
NoneNoname: resize is only the result of the window being resized and not the content change so that may cause this but I am not too sure the exact reason for this, probably there is a better solution I just yet to find it

also, it may seem fine at first to leave the deferred calls out, but in some cases, it will result (especially on mobile) the canvas getting oversized

elliesec: Yeah, leaving them in is fine for now, although it would be good to try to identify the cause at some point. Like I said, I'd be inclined to just leave a comment above the code to explain why it's there, and maybe someone will figure it out in the future.

NoneNoname: added the comment for it

elliesec:
Looks good to me now. I'll pull down the latest changes and give it another test run tomorrow just to double check.

BondageClub/Scripts/Drawing.js
```diff
@@ -615,27 +656,26 @@ function DrawInvertImage(Img) {
...
-				LineCount++;
-			} else line = testLine;
-		}
-
-		// If there's too many lines, we launch the function again with size minus 2
-		if (LineCount > MaxLine) {
-			MainCanvas.font = (parseInt(MainCanvas.font.substring(0, 2)) - 2).toString() + "px arial";
-			return GetWrapTextSize(Text, Width, MaxLine);
-		} else return;
+	if (MainCanvas.measureText(Text).width > Width) return;
```
jomshir98, line 659: This logic seems inverted to me

```suggestion
	if (MainCanvas.measureText(Text).width <= Width) return;
```

BondageClub/Scripts/Drawing.js
```diff
@@ -991,37 +1030,47 @@ function DrawProgressBar(X, Y, W, H, Progress) {
...
-			MainCanvas.width = MainCanvas.height * 2;
-			MainCanvas.canvas.style.width = "";
-			MainCanvas.canvas.style.height = "100%";
+		const Scale = (W <= H * 2) ? 2000 / W : 1000 / H;
+		const MainCanvasRect = document.getElementById("MainCanvas").getBoundingClientRect();
+		MouseMove = function MouseMove(event) {
+			MouseX = Math.round((event.clientX - MainCanvasRect.left) * Scale);
+			MouseY = Math.round((event.clientY - MainCanvasRect.top) * Scale);
 		}
+		Touch = function Touch(event) {
```
jomshir98, line 1050: `Touch` is a name of global interface we are overwriting; maybe worth renaming it to something else?

jomshir98 (changes requested):
Two small notes, but otherwise LGTM

jomshir98 (approved):

elliesec (approved):
Looks good to me - no obvious issues that I can see, and everything seems to work well in testing. Good call shifting the canvas resize out into CSS. Happy for this to go in.

Nina-1474 (approved):

BondageClub/Scripts/Drawing.js
```diff
@@ -403,26 +442,24 @@ function DrawImageResize(Source, X, Y, Width, Height) {
...
-		tmpCanvas.height = Img.height;
-		var ctx = tmpCanvas.getContext('2d');
-		ctx.drawImage(Img, 0, 0);
-		AlphaMasks.forEach(([x, y, w, h]) => ctx.clearRect(x - X, y - Y, w, h));
-		Canvas.drawImage(tmpCanvas, X, Y);
+		TempCanvas.canvas.width = Img.width;
+		TempCanvas.canvas.height = Img.height;
+		TempCanvas.drawImage(Img, 0, 0);
+		AlphaMasks.forEach(([x, y, w, h]) => TempCanvas.clearRect(x - X, y - Y, w, h));
+		Canvas.drawImage(TempCanvas.canvas, X, Y);
```
Nina-1474, line 462: Won't this line result in the image being drawn twice, once without the opacity setting and once with it?

NoneNoname: Yep it will, I fixed it, now it should work well

PR #1510, opened 2020-09-30
Miscellaneous asset fixes

Shiranui-Izayoi:
Just a few small asset fixes that I decided to do as a break from trying to make the spanking toys work as activities as well as get sounds into activities.

First up the diapers are my own fault since when I increased the Priority of the (Sock Slot) Reverse Bunny Suit, I didn't increase the diapers Priority as well meaning it went under the socks looking strange.

Second I replaced all the images of the Adult Baby Harness with the ones from the null pose aka the default. I chose to do it this way round as the normal one looked to be greyscaled, and from what I can tell their isn't actually any difference between them other than the colour.

I also added the Adult Baby Harness to the Over Head pose, and after a quick test it looks to work and look fine. I have known about this issue for a while and even reported it. Since it was never fixed I decided to do it myself. 

The issue was that when you restrained a character in any way that changed the pose of their upper body the Harness would change colours. Which was because of the aforementioned fact that only the original had been greyscaled.

So yer super small change, but meh. I am sure those that use it would like it.


ace-1331 (approved):

ace-1331:
Just a quick note in case you don't know/want to change this, Priority 23 conflicts with some straitjackets.
Due to the new way layers are ordered, you are guaranteed a consistent order regardless of the order things were put in unlike before. However, this means if an item on that layer is ever before the harness (alphabetically) it wouldn't be at the back like the rest of the items on Layer 23. 

I personally don't mind either way, this priority seems appropriate overall at the moment

Shiranui-Izayoi:
Ahhhh do you mean the leotard? In which case it was my intention to have the diapers be over the top. Since it was to my knowledge agreed a while back that the leotards should be skin tight. And thus go under a bunch of stuff. It was why at the time I changed the OG one to 18. 

Also all the leotards tend to clip badly with the diapers when they go over them. Same with the other stait jackets IIRC. So might be better for them to go over.

ace-1331:
Yes that's what I meant, and yes I agree that the leotards look better under the diaper. What I mean though is that now because they are all on layer 23, if there was a new item added to layer 23 that is before "Diapers" when sorted in alphabetical order, that item would show on top.

So let's say I add the item "ALeotard" on layer 23, then the diapers would be under.

Before it was a problem as that meant the order you put the item in was the order things were rendered in which made it inconsistent. 

It's not something that should be relied on, but the odds of something like that happening with this item is low, and can be changed when/if it happens.  That's why I didn't think it was a big deal.

Thought I'd let you know in case there's ever a problem with diaper priority later as it could be caused by this

Shiranui-Izayoi:
Hmmmm yer I see what you mean though honestly I am drawing a blank on anything they could call a leotard that would start with those letters. :)

But yes I do see your point worse comes to the worst, I will just adjust it again. Or the Leotards could go down a number depending on the best course of action. Honestly the larger the amount of priorities the better in situations like this since that means that their is more room to adjust without having to then also move everything else. But meh as you said fine for now... Now we wait to hear from the one person that found some super secret combo of items that make it work really well it happened last time with the leotard.

elliesec (approved):

PR #1511, opened 2020-09-30
Improvement: Introducing memoization and using it for extended items

SandrineBC:
# Summary
This PR introduces memoization to the `Common.js`. This feature will help to cache often used functions and return their results, without constant recalculations. This is best used with a small and foremost constant index, so the cache can be kept small. Objects are converted before using them for index building, so using them is possible, but using a constantly changing object, like the Player would not be good idea.
Memoization is put into use as well within this PR as it's applied to the requirements check within the extended item scripts, so that not applicable options now will be colored.
# Details
## What it does
Adds a new function `CommonMemoize` to `Common.js` to add memoization to a function, you can just do the following (example taken from `ExtendedItem.js`)
```JavaScript
const ExtendedItemRequirementCheckMessage = CommonMemoize(ExtendedItemRequirementCheckMessageMemo);
```
Then you can just call the memoized function as usual. To access the unmemorized function (for some last minutes checks), just call `ExtendedItemRequirementCheckMessageMemo` directly.
To clean the cache, you can use the build in `clearCache` function, like this (example taken from `ExtendedItem.js`)
```JavaScript
function ExtendedItemExit() {
	// invalidate the cache
	ExtendedItemRequirementCheckMessage.clearCache();
}
```

Additionally, all `xxxValidate` functions were changed, to return a string, instead of a boolean, so that the returned string can be used by the function within the extended item script, avoiding side effects by setting global variables within the items.
## Testing
* All changed items were tested in single and multi player environment.
* Before introducing memoization, some performance measurements were taken, showing that the use of this feature increases execution time by a factor of roughly 25 (example taken from the web validate function)
image


Nina-1474:
This conflicts with #1504 a fair bit since they both changed the ExtendedItem Validate functions. The main thing to try and maintain is passing the C parameter between functions. I assume you don't want to do that for the memoized function though, so if you must declare C anew anywhere, just make sure to set it as `var C = CharacterGetCurrent() || CharacterAppearanceSelection;`.

Nina-1474 (changes requested):
There are still some Validate functions returning booleans instead of strings. These will fail the memoized validation and show the text "true" instead of passing.

Nina-1474 (approved):
I found two issues that I pushed fixes for to save time and retesting:
- Exiting the extended item screen by selecting a new focus group on the current character was not calling the Exit function to clear the cache. I had a change for this exact case in a local branch so added it here.
- The Futuristic items all call their Validate functions outside the ExtendedItem script and these had not been updated to account for the new return type. Now they have.

Otherwise everything seemed to be working as intended.

elliesec (approved):
I've tested this out on all of the affected items, and it all seems to be working as expected. I made a couple of very minor changes (a typo fix, removed a commented code, and corrected a comment that was no longer entirely accurate), but it otherwise all looks good.

The only thing that I did notice is that the memo's cache is getting invalidated twice on most (all?) items when you click an extended option (due to the extended item script calling `ExtendedItemExit` and it also getting called as a result) of the `LeaveDialog` flag in the `PublishAction` functions. But I don't think that's something we can really get around easily, and as long as `ExtendedItemExit` is inexpensive and idempotent (which it currently is), this shouldn't be a problem.

Approving this, and will merge once we're out of code freeze.

PR #1514, opened 2020-09-30
Bugfix: Prevents players from colouring items they shouldn't be able to

elliesec:
## Summary

Removes the colour picker button from the dialog menu buttons when the item is locked and the player is unable to unlock the item.


NoneNoname (changes requested):
I think this will prevent you from coloring, Lover Belt, and the cuffs so in general items that have lock on the Asset itself

Edit: 
I selecting color before will still work, but there is no unlocked state for those items so you can only blindly guess the color you want on the item

elliesec:
Good point. The JSDoc on `DialogCanUnlock` was somewhat misleading. I've hopefully made that a little clearer now.

ace-1331 (approved):

PR #1515, opened 2020-10-01
convert 2d to 3d 

zerodawn1:
It contains: refresh function, color change function and a window resize function.
You should be able to generate the 3D model without any error messages.
I've changed the dress and strip function and added a jump over asset function to jump over non-existent elements.

to test the little update, test account : name = 222, pw = 2.


Ben987:
Lets close this for now.  We can open another 3D change later if this gets updated.

PR #1520, opened 2020-10-04
Futuristic Panel Gag

Ada18980:
- Added a new futuristic gag with three modes providing varying levels of gag strength. It can't be modified while locked.
- Added a harness variant with a colorable strap, same buygroup as the regular gag.

Tested:

- Trying to modify while locked
- Trying to modify while bound
- Trying to modify while bound and locked
- Maid removal
- Multi color picker
- Verified strings in English


Ada18980:
I realized that the localization may be broken, I did not localize any of the strings.

ace-1331:
> I did not localize any of the strings. Is this going to cause crashes for other language users or is it only an accessibility issue?

It will default to English when a line is not translated, so no it won't cause any crash

BondageClub/Screens/Inventory/ItemMouth/FuturisticHarnessPanelGag/FuturisticHarnessPanelGag.js
```diff
@@ -0,0 +1,69 @@
+"use strict";
+
+var InventoryItemMouthFuturisticHarnessPanelGagOptions = [
+	{
+		Name: "Padded",
+		Property: {
+			Type: null,
+			Effect: ["GagLight"],
```
ace-1331, line 8: This should have the blockmouth effect as well, otherwise it means anything that requires the mouth to be open will be possible such as the penetrate action.

BondageClub/Screens/Inventory/ItemMouth/FuturisticPanelGag/FuturisticPanelGag.js
```diff
@@ -0,0 +1,69 @@
+"use strict";
+
```
ace-1331, line 2: Since the second file is exactly the same, the functions should just point to the other gag's functions to reduce code duplication and make these items easier to maintain

BondageClub/Assets/Female3DCG/Female3DCG.csv
```diff
@@ -833,6 +833,8 @@ ItemMouth,KittyGag,Kitty Mask Gag
 ItemMouth,KittenHarnessPanelGag,Kitty Harness Gag
 ItemMouth,CarrotGag,Rubber Carrot Gag
 ItemMouth,MuzzleGag,Muzzle Gag
+ItemMouth,FuturisticPanelGag,Futuristic Panel Gag
+ItemMouth,FuturisticHarnessPanelGag,Futuristic Panel Gag w/ Harness
```
ace-1331, line 837: Can you change "w/" for "with"? Or "Futuristic Harness Panel Gag" is shorter if you want to save room

ace-1331 (changes requested):
Besides the 3 code comments below, my main comment is the following ->
I can't swap from ball to plug mode, I always have to go back to padded first, is this intended?

BondageClub/Screens/Inventory/ItemMouth/FuturisticHarnessPanelGag/FuturisticHarnessPanelGag.js
Ada18980: Agreed, I made the change.

BondageClub/Screens/Inventory/ItemMouth/FuturisticPanelGag/FuturisticPanelGag.js
Ada18980: Done.

Ada18980 (commented):
Hi Ace,

I agree with your 3 code commends and have committed them to the repo. Re: the inability to switch, I'm not sure why this is happening as I did not intentionally set that to happen.

Ada18980:
There, the issue that was preventing the user from switching from Ball to Plug was the fact that both were of "Plug" type whereas they should have been different

ace-1331 (approved):

Ada18980:
Double checked that the gag still works on the new commit.
@Ben987

PR #1524, opened 2020-10-05
Voice Changer Gag

Ada18980:
Added a new gag, an artificial voice changer gag. The purpose is to play messages (through emotes rather than through chat, since without additional hooks we can't get around the garble filter) during certain events. The current conditions are:
- Play a message on orgasm
- Play a message whenever the player goes from <5 Arousal to >5 Arousal
- Play a message when the player kneels
- Play a message when the player stands
- Play a message periodically with a random interval. The timing can be configured in units of 1 minute average between messages, and the actual time varies from 30 seconds to 90 seconds per interval
- Play a periodic message, but only when above 95 Arousal. This overrides the periodic message.
- Play a periodic message, but only when below 5 Arousal. This overrides the periodic message.

The gag also features three different appearances, the first is a ball gag with a speaker, the second is a harness ball gag, and the third is a sci-fi inspired muzzle gag based on my other pull request, though this one has a speaker instead of a screen.

Test cases:
- Tested with two accounts in a single chat room: Passed. Player name tags are correct and two characters with the same gag have different messages.
- Tested kneel and stand messages by standing and kneeling repeatedly and also with players being helped or forced to kneel. Being helped to stand up will not send the message, though curiously it will reset the kneeling status so it recovers immediately.
- Tested all strings
- Tested Harness and Muzzle configurations
- Tested locking and access with exclusive, metal, and combination locks (the gag can be reconfigured if you have access without taking it off. This is important)
- 40+ minutes of consecutive testing of timer functions


Ada18980:
Ack, I realized that the code for the futuristic panel gag got pulled in here as well

PR #1525, opened 2020-10-05
Speaker Gag

Ada18980:
Added a new gag, an artificial voice changer gag with a speaker. The purpose is to play messages (through emotes rather than through chat, since without additional hooks we can't get around the garble filter) during certain events. The current conditions are:

- Play a message on orgasm
- Play a message whenever the player goes from <5 Arousal to >5 Arousal
- Play a message when the player kneels
- Play a message when the player stands
- Play a message periodically with a random interval. The timing can be configured in units of 1 minute average between messages, and the actual time varies from 30 seconds to 90 seconds per interval
- Play a periodic message, but only when above 95 Arousal. This overrides the periodic message.
- Play a periodic message, but only when below 5 Arousal. This overrides the periodic message.

The gag also features three different appearances, the first is a ball gag with a speaker, the second is a harness ball gag, and the third is a sci-fi inspired muzzle gag based on my other pull request, though this one has a speaker instead of a screen.

Test cases:

- Tested with two accounts in a single chat room: Passed. Player name tags are correct and two characters with the same gag have different messages.
- Tested kneel and stand messages by standing and kneeling repeatedly and also with players being helped or forced to kneel. Being helped to stand up will not send the message, though curiously it will reset the kneeling status so it recovers immediately.
- Tested all strings
- Tested Harness and Muzzle configurations
- Tested locking and access with exclusive, metal, and combination locks (the gag can be reconfigured if you have access without taking it off. This is important)
- 40+ minutes of consecutive testing of timer functions
- Max character limit

The max character limit for any message is 200 characters, while the max stored message length is 1000 characters per field. Fields are separated with the '|' character like so: Meow|Woof. The gag will randomly select either Meow or Woof in this context.
If a message is greater than 200 characters then the gag will report "My gag is malfunctioning, please shorten the message."

Known issues:
- When helping someone to stand up, it does not play the stand up line. It works fine if you tie them using frogtie straps and untie them, though
- The gag has to use emotes instead of actions or chat. This is because actions require a group to be focused, and chat will be garbed as gag-speak. This prevents users who wish to hide automatic messages from hiding the gag's messages, which can be a spam concern especially in a room with 10 gagged members orgasming constantly--average of 10 messages per minute in addition to the orgasm messages every 3-4 minutes.
- When orgasming while modifying a speaker gag's settings, the text fields will be stuck. This may be a problem with the orgasm system. It is compounded by the fact that the screen does not turn pink when editing items on a person, but the issue is shared with the Combination Lock.
- Disabling features is done by publishing a blank message, but this is not immediately obvious to the user
- There is no default values for the gag, so it does nothing until a user takes advantage of its features

I've also removed the FuturisticPanelGag files that were mistakenly added in the first PR. Thanks Ace for the help in figuring that out!


ace-1331:
I have several concerns about this beyond the listed known issues so I'll have a look at it later, but just wanted to drop by to say that #1527 fixes the menu issue when orgasming

Nina-1474:
Suggestions/thoughts:
- Instead of each message having its own Save button, there should just be a single Save/Confirm button at the end. There's plenty of space at the top to move the controls up and give more space at the bottom. That way the user can enter/change/remove the messages all they want then confirm the full set of changes once done. This would 1) send the final configuration in one server call instead of many and 2) prevent the flood of chat messages when the user wants to set everything. It also means there aren't messages and changes being sent out while the player is still viewing the item screen, possibly unaware of them, making this more consistent with every other item.
- That would also allow the buttons to be replaced with labels without the "Save" part. At the moment, the text gets very small and I expect may be near-unreadable on mobile. I'd even give the label more horizontal space as well. Particularly since the word "Msg" stands out as being a bit unprofessional to me, for lack of a better way to put it.
- I'd wonder if there's a better delimiter than "|" you could use. It's quite an awkward key to get to on mobile. Or maybe there could be a small button on screen to type it into the focused box?
- Is CharacterRefresh needed in the `persistentData.MsgTime < time` block?
- Is the BeforeDraw setting and function needed?
- Small detail but the usual method of checking if C is the Player is currently always `C.ID == 0`.

Ada18980:
> Suggestions/thoughts:
> 
> * Instead of each message having its own Save button, there should just be a single Save/Confirm button at the end. There's plenty of space at the top to move the controls up and give more space at the bottom. That way the user can enter/change/remove the messages all they want then confirm the full set of changes once done. This would 1) send the final configuration in one server call instead of many and 2) prevent the flood of chat messages when the user wants to set everything. It also means there aren't messages and changes being sent out while the player is still viewing the item screen, possibly unaware of them, making this more consistent with every other item.
> * That would also allow the buttons to be replaced with labels without the "Save" part. At the moment, the text gets very small and I expect may be near-unreadable on mobile. I'd even give the label more horizontal space as well. Particularly since the word "Msg" stands out as being a bit unprofessional to me, for lack of a better way to put it.
> * I'd wonder if there's a better delimiter than "|" you could use. It's quite an awkward key to get to on mobile. Or maybe there could be a small button on screen to type it into the focused box?
> * Is CharacterRefresh needed in the `persistentData.MsgTime < time` block?
> * Is the BeforeDraw setting and function needed?
> * Small detail but the usual method of checking if C is the Player is currently always `C.ID == 0`.

I agree with all of your comments. This item isn't super high on my priority list, but when I do get around to salvaging it (it's one of the very first of the futuristic items before all the standardization and shared code) I'll be sure to take a loot at redoing the interface and especially looking at the dynamic draw stuff because Im not confident I implemented it well in the this item.

PR #1528, opened 2020-10-05
Remove all `+ "Dark"` background images

NoneNoname:
depends on #1507


Ben987:
Please make sure to test that one on slower computers, we've tried to remove the dark images before but players with slower computers had in-game slow-downs.  Thanks!

NoneNoname:
We already use this mechanism since #625, when the player has a blindfold, or if we are in a ChatRoom so in most cases this only removes the need to load an extra image.

I also built it on top of my optimization pr that remove that removes a big chunk of processing from the `DrawProcess()` namely the resize operation and the lookup for the current draw function, any of these two operations take longer than drawing an opaque rectangle on top of the actual background

This will also make it more simple for other contributors to add background since now they do not have to provide a dark version only a single image

Ben987:
As long as you test it with slower computers so we don't get any complaints I'm good with it. :)

PR #1534, opened 2020-10-07
Futuristic chastity belt

Ada18980:
Added a sci-fi style chastity belt. Functionality is identical to the Polished chastity belt. It has 3 layers that can be colored: The display on the front, the body, and the mesh backing behind the screen.

It also has a symbol on the front as a lock, instead of a physical lock.


ace-1331 (approved):

Ben987:
Players that loves sci-fi will appreciate all these new items, thanks :)

PR #1547, opened 2020-10-10
Fixed coffin interaction

ace-1331:
- fixed the action text for the coffin


aldarris:
For the proper grammar, it should be 'DestinationCharacter' (which translates to 'her') rather than 'TargetCharacter' (which translates to 'herself'):
'CoffinSetOpen,,,SourceCharacter opens the lid on DestinationCharacter coffin.,,'

PR #1549, opened 2020-10-10
Bunny & Cow tail strap variant

ace-1331:
- added a cow&bunny tail strap variant as requested

They were the only one that were only available as plugs


Shiranui-Izayoi:
Hey wouldn't it be a good idea to add a blank image for the bunny tail? Since otherwise when the game tries to load it when equiped it will fail and create errors right?

elliesec:
> Hey wouldn't it be a good idea to add a blank image for the bunny tail? Since otherwise when the game tries to load it when equiped it will fail and create errors right?

The `Visible: false` flag in the asset definition should prevent the game from trying to render an image for it.

Shiranui-Izayoi:
Ahhh ok then. I wasn't aware that that flag existed. I just saw a lack of image and thought I should mention it.

ace-1331:
Yes, `Visible: false` makes it so no asset at all is looked for when drawing a character so no need for empty images, those items just need a preview

Ben987:
Yes!  That's a good idea.  Cheers!

PR #1550, opened 2020-10-11
Import/Export Colors

ace-1331:
- add the ability to quickly copy/paste a color for mobile user friendliness

Made sure to only display these if the clipboard api is available


Ben987:
I won't be able to test this one but this looks very strait-forward.  I'm sure mobile users will like it.  Thanks a lot!

PR #1551, opened 2020-10-11
Beta bugfix: Disallow items on movie studio props

wultir:
Tiny bug (or at least weirdness) in the movie studio. The cross and drawers have `AllowItem` set to true allowing you to add restraints to them.

This simply sets it to false on both.


Ben987:
Absolutely!  Thanks.

PR #1553, opened 2020-10-11
Small coffin fix, and cryo capsule.

aldarris:
Coffin was erasing the emoticon, and another alfa mask was required to cut off certain kinds of hair.
Cryo capsule is practically cut and paste.


BondageClub/Screens/Character/Player/Dialog_Player.csv
```diff
@@ -1280,8 +1280,13 @@ FamiliarSetParrot,,,SourceCharacter summons Pirate Familiar on TargetCharacter.,
 SelectCoffinType,,,Select coffin closed or open,,
 CoffinTypeOpen,,,Open,,
 CoffinTypeClosed,,,Closed,,
-CoffinSetOpen,,,SourceCharacter opens the lid on the coffin TargetCharacterName is in.,,
-CoffinSetClosed,,,SourceCharacter closes the lid on the coffin TargetCharacterName is in.,,
+CoffinSetOpen,,,SourceCharacter opens the lid on the coffin TargetCharacter is in.,,
```
ace-1331, line 1283: There's a mismatch between the text and what's passed in the dictionnary (TargetCharacterName vs TargetCharacter)

Target character outputs "X closes the lid on the CryoCapsule herself is in" so it is not correct anyway.

Might want to look at the action text to have them reflect what you intend to say.

ace-1331 (changes requested):
Other than the 1 comment about the actions, everything else looks good

aldarris:
TargetCharacter instead of TargetCharacterName is the fix.  In the current beta, it prints 'TargetCharacterName' to the chat isntead of target character's name.

aldarris:
Oh, I see.  The 'TargetCharacter' has to match in js file as well.
Fixed.

ace-1331:
I was just saying "(X opens the lid on the cryogenic capsule **herself** is in.)" doesn't sound right like 
"(X opens the lid on the cryogenic capsule **she** is in.)" or like most items do, "(Gray opens the lid on her cryogenic capsule.)"

But I'm not a native english speaker so I'll leave that up to everyone else.
However, I noticed another thing, the lower Alpha mask on the capsule visually cuts the feet off in its open state because it goes a little lower than the coffin. It could probably be lowers a bit more to look better.

aldarris:
Fixed the cutout (it was frame as well as alpha).
Looked over the ChatRoomMessage function, no idea how to replace 'herself' with 'she'.  Not without some major refactoring.

aldarris:
Grammar is fixed.

ace-1331 (approved):

Ben987:
Sweet!  This will fit well with all the new sci-fi stuff.  Thanks a lot :)

PR #1555, opened 2020-10-12
Fix: Simple Hogtie Prerequisites

Nina-1474:
The Simple Hogtie option for Hemp Rope on arms never had Prerequisites allowing it to be combined with incompatible poses like suspension and horse. The standard Hogtie prereqs have been copied over.


Ben987:
Thanks for the fixes!  I will add them to the second beta.  Cheers.

PR #1556, opened 2020-10-12
Foam Sword Complete

EmilyR42:
- Fixed avatar's response to the foam sword.
- Removed extra file that was causing issues.


ace-1331 (dismissed):

Ben987:
Sure!  This is a fun idea.  Thanks a lot :)

PR #1557, opened 2020-10-12
Beta bugfix: Don't set auto AFK if it's already set.

wultir:
The automatic AFK emoticon does not get cleared if

1. Starting without an emoticon active
2. Going AFK for two or more "cycles" of the AFK check (>10 minutes)

On the second check `if (InventoryGet(Player, "Emoticon") && InventoryGet(Player, "Emoticon").Property && AfkTimerOldEmoticon == null)` will be true which sets the `AfkTimerOldEmoticon` to the AFK emoticon making the reset ineffective.

The fix checks `AfkTimerIsSet` to avoid setting it again. From what I can tell everything for auto AFK should be set by that point so it doesn't need to set anything again.


ace-1331 (approved):

Ben987:
Absolutely!  Thanks a lot.

PR #1558, opened 2020-10-12
Chat Search Friends Position fix

Nina-1474:
The 'Friend in room' hovering text for chat room buttons is mispositioned sometimes. 
When the room button is low enough, currently if below Y = 352 i.e. row 4, then the hovering text is positioned at row 4. This is presumably to ensure there's enough space to display a long list of friends without overlapping the bottom buttons and/or running offscreen. However for a short list of friends, the text often appears far above the room button and looks incorrect.
Now the height of this text is calculated so that it's only moved up when it will go too low, and ends up still beside beside the room button.


ace-1331 (approved):

Ben987:
Thanks for that one, I did not notice it before.

PR #1559, opened 2020-10-13
Vibe VFX

Ada18980:
Added a pink glow around the arousal meter that's drawn when the character is being vibrated/aroused by actions, and a blue glow that appears when the character is aroused but no longer stimulated. The purpose of this is to show players that their arousal is going up, without requiring them to read the chat log or manually inspect their items. This can help to reduce 'surprise orgasms,' and also make it easier to roleplay based on arousal even with automated messages from vibrators turned off, as many players are likely to do now that the vibrator modes are sending out a fair amount of chat messages.

Also adds a new "Graphics" subpage in the settings with a flower icon, used for settings that are pure eyecandy without practical effect or relation to kink.

There are four options:
- Inactive (doesn't show on your screen)
- Solid glow (Pink when vibed / Blue when cooling down)
- Temporary Animation (5 second vibe/blinking, then solid)
- Constant Animation (Vibe/blinking never goes away)

Test cases:
- Different vibrator intensities
- Actions that generate arousal
- Actions that subtract arousal
- Vibing someone that doesn't like to be vibed
- Orgasming (the visuals should disappear and reappear)
- Multiple accounts (the setting should be player-side so a player who disabled it doesn't see it, but changes in vibration mode should be character-specific so even if a player doesn't see it, other players will see the animation if they have temp animations or constant animations enabled)
- Test all 4 options in preferences screen


ace-1331:
I won't be reviewing that one myself because I couldn't care less about the arousal system so I have no clue what people want and how this fits into it, but after looking through the code, it looks like it wont add too much data or add any syncs to it (they were my two worries)

Overall Ill let others who know the system better look into the details, but it seems ok.

Ada18980:
Updated to use the new settings subpages

Ben987:
Alright!  Thanks for that one, I'm curious to test it.

PR #1560, opened 2020-10-13
Advanced Futuristic Chastity

Shiranui-Izayoi:
Added custom extended functionality to the newly added Futuristic Bra & Belt with custom names for the buttons and descriptions to avoid confusion due to the shock and vibe otherwise having the same words.

In addition wrote up custom dialogue to the actions to make the item feel more unique and distinct. The shared functions for the Bra and Belt are contained within the FuturisticChastityBelt.js file, as I couldn't figure out a better place to put them. Also added is a validation to prevent people changing the settings of the bra and belt if they can't unlock it.

While it is possible that the functions could be made generic or done using ExtendedItem scripts I do not have the knowledge to do this. And from what I can tell the ExtendedItem scripts cannot currently handle it. Though I am fine if it is changed in the future for the sake of making it easier to fix etc.

I have tested it and it all seems to work fine.

Ow and also please ignore the closing and opening along with the edit. I accidentally hit enter and it submitted it before I was done typing and I couldn't find the delete button.


Shiranui-Izayoi:
> Hey, just my 2 cents as the one who made the futuristic chastity belt and bra.
> I have some future plans in the mix for the items themselves, and I don't know what I'd feel about deleting your work to make room for those. It would be a lot easier if you made a separate item, but I don't know how people would feel about having two items with the exact same appearance so... well... yeah.

Well I will say that there is room. As you have seen from the screen shots. If you want we can work together to make it work. Or I can just make it a seperate item. I don't think people would have an issue with two items looking the same if they had vastly different uses.

One thing that could be done to give more room to the items would be to move all the buttons up a bit and potentially remove the main item image since it doesn't really serve any use. If we removed the image and moved the current buttons up 200 ish then we would have plenty of room.

Finally another possible idea is to create a dual mode for the bra and belt where you could select one of two "software" upgrades that would then show different menus allowing 1 item to do two separate things. I am up to discuss more on Discord if you want? Worst comes to the worst I will just copy assets and make a second item as said.

Shiranui-Izayoi:
If the only issue is the potential use of these items later by the original developer I can create copies of these and submit them as standalone items.

Otherwise if there are any issues that the devs have with it please let me know so I can fix them. I am up to removing a lot including getting rid of any of the custom actions, names etc. Including the Orgasm Lock. All I really care about is the vibe and shock functions since it is interesting.

I am currently making a standalone version just in case that is the issue. And will submit it seprately soon.

ace-1331:
**Marked this PR and #1571 as "on-hold" to avoid confusion and make sure they are not *both* merged, but also to make sure its decided which one is taken before any of them is merged.**

___

We'll take the time to discuss this more after the beta when appropriate to clear it up, but if an item (especially a recent one) that was made by someone is modified soon after it is included, I'm strongly inclined to let the contributors decide what happens with their items. We'll wait for Ada's input on that side. (Duping the item with the exact same assets doesn't solve this issue)

As for the functionalities themselves, to make the item easier to maintain and include in future extended item expansions, it would be much nicer if the intensity levels did use the common text (it is also a lot less confusing for the end user). Also, the ability to lock/unlock orgasm is a new effect that is not realistic and does actually enter sci-fi which is new. If it is kept, I personally would open up a discussion about it for community feedback before approving it.

Shiranui-Izayoi:
> 
> 
> **Marked this PR and #1571 as "on-hold" to avoid confusion and make sure they are not _both_ merged, but also to make sure its decided which one is taken before any of them is merged.**
> 
> We'll take the time to discuss this more after the beta when appropriate to clear it up, but if an item (especially a recent one) that was made by someone is modified soon after it is included, I'm strongly inclined to let the contributors decide what happens with their items. We'll wait for Ada's input on that side. (Duping the item with the exact same assets doesn't solve this issue)
> 
> As for the functionalities themselves, to make the item easier to maintain and include in future extended item expansions, it would be much nicer if the intensity levels did use the common text (it is also a lot less confusing for the end user). Also, the ability to lock/unlock orgasm is a new effect that is not realistic and does actually enter sci-fi which is new. If it is kept, I personally would open up a discussion about it for community feedback before approving it.

Sure no problem I can remove the buttons for the Lock Orgasm for now but leave in the functionality just in case so it can't be accessed. And I will change the button names back to normal deleting the custom ones.

Shiranui-Izayoi:
Closing the pull requests due to them no longer being needed.

PR #1530, opened 2020-10-06
Ability to suspend maid service

Ada18980:
This change adds an extra dialogue option in the Maid Quarters, allowing you to prevent the maids from helping you for a number of time periods:
- 10 minutes
- 30 minutes
- 1 hour

The effects of disabling maid service are:
- No help from the hallway maid (she will refuse to release you and tell you how much time is remaining)
- No help from photography club or the maid quarters maid 
- No chastity removal from the management

There is a full dialogue tree (english only, currently) with several chances for the player to back out before deciding to disable maids, and there are no options that are indefinite; the longest is 1 week. 

When talking to the maid, the player can also "Shake their head and plead for more" which allows them to refresh the timer for the maids being disabled. This is helpful for players who don't wish to remove their gags in order to talk to the maid in the maid quarters (or can't)

Test cases:
- Tested asking for maid work while both gagged and not gagged
- Tested photography club (refuses to help)
- Tested asking for chastity removal in Management (refuses to help)
- Tested being both gagged and ungagged when being helped by maid
- Tested owner locks and lover locks with maid (it should override it and prevent the player from removing owner locks while maid help is disabled), gagged and ungagged
- Tested being able to remove chastity after the maid status was removed
- Tested the maid status expiring after 10 minutes
- Tested being able to cancel owners and lovers while the maids are disabled (although not when gagged).

Known issues:
- Being gagged with an owner lock will softlock you unless your owner helps you. It might be possible to get stuck for a week in the worst case scenario, but disabling the maids for a week is entirely the user's own choice.
- A week seems like it might be too dangerous and may get a lot of console help requests. The three days option might make more sense to cap it. Alternatively, disabling the maids for a week could be made more difficult to reduce the incidence of users choosing the option unwisely.
- Players might get softlocked for a week if their owner is never online and they chose to be helpless for a week, especially if wearing an owner locked gag
- Combination locks may be dangerous for users without owners, as there will be no easy way to remove them without waiting it out.
- There is no way currently for the player to disable maids while gagged, aside from extending it by talking to the hallway maid and nodding.
- Random kidnappers will remove items with lover's locks. This isn't a bug per se but may reduce enjoyment for people who prefer to be trapped in lover's items. This issue is outside the scope of this PR but is worth noting.

PR #1564, opened 2020-10-13
Maid Button

Ada18980:
This PR adds a button that the user can press to summon the maids (with a bell ring sound effect!). The button simply starts the timer for the maids that normally starts if the user is restrained, and does not add additional functionality to the maids.

There is also a toggle for the automatic maids appearing when the player is restrained.

The purpose of this is to reduce annoyance from the maid bothering the player while the player doesn't want help, and also to help players who have locks on or restraints that maids will not automatically come to remove (such as ankle shackles or combination locks on chastity)

Button icon by Nina


Ada18980:
There, updated it to the behavior requested in discord.

ace-1331 (approved):
Fixes a few scenarios where the maid didn't come + gives people an option to ignore the maid more easily, overall it seems to work smoothly in all cases and I think even if it is not on the hardcore-ish side of thing, people will enjoy it.

We can have a look at #1530 and have more room for changes once this goes through

Ben987:
I'm curious to try that new maid button.  I don't think there's a pitfall but we'll see when beta rolls out.  Thanks!

PR #1565, opened 2020-10-13
Beta bugfix: Makes ItemColorExit more resilient

elliesec:
## Summary

This makes 2 changes which can cause errors on character update within the colour picker:

* Refines the conditions upon which `ItemColorExit` is called upon character refresh so that it is only called when needed
* Updates the `ItemColorExit` function to be more resilient to in cases where `ItemColorLoad` has never been called

These changes should fix errors like this from occurring:

### DIFF
--- a/BondageClub/Screens/Character/ItemColor/ItemColor.js
+++ b/BondageClub/Screens/Character/ItemColor/ItemColor.js
@@ -73,16 +73,9 @@ let ItemColorGroupNames;
  * @returns {void} - Nothing
  */
 function ItemColorLoad(c, item, x, y, width, height) {
+	ItemColorReset();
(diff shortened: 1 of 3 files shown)


ace-1331:
As discussed on discord, seems like there are still some issues after these changes. Let me know when you've made more changes so we can look further.
(Posting here so other approvers know not to merge this yet)

elliesec:
Okay, I've now added some extra logic to ensure all the colour picker elements and variables are properly reset after exiting, and also made the `ElementPosition` function a little safer, with a log warning if there is no element with the provided ID.

ace-1331 (approved):

Nina-1474 (approved):

PR #1568, opened 2020-10-14
Fix: Online drinks

ace-1331:
- fixed a maid dialog soft lock that happens when you have not completed the minigame yet the tray was removed either by a NPC or a player (by accident or due to griefing)

I fixed it by adding a dialog where the maid gives you another drink tray, that way you can go on and continue instead of resetting. This prevents having to refresh and general griefing around the maid tray.

Although not particular to the beta, this is simple enough to squeeze it in


elliesec (approved):

ace-1331:
Good point, I'll include the updated changelog in here right now

Nina-1474 (approved):

PR #1569, opened 2020-10-14
Extended Latex Corset

Shiranui-Izayoi:
I am submitting this as a seprate pull request and plan to edit my original one so it doesn't have this. Since the original one looks like it will be delayed due to the original author having some issues that could take a bit to work out. I will quote what I said in the original pull request for this just in case.

> Along with that I have also split the LatexCorset1 both Bra and ItemTorso versions into 2 layers and made them extended. This was done in order to allow it to be more easily worked into outfits. Since it was made with the Latex Socks in mind it rarely works with anything else normally having the Garter part simply hanging there, mainly due to the fact that the Latex Socks have a rarther unique height and shape.
> 
> Currently the modifications to the LatexCorset1 Assets was done by me so honestly it could be crap, but I felt it was a good idea to submit this now and if I can get my hands on assets that were done better I will either edit this request or submit a new one depending.
> 
> Also quick edit. I added the LatexCorset1 to a few poses that didn't have it while it would be invisible I just gave it an invisible image so that it doesn't draw errors by not being able to find a matching image.


BondageClub/Screens/Inventory/ItemTorso/LatexCorset1/LatexCorset1.js
```diff
@@ -0,0 +1,30 @@
...
+	{
+		Name: "Garter",
+		Property: {
+			Type: "Garter",
+		},
+	},
+];
+
+// Loads the item extension properties
+function InventoryItemTorsoLatexCorset1Load() {
```
ace-1331, line 18: Since these two files are the same, please just refer to one or the other from one of the files to make it easier to maintain in the long run. (Similar to how shock collars, belts, etc. do)

BondageClub/Assets/Female3DCG/Female3DCG.js
```diff
@@ -1520,7 +1526,13 @@ var AssetFemale3DCG = [
...
 			{ Name: "Corset5", Fetish: ["Lingerie"], Priority: 22, Value: 20, Difficulty: 8, Time: 15, RemoveTime: 10, AllowLock: true, BuyGroup: "Corset5", Prerequisite: "AccessTorso", Expose: ["ItemNipples", "ItemBreast", "ItemNipplesPiercings"] },
 			{ Name: "LeatherBreastBinder", Fetish: ["Leather"], Value: 30, Difficulty: 50, Time: 15, RemoveTime: 10, AllowLock: true, Prerequisite: "AccessTorso" },
-			{ Name: "LatexCorset1", Fetish: ["Lingerie", "Latex"], Priority: 21, Value: 40, Difficulty: 8, Time: 15, RemoveTime: 10, AllowLock: true, BuyGroup: "LatexCorset1", Prerequisite: "AccessTorso", Expose: ["ItemNipples", "ItemBreast", "ItemNipplesPiercings"] },
+			{ 
+				Name: "LatexCorset1", Fetish: ["Lingerie", "Latex"], Priority: 21, Value: 40, Difficulty: 8, Time: 15, RemoveTime: 10, AllowLock: true, BuyGroup: "LatexCorset1", Prerequisite: "AccessTorso", Expose: ["ItemNipples", "ItemBreast", "ItemNipplesPiercings"], Extended: true, HasType: false,
```
ace-1331, line 1530: Both items are missing allowtype on the item itself, so the type does not stick

BondageClub/Assets/Female3DCG/Female3DCG.js
```diff
@@ -371,7 +371,13 @@ var AssetFemale3DCG = [
 			{ Name: "StarHarnessBra", Fetish: ["Leather"], Priority: 20, Value: 40, Expose: ["ItemNipples", "ItemBreast", "ItemNipplesPiercings"] },
 			{ Name: "HeartTop", Fetish: ["Lingerie"], Priority: 20, Value: 35, Hide: ["ItemNipples", "ItemNipplesPiercings"] },
 			{ Name: "ChineseBra1", Fetish: ["Lingerie"], Value: 35, Hide: ["ItemNipples", "ItemNipplesPiercings"] },
-			{ Name: "LatexCorset1", Fetish: ["Lingerie", "Latex"], Priority: 21, Value: 40, BuyGroup: "LatexCorset1", Expose: ["ItemNipples", "ItemBreast", "ItemNipplesPiercings"] },
+			{ 
+				Name: "LatexCorset1", Fetish: ["Lingerie", "Latex"], Priority: 21, Value: 40, BuyGroup: "LatexCorset1", Expose: ["ItemNipples", "ItemBreast", "ItemNipplesPiercings"], Extended: true, HasType: false,
+				Layer:[
+					{ Name: "Base", HasType: false },
```
ace-1331, line 377: Since this is someone else's item, it seems appropriate to keep the default behavior it had before. 
The with-garter type should be the default one

ace-1331 (dismissed):
Left some comments that need to be addressed before looking at this further.

BondageClub/Screens/Inventory/ItemTorso/LatexCorset1/LatexCorset1.js
Shiranui-Izayoi: Ahhh sure I kinda forgot I could do that. I just thought that since they both use extendeditem script I thought it was fine. Especially since one is classed as a clothing item and one isn't I will just refer to it in a sec.

BondageClub/Assets/Female3DCG/Female3DCG.js
Shiranui-Izayoi: I had forgotten that sorry. I will add that now.

BondageClub/Assets/Female3DCG/Female3DCG.js
Shiranui-Izayoi: I am not actually sure how to acomplish this I am afraid. The way I have done this is that the top of the corset and the garter part is only that end part. I am unsure how to make it so that it the garters show up by default then go away when you select a type if you could offer any advice on this it would be most appriciated. That said considering how easy it is to just click it I don't think it not looking like that by default is much of an issue.

BondageClub/Screens/Inventory/ItemTorso/LatexCorset1/LatexCorset1.js
Shiranui-Izayoi: Ok done the Bra now loads from the ItemTorso version.

BondageClub/Assets/Female3DCG/Female3DCG.js
Shiranui-Izayoi: They now have AllowType on the main item so it should stick.

Shiranui-Izayoi:
Ok addressed two of the three comments since I am unsure how to make the garters appear by default after trying. I will take another go at it soon when I can think some more.

BondageClub/Assets/Female3DCG/Female3DCG.js
Shiranui-Izayoi: Finally managed to fix this.

Shiranui-Izayoi:
Ok I managed to restore the deafult behaviour in the process also made it so the garters layer can no longer be coloured seprately to the Base layer.

While the solution is potentially hacky it seems that trying to use layers for clothing works a fair bit differently than for non clothing items. Due to this the Bra version requires multiple extra assets. Mainly 2 different versions of the garters image one that is applied when initally equipped which is then removed when the extended menu is opened. The second one is then applied once the extended menu is opened and will vanish when no garters is selected.

I am not sure if there is a better way to do this if so feel free to edit the request directly to do so. But this is the only solution I could come up with that worked. It shouldn't add many if any unneeded resets but I cannot be 100% sure.

That said in order to keep the items default behaviour intact there is no other solution obvious to me.

Hopefully this can now be reviewed and approved in short order with these changes. I look forward to hearing from you.

Shiranui-Izayoi:
Can this be reviewed or whatever since all requested changes were made.

ace-1331:
Since the chatroom filters were  announced, I'm no longer contributing so I won't be doing reviews.

Someone else will get to it eventually, but there's a lot of PRs in general, it's normal if it takes a little while.

Shiranui-Izayoi:
Huh I was not aware that you had stoped. While I was aware that the filters are causing a lot of community outrage I did not realise that it was that bad that you effectivly quit.

elliesec:
I imagine it was more of a last straw kind of thing, which is understandable. As Ace said, we'll get to it eventually - I haven't been very active with club dev recently either, but I'm planning on working through some reviews this weekend if I get the chance. Just be aware that there are also quite a few PRs that have been around for quite a while and are working their way through the process, and we're only just beginning to review and merge things again after pausing for the release.

Ayesha-678 (approved):
Looks good asset wise.

Shiranui-Izayoi:
Hey I was wondering if this could be approved any time soon. Maybe before the next version comes out or something considering it has had 1 approval for a while now and just needs 1 more.

I am sorry if this seems pushy or anything.

Nina-1474 (changes requested):
The way to set the default option is set the first Options array entry to have `Type: null`.
That way you don't need duplicate files like LatexCorset1_LargeGarter_Garter.png, and you don't need that custom GartersByDefault function (which actually has a screen-lock bug that was fixed for the 'null' default setup).

Shiranui-Izayoi:
> 
> 
> The way to set the default option is set the first Options array entry to have `Type: null`.
> That way you don't need duplicate files like LatexCorset1_LargeGarter_Garter.png, and you don't need that custom GartersByDefault function (which actually has a screen-lock bug that was fixed for the 'null' default setup).

Strange because I actually originally tried it with the null option, but I found that the cloth version then wouldn't work correctly. I can test it again with the most recent updates to see if that has changed.

It is actuallty for that reason that the cloth version has so many assets and works in such a round about way. Because I could not get the garters to appear by default using type null.

Nina-1474 (approved):

PR #1570, opened 2020-10-14
Improvement: Adjustments to the inflatable straight leotard

SandrineBC:
# Summary
This PR adjusts the difficulty on the inflatable straight leotard and hinders the bound player to leave the room, if the inflation reaches max level, simply, because they don't pass through doors anymore


Ben987:
Pretty funny, I like it :)

PR #1571, opened 2020-10-14
Sci-fi Chastity Standalone

Shiranui-Izayoi:
Standalone version of the previous pull request to add more functions to the Futuristic Chastity. This was done in order to allow the author to make the future updates he wants.

Assuming that was the main issue then please use this one instead. I will copy and quote my original pull request on this. I have copied the assets and changed the name Sci-fi as well as modified the colour on the preview image to distingish them. Also created a buy group and added the Sci-fi versions to them so that anyone with the Future stuff also gets the Sci-fi stuff.

> Added custom extended functionality to the newly added Futuristic Bra & Belt with custom names for the buttons and descriptions to avoid confusion due to the shock and vibe otherwise having the same words.
> 
> In addition wrote up custom dialogue to the actions to make the item feel more unique and distinct. The shared functions for the Bra and Belt are contained within the FuturisticChastityBelt.js file, as I couldn't figure out a better place to put them. Also added is a validation to prevent people changing the settings of the bra and belt if they can't unlock it.
> 
> While it is possible that the functions could be made generic or done using ExtendedItem scripts I do not have the knowledge to do this. And from what I can tell the ExtendedItem scripts cannot currently handle it. Though I am fine if it is changed in the future for the sake of making it easier to fix etc.


Ada18980:
Im fine with it being a separate item but I don't really want there to be confusion as to which futuristic belt is which. There aren't a whole lot of items which look the same but function differently, and Im not sure the devs would agree either with a fork like this.

I could maybe design a second asset for you to eliminate the confusion.

Shiranui-Izayoi:
Closing the pull requests due to them no longer being needed.

PR #1573, opened 2020-10-14
Fix: asylum bedroom softlocks

ace-1331:
- fixed a bug where you would be softlocked on the sybian when relogging in the asylum

This basically happens to things that have IsRestraint set to false and RemoveAtLogin set to true while you have the check to keep everything on you when logging WHILE you log into the asylum and that item needs to have something that conflicts with clothes in general (patient clothes) or you have >50 rep and cant use your arms.

After verification, it only happens to the sybian currently, however to prevent this from happening ever again, I added a check to bypass "keep all" in the login process if you are about to be sent to the asylum.

This change will only make it so you no longer keep teddy bears or the sybian when relogging in the asylum

### DIFF
--- a/BondageClub/Screens/Character/Appearance/Appearance.js
+++ b/BondageClub/Screens/Character/Appearance/Appearance.js
@@ -37,9 +37,8 @@ function CharacterAppearanceBuildAssets(C) {
 
 }
 
-// 
 /**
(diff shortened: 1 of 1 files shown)


Ben987:
Thanks!  It will probably be useful for other items later.

PR #1574, opened 2020-10-15
Dynamic Functions Resilience

ace-1331:
- added better error handling of non-existing dynamic functions
- added `G` to the list of parameters that can be read from before/after draw to adjust dynamically drawn stuff based on the size

This mimics the behavior of extended item to prevent crashing when testing new items, and it will correctly display the name of the function at fault. (This is not something players will see, it is something to help devs use the framework)


ace-1331:
Closing this since I'm no lonfer wasting my time contributing. Someone can pick up the branch if theyd like

PR #1575, opened 2020-10-15
Fix Diaper Harness Layering

mrbakucamp:
During a layering adjustment, the diaper harness was set to a layer below the diapers. This PR resets it to the same layer as the wand belt.


ace-1331 (changes requested):
All looks good, but it conflicts with the 2 items who hid `ItemVulvaWandBelt`. It's just a matter of adding the diaper harness to the two wedding dress hideitems' list and it'll be all good (`WeddingDress1` and `WeddingDress2`)

ace-1331 (approved):

PR #1576, opened 2020-10-15
Gloves Temp fix

Codefreak64:
so plans have changed, this is a TEMPORARY fix and will be reverted when a proper glove fix is out (side clipping)


Nina-1474:
This makes gloves appear under ClothLower and Hands themselves.
I don't think this issue can be fixed by simple priority changes without causing worse issues. I'd say we need to leave it for now until the proposed gloves priority-split.

Ben987:
Very well, lets not apply this fix for now until we have a better solution.  Thanks everyone.

PR #1577, opened 2020-10-15
Beta fix: Player appearance backup in the movie studio

elliesec:
## Summary

Previously, the player's appearance would get backed up as soon as she loaded the movie studio, meaning that if she was wearing restraints that she subsequently removed whilst inside the studio, these would get re-applied on finishing the movie. This moves the appearance backup to just before starting the movie by adding a `Backup` parameter to the `MovieStudioChange` function.


ace-1331 (approved):

Ben987:
I'll simply add a CharacterRelease to fix that possible bug.  Thanks!

PR #1581, opened 2020-10-16
Futuristic Chastity Belt Update

Ada18980:
Adds 3 new features to the futuristic chastity belt:
- Shock the player for orgasm
- Shock the player for touching pelvis or butt
- Shock the player for reaching 60% struggle progress in another item

Being shocked while struggling will take the player back to the chat room (I added a new hook in Drawing.js to allow this to be set during dynamic item draw without crashing the game). I implemented the above using a dynasmic item script that polls every 950ms and checks based on the parameters of Dialog.js (for punishing struggling), and the current orgasm stage (for punishing orgasms)

Test Cases:
- Ensure shock only happens when removing or swapping items and not putting items on
- Ensure shock only happens in chatrooms
- Ensure chatroom message toggle works
- Ensure open/closed back functionality still works
- Ensure locking feature prevents people from editing
- Test with other players in room

**This has a dependency on my other PR (Futuristic Item Overhaul). This is because the chastity belt depends on files modified by the Futuristic Item Overhaul: https://github.com/Ben987/Bondage-College/pull/1580**


Ada18980:
I removed the punish speech feature as I feel like it's not ideal for the belt to have the item.

Also, I'd like feedback on whether the punish struggling feature should have a lower cooldown (5-10 seconds) or stay at a 30 second cooldown. I think some players would like to have the chastity belt make struggling out of anything impossible, but since it can be locked on it might be tedious. On the other hand, the maids will remove combination locks and allow the user to change the settings to stop punishing struggling.

PR #1584, opened 2020-10-17
Bugfix: Skin colour change fix

elliesec:
## Summary

This fixes a check that would prevent skin colour from being changed in the appearance screen so that skin colour is changeable again.


Ben987:
Thanks!  I will prepare a hotfix.

PR #1587, opened 2020-10-19
fix cloak

ace-1331:
- fixed not being able to kneel with the cloak


Ben987:
Thanks!  I will prepare a hotfix.  Cheers :)

PR #1588, opened 2020-10-19
Automatic Shock Collar/Unit

Ada18980:
EDIT: Adds a new Automatic Shock Unit and Automatic Shock Collar with four sensitivity modes in addition to the existing features:
 - Off: Same as regular shock unit
 - Low: Shocks the player if they say something in all caps
 - Med: Shocks the player if they say a sentence longer than 25 characters
 - High: Shocks the player if they say anything at all

The shocks should not trigger if the player uses an emote or uses OOC.


NoneNoname:
I think for this a type should be used with ExtendedItems, since I imagine there are few who do not want this option on themself and would like to block this or just make it as a separate new item

Ada18980:
I agree that some people would not want to allow this on themselves. I was considering making a futuristic version but I thought that people would like a non-futuristic version. 

That said, I'm not sure I want to make an owner-only version as I think it opens more RP opportunities when others can alter your toys and play with you in more ways, even when you're locked up in owner restraints. I recognize some people would want that option, though.

I'll put the features into an 'Automatic Shock Unit' most likely.

PR #1591, opened 2020-10-20
Bugfix: Various minor asset fixes

elliesec:
* Removes a duplicate `Fetish` property on fox ears 1
* Removes an additional `Fetish` property on the cow bell
* Removes a redundant `DefaultColor` property on the prison lockdown suit
* Corrects the audio on the rope toe tie


Ben987:
Yes!  Thanks a lot :)

PR #1598, opened 2020-10-21
New gagged self-actions

Ada18980:
Allows the player to use up to 5 new activities on themselves when wearing a gag:
 - Moaning into gag
 - Mumbling into gag
 - Screaming into gag
 - Whimpering into gag
 - Groaning into gag

Also, there is a new action available for other players who can kiss the player, to be used on gagged players:
 - Kiss on Gag

The activities have been tested with a second account and the self-actions cannot be used on other players, only on the self and only while gagged.


Ben987:
Sounds good, I will test it right now.

PR #1599, opened 2020-10-22
Preferences Homepage

Nina-1474:
The preferences icon in your Character Profile will now open to a 'homepage' dedicated to links to preference subscreens. This will give us the necessary space to add more buttons for new subscreens as we had run out of room using a vertical column under the Exit button. It also means we can display the page name for easier navigation. There's space for three columns of 7 buttons (one 1 column used so far).
The default page of settings has been moved into a 'General' subscreen. This allowed some tidy-up of the load & exit steps of the other screens.


Ben987:
Doing sub-screens sounds like a good idea.  Can you post a few screenshot to show the new look?  Thanks a lot.

Nina-1474:
What it currently looks like:

Preferences

An example with a second column, adding an additional menu in progress from #1432:

Preferences2

Ben987:
That looks great!  I'm sure it will help players a lot.  Thanks for that one.

PR #1600, opened 2020-10-22
Futuristic armbinder

Ada18980:
Added a new futuristic armbinder. It's an extended item which can be configured to either place the arms strictly behind the back or loosely (which toggles whether or not the armbinder is visible).

It has sound effects like standard futuristic items.


Ben987:
Awesome!  I will test it soon.  Cheers :)

PR #1601, opened 2020-10-22
Futuristic leg, arm, and ankle cuffs

Ada18980:
Added new futuristic versions of the leg, arm, and ankle cuffs, with futuristic sound effects and access denied screens (tested fully)


Ben987:
This is huge!  You're a machine.  Thanks for that one :)

PR #1603, opened 2020-10-22
Bugfix: Colour picker error

elliesec:
## Summary

This fixes an issue that would cause the colour picker callback to be called continuously due to a rounding error in the conversion between CSS RGB values and HSV values, which would result in errors on attempting to leave the colour picker when certain colour picker values were entered into the colour picker input.

The reason this happens is that the colour picker does an internal colour equality check between it's stored HSV value (which was converted from the input value) and the current input value. For some colours, converting from CSS --> HSV --> CSS would not return the exact same CSS colour due to rounding errors. For example, converting the CSS value `#3b1866` to HSV and back would return the CSS value `#3b1766`, causing the equality check to fail every time, meaning the colour picker callback would get called every frame.

This bugfix changes the colour picker to test against the last CSS colour entered into the picker directly, rather than passing it through HSV conversion and back before comparison, which should prevent this sort of issue occurring in the future.

## Steps to reproduce

1. Equip an item that only permits a single colour
2. Change the colour of that item in the colour picker
3. Enter the colour `#3b1866` (this is an intermittent bug that only occurs with certain CSS values, but this colour is known to reproduce the bug)
4. Note that the item colour does not update properly
5. Click the accept button
6. Note that the console now displays an error:
```
Uncaught TypeError: ItemColorState is null
    ItemColorOnPickerChange https://www.bondageprojects.com/college/R61/BondageClub/Screens/Character/ItemColor/ItemColor.js:199
    later https://www.bondageprojects.com/college/R61/BondageClub/Scripts/Common.js:384
    setTimeout handler*later https://www.bondageprojects.com/college/R61/BondageClub/Scripts/Common.js:381
```

### DIFF
--- a/BondageClub/Scripts/ColorPicker.js
+++ b/BondageClub/Scripts/ColorPicker.js
@@ -10,6 +10,7 @@
 
 var ColorPickerX, ColorPickerY, ColorPickerWidth, ColorPickerHeight;
 var ColorPickerInitialHSV, ColorPickerLastHSV, ColorPickerHSV, ColorPickerCallback, ColorPickerSourceElement;
+var ColorPickerCSS;
 
 var ColorPickerHueBarHeight = 40;
 var ColorPickerSVPanelGap = 20;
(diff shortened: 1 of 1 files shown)


Ben987:
I don't think I've seen this error before, but thanks a lot for the fix :)

PR #1604, opened 2020-10-22
Graphical fixes for the futuristic bra

Ada18980:
To make the band better surround the body at various sizes


Ben987:
Thanks a lot!  I will try it right now :)

PR #1606, opened 2020-10-23
Automatic Maids Change

Ada18980:
A fix/change to the maid system as follows:
 - If you ring the maid button, the maid will continue to come even if you free yourself before she arrives (the purpose of the maid button is to be able to call a maid even if the maid normally wouldn't come)
 - If you did not ring the maid button, but AutoMaids is turned on, and a maid starts coming but you free yourself first, the maid should stop coming.


Ben987:
Great!  I will test it soon.  Cheers :)

PR #1609, opened 2020-10-23
Help Kneel/Stand Fix

Nina-1474:
When a character had a self-applied pose in a chatroom, the dialog option to help them kneel would always make them stand even if they were already standing. 
This is because it still held the old assumptions where ActivePose being null was equivalent to standing and ActivePose being non-null was equivalent to kneeling. This has been updated.


Ben987:

PR #1610, opened 2020-10-23
Strait Dress & Chains fix

Nina-1474:
Chains on feet when wearing the Strait Dress will poke out at the sides. These will be hidden.


Ben987:
Excellent.

PR #1612, opened 2020-10-23
Succubus Heart Tail Fix

Shiranui-Izayoi:
With the help of RuiLove for the heart part.

Made it so there is no longer missing pixels in the tail heart. And made the tail pattern thicker so it can be seen. Previously the tail had missing or mostly transparent pixels see the image below. Fixed it so that is no longer the case.
unknown
With the fixed assets the tail looks like this
Desktop Screenshot 2020 10 23 - 13 47 33 18
And in game it looks like this
Desktop Screenshot 2020 10 23 - 13 51 11 02


Ben987:
Thanks!  I've never noticed that one before.

Shiranui-Izayoi:
No problem I only really noticed it when I turned the tail spade dark. Then asked about it and checked the files. Glad it could be fixed.

PR #1613, opened 2020-10-23
Inherit Colour improvement & Latex Hood Open Hair split

Nina-1474:
Changes:
- Tweaked the colour inheritance to allow an asset layer inherit colour *and* be colourable.
- Set HairBack to have a Default option that inherits the colour of HairFront. (ParentColor had to be swapped between the two to avoid half the NPCs having grey hair!).
- Split Latex Hood Open Hair into the hood and hair for multicolouring, with the hair copying the character hair colour by default.


Ben987:
Great Idea!  Thanks.

PR #1614, opened 2020-10-23
Block Category Hovertext fix

Nina-1474:
The latest changes to ChatSearchNormalDraw will have minor issues in a full page of rooms.
The initial setting of ListY needs to be aware of every text row to come so that it can start high enough to give them all space. At the moment it's based solely on Friends, so an additional Block row near the bottom will overflow too far down.

I've tweaked the logic a little bit so that:
1) The BlockCategory length is considered in ListHeight so that the starting ListY position is higher if necessary.
2) ListHeight properly handles the case where there's a Block row and no Friends.
2) ListY is incremented after each DrawTextWrap by one row height, to avoid messy calculations to determine each row's position.


Ben987:
Sure!  Thanks for that one also, you're on fire :)

PR #1617, opened 2020-10-26
Handheld Toy fixes

Nina-1474:
Changes:
- Fix to hide the handheld toys item when there are no individual toys to select (I reverted the previous fix).
- Making both the player and target character's toys available, to be more consistent with general item behaviour.
- Adjusting the preview image sizes as they were smaller than the standard size, making them appear slightly off-centre.

Note: There is still an existing bug where if the player doesn't own the crop, the default options appears as the actual first available option on the player's screen, but as the crop on other screens. I haven't figured out how to avoid this yet.


Ben987:
This is a much better fix.  Thanks a lot.  Let me know if you find a solution for the crop problem, no rush :)

PR #1618, opened 2020-10-26
Changed references in maid disabling script from CurrentTime to CommonTime()

Ada18980:
Changed references to CurrentTime to CommonTime() as it is more reliable.

Also added 100 seconds to the bug detection timer in Mainhall.js. This timer is meant to protect against people being locked out of maids for far longer than necessary. It shouldn't ever happen unless there is a bug, and is mainly meant to protect against being indefinitely bound. Adding 100 seconds should not do anything, but it might stop a rare bug I encountered that I am unable to repro, in which the bug notice appeared but disappeared after I called the maids again.


Ben987:
Maybe you should keep CurrentTime and not use CommonTime()

CurrentTime is server based so it won't be affected if you change device and cannot be cheated out if you change your computer clock.  It can get out of sync for 30 seconds if you keep your browser out of focus but it should not matter much.

Ada18980:
If that's the case, we can close this PR.

PR #1620, opened 2020-10-27
OOC chat and whisper garbling

jomshir98:
Adds `/ooc` command to speak out of character (in whispers too)
Also triggers when message starts with "(" and there are no other parentheses in the message

Example on the [Discord](https://discord.com/channels/554377975714414605/554378904308023297/770780604512862248)

This also changes whispers to be garbled, if you are gagged

Implementation details:
The `/ooc` is sent as part of the normal message
Recipient detects `/ooc` and strips it (no change needed on the server)
The command does 2 things: prepends sender's name with `[OOC] ` and skips garbling the message


jomshir98:
Unresolved questions:

- Should this change whispers to be garbled? 
- Should it trigger on "(", or be just a command?

Ben987:
Yes, a OOC tag could be useful.  Please don't add "OOC" directly in the code though, add a text from the CSV file so translators can have OOC in French, German, etc.

For the garbled whisper, I'll do a quick survey on Discord first to see if players want it or not.

Thanks!

Nina-1474:
FYI: For work-in-progress pieces, there's a 'Convert to draft' option under the Reviewers section.

PR #1621, opened 2020-10-28
Latex Elbow Gloves

EmilyR42:
- Added Latex Elbow Gloves item


Ben987:
Thanks!  But you seem to be pushing more than only gloves, there's a background and a foam sword.  Make sure you only add the gloves with your pull request.  And make sure there's no conflicts with the current master branch.  Thanks a lot!

PR #1623, opened 2020-10-28
Vibrators Edged Fix

Nina-1474:
The Edge feature of TapedClitEgg wasn't working for two reasons:
- AllowEffect wasn't updated to include 'Edged' when the new vibrator modes were added to two items.
- The IsEdged check only looked at items belonging to groups which allowed orgasms. It instead needs to check the settings of vibrating items' ArousalZone group, which usually defaults to the item's group but could be different. In this case, TapedClitEgg is in `ItemVulvaPiercings` which does not have arousal settings, but it has an ArousalZone of `ItemVulva` which does.


jomshir98:
I don't think that items should have `Edged` effect: It would make them edge on every mode.
Last I checked the `Edged` effect was applied correctly, by [this code](https://github.com/Ben987/Bondage-College/blob/25543e5308f3638ba59bf6f4850b9e06d2812949/BondageClub/Scripts/VibratorMode.js#L139)

Nina-1474:
You would be quite right if I added it to `Effect`, but I added it to `AllowEffect`. That list is used to validate the item on the target's side and remove invalid effects to prevent griefing.

Ben987:
Thanks a lot for that fix!  Cheers :)

PR #1624, opened 2020-10-29
Spreading Items Kneel Fix

Nina-1474:
The Metal Leg Spreader and Wooden Cuffs in ItemFeet both had issues when the target is kneeling.
If the character was kneeling but not forced into the position (e.g. by collar & short chain), then the NotKneeling prerequisite passed and the item could be applied. However the item could not be removed again because it had an AllowedActivePose of Kneel which failed the NotKneeling prerequisite.
The AllowedActivePose setting has been removed. This does mean that applying the item forces the character to stand up, but that's at least consistent with other items that have the NotKneeling prerequisite.


Ben987:
Awesome!  Thanks :)

PR #1625, opened 2020-10-29
Blocked Handheld Toy - Inventory Order fix

Nina-1474:
When a player tries to use a handheld toy on a target who has blocked that specific toy, the item still appears as the first option in the group with a white background, as if usable. Clicking on it does nothing.
A fix has been put in so that the item now moves to the end of the list highlighted red, the same as other items.


Ben987:
Very nice, thanks again.

PR #1626, opened 2020-10-29
Leather Neck Items Split

Ayesha-678:
Splits and grey-scales the leather neck items.

**Items Split**
Leather Collar: Collar, Ring
Leather Posture Collar: Collar, Left Ring, Middle Ring, Right Ring
Leather Choker: Leather, Metal


Ben987:
I love it!  Thanks a lot :)

PR #1627, opened 2020-10-30
Immersion Settings

Ada18980:
Moves Sensory Deprivation, Keep restraints while relogging, and disabling inspection while blind, to a new Immersion menu (the icon for this one already seemed to exist, thank you whoever made it!).

This new menu contains an option (default: false) which, if set to true, prevents all the other items in the menu from being changed if the player is restrained (or blind).

Also includes two new settings:
 - A new level of sensory deprivation that prevents whispers
 - Block the player from sending OOC while gagged (can still send emotes) 

Safeword is separate since it is a pretty important feature and remains in the General settings.

All features currently default to false, and the feature to disable the menu while bound is independent of any other features so players can still choose the same options and settings as before.

There is also nothing to block emotes since emotes are a critical part of roleplay.


Ben987:
Thanks!  This one creates a few conflicts with other pull request, could you review and test?  Cheers.

Ada18980:
Merged, though Vibe VFX still is on the general screen so I need to take care of that.

Ada18980:
Tested and ready for review.

Ben987:
Very good!  This could all be used in a future "hardcore" mode.  Thanks a lot :)

PR #1628, opened 2020-10-30
New actions

Ada18980:
Adds wiggling, struggling, and nodding as actions one can perform on self. 
Also adds "Sit" as an action you can do to sit in someone's lap if they are kneeling


Ben987:
These new actions sound fun, thanks a lot :)

PR #1630, opened 2020-10-31
Latex Elbow Gloves (PR fix)

EmilyR42:
- Added (ONLY) the Latex Elbow Gloves item.


Ben987:
Thank you very much.  I will test it soon.  Cheers!

PR #1633, opened 2020-11-04
Bugfix: Fix for vibrators removing locks from an item

elliesec:
## Summary

This fixes an issue where a lockable item that also uses the new vibrator mode script could have locks removed by updating the vibrator mode (or via automatic vibrator updates). This happened because the `Effect` array was being overwritten, effectively removing any lock on the item. This is currently only an issue with the mermaid tail as far as I'm aware, as that is the only lockable item that uses the new vibrator script.

## Steps to reproduce

1. Equip a mermaid tail
2. Lock the mermaid tail
3. Change the vibrator mode on the mermaid tail
4. Note that the lock has been removed


Ben987:
Thanks!  That was a tricky one.  Cheers :)

PR #1634, opened 2020-11-04
Extended Items Lock fix

Nina-1474:
A small fix to #1449 to prevent an error when a lockable extended item doesn't have an effect. To see this apply Leather Ankle Cuffs on feet and try to change the option.
I also removed an unneeded C declaration that could actually cause issues for extended clothes.


Ben987:
Good catch!  Thanks for that one.

PR #1635, opened 2020-11-05
Sensory deprivation setting: Light

jomshir98:
While being blind is great, seeing your character blindfolded can be as well.

This PR adds new sensory deprivation setting level: Light
Effects of this setting are:
- `BlindHeavy` effect is treated as `BlindNormal`
- Character's names are not hidden

Technical details:
- Adds `Character.GetBlindLevel` function, that works like `Character.GetDeafLevel`


Ben987:
Thanks!  If a player doesn't change any settings and keeps it as default, does she gets the same experience as in the previous versions?

jomshir98:
Yes, unless I made an error, there should be nothing changed. Also the default mode is still "Normal"

PR #1636, opened 2020-11-07
Feature: Changelog Update

elliesec:
## Summary

This updates the changelog with all the changes since the last changelog update. I've also written a short Node script that converts the changelog into an HTML file, which should make it easier to read than passing a link to a text file. We could in the future attach this to a Github action to automatically update the HTML whenever the changelog changes, but this works well enough for now.


Ben987:
Thanks!  I will test it right now :)

PR #1637, opened 2020-11-07
Unlocked Gag Garble fix

Nina-1474:
This is a part 2 for the fix from #1274. When a non-extended item gag is locked then unlocked, it no longer garbles text. 
The original PR restored the garbling while the gag was still locked, but not when it was unlocked again.
Now Property.Effect, which will always be non-garbling for non-extended gags, will only override Asset.Effect if Type is undefined, which ensures this will only happen for extended gags.


elliesec (approved):

Nina-1474:
Reverted the fix from #1334 since it's now covered by this and since the old fix would affected any extended gag types that remove the gag effect (none in the game yet).

Ben987:
Nice!  Thanks a lot.

PR #1638, opened 2020-11-07
GP9 Gas Mask

Nina-1474:
A new gas mask in the hoods group.
Credit for this change goes to **EmilyR**. I'm just checking this in on her behalf.

PR #1639, opened 2020-11-08
GP9 Gas Mask Lock

Nina-1474:
A non-empty lock image for the gas mask from #1638.


Ben987:

PR #1640, opened 2020-11-08
Latex Room & Open Face Hood

EmilyR42:
- Added Latex Room background for both regular and asylum chat rooms
- Added Open Face hood as both a hood restraint and cosmetic mask


Ben987:
Right on time for the next beta!  Thanks a lot :)

PR #1642, opened 2020-11-09
Self-Bondage Difficulty for Extended Types

Nina-1474:
When an extended item type has no SelfBondageLevel setting, it defaults to 0. This allows some types to be easier to apply to oneself than to someone else which isn't very consistent. The ItemArms HempRope options being a prime example.

Now the self-bondage type difficulty copies the Asset's overall self-bondage setting or the type's BondageLevel, whichever is higher, when a SelfBondageLevel has not been set.

Also includes a fix to the BondageBench settings.


Ben987:
Thanks for the fix!  We'll test it in the second beta.  Cheers :)

PR #1643, opened 2020-11-10
College Uniform Split

Nina-1474:
Split the college uniform into Shirt, Tie and Skirt for multi-colouring.


Ben987:
Thanks!  That one has been on my list for a while.  

But since this outfit is given in some parts of the code, shouldn't we change that code to give both the upper and lower part?  Same goes when you import a Bondage College character.

Nina-1474:
What's the 'lower part' exactly? Is it a separate asset? CollegeOutfit1 is still a single asset in the Cloth group, so the InventoryWear references to it in code will still give the same result.

Ben987:
My mistake!  I first thought you were separating the shirt from the skirt to do outfits mixes.  This is great, thanks :)

PR #1644, opened 2020-11-10
Striped Socks Split

Nina-1474:
Split the striped socks to allow different colours for the alternating stripes.


Ben987:
Awesome!  Thanks for that one :)

PR #1646, opened 2020-11-10
Latex Elbow Gloves - Graphical tweak

EmilyR42:
- Cropped default position Latex Elbow Gloves to clash less with clothing items


Ben987:
Thanks for that fix.  Cheers :)

PR #1647, opened 2020-11-10
Bugfix: Vibrator mode error

elliesec:
## Summary

This fixes an issue where the vibrator mode script would cause an error when opening the vibrator menu on several of the vibrators using the `VibratorMode` script while not locked.

This is the stack trace of the error in question, as reported on Discord:

Stack trace


Ben987:
Great!  Thanks for the quick fix.

PR #1648, opened 2020-11-10
Fix: Automatic OOC whisper history

jomshir98:
Fixes bug causing gag auto punishment to trigger


Ben987:
Thanks for the quick fix :)

PR #1649, opened 2020-11-10
Chat Search - Game Label fix

Nina-1474:
Similar to part 1 of #1614, this considers the Game label in the chatroom button popup's start position to avoid going too far down when near the bottom.


Ben987:
Sure!  Good idea :)

PR #1650, opened 2020-11-11
Auto gag fixes

Ada18980:
Changes the automatic gags (shock unit and futuristic) to not trigger on messages with no numbers all the time
Also snuck in a change to have it trigger on using exclamation points, since not everyone uses all caps


Ben987:
Alright, thanks!

PR #1651, opened 2020-11-11
Pilot suit no longer hides butt

Ada18980:
To prevent it from hiding tails


Ben987:
Good catch, thanks a lot.

PR #1653, opened 2020-11-11
Futuristic Armbinder Lock fix

Nina-1474:
The lock images weren't appearing due to a lowercase 'L' in the filenames.


Ben987:
Good catch.  Thanks for these fixes.

PR #1655, opened 2020-11-11
Hidden Item Hiding fix

Nina-1474:
A long standing bug is hidden items still being able to hide other items, causing graphical oddities. Two examples:
- Latex Strait Leotard > hides Wedding Dress > hides BodyLower => results in the lower legs disappearing
- Latex Blindfold > hides Open Face Hood (mask version) > hides HairFront => results in the hair disappearing

The visible check has now been extended to only let another item's hide properties take effect if that item is itself visible. 
A safeguard has been included to prevent the (unlikely) possibility of two items somehow hiding each other and getting stuck in an infinite loop.

### DIFF
--- a/BondageClub/Screens/Character/Appearance/Appearance.js
+++ b/BondageClub/Screens/Character/Appearance/Appearance.js
@@ -336,13 +336,14 @@ function CharacterAppearanceSortLayers(C) {
 }
 
 /**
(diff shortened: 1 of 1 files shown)


Ben987:
Thanks for that extra fix.  Cheers!

PR #1656, opened 2020-11-12
Box Tie Armbinder Alpha Fix

Nina-1474:
Tweaking the alpha to avoid cutting off clothes over larger boobs, the same exact fix as the Bolero from #1611.


Ben987:

PR #1657, opened 2020-11-13
Bugfix: Locks blocking type switching

elliesec:
## Summary

As an unintentional result of #1449, it is not possible to change the type of items that are locked without unlocking them unless they explicitly set `Property.AllowLock = true`. After discussing with @Nina-1474, it seems like more often than not, this doesn't make sense as a default behaviour (for example, the plug gags, latex leotards, inflatable body bag, etc.). This change therefore reverses that default behaviour so that by default `Property.AllowLock` is presumed to be `true` unless it is explicitly set to false. This maintains the behaviour of the bondage bench as defined in #1449 but reverts any other extended item back to the way it was before that change. 

There may be some items for which this behaviour makes sense, but I'm not going to change any items now. I will work through the asset catalogue and make these changes where it seems appropriate in a separate PR, which I'd probably rather _not_ have merged before the release, as these would be item changes rather than fixes.

## Steps to reproduce

1. Equip a plug gag/latex leotard/inflatable body bag
2. Lock the item
3. Note that when entering the extended item screen for the item, you cannot plug/polish/inflate the item

### DIFF
--- a/BondageClub/Screens/Inventory/ItemDevices/BondageBench/BondageBench.js
+++ b/BondageClub/Screens/Inventory/ItemDevices/BondageBench/BondageBench.js
@@ -6,6 +6,7 @@ var InventoryItemDevicesBondageBenchOptions = [
 		Property: {
 			Type: null,
 			Difficulty: 0,
+			AllowLock: false,
 			SetPose: ["LegsClosed"],
 			Effect: ["Mounted"],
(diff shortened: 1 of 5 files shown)


Ben987:
Thanks for that fix!  Yeah, lots of changes have unexpected side effects.

PR #1658, opened 2020-11-13
Clamshell Hogtied Images fix

Nina-1474:
The Hogtied versions of the Clamshell bra were using placeholder images.


Ben987:
Good one!  Thanks.

PR #1659, opened 2020-11-13
Large Asian Overhead fix

Nina-1474:
The OverTheHead pose for the Large Asian BodyUpper image had a small gap at the bottom which broke it away from the lower body. This is now filled.


Ben987:
Awesome!  Thanks again.

PR #1660, opened 2020-11-13
New backhairstyles

Buizel333:
New backhairstyles done by merging some already existing styles into new styles.


Ben987:
Thanks for the hair mix and match, I hope it won't be too long to find a look using the back and forward buttons.  Cheers.

PR #1662, opened 2020-11-13
Futuristic Collar 'Copy Colors' Feature

Ada18980:
The futuristic collar can now copy its own colors (display, band, and mesh) to other futuristic items with a new button interface.

This is a new feature, so it might make sense to keep this out of the second beta.


Ben987:

PR #1663, opened 2020-11-14
Password Lock

Ada18980:
Adds a new lock, similar to the combination lock. However, instead of being limited to only numbers, it is limited to only 6 letter passwords made of letters, and the user can set a hint. Once a password is set, it cannot be changed.

Additionally, futuristic items will use the password entered by players and can be unlocked if the password entered in the field is the same.

Also changes the main hall maid to remove the lock from chastity belts.


BondageClub/Scripts/Server.js
```diff
@@ -391,6 +409,9 @@ function ServerAppearanceLoadFromBundle(C, AssetFamily, Bundle, SourceMemberNumb
 							if (C.Appearance[A].Property.LockedBy != null) NA.Property.LockedBy = C.Appearance[A].Property.LockedBy;
 							if (C.Appearance[A].Property.LockMemberNumber != null) NA.Property.LockMemberNumber = C.Appearance[A].Property.LockMemberNumber; else delete NA.Property.LockMemberNumber;
 							if (C.Appearance[A].Property.CombinationNumber != null) NA.Property.CombinationNumber = C.Appearance[A].Property.CombinationNumber; else delete NA.Property.CombinationNumber;
+							if (C.Appearance[A].Property.Password != null) NA.Property.CombinationNumber = C.Appearance[A].Property.Password; else delete NA.Property.Password;
```
Nina-1474, line 412: I'm guessing this shouldn't be `NA.Property.CombinationNumber` every time.

Nina-1474 (dismissed):

BondageClub/Scripts/Server.js
Ada18980: Whoops

Nina-1474: The line I highlighted was just one of them. There's another few farther down.

Ada18980: Fixed! Thanks >.<

Ben987:
Alright!  Time to test this one, thanks a lot for that new lock :)

PR #1664, opened 2020-11-14
Bugfix: Issue with non-lockable items

elliesec:
## Summary

This is a follow-up fix to #1657, which broke the adding/removing of non-lockable items. I've tested this one pretty thoroughly now and I'm pretty sure it covers all bases.

## Steps to reproduce

1. Character A applies a non-lockable item (eg. duct tape) to player B
2. When the progress bar is full, the game crashes, and an error is displayed in the console (see below)

Stack trace


Ben987:
Thanks for the new fix!  Let's hope this is the good one :)

PR #1665, opened 2020-11-15
Futuristic Armbinder Layer Names

Nina-1474:
The layer names for the Futuristic Armbinder were missing.


Ben987:

PR #1666, opened 2020-11-16
Improvement: DOM buttons added to the information sheet

SandrineBC:
# Summary
To improve and further standardize the behavior of the BC GUI, DOM buttons are introduced and added to the information sheet. With this, buttons can be styled via a centralized CSS file and things like changing the color or tool tips don't need to be added via the common draw function within the application.
image

# Details
## What it does
Three files are changed with this PR:
### `CSS/Styles.css`
Behavior definition for the new buttons and the tool tips
### `Elements.js`
A new function is added for the creation of the buttons. Within this function the position of the tool tips is defined.
### `InformationSheet.js`
The new buttons are added in the `Run` function instead of the `Draw` function, which should reduce processor usage

# To be done
If this PR is approved, DOM buttons should be used in the various screens. What probably means further adjustment to the styles and the position of the tool tips


Ben987:
Having a CSS to control all button styles is a good idea, but are you sure this is worth it?  It's a lot of work for no gain for players.  

And there's a few drawbacks with using this: 
- browser compatibility
- extensions for the visually impaired will mess the game, it already does it for text boxes
- if the button gets focus, the keyboard shortcuts might not work
- we'll need to add code to remove the buttons everywhere
- the back-next buttons will need to stay as they are now so we'll have double-standards
- the buttons that changes color based on conditions won't work either

And all the possible bugs that might be created by this, with no real player side gain.  

Personally I would not that change this.  Thanks!  Let me know what you think.

elliesec:
I think the ability to use DOM buttons brings several benefits:
* Centrally controlled button styling using CSS
* The ability to tie button handler callbacks directly to the buttons (no more issues where someone moved a button in the draw function, but forgot to modify the click function)
* Removes tooltip drawing issues (won't have buttons or other UI elements drawn on top of tooltips any more)
* Some small performance improvements from not having to draw buttons every frame

That said, I do think Ben's concerns are valid, but I also believe they can be worked around/mitigated:
* Browser compatibility shouldn't be an issue - all browsers support DOM buttons
* Extensions for the visually impaired modifying the elements will happen. But I think that's fine - that's what they're there for - to resize and re-color UI elements. If a player has a visual impairment, forcing them to view the game in a fixed way is arguably worse.
* The keyboard focus is a valid point. We can work around it easily enough by attaching listeners to the window rather than the canvas - but this is probably something we'd want to make sure is working before this PR is merged.
* Adding the code to remove the buttons is a slight overhead in terms of code size, but I think the benefits outweigh this
* We could quite easily implement back/next buttons in DOM with a bit of CSS to style them correctly
* Colour conditions on buttons could still be implemented as they are now

SandrineBC:
Can you tell me, where we have keyboard shortcuts?
I'll have a look at them and add the handling to this PR.

elliesec:
I've taken a look at this now, and as far as I can see, there are no issues with the buttons blocking key event handlers right now, provided the button callbacks don't explicitly stop event propagation. If this becomes an issue in the future (but I don't see that it should), the best plan of action would be to rework the global keydown handlers, because it is entirely possible to have the button handlers and global key event handlers working together.

I've made a PR into the branch with a handful of suggested refinements, but overall, I'm pretty happy that this works as expected (with the added bonus of allowing keyboard-only navigation to screens with DOM buttons implemented).

BondageClub/Screens/Character/InformationSheet/InformationSheet.js
```diff
@@ -176,6 +203,9 @@ function InformationSheetSecondScreenRun() {
  * @returns {void} - Nothing
  */
 function InformationSheetClick() {
+/* 	var C = InformationSheetSelection;
```
elliesec, line 206: Probably want to delete the contents of this function now.

elliesec (changes requested):
Just the one comment on this. Otherwise, it all looks good to me (and I've tested and everything works great).

Incidentally, I ran a performance profiler over the club yesterday, and `DrawButton` is actually fairly high up in the list of places where processor time is spent, so I think if DOM buttons become more widespread, it could have a pretty positive impact on the game's performance.

elliesec (approved):
All looks good to me now!

elliesec:
@Ben987 - Once we're out of the hotfix period, I think this is worth merging. Aside from the fact that the DOM buttons have a beneficial effect on performance, they also allow players to navigate the game with the keyboard (well, just the information sheet right now), and they'd allow us to add tooltips to buttons where we currently can't (a prime example is the color picker, where I'm frequently confused by which button is copy and which is paste, but a tooltip with the current buttons would get rendered under the input element).

elliesec:
Just thought I'd add some more concrete examples of where this would be useful:

I've tried adding tooltips to the color picker buttons, but haven't submitted that, because of this:
tooltips

Additionally, I'm currently working on a library room, where players can read books and view the game changelog inside the game - DOM buttons there would allow me to avoid this:
tooltips

PR #1669, opened 2020-11-16
Slow player from leaving indoor rooms w/ hands covered

Ada18980:
Players that are not immune to slow will now have to try the doorknob potentially multiple times before escaping a room with an indoor background.

Clicking the leave button while wearing one of these items will give a public message saying that you cannot get the door open; the message changes (there are four messages total with increasing futility). After 3 attempts, the player will be able to leave.

Also, this only applies to rooms with 'indoor' backgrounds, determined using background tags.

HandsCovered is a new effect I have manually added to ItemArms and ItemLegs items and the inflatable body bag. I did not put it on boxes or cases because I want to leave open the possibility of using HandsCovered for things like lockpicking for a future escape minigame (idea being that if your hands are covered you can't hold a lockpick and can't pick locks, but if you are in a box and have a lockpick then you can free yourself from restraints you are wearing)


BondageClub/Screens/Online/ChatRoom/ChatRoom.js
```diff
@@ -20,7 +20,81 @@ var ChatRoomStruggleAssistBonus = 0;
...
 var ChatRoomSlowtimer = 0;
 var ChatRoomSlowStop = false;
-
+var ChatRoomDoorTimer = 0;
+var ChatRoomDoorfailtimer = 0;
+var ChatRoomDorFailLevel = 0;
+
+var ChatRoomLastBackground = ""
+var ChatRoomBackgroundIndoors = false
+var ChatRoomIndoorKeywords = [
```
Nina-1474, line 29: This would be better as a new property on BackgroundsList in Background.js rather than a seperate list in a separate file that would be easy to forget about when adding new backgrounds.

Ada18980:
I feel silly not realizing that we already had an indoors tag. Thanks for that!

Ada18980:
@Ben987 I have integrated this into the new difficulty system. Have a look :)

PR #1670, opened 2020-11-16
Immersion Option: Return to chat room on relog

Ada18980:
Adds an immersion setting which returns the player to the chat room if they disconnected from the chat room and log back into the main hall. This does not allow players to join a banned room, and does not allow players to instantly rejoin a room upon being kicked.

I've placed it in the Immersion options because if a player is restrained, they are stuck in that room (which is an intended side effect). However, this feature is also useful for anyone who doesn't lock themselves out of the immersion menu when bound, for convenience reasons.

There is a second option for recreating the room if it doesn't exist. It adds the player to the admins if they are not already, in order to allow the player to make the room public (the player could otherwise make themselves admin by making an alt account and remaking the room first, then rejoining on their main, and we dont want to encourage that exploit.

The player, if trapped for more than 30 minutes, can call the maids for help, but the maids will restrain the player quite strictly afterward. They should not come if the maids are disabled.


Nina-1474:
A few things that stand out on first look:
- This is adding a lot of ServerSend calls. I don't know how "heavy" a call AccountUpdate is but I'd wonder about the increase in server traffic from merely room hopping. Integrating it into join/admin/leave calls might be better though that would involve server code changes.
- Does this work for Asylum and LARP rooms as well?
- What happens when an admin locks a room and leaves? Does everyone have to either disable the setting or safeword out? If so is that just an accepted risk of the setting?
- LastChatRoom being a member of ImmersionSettings looks a bit odd, since it's a constantly updated value rather than a relatively static setting the player controls.
- The same 5 lines to update the account are repeated a few times. These could be wrapped up into a function.

Ada18980:
> A few things that stand out on first look:
> 
> * This is adding a lot of ServerSend calls. I don't know how "heavy" a call AccountUpdate is but I'd wonder about the increase in server traffic from merely room hopping. Integrating it into chat room join/admin/leave calls might be better though that would involve server code changes.
> * Does this work for Asylum and LARP rooms as well?
> * What happens when an admin locks a room and leaves? Does everyone have to either disable the setting or safeword out? If so is that just an accepted risk of the setting?
> * LastChatRoom being a member of ImmersionSettings looks a bit odd, since it's a constantly updated value rather than a relatively static setting the player controls.
> * The same 5 lines to update the account are repeated a few times. These could be wrapped up into a function.

Re: Serversend calls: Yes, it is sent a lot. It adds an extra serversend whenever leaving a chat room and whenever getting kicked from a chat room, as well as a serversend call when you log in and it can't find the room. I think the most common case this is adding is the serversend when leaving a chat room--I don't know how much load this will cause but to put it into perspective, opening and closing the options page causes a serversend.

Re: Asylum and LARP: It should not work for LARP to avoid breaking games, and for asylum I do not believe it will work because it adds code to the mainhall rather than the asylum room. Do you think it should work for asylum?

Re: Locked room: Since you cannot join locked rooms from the regular char room search, you won't be pulled back into a locked room. Only a private room.

Re: LastChatRoom, it doesn't have to be a part of immersionsettings, I'll go and change that.

Re: Function, Will do.

Ada18980:
Pushed a new commit. Now:
1) Works for asylum bedroom as well
2) ChatRoomSetLastChatRoom takes over the 5 lines
3) LastChatRoom removed from immersionsettings.

Nina-1474:
> to put it into perspective, opening and closing the options page causes a serversend.
I did know this but the difference is players don't hop in and out of preferences often whereas they do with chat rooms. It hopefully won't have any noticeable impact anyway. Just something to keep an eye out for.

Will the automatic room creation happen on relogging as well? Just wondering what might happen after a server crash with a full room of people try to reconnect at the same time.

Ada18980:
> Will the automatic room creation happen on relogging as well? Just wondering what might happen after a server crash with a full room of people try to reconnect at the same time.

I'm currently working on the feature; probably yes, but I'm open to having it be taken out for stability reasons. The automatic creation only happens once on login and if the room was already created, it will auto-join. 

In the latest version, auto-recreating the room is an additional option so it's not mandatory.

Ada18980:
Okay. I think the admin setting feature is fixed. It waits until there is chat room data to copy before setting the chat room parameters, which is probably safer against server lag.

Ada18980:
I renamed kicking to having the admin "call the maids to escort [PLAYER] out." I think this makes sense to include with this feature because
1) The only way to get out of a room currently if you are not the admin is to have an admin kick you out, or wait for it to be disbanded
2) If a player pleads for you to help them out but they are in a cage, it makes sense to call maids to lift the cage out. This should improve transparency when a player wants out and asks someone to kick them.

Im also removing the extra conditions (safeword or maids disabled) on recreating rooms because the player is now added to the admin list.

Ada18980:
Added a new feature to call the maids to help out if you are trapped for 30 mins or more in the same room. The maids will free your legs but they will punish you quite strictly afterward!

Ben987:
I'm not sure if this one will create too many questions or problems.

Let's wait until hardcore mode is ready, maybe there's a way to integrate that change in it.

Thanks!

Ada18980:
> I'm not sure if this one will create too many questions or problems.
> 
> Let's wait until hardcore mode is ready, maybe there's a way to integrate that change in it.
> 
> Thanks!

Sounds good!

PR #1671, opened 2020-11-17
if OOC disabled, also block whispers while gagged

Ada18980:
Short fix that makes it so that whispers are blocked if you have disabled OOC while gagged. This is because garbled whispers have been reverted


Ben987:
Yep!  That makes sense.

PR #1672, opened 2020-11-17
Slow down the player when in LegsClosed or Kneeling pose (if player can otherwise still leave). Blocked by slow immunity

Ada18980:
This adds an option in the immersion menu to slow the player down in more cases. If disabled, it leaves everything as-is according to the current behavior.

The implementation basically checks if the player's pose is LegsClosed or Kneel, and applies the IsSlowed effect if so.


Ada18980:
@Ben987 Changed it to work with slow immunity. Testing with some new slave accounts~

Ben987:
I've just tested your code and it works great, thanks!  Check commit: https://github.com/Ben987/Bondage-College/commit/d779b8247122e08d86f0ce04240ac07576194f2d

PR #1674, opened 2020-11-17
Futuristic Collar now locks itself

Ada18980:
When you click "lock" it will lock the futuristic collar as well with the selected type of lock. unlock functionality is unchanged as you can always take the collar off manually, and most doms might prefer to keep the collar owner-locked permanently.


Ben987:
Alright!  Thanks.

PR #1675, opened 2020-11-18
Feature: Ability to pick locks

Ada18980:
Oh boy, lots of changes in this one.
1) Instead of Inspect Lock and Unlock being in the menu at the same time, there is a "Lock Menu" which contains 3 options: Inspect, Pick, and Unlock
2) Lock picking minigame was added. it is very similar to the minigame for picking locks from The Elder Scrolls: Oblivion. Higher challenge means fewer tries to find the right pins. It's mostly random and low-skill but there is some memory element as having a higher lock picking skill will make the later pins go higher in the cylinder and earlier pins will stay lower when clicked. Challenge increases based on lock time and also how bound the lock-picker is, but if her arms are covered (like, say, an armbinder) then it is impossible to even attempt to pick locks.
3) Owner, Lover, Mistress locks are not pickable. The user can try, but there will be several pins that are impossible to set no matter what, even if the user gets extremely lucky.
4) Some items have a "NotSelfPickable" effect such as yokes. It doesn't make sense to be able to pick these items because your hands are separated from the locking mechanism. You will need help to get out of these.
5) Difficulty scales with being bound:
+2 for arms bound
+50 if the item is NotSelfPickable
+50 if the player's hands are blocked (tape, mittens, etc) However, if you have keys to the lock in question, it only adds 4 difficulty, along with 2 if your arms are also bound. This means that skilled lockpickers can turn the key in the easier locks.
+1 if gagged (but won't make it impossible to pick)
+1 if legs are bound (but won't make it impossible to pick)
+1 if feet are bound (but won't make it impossible to pick)
+1 additionally if feet are covered by something like a legbinder or straitdress (but won't make it impossible to pick)
The last few penalties are mainly there to increase the difficulty of picking while trussed up excessively; it should be a real challenge since you can barely maneuver or shift your position to reach locks.
6) You unlock the ability to pick the following locks (assuming 0 bondage level on the locked item):
1: Intricate
3: Exclusive
4: High Security
5: Combination
Bondage skill effectively subtracts 5 Lockpicking skill at Bondage 10, so a Bondage 10 combination lock requires Lockpicking 10, a Bondage 10 Intricate lock requires Lockpicking 6, and so on.
You can't pick locks if your hands are covered (e.g. by an armbinder)
7) Intricate locks can now only be opened by the person who locked them on. There is also a new High Security padlock.  It can only be opened by the person who locked it on, and can be picked with a lock picking skill of 4 or higher. The person who locked it on can also lend the key to others by writing others' member numbers in the lock's settings, allowing others to have access to it and be able to unlock it. There are switches to allow you to automatically append your owner, lovers, and whitelist to the key list when you save.
8) You can now give away your keys for the high security padlock and intricate padlock as a player interaction. 
9) Lock picking requires a lockpicks item. You can give lockpicks to a player if they are locked, allowing them to pick locks even if they dont have the item. This lasts until the player relogs or leaves the chat room, but owners can still take it away by confiscating keys. Owners cant prevent the sub from having lockpicks lent to them, however.
If you fail a lock with the lent lockpick, the pick breaks and you need to borrow another one.


Ben987:
That's a huge one, could we introduce the lock picking mini game only to the intricate lock first?  We'll see if it works well and if players like it.  If it's fun, we could expand it to more locks.  Thanks!

Ada18980:
> That's a huge one, could we introduce the lock picking mini game only to the intricate lock first? We'll see if it works well and if players like it. If it's fun, we could expand it to more locks. Thanks!

Thats easy, just remove Pickable and PickDifficulty from the other locks. I can do it when I get back home

Ada18980:
@Ben987 There, I removed it from the other locks and made a few small bugfixes. Also it's merged with master repo, the git merge was really confusing but it was mainly the dialog change you added before.

SandrineBC:
I love that feature. Could you increase the difficulty, if the player is extremely aroused? Picking a lock with shaking hands and so on?

Ada18980:
@SandrineBC Yes! I went and added something to make picking locks more kinky. It doesnt increase difficulty directly, but it does cause pins to reset so you're on a sort of time limit

Ben987:
I'd like to test it for the next version but there are two conflicts left.  Can you resolve them and test again?  Thanks a lot.

Ada18980:
@Ben987  Fixed!

Ben987:
This is huge!  Thanks a lot, it's been on my list for a long time.  Cheers :)

PR #1677, opened 2020-11-19
Offline friends&submissives

jomshir98:
~~Depends on [Bondage-Club-Server#55](https://github.com/Ben987/Bondage-Club-Server/pull/55)~~

Changes the friendlist to display offline friends and submissives.

Friends are sorted by being online and than by name
Friends have type: "Owner", "Lover", "Submissive" or "Friend" - first one applies
On the delete tab: Friend type is shown instead of room.
Friendlist refreshes every 5s while open

Info is collected when friends are online


Ben987:
Instead of changing the server for that one, could we keep a dictionary on the client side to map the MemberNumber with the character name?  That dictionary of character name could increase or decrease each time a player is added or removed from the friend list.  And could be updated when you meet that person in a chat room.

Changing the server is always tricky and created many crashes in the past, so I'm very cautious.   And the less load we add on the server side, the faster it goes.  Thanks a lot.

jomshir98:
@Ben987 Done.
I changed the code not to rely on server, the information will be collected when friend is seen online.
Note that current solution doesn't check situation when friend removed us; the friend is removed only after we remove them from our list.

Ben987:
Thanks for this!  I love that you're compressing the strings.  I will test it soon.

PR #1678, opened 2020-11-19
Ab dress1 colors

wildsj:
Added layers to the puffy dress so then can be colored individually.


Ben987:
Great idea!  Thanks for the update.

PR #1679, opened 2020-11-20
Rope/Gag/Lock Emoticons

Nina-1474:
Three new thought-bubble emoticons displaying rope, a gag and a lock.

Credit to **Verity** for the images.


Ben987:
Thanks you two!

PR #1680, opened 2020-11-20
Automatic shock collar and futuristic panel gag trigger fix

Ada18980:
This fix makes the auto shock collar and the futuristic panel gag no longer have the "Automatic" tag, since players reported it as a bug when the messages got filtered. It may also hinder communication in RP. So the solution, I think, is to make them no longer automatic since they are triggered by intentional actions


Ben987:
Thanks for the fix!

PR #1681, opened 2020-11-20
Auto shock collar and futuristic panel gag do not trigger on whispers

Ada18980:
This is because it's confusing to other people when it happens


Ben987:
Sure, thanks for the item update.

PR #1682, opened 2020-11-20
Openable front for the futuristic chastity belt

Ada18980:
Adds a new option to open the front of the futuristic belt (useful because it has settings you would otherwise need to reapply, and you would have to set the colors. Also snuck in a graphical change that makes it look slightly less flat in my opinion.

One might argue it's not a *chastity belt* if you can open the front, but it's meant to be used with a lock, so I think this is okay. It also isn't used by NPCs.


Ben987:
Awesome!  Thanks for the item update.

PR #1683, opened 2020-11-20
New facial expression menu

jomshir98:
Creates new facial expression menu, where instead of scrolling endlessly through emotes user first picks category and than expression.

test

Should there be more emotes than can fit, I will create pagination. But for that at least seven more symbols would have to be added into emoticon category


Ben987:
It looks good!  I will test it right now. Thanks :)

PR #1687, opened 2020-11-20
Futuristic gag deflation speed

Ada18980:
By popular request, I made the gag deflate one level at a time instead of all at once.


Ben987:
Alright!

PR #1688, opened 2020-11-20
New facial expression menu: remove hover trigger

jomshir98:
Removes the hover trigger for the facial expression menu categories


Ben987:
Awesome!  Thanks.

PR #1689, opened 2020-11-20
Tweak: Futuristic Panel Gag feature update

Ada18980:
The gag can now have the deflation timer changed, and it can be manually pumped one level (to deflate it, just manually set it lower)


Ben987:
Very good!  Thanks for that one.  Cheers.

PR #1690, opened 2020-11-20
Tweak: Futuristic Chastity belt triggers outside of chatrooms

Ada18980:
Makes the item a bit more interesting in singleplayer. NPCs wont put it on, though, since they cant configure it.


Ben987:
Great, thanks for that one!

PR #1691, opened 2020-11-21
Tweak: Room size while blind

Ada18980:
When the player's sensory deprivation settings are set to maximum (no whispers while blind) and the player also cannot examine other players while blind, only the player will be rendered in a chat room regardless of the number of players in the room.


Ben987:
Sure, that's a good idea, thanks :)

PR #1692, opened 2020-11-21
Fix: Can now kiss someone on the gag even if they have multiple gags

Ada18980:


Ben987:
Have you tested that you cannot kiss if there's a hood of if someone is in a box?

Ada18980:
You can kiss someone on the gag through the hood now. Perhaps there isn't a good way of doing this.

PR #1693, opened 2020-11-21
Player Leashing

Ada18980:
Adds a requested command in the immersion settings: The user can set themselves to be able to be yanked into a different chatroom by someone on their friend list when they receive a beep. The requirements are:
1) The person being yanked must have an item with "Leash" in its name or type
2) The person being leashed must not be tethered, mounted, or enclosed
3) The person being leashed must have the option enabled in Immersion settings
4) If the neck restraint slot is locked with a lover lock, only lovers can pull the player
5) If the neck restraint slot is locked with an owner lock, only owners can pull the player


Ben987:
Leashing and dragging through beeps isn't very intuitive.  Maybe there's a better way to do that, I'm not sure, let's think about it.

Ada18980:
> Leashing and dragging through beeps isn't very intuitive. Maybe there's a better way to do that, I'm not sure, let's think about it.

What about a button that attaches the leash to the target player, and then leashes them out when entering a new room? If someone else leashes the player then they're no longer leashed on your end

Ben987:
I will test it right now.  Cheers!

PR #1694, opened 2020-11-21
Snorkel Edits

EmilyR42:
- Improved snorkel positioning and recoloring
- Altered snorkel layers to support these changes
- Allowed access to nose items.
(Done with permission from Z-Freak/Natsuki.)


Ben987:
It looks good!  Thanks a lot.

PR #1695, opened 2020-11-22
Fixes a duplicate dialog option in the maid when you have owner locks

Ada18980:


Ben987:
That's a strange one.  That dialog option would lead to nowhere.  Thanks for the fix :)

PR #1696, opened 2020-11-22
Fix: Duplicate character IDs causing disappearing activities

wultir:
Multiple players sometimes get assigned the same ID by the client and when they are in the same room they can interfere with each other causing items and activities to dissapear.
It can probably cause other issues too but that is the one I noticed that caused me to investigate this.
V3ejoP21Yn
Reselecting the area will show the activities again but in actual online situations this can be triggered in quick succession making the activities unusable.
### Steps to reproduce
1. Log in with three characters (let's call them A, B and C)
2. Create a room with A
3. Have B and C join the room in order
4. Their character IDs in A's client should now be A: 0, B: 2, C: 3 (1 is taken by the main hall maid)
5. Log out with B, log back in and join the room.
6. IDs will be A: 0, B: 3, C: 3.
7. Have A select an area with activities on C
8. Change B's arousal to trigger an expression change (and thus a CharacterRefresh call).
9. Activities for C will be cleared in A's client
### Fix
This happens because ID is based on the length of the `Character` array and removing a character then adding them back when logging out and back in causes them to reuse the last ID.
This PR instead changes the ID to be based on an incrementing global counter that is reset when the array is cleared (on login). Not the only possible fix but seemed like the cleanest to me.

### DIFF
--- a/BondageClub/Screens/Character/Login/Login.js
+++ b/BondageClub/Screens/Character/Login/Login.js
@@ -91,6 +91,7 @@ function LoginLoad() {
 
 	// Resets the player and other characters
 	Character = [];
+	NextCharacterId = 1;
 	CharacterReset(0, "Female3DCG");
 	LoginDoNextThankYou();
 	CharacterLoadCSVDialog(Player);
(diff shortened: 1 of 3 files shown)


Ben987:
I've never seen that one before but thanks a lot for the fix :)

PR #1698, opened 2020-11-22
Removed some methods to free yourself, when you are on the no-help list

Fox906090:
Heyho,
like I wrote on Discord a few days ago, I found some methods to free yourself, even if the player is on the no help list.
In the following pull request I tried to fix all I'm aware of, exept the ones in the private room. There would have been some  dead ends, in which your feet are tied and you are stuck in it.  Maybe one of you has an idea how to get around this problem ^^
The methods I fixed:
- With a high enough gamble skill you could pay 25 $ to the right character in the gamble hall to get free
- You could still pony your way out of your bondage (arms zone got overwritten by armbinder)
- When the police officer of the prison section got you (bad girls gang), she has overwritten your bindings with her shackles
- If you were bound by just handcuffs, you could get out of them by going through the prison
- You could get out, if you joined the nursary
- Mistress Sophie untied you in the end of the "Looking for Sarah" event
- The maid in the introduction  room could have overwritten some of your bindings with her nylon ropes, if your arms were free


Ben987:
Disabling the rescue maids doesn't mean that nobody in the club can free you though.  Can you make a pull request targeting only the maids interactions?  Maybe we could add another toggle to disable all NPC messing with restraints, but I'm not sure this is needed.  Thanks!

Ada18980:
The intent behind the disabling maids feature was that there are no easy outs if you have owner locks or lover locks, but you can still interact with the club. It's just riskier.

I think it would be good to disable these interactions only if the player is wearing locks.

Ben987:
If it's an owner/lover lock problem then it's different than the no help from maids.  We need to make sure that no NPC can remove these, it's a bug.

Ben987:
@Fox906090 , in all the NPC release paths that you've found, which one allowed you to get out of owner or lover locks?  Can we prepare a fix only for these paths?  The maid flag doesn't matter in that case, it should never be possible to remove these locks.  Thanks a lot.

Fox906090:
I didn't specially test these options if you are Lover/Owner locked. Almost all of these possibilities are cases, where you are released by "CharacterRelease(Player)" The others are overwrite issues. There might be way more of these in the club. Furthermore there are many issues, if for example your shoes are owner/lover locked, because Player.IsRestrained() is false in this case. Theoretically I could adjust the no-help list for the maids only and check for owner/lover locks in the other cases. I'll do it in the next few days ^^

Ben987:
The most important thing is that NPCs cannot unlock owner & lover locks.  If there's already a validation that comes before the CharacterRelease(Player) then it's fine.  But if there's no validation we need to apply a custom fix like you did.

Being on "no rescue maid help" doesn't mean that other NPCs cannot release you.

Thanks a lot!

Ben987:
Let's close this for now, this is fairly outdated.  If you find situations where NPCs can remove owner or lover locks, please let me know.  Thanks!

PR #1699, opened 2020-11-23
Chinese Long Dress Multicolor

EmilyR42:
- Added layers and multicoloring to Chinese Long Dress


Ben987:
Great idea!  Thanks a lot :)

PR #1700, opened 2020-11-24
Fix console error for futuristic collar

Ada18980:


Ben987:
Thanks for the quick fix!

PR #1701, opened 2020-11-25
Increased max heartrate on Futuristic Chastity Bra

Ada18980:
I was told the heartrate seems quite low on the Futuristic Chastity Bra, so I increased the scaling so it's higher at max arousal.


Ben987:
Alright!  Thanks

PR #1702, opened 2020-11-25
Feature: Darken room if both eyes are closed

SandrineBC:
# Summary
If the character closes both eyes, the maximum blind level is applied, so that the room gets dark.
This does not happen, when the character blinks, but only if the eyes are closed on purpose.

# Details
## What was changed
A simple if statement added to `GetBlindLevel`in `Character.js

## Testing
This was tested on my local copy of the game and works in the single player, as well as the multi player rooms.
If the player keeps one eye open, by winking, the effect does not trigger.


Ben987:
Sure, that's a fun idea.  Thanks!

PR #1704, opened 2020-11-26
Fix: All diapers count as being diapered in the nursery

SandrineBC:
# Summary
Fixes that only the plain diaper count as being diapered when it comes to enter the nursery.

# Details
## What it does
Changed the `PlayerDiapered`Check in `Nursery.js`

## Testing
Tested that on my local instance of the game. The nurses accept other diapers now as well, when the player wants to enter the nursery for a second time.


Ben987:

PR #1705, opened 2020-11-27
Leashing option default and dialog

Ada18980:
1) Player leashing is on by default
2) Changed the text to reflect RP better
3) Moved the checkbox so it's not way off
4) Fixed a bug in HoldLeash that required the dom to leash a player twice before leashing would work
5) Added logic to ping players joining the room that are on the leashlist to make sure they are still leashed, removing from dom side if they are not (in case something happened in a different room or they got unleashed.


Ben987:
Thanks for the quick fixes!  I will test it soon.

PR #1706, opened 2020-11-27
Fix: New characters and blindness check

SandrineBC:
# Summary
This fixes a bug that was introduced with #1702 
When a new character is created, the check in `GetBlindLevel` raised a null pointer exception. This is now prevented.


Ben987:
Thanks for the quick fix!

PR #1707, opened 2020-11-27
Futuristic collar fixes

Ada18980:
Some bugfixes for the futuristic collar, including being able to put a mistress lock on someone without having one yourself.


Ben987:
Thanks for the fix!  Cheers.

PR #1708, opened 2020-11-27
Hide focus zones grid while coloring an item

wultir:
Hides the focus zones grid while in item-coloring mode to avoid it interfering visually with the item being colored, and also somewhat cleaner look.

Not sure if this would be better as an option instead but I'm happy to turn it into one if that's the case.


Ben987:
Sure, that's a good idea.  Thanks!

PR #1711, opened 2020-11-29
Heavy Latex Corset

EmilyR42:
- Added new Heavy Latex Corset item


Ben987:
Great!  Thanks a lot for that new item :)

PR #1712, opened 2020-11-29
Arousal bar animation disappears when progress = 0

Ada18980:
This is to stop the blue blinking from blinking all the time


Ben987:
Good one!  Thanks.

PR #1714, opened 2020-11-29
Allow starting maid work with the exposed uniform version

wultir:
Adds `MaidOutfit2` ("Maid Exposed") to the valid uniforms to start maid work.
Also skips replacement of the currently worn uniform slots when starting (some) jobs which allows this to work and has the added benefit of keeping custom colors of your currently worn maid outfit when the work starts.


Ben987:
I love it.  Thanks for that little change :)

PR #1716, opened 2020-12-01
Crotch rope now has an "Over Panties" option

Ada18980:
Added as an extended item option, had to split up the rope into two layers. Layers still copy colors and work with multicoloring


Nina-1474:
Instead of creating a new layer, it'd be much neater to set OverridePriority in the new hemp rope option. See the RopeCuffs option for hemp ropes on arms for an example.

Ada18980:
Thanks @Nina-1474 ! I changed it to use override priority instead.

Ben987:
Thanks!  This is a good idea.

PR #1720, opened 2020-12-02
Fix: Missing ignored rooms in chat room filter

wultir:
Because looping through the ignored rooms that are not in the latest search results starts at `ChatSearchResultOffset`, which is 24 on page 2, they do not get drawn in the filter list when they get pushed to page 2 or higher.

Changes the initial offset to be calculated based on page offset, already shown rooms and total rooms in the search results (`ChatSearchResultOffset + shownRooms - ChatSearchResult.length`).


Ben987:
I've never seen that bug but thanks for the fix :)

PR #1721, opened 2020-12-02
Fix: Don't show difficulty for NPCs.

wultir:
Removes difficulty from NPC information sheets which is currently shown and overlaps their lover info:
NPC Difficulty

(also, since I had to move the `OnlinePlayer` declaration anyway I moved it a bit further up to reuse it in the previous check)


Ben987:

PR #1724, opened 2020-12-03
Room admin icon

jomshir98:
Adds an icon for room administrator

The icons also get smaller with the character (when there are more than two persons in room)


Ben987:
Sure!  That's a good idea.  Thanks a lot.

PR #1725, opened 2020-12-04
Extreme Mode - Automatically limit combination and password padlocks

Ada18980:
Instead of blocking them outright. Owners and players on the whitelist will be able to put them on.


Ben987:
Sounds good to me!  Thanks.

PR #1727, opened 2020-12-05
1 New gag

ace-1331:
- added 1 new gag

Credits to evals for the assets.


Ben987:
Thanks a lot you two.  We always need more gags :)

PR #1729, opened 2020-12-06
Default colors for Bed and Covers

TheGnarp:
As discussed with Ayesha, just some basic default colors for the bed and the covers/blanket since most players probably don't always want a pure white experience, and the number of clicks to give it a wooden frame and some simple colors was quite large.


Ayesha-678 (approved):

Ben987:
Yes!  Thanks a lot for that one, cheers :)

PR #1731, opened 2020-12-06
Futuristic Collar Open Permission Setting

Ada18980:
Allows the user of a futuristic collar to set whether or not futuristic items can be modified by other players or not, even if they are locked. A normal collar prevents this--the only way to allow futuristic items to be modified is if there is no collar on the player or there is a futuristic collar that the player can access. This makes it impossible for most players to bypass owner locks unless the owner disables collars for the sub and does not put her own collar on.

This will allow players to modify items that have lover locks on them, because I did not want to add scripts for some complex lock hierarchy. It does not allow people to unlock or remove the items or change their color, however, only alter settings like connecting/disconnecting cuffs, which is a normal feature on most items. To enhance security, lovers should lock a collar on with a lovers lock.


Ben987:
Very nice.  I will test it soon.  Cheers :)

PR #1732, opened 2020-12-06
Barefoot Sandals

Ayesha-678:
Added a new item: Barefoot Sandals.
More to come.


Ben987:
Yes!  Absolutely.  Thanks a lot for this new item :)

PR #1733, opened 2020-12-06
Bugfix: Item colour resilience

elliesec:
## Summary

This adds some extra checks to `CommonDrawAppearanceBuild` and `ChatRoomSyncItem` to validate that an item's `Color` is defined properly (either `null`, `undefined`, a string or an array of those types). This isn't something that should be possible through normal gameplay, but players can currently use the console to set an item's `Color` to anything they like, for example an object like `{}`, which results in the error below (screenshot taken from discord, where players have been reporting someone's console usage causing these issues):

Stack Trace

This check should prevent people from breaking other people's games via this particular loophole.

## Steps to reproduce

Character 1 and Character 2 in the same chatroom.

1. Character 2 applies an item (e.g. something in `ItemArms`) to Character 1
2. Character 2 then runs the following in the console:
```js
InventoryGet(ChatRoomCharacter[0], "ItemArms").Color = {};
ChatRoomCharacterItemUpdate(ChatRoomCharacter[0], "ItemArms");
```
or:
```js
InventoryGet(ChatRoomCharacter[0], "ItemArms").Color = false;
ChatRoomCharacterItemUpdate(ChatRoomCharacter[0], "ItemArms");
```
These currently cause the game to crash for Character 2.

### DIFF
--- a/BondageClub/Screens/Online/ChatRoom/ChatRoom.js
+++ b/BondageClub/Screens/Online/ChatRoom/ChatRoom.js
@@ -1370,9 +1370,11 @@ function ChatRoomSyncItem(data) {
 			if ((data.Item.Name == null) || (data.Item.Name == "")) {
 				InventoryRemove(ChatRoomCharacter[C], data.Item.Group);
 			} else {
+				var Color = data.Item.Color;
+				if (!CommonColorIsValid(Color)) Color = "Default";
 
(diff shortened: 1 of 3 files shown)


Ben987:
Thanks for that fix!  I've never seen that scenario before but we need to make sure that troll scripters cannot crash the game :)

PR #1734, opened 2020-12-07
Reindeer Hairband

EmilyR42:
- Added a new hat item for the holiday season


Ben987:
The timing is perfect.  Thanks a lot :)

PR #1736, opened 2020-12-07
Fix: Exception when triggering shocks on self

wultir:
The shock trigger function on the shock collar (`InventoryItemNeckAccessoriesCollarShockUnitTrigger()`; which is also used for the dildo, clamps and plug) calls `ChatRoomPublishCustomAction()` with `LeaveDialog` set to `true` which causes it to clear `DialogFocusItem` and in turn throw an exception when trying to update arousal if the player shocks themself resulting in no arousal or facial expression updates.

Changes `LeaveDialog` to `false` in the call and to call `LeaveDialog()` manually later if in a chat room.


Ben987:
Thanks!  I've never seen that error before but the fix is appreciated :)

PR #1738, opened 2020-12-07
Rhinestone Sandals converted to multicolor item

TheGnarp:
Always frowned when the soles changed color with the stones on the top, so I updated the item to support multi color layers.
My first real contact with layer masks, so please take a closer look when you review the changes.
Had to use this trick supposed by Ace to hide the top layer of the shoes when in a hogtie, since the order of the layers is then reverted.
(That is, in a hogtie, you look at the soles and the top of the shoes is not visible behind them, although, for the shoe asset, the order didn't actually change and it would try to display the top of the shoes over the soles ... complicated - at least for me - *sigh*)

I'm really happy about how often I see these sandals nowadays in the club!!
Freedom for the tootsies!!! ^_-


Ben987:
Excellent!  Thanks for that item upgrade.  Cheers :)

PR #1743, opened 2020-12-11
Improvement: Preference init inconsistencies

ace-1331:
made it so the settings are synced on login if they were changed due to default settings (happens when there are new settings in update and such)

This caused desync problems online


Ben987:
Good catch, it will be useful for new releases.  Cheers!

PR #1745, opened 2020-12-11
Fix: removed allfours from the pose menu

ace-1331:
removed the pose from the pose menu as it was enabled by mistake.


Ayesha-678 (approved):
My bad for enabling it ^^;

PR #1752, opened 2020-12-11
Fixed clipping issues with mittens and also the pilot suit, and lets mittens change poses

Ada18980:
List of changes:
1) Pilot suit now hides upper and lower body, along with the regular catsuit (but not translucent ones)
2) Futuristic mittens can now change poses from the pose switch menu. Alphamask included to avoid conflicting with gloves
3) Regular upperbody sprites no longer have hands, since the hands layer already exists now. Dont need 2 hands.


Nina-1474:
You're changing the colour scheme of the body images I think. Here's what the change looks like in Firefox (no change when viewing in Chrome):
BodyColour

Ada18980:
> You're changing the colour scheme of the body images I think. Here's what the change looks like in Firefox (no change when viewing in Chrome):
> BodyColour

I'm not sure how this is happening? I don't see any difference when I'm viewing in Chrome. Also, I'm not sure the colors on the arms could even be changed considering that the only part I copied over was the belly from the BackElbowTouch pose.

Ada18980:
I also don't see any difference on my side in firefox. Does anyone else see this? I'm curious as to what's causing it. Maybe I'm not saving in the correct bit depth?

Nina-1474:
The same thing happened to me when I was removing the heads from the bodies. I had to change a setting somewhere in GIMP so that it didn't convert the edited image to a new colour profile, or something like that. Maybe your image editor is doing something similar. Although I'm struggling right now to find where in both the program and the files this data/setting is stored.

Ada18980:
I'm using Paint.net with the auto detect bit depth.

Ada18980:
@Nina-1474 Can you check again on your setup? I saved with 32 bit depth now. I wonder if that will fix it.

Nina-1474:
No luck I'm afraid. They still look different under 'Files changed' for me.

Ada18980:
You say this only happens with Firefox?

Nina-1474:
I get the same value but that's not the issue. It looks fine on one browser but muted on another (for me at least) so there's something other than the colour values that different browsers are handling differently.
I opened and re-exported one image and it made it look the same in Firefox again. I could do that with all of them and push it to your branch if you're ok with that?

Ada18980:
> I get the same value but that's not the issue. It looks fine on one browser but muted on another (for me at least) so there's something other than the colour values that different browsers are handling differently.
> I opened and re-exported one image and it made it look the same in Firefox again. I could do that with all of them and push it to your branch if you're ok with that?

I wouldn't mind that. In fact I'd appreciate it; I wish I could repro it on my end.

Nina-1474:
I'm using Firefox version 83.0 (64-bit) just in case it helps.

Nina-1474:
Looking all good now!
BodyColour2

Ada18980:
Awesome!

Ben987:
@Ada18980 , are you sure that changing the base character assets doesn't create any bugs?  I'm uneasy to change the base files of the game.  Also, did you use 3DCG to make them?  If so, can you upload the pose file to it?   Thanks a lot.

Ada18980:
Hi Ben, I did not use 3dCG to modify them, I simply filled in the midriff with the midriff from the elbows bound pose.

The original intent for this was to allow some types of mittens to change poses, but I believe Ace may have pushed a change that makes this redundant. I can split off the other changes into a separate PR and close this one when I get a chance

Ada18980:
Made a new PR to include catsuit changes, closing this one.

PR #1756, opened 2020-12-12
Futuristic Chastity Belt Reskin

Ada18980:
Added a reskin of the futuristic chastity belt: The futuristic chastity panties.

They are identical in scripts, with the new FuturisticChastityBelt2.js calling the functions of the original.


Ben987:
Great!  I'm sure it will be popular.  Thanks :)

PR #1762, opened 2020-12-12
Bring MouseEars1/2 and ElfEars values in line with other ear items

wultir:
I'd say this one is subjective whether it's a bug or not but MouseEars1, MouseEars2 and ElfEars ear items have a value of 20 instead of -1 like other items in the HairAccessory2 (Ears) item group causing them to show up twice when you ask to see items you don't already own.

Changes them to behave in the same way as the rest of the group by setting their values to -1.


Nina-1474:
I can say that the ElfEars at least was a mistake on my part, and should indeed have been -1.

Ben987:

PR #1764, opened 2020-12-13
Pose Menu in Photography room

Nina-1474:
Enabling the pose-changing menu in the Photographic room as unlike other single-player rooms, there's a clear benefit and no apparent issues with enabling it.


ace-1331 (approved):

Ben987:
Sure!  It makes perfect sense.  Thanks for that one.

PR #1765, opened 2020-12-13
Many missing assets

ace-1331:
- added several missing locks/pose images for the heavy latex corsets (blank images)
- properly hid the entire anklet groups in the relevant poses to prevent 404s in several poses that are not required as well as prevent them from floating in the air
- corrected the groups so they'd have the same values. (Some were extra)


Ben987:
Thanks for fixing all these missing assets :)

PR #1766, opened 2020-12-13
ItemHood in Arousal menu

Nina-1474:
Small followup fix to #1763: Since ItemHood now has arousal activities it needs a line in the Preferences screen text file for the Arousal menu.


Ben987:
Thanks!

PR #1767, opened 2020-12-13
Mittens with pose menu

ace-1331:
Enabled the upperbody pose menu with mittens by no longer considering it to be a pose that force the arms in one way or the other.

No side effect was found


Ben987:
Very well, thanks!

PR #1769, opened 2020-12-14
Gwen hood adjustments

ace-1331:
uses masks instead of hide for better compatibility, this should be the final adjustments, I hope haha


Ben987:
Great!  Thanks a lot.

PR #1770, opened 2020-12-14
Rope Over Panties message fix

Nina-1474:
Fix to #1760 so that the over panties option chat message displays.


Ben987:
Awesome!  Thanks for that one.

PR #1771, opened 2020-12-14
Text fit profile screen

ace-1331:
- Improved draw text fit to allow the option for a background like draw text
- Made it so the first column of a profile wouldn't overlap with the second


Ben987:
Great idea!  It will also help translators that could not fit the foreign text properly.

PR #1772, opened 2020-12-14
Photography Room - NPC Poses

Nina-1474:
Expanding the 'Please change your pose' dialog branch for the NPC in the Photographic room from only toggling kneeling to instead allow all the same upper & lower body poses that the player has in the pose menu.


Ben987:
Great idea!  I'm sure git4nick will like it.

PR #1773, opened 2020-12-14
Remove offline friends from first tab

jomshir98:
This removes offline friends from first tab, turning it into a "Online friends" tab only.
All friends are still visible on delete tab named "All friends"

_Reason: seeing offline friends made several people uncomfortable_


Ben987:
This is a great idea, I will test it soon.  Cheers!

PR #1774, opened 2020-12-14
Missing asset

ace-1331:
An asset didn't have its corresponding name changed after it was split for multicoloring


Ben987:
You're the best at finding all these wrong assets, cheers!

PR #1775, opened 2020-12-14
Suit Feet when Hogtied fix

Nina-1474:
The hogtied pose for catsuits/pilotsuits/etc all included the feet in the upper body Suit image instead of the lower body SuitLower equivalents. This means that if the Suit selection was different from the SuitLower selection, by having either one empty or by colouring them differently, the feet would incorrectly show the upper body part's settings.
A Hogtied folder has been added for SuitLower and all the feet portions of the suits have been moved into here.


Ben987:
Thanks a lot for that fix!  Cheers :)

PR #1776, opened 2020-12-14
Bugfix: Fixes floating characters when leaving the dialog screen via the escape key

elliesec:
## Summary

This fixes a bug which leaves characters floating in the air after using the "up" button to show all of the zones on a character that is kneeling/hogtied/etc and then leaving the dialog screen via the escape key.

## Steps to reproduce

1. Enter a chatroom and kneel
2. Click on the kneeling character and hit the ^ button to show all of the character's zones
3. Hit the escape key on your keyboard
4. Note that the character is left floating in the air


Ben987:
Thanks!  I've never seen that one before but it certainly needs a fix :)

PR #1777, opened 2020-12-14
Anklets in Random Clothes

Nina-1474:
At the moment the band anklets are appearing on every NPC. Changing the group settings to be more in line with other clothing accessories such as hats and necklaces.
Note: There does seem to be issues with how these group settings are used in general which I'm going to look at later.


Ben987:
Yes, absolutely.  Cheers!

PR #1778, opened 2020-12-14
New Item: Medical Bed Restraints

elliesec:
## Summary

Adds a new set of Medical Bed Restraints items for feet, legs and arms - assets created as always by Cecilia.

The items are not available through the shop, and have been tied into the Asylum. They are always available when inside the Asylum (or in Asylum chatrooms), and can be obtained from the nurse in the Asylum entrance once the player has reached 100 nurse/patient reputation. Once obtained, the player may use them outside of the Asylum as well as inside it.

Below is a preview of the new items.

segufix-8

## Changes

* Added two new asset properties:
    * `HideItemExclude` - like the `HideItem` array, but this will exclude items otherwise hidden by their group in the `Hide` array. For example, this would hide all clothing items _except_ for the Mistress Top:
    ```js
    { ..., Hide: ["Cloth"], HideItemExclude: ["ClothMistressTop"], ... }
    ```
    * `AvailableLocations` - An array of strings representing locations where an item will _always_ be available, regardless of whether or not the player owns it. At the moment, the only supported value in this array is `Asylum`.
* Modified the Asylum entrance nurse's dialogue to allow the player to obtain the new restraints once their nurse/patient reputation has reached 100
* Modified the `DialogInventoryBuild` function to allow items with `AvailableLocations: ["Asylum"]` to be always useable within the Asylum or Asylum chatrooms
* Added logic to the login screen to add/remove the items from players on login, as appropriate


Ben987:
I love it!  And it's a nice incentive to reach a 100 rep in the Asylum.  Cheers :)

PR #1780, opened 2020-12-14
GP-9 Gas Mask fixes

EmilyR42:
- Enabled access to ear and neck slots while wearing the GP-9 gas mask


Ben987:
Absolutely!  Thanks :)

PR #1781, opened 2020-12-15
Bugfix: Fixes chatroom messages on the futuristic gags

elliesec:
## Summary

The futuristic gags were producing incorrect chatroom messages when the "Trigger Inflation" button was pressed.

## Steps to reproduce

Two characters: "Character 1" and "Character 2".

1. Add a Futuristic Panel Gag/Futuristic Harness Panel Gag to Character 1
2. As Character 2, enter the extended item menu for the gag and hit the "Trigger Inflation" button
3. Note that the chatroom message says "Character 2's gag inflates into a mouth-filling ball.", and expression changes are applied to the wrong character as well.


Ben987:

PR #1782, opened 2020-12-15
Remove On Remove - Type fix

Nina-1474:
The function that automatically removes certain items when certain others are removed has the ability to check the type of extended assets, currently only used to remove round nipple piercings with the chain or weighted chain when removing a collar. 
This wasn't working and always removed the item regardless of type, i.e. it removed the piercings even without a chain.
The function has been fixed so now it won't remove the item if it does not have the type matching the type configuration.


Ben987:

PR #1783, opened 2020-12-15
Bugfix: Fixes chatroom messages for the futuristic collar

elliesec:
## Summary

Chatroom messages for the futuristic collar were being sent using `DestinationCharacter` when the text in the CSV uses `DestinationCharacterName`. This was resulting in messages like "Character'sName bindings tighten around her body with a hiss as they lock themselves automatically."


Ben987:

PR #1784, opened 2020-12-15
Beta: Custom eyes closed blindness

ace-1331:
- added a toggle to cycle through blindness levels when your eyes are closed to allow for more freedom than either fully open or completely blind


Ben987:
Good idea!  Thanks for that one.  Cheers :)

PR #1785, opened 2020-12-15
Fix: Error for leashed player when leash holder fails to enter a room

wultir:
When the player holding a leash fails to enter a room, either because it's full or because they try to create a room with a name that already exists, the leashed player gets sent a `Leash` beep with no room set, causing `ChatRoomJoinLeash` to be set to null and in turn cause an error in `ChatSearchQuery()` until `ChatRoomJoinLeash` is changed again.

Adds checks both in `ChatSearchQuery()` which fixes the resulting error and in `ServerAccountBeep()` which avoids causing it in the first place.


Ben987:
Thanks a lot for the fix!

PR #1787, opened 2020-12-15
Fix crash when friendnames are corrupted

ace-1331:


Ben987:
Thanks!  I haven't seen it happen but it's better to catch any possible error.

PR #1789, opened 2020-12-15
Login NPC Overwrite fix

Nina-1474:
When a login crashes, it's possible on successfully logging in later to find that the character's appearance and bindings are completely changed.
This is because the Player variable is used as a placeholder for the randomly generated login screen NPC, and ends up keeping that appearance. This uses a dedicated character for the NPC instead to avoid the issue.

I'm holding this change back from this release given that we're late in the beta period, the bug is not critical if it occurs, and if there's any hidden issues with this change it could potentially be quite disruptive.


Ben987:
Sure!  I've seen that one a few times before.  Thanks.

PR #1791, opened 2020-12-15
Improvement to pose system: Partial movement

ace-1331:
Add a new property for item to whitelist active poses which allows to use some, but not allow active poses from the menu. It also changes CanKneel to verify if kneeling is actually possible (as solely having LegsClosed no longer means someone cannot kneel).

This fixes the current technical limitations the pencil skirt and wedding dress were facing as well as allowing restraints and clothes to set a pose while still allowing partial movement


Ben987:
Yes!  Thanks for that one.  I will try it soon.

llll434:
There are issues with the restraints now, for example you can't kneel down or stand up with detached cuffs or restraints you could add while kneeling like the legbinders making you stand all of the sudden and others can't help you down again aswell.

PR #1792, opened 2020-12-15
Scroll Chat Down with Esc Key

Nina-1474:
Allowing the Escape key to move the chat scroll position to the end similar to Discord, as an alternative to lengthy page-downs or scrolling.


Ben987:
Sure, that should be handy.  Thanks for that feature.

PR #1793, opened 2020-12-15
Bugfix: Allow selection of legs/feet, while keeping floating character fix

jomshir98:
Moves the fix to `DialogLeave` to allow "Show all zones" to work properly


Ben987:
Excellent!  Thanks for the fix.

PR #1794, opened 2020-12-15
New FixedHead effect

ace-1331:
New effect to block moving your head, which influences the new wiggle and nod activity.

Currently only applied to posture collars


Ben987:
Sure, this is a good idea, I'm curious to try it.  Thanks!

PR #1796, opened 2020-12-16
Spread eagle ropes (Arms, Feet) as extended item for hemp rope

TheGnarp:
The hemp rope for arms and the ones for feet now include a new bondage type "Bed Spread-Eagle".
If the target character is "mounted" on a bed, these options can be used to spread-eagle her nice and cozily.

Special thanks to @elliesec, since those poses required repositioning the ropes assets dynamically in the HempRope.js files, and i wouldn't have figured that out myself. One question, not even five minutes of waiting and a very detailed reply later, it worked like a charm.

Reviewers: I'm still not 100% fluent with prerequisites or effects and might have missed something.


Ayesha-678 (approved):
Cool! I was working on something like this too, but yours is much better!

Ben987:
Great idea!  I will test it soon.  Cheers :)

PR #1798, opened 2020-12-16
Feature: Changes the "OnBed" prerequisite to use a new "OnBed" effect

elliesec:
## Summary

As @Ayesha-678 has a new medical bed item on the way, this PR adds an `OnBed` effect and changes the `OnBed` prerequisite to check for that effect, rather than just checking for the current bed item. This will allow existing bed restraints to work with the medical bed, and any other future items that want to use the effect.


Ben987:
Yes!  This will be useful for new items.  Thanks.

PR #1801, opened 2020-12-17
Fix: Leash actions with lover/owner lock on leash

wultir:
Fixes a few checks for lover/owner lock on leash when holding or pulling into a room.

- Currently tries to get the lock from `Player.Appearance` instead of the target player which makes it unlikely that the lock is found on the sender side which allows the sender to use the hold leash action even if owner/lover locked (unless prevented by other permissions on the leashed player).
- The check when the target receives the leash messages checks against the local player and not the sender, causing the check to succeed even if lover/owner locked.


Ben987:
@Ada18980 , can you validate this fix and let me know if it works as you want it?  Thanks a lot.

Ada18980:
I have tested it and it works as expected! Just need @wultir  to resolve conflicts.

wultir:
Updated with conflicts resolved.

Ben987:
Thanks @wultir , there's still a conflict with Ada's latest pull request though.  Can you resolve it and test on your side?  Let me know when it's ready.  Cheers :)

wultir:
@Ben987, I've fixed the conflicts with the latest changes. Should be good to go now.

Ben987:
That was really quick.  Thanks a lot!

PR #1802, opened 2020-12-18
Clears friendlist, room admin, and other inputs when leashed out

Ada18980:
Adds a ChatRoomClearAllElements function which is called when leaving the chat room


Ben987:
Thanks!  I will prepare a hotfix.  Cheers :)

PR #1803, opened 2020-12-18
R63 fixes

ace-1331:
- fixed whispers getting stuck
- fixed the try catch syntax not being liked  by mysterious browsers


Ben987:
Thanks a lot!  I will prepare hotfixes.

PR #1804, opened 2020-12-19
Corset slot

Ruilov3:
Migrated the existing corsets from the Bra slot to their own separate Corset slot to reduce clutter, allow more customization options and in preparation for a future expansion on corsets


Nina-1474:
I'd say the new group should have `Default: False` to avoid always appearing on every random NPC, and `Underwear: true` to be consistent with the original Bra behaviour.

Ruilov3:
wops :D done and done

Ben987:
This is a good idea.  Players will lose the previously bought items but it's worth it.  Thanks a lot!

PR #1809, opened 2020-12-20
Bugfix: Makes the login credits independent of framerate

elliesec:
## Summary

I've now obtained a 144Hz monitor, so I can test out framerate-based differences in behaviour across the game. I've noticed that the credits on the login screen whizz past substantially faster at 144Hz than at the standard 60Hz. This change makes the credit roll speed independent of framerate.


Ben987:
Thanks for that fix!  Cheers.

PR #1790, opened 2020-12-15
Leash improvements

Ada18980:
Leashing is now determined by the "Leash" effect, which items can use as needed. This allows for items which leash and also tether to be used to leash, and items like a dolly or pet carrier are possible.

PR #1810, opened 2020-12-20
Fix: Leash actions on character in boxes/cages/lockers

wultir:
The leashing code checks for the `Enclosed` effect instead of `Enclose` which items like boxes, cages and lockers use so leash actions are not blocked while in those items.

Might conflict with #1790 since it changes a line that's moved in that PR.


Ben987:
Yes!  Thanks for the fix :)

PR #1811, opened 2020-12-21
Login Character Dialog fix

Nina-1474:
Adding a dummy dialog csv file to prevent 404s.


Ben987:
Thanks for the quick fix!

PR #1813, opened 2020-12-21
Improvement: Optimisation of prerequisite checking

elliesec:
## Summary

This PR makes some improvements to the efficiency of prerequisite checking. As prerequisites are checked in a lot of places in the game, some performance optimisations here should help with performance on slower machines in several places in the game. In terms of functionality, all prerequisites behave as they did before. Combined with #1511, prerequisite checking in extended items should become substantially more efficient.

### Changes

* Changed the numerous `if` statements into cases inside a `switch` statement. Previously, the `if` format meant that if `Prerequisite` was further down in the function, each `if` would be checked before getting to the prerequisite (time scales linearly with the number of prerequisites). The `switch` statement will jump immediately to the relevant `case` (constant time).
* Reworked prerequisites that used multiple `InventoryGet` calls, so that `InventoryGet` is called at most once per prerequisite check. Since `InventoryGet` loops through the player's appearance it is relatively expensive for a simple getter.
* Refactored some common prerequisite checks into utility functions to improve readability.
* Removed the `CannotBeHogtiedWithAlphaHood` prerequisite from two web types where it was not needed.

## Testing

I've tested every prerequisite that has been changed by this PR (both negative and positive cases) to verify that they are all still functioning as they were before.


Ben987:
Sure!  A little more speed is always welcome.  Lets hope it won't create any new bug.  Thanks!

PR #1817, opened 2020-12-21
Funnel Gag

EmilyR42:
- Added a new funnel gag with options to add or remove the funnel


Ben987:
Wicked!  I'm sure that some players will like it.  Thanks :)

PR #1819, opened 2020-12-22
Bugfix: Fixes condition for the character actions dialogue option

elliesec:
## Summary

The "Character Actions" dialogue option would not always show up when the character was leashed, as the leash condition was missing from the `ChatRoomCanPerformCharacterAction` function.

## Steps to reproduce

1. Apply wooden cuffs (or any item that prevents kneeling), and a leashing item to a character
2. Note that the (Character Actions) dialogue option does not appear


Ben987:
Good catch!  Thanks for that one.

PR #1820, opened 2020-12-22
Leash Dialog fix

Nina-1474:
The 'Character actions' dialog option doesn't appear when the only options it contains are the leash options.
This is a slightly more comprehensive version of the #1819 fix just merged.


elliesec:
As Nina mentioned, this covers additional cases to #1819 - I was going to close that in favour of this, but it got merged too quickly!

Ben987:
Awesome!  Thanks you two.

PR #1821, opened 2020-12-22
Bugfix: Fixes validation functions on futuristic mittens and breast harness

elliesec:
## Summary

This fixes some validation function calls on the Futuristic Mittens and Breast Harness, whose validation functions weren't changed by #1511 (likely due to merge order). This causes these items to appear locked when they aren't, and blocks access to their extended item screens.

## Steps to reproduce

1. Equip Futuristic Mittens or Breast Harness.
2. Note that the extended item screen for the items declares that they are locked.
3. Note that clicking the unlock button denies access


Ben987:
Thanks for the fix!  I'm sure Ada will like it.

PR #1815, opened 2020-12-21
New Item: Elegant Heart Necklace

Ruilov3:
small, but shiny

PR #1822, opened 2020-12-22
Fix: Missing layer names for the Elegant heart necklace

Ruilov3:
#1815


Ben987:
Thanks for the quick fix!  Cheers.

PR #1825, opened 2020-12-23
Colour Name Validation

Nina-1474:
At the moment it's possible to set an invalid colour on another user's item via console and cause an error on the item for them when it's being rendered.
The drawing code now checks that any named colour (currently only White/Asian/Black) is in the group's colour schema and otherwise replaces the invalid colour with Default. Some additional changes were needed for inherited colours as they can pass down a named colour not in the target group's schema. The head & hands settings for preset NPCs were also updated to use the Default colour inheriting to pass this validation.


Ben987:
I've never seen that error before, but it's always a good thing to prevent crashes.  Thanks a lot!

PR #1799, opened 2020-12-17
Feature: Ability to change text fonts

llll434:
This PR adds the feature to be able to change the text Font with a Selector in the GraphicsSubSettings.
The added fonts are some of the most web save and often used fonts and should be available on nearly all devices.

PR #1827, opened 2020-12-24
Improvement: Font system refinements

elliesec:
## Summary

This is something of a follow up to #1799. I was originally going to request some changes on that PR, but it got merged before I did. This PR makes some refinements to the font system which should make it a bit more resilient and performant generally, and should also make it more likely that players will see distinct font options as they click through the available choices (rather than half of them displaying as the browser default). A summary of changes:

* Adds appropriate fallback font stacks for each font, so that the chosen font more closely resembles what the user requested if their chosen font is not available on their device. This means, for example, that if "Courier New" is not available on someone's machine, it will still fall back to a monospace font, rather than the browser default (which is generally a serif font like Times New Roman).
* Adds a Helvetica Neue/Helvetica/Arial/sans-serif font stack, as this is a reasonably commonly used font stack, particularly on Apple devices
* Makes use of @SandrineBC's recently-added memoization function for efficiency purposes. On most screens, the `CommonGetFont` or `CommonGetFontName` functions are called on every frame. Memoizing them makes it much more efficient to repeatedly call them, meaning the font computation is generally only ever done once per screen. The memo caches are cleared when changing font in the preference screen, and when changing screens (via `CommonSetScreen`) to ensure the font is always in sync with the player's actual preferences.
* Changes some text in the chat search to use`DrawTextFit` rather than `DrawText`, as it was not displaying well in certain fonts
* Improves the JSDoc

### DIFF
--- a/BondageClub/Screens/Character/Preference/Preference.js
+++ b/BondageClub/Screens/Character/Preference/Preference.js
@@ -48,7 +48,7 @@ var PreferenceVisibilityBlockList = [];
 var PreferenceVisibilityResetClicked = false;
 var PreferenceDifficultyLevel = null;
 var PreferenceDifficultyAccept = false;
-var PreferenceGraphicsFontList = ["Arial", "Times New Roman", "Papyrus", "Comic Sans", "Impact", "Verdana", "Century Gothic", "Georgia", "Courier New", "Copperplate"];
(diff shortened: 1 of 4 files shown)


Ben987:
Absolutely!  Thanks for the upgrade, I will try it soon :)

PR #1828, opened 2020-12-24
Add additional Y space on top of CharacterCanvas

jomshir98:
This will make it possible to create items "attached" to ceiling even with characters smaller than full high.

This change should be fully transparent, it's only effect should be, that negative Top values will become meaningful for item assets.
There shouldn't be any visible change at all.

The reserve space was calculated so smallest character in kneeling pose still reaches top of room


elliesec:
I mentioned this in Discord, but it's probably worth discussing this change with @Nina-1474, as I think she had something in the works that may well cover this case, although I suspect she may be having a festive break right now.

Nina-1474:
As discussed I do have a similar feature done as part of a larger project. I've just submitted it now under #1844.

(At the time my intention was actually to support extra drawing space without needing to change the player canvas size *at all*, but due to limitations with WebGL scaling and canvas drawing smoothing, I had to end up doing something pretty similar to this.)

jomshir98:
Closing in favour of much better #1844

PR #1829, opened 2020-12-25
Bugfix: Fixes alpha masks for the open crotch straitdress

elliesec:
## Summary

Fixes alpha masks for the open crotch straitdress. Previously the masks were only being applied to the lower body group, which caused clipping issues when the lower body size was larger than the upper body size. See the screenshot below for an example.

Bug


Ben987:
Thanks for fixing that one.  Cheers :)

PR #1830, opened 2020-12-26
Bugfix: LoverOnly/OwnerOnly items getting removed on relog

elliesec:
## Summary

As a result of some of the changes in #1387, items marked as `LoverOnly` or `OwnerOnly` get removed on login, due to a check in `ServerAppearanceLoadFromBundle` against the source member number (which is undefined when logging in). This PR reinstates the missing checks for a login.

## Steps to reproduce

1. Equip either the Lovers Vibrator or Love Chastity Belt on a character
2. Have that character log out and log back in
3. Note that the vibrator/belt has been removed by the appearance bundle load


Ben987:
Nice!  Thanks for the fix.

PR #1831, opened 2020-12-26
Allow room owners to block leashing

Ada18980:
This adds a "Player Leashing" toggle to the room block menu. This does not block leashes by itself but simply prevents the "Hold On To Her Leash" option from appearing on players. This will not stop a dom from having a leashed sub and bringing her into the room and taking her out (unless one of them disconnects in the process), but it will stop people from grabbing the leash of another player without first asking the room owner to turn off the block.

### DIFF
--- a/BondageClub/Screens/Online/ChatBlockItem/ChatBlockItem.js
+++ b/BondageClub/Screens/Online/ChatBlockItem/ChatBlockItem.js
@@ -1,6 +1,6 @@
 "use strict";
 var ChatBlockItemBackground = "Sheet";
-var ChatBlockItemList = ["ABDL", "SciFi"];
+var ChatBlockItemList = ["ABDL", "SciFi", "Leashing"];
 var ChatBlockItemCategory = [];
 var ChatBlockItemReturnData = {};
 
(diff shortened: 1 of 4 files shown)


elliesec:
I think it would be prudent to also modify the `ChatRoomMessage` function where the leashed player receives the `HoldLeash` message (or in the corresponding `ChatRoomCanBeLeashedBy` function). As is, this won't stop someone from using the console to call `ChatRoomHoldLeash` or send the `HoldLeash` message manually. A check on the leashed player's side would prevent people from circumventing the restrictions.

Ada18980:
> I think it would be prudent to also modify the `ChatRoomMessage` function where the leashed player receives the `HoldLeash` message (or in the corresponding `ChatRoomCanBeLeashedBy` function). As is, this won't stop someone from using the console to call `ChatRoomHoldLeash` or send the `HoldLeash` message manually. A check on the leashed player's side would prevent people from circumventing the restrictions to grief other players.

I agree. I'll get on that~

Ben987:
Sure!  This will be useful.  Thanks for that one.

PR #1832, opened 2020-12-27
Fix for spread pose on futuristic ankle cuffs

Ada18980:
Fixes a graphical glitch that results in floating lock icons


Ben987:
Thanks for the fix!

PR #1833, opened 2020-12-28
Controller Support

fleisch11:
it's not perferct, but it's the best i can do
(it only works properly using chrome)
(i am not good at creating icons)
(tested with a switch pro controller)


Ben987:
So it only works with one specific controller and only on one browser?  Maybe you could create a Chrome extension if it's that specific, we probably won't be able to support it in the game.

fleisch11:
> So it only works with one specific controller and only on one browser? Maybe you could create a Chrome extension if it's that specific, we probably won't be able to support it in the game.

no, i just dont have another controller to test it with. and most people use chroe anyway

fleisch11:
closed for a bit (making some last minute changes)

PR #1834, opened 2020-12-28
Lowers cooldown of futuristic chastity belt after attempting to struggle

Ada18980:
It's still possible to struggle out of a lot of items this way, I wanted to make it harder for any item with a challenge rating of 5 or so and probably impossible for challenge rating 6~


Ben987:
Sure!  Thanks for that one.

PR #1839, opened 2020-12-31
Fixed a bug in returning to chatrooms while being club mistress

Ada18980:
Would cause the maid rescue to fail due to the maid not being initialized properly at the start of the game. Added a 1 frame delay before it sends you


Ben987:
Thanks for the fix!

PR #1840, opened 2020-12-31
Bed Spread-Eagle Relog Fix

Nina-1474:
An error occurs when a player is tied to the bed with the spread-eagle option for the arms and/or leg ropes, and relogs while still tied. The bed is removed but the ropes are left mid-air, blocking the devices slot.
The bed has been given RemoveItemOnRemove settings for these and the login item-removal has been updated so that this setting is used at that point.


Ben987:

PR #1841, opened 2020-12-31
Kicked in Chatroom Subscreen Fix

Nina-1474:
If a player is viewing a subscreen in a chat room (wardrobe, profile, etc) and gets kicked or banned, on exiting the subscreen they still see the chat room despite no longer being in it. This fix re-uses the disconnect check so that they properly jump to the chat search page regardless of what subscreen they're in.


Ben987:
Yes!  Thanks for the fix.  Cheers :)

PR #1842, opened 2021-01-03
Bugfix: Vibrator mode fixes

elliesec:
## Summary

This fixes some cases where the Deny and Edge vibrator modes were not working as intended:

1. Adds stricter checks for vibrating items in the `CharacterIsEdged` function to prevent items like shock toys from counting as "vibrating"
2. Adds a `CharaterLoadEffect` call to the `VibratorMode` code to ensure the player's effect array is up to date on vibrator changes


Ben987:
Yes!  Thanks for the fix.  Cheers :)

PR #1843, opened 2021-01-04
Headphones/Leather Hood Multicolor

EmilyR42:
- Added a new headphones item in the ear slot with adjustable deafness settings
- Altered various hoods to hide/show them
(Note: Since this is an ear item, there's no way to apply them on top of hoods. The headphones need to be put on first. Currently looking for a workaround.)


Ben987:
They look great!  Thanks a lot :)

PR #1848, opened 2021-01-05
Improvement for Item Activities

tickler2000:
The "AllowActivityOn" field is now also used for Item Activities.
So Items hold in hand (Crop, Flogger, Feather, ...) can now be used on the body zones specified in this field, even when the zone is blocked for items.

I also fixed a bug, where AllowActivityOn was not checked correctly.

I also added AllowActivityOn for the PetCrawler-Asset and extended it for the BitchSuit-Asset and BitchSuitExposed-Asset.


Ben987:
Great idea.  Thanks a lot!

PR #1849, opened 2021-01-05
Seperate part colors for harness pacifier

wildsj:
Harness pacifer split into different sections for individual color selection.

No further fuctionality added or changed.


Ben987:
Very nice!  Thanks for that upgrade :)

PR #1850, opened 2021-01-06
Fixed Bug regarding the AllowActivityOn field

tickler2000:


Ben987:
Thanks for the fix, I will try it soon.

PR #1311, opened 2020-08-20
Colorable blush, emoticons, and fluids

ADRCorp:

PR #1853, opened 2021-01-07
Improvement: Expression coloring from the expression menu

elliesec:
## Summary

A while back #1311 added functionality to the UI to allow blush, fluids and emoticons to be colored via the appearance menu. However, exposing this via the appearance menu was somewhat unintuitive due to the way that most expressions work (i.e. one asset with several expressions), meaning that the appearance menu ended up with several unchangeable (except for color) entries.

This PR removes blush, fluids and emoticons from the appearance menu, and instead adds the ability to color expressions (where supported) through the expression menu (see screenshot below). This will allow blush, emoticons, fluids and the mouth to be colored from the expression menu (I excluded eyes because eyes are now split into two separate groups). This should make expression coloring more intuitive and convenient.

Because of the limited space at the top of the expression menu, the PR also combines the two wink buttons into a single wink/blink toggle that cycles through the 4 wink/blink combinations.

Screenshot of the functionality:
expression-color


elliesec:
This causes some conflicts with the next button in chatrooms. Converting to draft until that's resolved.

PR #1854, opened 2021-01-07
Bugfix: Fixes some font issues

elliesec:
## Summary

This fixes two font issues:
1. Fonts were not cached correctly when calling `CommonGetFont()` due to a bug with number arguments in `CommonMemoize`
2. The Kidnap minigame used to use the "Arial Narrow" font, which is a completely separate font, meaning `CommonGetFont()` would have " Narrow" appended to its result, which would not work. I've emulated this for generalised fonts as well as possible by changing the font weight to 300 (light) for those instances. It won't reproduce the previous look when using Arial exactly, but it will be consistent with the player's preferred font choice.


Ben987:
Thanks for the fixes!  Cheers.

PR #1855, opened 2021-01-07
Bugfix: Minor asset definition fixes

elliesec:
## Summary

A handful of small fixes to some asset definitions:
* Removes a duplicate `Clothing: true` property on the Corset asset group
* Removes duplicate `Random: false` properties from the Futuristic Cuffs, Ankle Cuffs and Leg Cuffs
* Removes the explicit `Priority` property from the Reverse Bunny Suit in the socks group - this was causing it to render above most items in the `ClothLower` group, and there don't seem to be any issues with it simply inheriting the socks group's priority


Ben987:
Awesome!  Thanks for the fixes.

PR #1856, opened 2021-01-07
Fix softlock when disconnected during login

jomshir98:
Properly clears the `LoginSubmitted` flag on server disconnect to avoid softlock when server crashes during player's (re)login.
( Also server disconnects are properly shown instead of "Validating name and password" )


Ben987:
I've never seen it myself, but thanks for the fix!

PR #1857, opened 2021-01-08
Flippers

EmilyR42:
Added a new par of swimming flippers as a shoe item. Relies on Nina's height rework.


Ben987:
Thanks!  It's a fun idea :)

PR #1858, opened 2021-01-08
Photography NPC Pose fix

Nina-1474:
The 'Relax your arms' option to make the Photography NPC return her arms to normal position was referencing the wrong thing.


Ben987:
Awesome!  Thanks :)

PR #1860, opened 2021-01-08
New Item: Vac-Cube

elliesec:
## Summary

Adds a new Vac-Cube item. Awesome new assets courtesy of Cecilia as always.

vac-cube


Ben987:
Thanks a lot you two, it's a fun idea :)

PR #1862, opened 2021-01-09
Feature: Adds new alternative light & dark chatroom themes

elliesec:
## Summary

This PR adds alternative light and dark chatroom themes to the game. I've been using these personally for ages, as I prefer the slightly cleaner look, but thought that maybe it's something other people might like to be able to try out as well. They're not radically different from the current light and dark themes, but they do modify the size of certain and positioning of certain chat elements. I've been using them personally for a long time, and have tested them with the various other chat preferences and not seen any problems with them (I've also verified that the new CSS doesn't interfere with the current light & dark themes).

"Light 2" Theme:
light-theme-2

"Dark 2" Theme:
dark-theme-2


Ben987:
Absolutely, this is a good concept.  Thanks!

PR #1863, opened 2021-01-09
Buttplug Tail Colour Splits

Nina-1474:
Splitting the ItemButt versions of the Kitty tail and Fox tails for multi-colouring which had already been split in TailStraps.


Ben987:
Very good!  Thanks for that one :)

PR #1864, opened 2021-01-09
Stable Exam End Fix

Nina-1474:
Bugfix for the removal of pony-exam items in the stable.


Ben987:
Yes!  Good catch there.

PR #1865, opened 2021-01-09
Height Rework Fixes

Nina-1474:
1. In rare cases when a character is particularly tall (for example is max height, wearing tall boots and has a ponytail), a portion of the top of their head would not be displayed. This cutoff was intentional so that ceiling restraints in the lower row of a chatroom would not spill up onto the upper row. However outside of chatrooms, there are cases where there is no 'ceiling' and we'd rather always display the character in full. The DrawCharacter calculations have been enhanced to make sure it only applies this cutoff where it needs to.
2. The asset zones in the arousal preferences screen were sometimes not lining up with the character. The zone functions have been corrected and further refactored.
3. Full-body alpha masks for WebGL drawing were misaligned. Now corrected.

I've redone the full set of parellel tests: poses, character heights, chatroom sizes, all NPC rooms, private rooms, assets of all types, etc. Also the new ceiling chains. The issues have been fixed and everything else still looks good.

### DIFF
--- a/BondageClub/Screens/Character/Appearance/Appearance.js
+++ b/BondageClub/Screens/Character/Appearance/Appearance.js
@@ -19,7 +19,7 @@ var CharacterAppearancePreviousEmoticon = "";
 var CharacterAppearanceMode = "";
 var CharacterAppearanceCloth = null;
 
-const CanvasUpperOverflow = 600;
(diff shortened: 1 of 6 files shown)


Ben987:
Thanks for the fixes!  Right on time for the beta :)

PR #1866, opened 2021-01-10
Water Cell improvement

Nina-1474:
Tweaking the Water Cell used in the magic show so that the bottom (which is no longer cut off) always touches the ground regardless of the character height.
Also added alpha masks for the sides.


Ben987:
Thanks for the fix!  I'm sure that git4nick would like it.

PR #1867, opened 2021-01-10
Missing devices toy text

Nina-1474:
The extended item screen's 'Select a type' text was missing for the LittleMonster and Familiar items.


Ben987:
Sure!  Thanks

PR #1868, opened 2021-01-10
Suspension Hogtie Tails Visible

Nina-1474:
Making tails visible for Suspension Hogties same as regular ones.


Ben987:
Your timing is perfect :)

PR #1871, opened 2021-01-11
Beta Fix: Fix for extended gags not gagging

elliesec:
## Summary

The new extended gag types were missing gagging effects, causing them not to gag when applied.


Ben987:
Thanks for the fix!

PR #1872, opened 2021-01-11
Beta Fix: Fixes broken extended item menu for the clothing latex corset

elliesec:
## Summary

When clothing corsets got split out into their own group, the extended item menu files for the latex corset weren't moved or renamed accordingly.


Ben987:
Great!  Thanks again.

PR #1874, opened 2021-01-11
Fixes LastChatRoom being forced on hardcore players

Ada18980:
Fixes a typo on setting the last chat room which causes the option to be forced on players whenever they open the preference screen. Should only affect Extreme mode now.


Ben987:
Thanks for the fix!

PR #1875, opened 2021-01-11
Fix for only the last chat room appearing in the chat search on exit

Ada18980:


Ben987:
Thanks for that one also.

PR #1876, opened 2021-01-11
Extended ball gag chat message fix

Ada18980:
Fixes the chat messages for changing ball gag types so it doesn't say "tightens" for both setting it to normal and tight


Ben987:
Alright!  Thanks :)

PR #1877, opened 2021-01-11
Fixes player being trapped on relog if the room they rejoin is full

Ada18980:
Fixes this by creating a new chatroom and adding the player's member number to the end.

It also fixes a bug with a newly created room being the wrong size


Ben987:
Thanks for these quick fixes.

PR #1878, opened 2021-01-11
All Fours Tails fixups

Nina-1474:
There were a few cases of the AllFours versions of tails not being updated to match their standard versions. To fix this and reduce image duplication, the AllowPose settings have been moved from group-level to asset-level and fixed up where needed.


Ben987:

PR #1879, opened 2021-01-11
Hands Missing Fix

Nina-1474:
Adding the colour schema in for Hands, so that they pass validation and don't disappear.
While this shouldn't be necessary, since they're hidden from the player and should only need Default + InheritColor from BodyUpper, all saved wardrobe outfits created back when Hands were editable will have them saved as Color = White/Asian/Black, meaning the schema is needed.


Ben987:
Yes!  Thanks for that fix.

PR #1880, opened 2021-01-11
Futuristic Full Harness

Ada18980:
Adds a simple new version of the futuristic harness that borrows some straps from the futuristic armbinder


Ben987:
Sure!  This is simple enough, we can slide it in the beta version.  Thanks.

PR #1881, opened 2021-01-11
Ceiling Chains fixes

Nina-1474:
Adding a prerequisite and effect to prevent the combination of ceiling shackles and being suspended upside-down.

Added the Freeze effect to prevent leaving rooms and kneeling.


jomshir98:
Thank you for the fix!

Ben987:
Yes!  Thanks for the fix :)

PR #1882, opened 2021-01-11
Bugfix: Expression reset button now resets expression colors as well (where applicable)

elliesec:
There was previously no easy way to reset expression item colors. I've added this functionality into the expression reset button for now. In the future, we might want to look at integrating "reset to default" functionality into the color picker itself.


Ben987:
Sure!  That's a good idea.  Cheers :)

PR #1883, opened 2021-01-11
DrawCharacter on iOS fix

Nina-1474:
A small fix to the new drawing method for how the width is calculated so that the height ratio is applied to the destination width instead of the source width. It worked fine on most browsers but on iOS using a source width greater than the canvas width failed to draw.


Ben987:
Thanks!  Can't forget the iOS users :)

PR #1885, opened 2021-01-11
Improvement: Timed Conditions

SandrineBC:
# Summary
Introducing timed Conditions, so that a player can use an item (Nursery Milk for example) and the condition applied with that item lasts, even if the item is removed.

### DIFF
--- a/BondageClub/Scripts/Character.js
+++ b/BondageClub/Scripts/Character.js
@@ -42,6 +42,7 @@ function CharacterReset(CharacterID, CharacterAssetFamily) {
 		HeightModifier: 0,
 		HeightRatio: 1,
 		HasHiddenItems: false,
+		Conditions: [],
 		CanTalk: function () { return ((this.Effect.indexOf("GagVeryLight") < 0) && (this.Effect.indexOf("GagLight") < 0) && (this.Effect.indexOf("GagEasy") < 0) && (this.Effect.indexOf("GagNormal") < 0) && (this.Effect.indexOf("GagMedium") < 0) && (this.Effect.indexOf("GagHeavy") < 0) && (this.Effect.indexOf("GagVeryHeavy") < 0) && (this.Effect.indexOf("GagTotal") < 0) && (this.Effect.indexOf("GagTotal2") < 0) && (this.Effect.indexOf("GagTotal3") < 0) && (this.Effect.indexOf("GagTotal4") < 0)) },
 		CanWalk: function () { return ((this.Effect.indexOf("Freeze") < 0) && (this.Effect.indexOf("Tethered") < 0) && ((this.Pose == null) || (this.Pose.indexOf("Kneel") < 0) || (this.Effect.indexOf("KneelFreeze") < 0))) },
 		CanKneel: function () { return ((this.Effect.indexOf("Freeze") < 0) && (this.Effect.indexOf("ForceKneel") < 0) && ((this.Pose == null) || (CharacterItemsHavePoseAvailable(this, "BodyLower", "Kneel") && (this.Pose.indexOf("Supension") < 0) && (this.Pose.indexOf("Hogtied") < 0)))) },
@@ -1001,6 +1002,11 @@ function CharacterSetFacialExpression(C, AssetGroup, Expression, Timer) {
 			}
 		}
 	}
+}
+
+function CharacterSetTimedCondition(C, Effect, Timer) {
+	CharacterAddEffect(C, Effect);
+	TimerCharacterRemoveConditionSet(C, Effect, Timer);
 }
 
 /**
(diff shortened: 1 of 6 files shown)


elliesec:
(just edited the server PR link for you @SandrineBC)

BondageClub/Scripts/Server.js
```diff
@@ -155,6 +155,8 @@ function ServerPlayerIsInChatRoom() {
 
 /** Sends a message with the given data to the server via socket.emit */
 function ServerSend(Message, Data) {
+	console.log(Message);
+	console.log(Data);
```
Nina-1474, line 159: I presume these were just meant for testing.

SandrineBC: Yes. I thought I purged them all with my first commit. Good point.

BondageClub/Scripts/Character.js
```diff
@@ -390,7 +392,8 @@ function CharacterLoadOnline(data, SourceMemberNumber) {
 			else
 				for (let C = 0; C < ChatRoomData.Character.length; C++)
 					if (ChatRoomData.Character[C].ID == data.ID)
-						if (ChatRoomData.Character[C].Appearance.length != data.Appearance.length)
+						if ((ChatRoomData.Character[C].Appearance.length != data.Appearance.length) ||
+							(ChatRoomData.Character[C].Conditions.length != data.Conditions.length))
```
elliesec, line 396: Same comment as above - `data.Conditions` could be `undefined` after a player's first login.

BondageClub/Scripts/Character.js
```diff
@@ -325,6 +326,7 @@ function CharacterOnlineRefresh(Char, data, SourceMemberNumber) {
 	Char.LimitedItems = Array.isArray(data.LimitedItems) ? data.LimitedItems : [];
 	if (Char.ID != 0) Char.WhiteList = data.WhiteList;
 	Char.Appearance = ServerAppearanceLoadFromBundle(Char, "Female3DCG", data.Appearance, SourceMemberNumber);
+	Char.Conditions = data.Conditions;
```
elliesec, line 329: The first time a player logs in after this change, their data bundle's `Condition` property will be undefined. This probably needs a check like the one for `LimitedItems` above.

elliesec (changes requested):
Looks to be working pretty well so far, but I had to fix the two comments to make it work without errors. I'll give it some further testing over the next couple of days, but it's looking promising from what I've seen so far!

BondageClub/Scripts/Character.js
SandrineBC: I added a check, like the one you proposed. I tested with new characters before and used existing ones and ran in no problems with them as well, but perhaps it's better to play it safe.

elliesec: Interesting - I was definitely seeing errors in console for `Character.Conditions` being undefined on existing characters. Putting a fix in here _should_ catch everything I think (the error in question was being thrown over in line 396) - I wasn't able to proceed without the game crashing until I put a check in here.

BondageClub/Scripts/Character.js
elliesec: I believe this will be okay with the fix in `CharacterOnlineRefresh`. `CharacterRefresh` initialises `Character.Conditions` to an array anyway, so provided there's nowhere else that overwrites that, we should be safe to assume that `Character.Conditions` is now always defined.

elliesec:
Now that we have asset type checks, you'll need to add type definitions for the new `Consumable` and `Condition` properties to `BondageClub/Tools/Node/AssetCheck_Types.js` to get the checks passing. Otherwise this looks good to me.

SandrineBC:
Like that?

jomshir98:
@SandrineBC Works, but it would be better to have more exact type than "Object array"; for example look on `OverrideHeight`or `RemoveItemOnRemove`.
The reason Layer is only set as `[Object]` is because it is checked separately

elliesec:
Pretty much - I'd go as far as to express the shape for conditions too. Something like:
```js
...
Condition: [{ Effect: Maybe [String], Timer: Number }]
...
```

SandrineBC:
Are we happy now?
😸

elliesec (approved):
Yep, I think so! All looks good to me!

NoneNoname:
This will allow adding all type of effect on others (Blind, Deaf, etc) which will be unremovable by normal ways, better way would be if the condition is only created as the item is added on the player then it updated from the player to the room and never from anyone to the player

elliesec:
> This will allow adding all type of effect on others (Blind, Deaf, etc) which will be unremovable by normal ways, better way would be if the condition is only created as the item is added on the player then it updated from the player to the room and never from anyone to the player

Yeah, this is a good point. It would be better if conditions only ever came from the player's client, rather than allowing players to add conditions to other players directly. As is, there's nothing stopping me from adding the baby talk effect to someone whenever I want, for as long as I want. It's currently mitigated by the fact that conditions disappear on login, but if we ever want to support persistent conditions, it would be safer for them to come from the player's side.

PR #1886, opened 2021-01-11
Beta Improvement: Adds AllowActivityOn to the Vac-Cube

elliesec:
Not really a fix, but a simple addition to the vac-cube, now that we have the `AllowActivityOn` property. Opens up several zones to activities when in the vac-cube.


Ben987:
Absolutely!  Thanks :)

PR #1887, opened 2021-01-11
Auto-Remake rooms now remakes private, adding player as admin

Ada18980:
Due to this:
https://cdn.discordapp.com/attachments/797897462694019093/798246477809057872/unknown.png

It became pretty clear that remaking private rooms as public will lead to a lot of spam (and also players being separated!). So instead of remaking them as public, they are instead remade as private with the caveat that the remaker is now an admin. So they can change the room parameters if they want to make it public, or choose to stay private, respecting people's wishes.

This has the side effect of making players less helpless, as the admin player can always make a private room public, wait for people to join, lock the room, and then F5 (since players cannot rejoin a locked room), but this was always possible if you were the room creator in the first place. It was also possible to use an alt account to get out of a solo room this way before.

The player is not made an admin after rejoining a public room


Ada18980:
There is also a problem with this if the original room doesnt get recreated, all of the players will be rejoined into different rooms.

Might be worth classing as a beta fix for that alone.

PR #1888, opened 2021-01-11
Hides the body when wearing a catsuit

Ada18980:
This is to prevent clipping due to the catsuit having a different model from the actual lower body. It covers the whole body so there is no reason to render the body layer.


elliesec:
Might be worth double-checking if you haven't already that there are no items that hide suits, but expect the body to be visible, in case we run into another wedding dress/leotard combination. It seems unlikely that anything would do that - but it's an edge case worth checking.

Ada18980:
The allfours pose will probably make the body disappear with this change.

PR #1889, opened 2021-01-11
Fix: Ceiling Shackles: Make the chain longer

jomshir98:
Make the chain longer to allow kneeling/sybian pose, without the chains cutoff mid-air


Ben987:
Sure!  Thanks for that one.

PR #1890, opened 2021-01-11
Futuristic Collar now allows player to block remotes on wearer

Ada18980:
There is a new checkbox on the collar that lets players block the use of remotes on the wearer. This simply disables any item that requires a remote, except the lover's vibrator (since that one has a very special use).


Ben987:
Alright!  Thanks for that new feature.

PR #1892, opened 2021-01-12
Spread Legs Boot fix

Nina-1474:
Boots and feet go wonky when spread-eagle tied by the legs to the bed. 
The setting to hide boots has been moved from X-Cross (which was the only Spread applying item before) into the pose itself.


Ben987:
Great!  Thanks for the fix.

PR #1894, opened 2021-01-12
Kicked Room Rejoin fix

Nina-1474:
Adding back in the function to clear the room data when kicked to prevent immediately re-entering it.


Ben987:
Absolutely!

PR #1896, opened 2021-01-12
Extreme Difficulty - Locks accessible to Whitelist

Nina-1474:
The Extreme difficulty currently forces the Combination and Password locks to be limited permissions to prevent griefing. However when on the 'Owner, Lover and whitelist' only permissions, limited items become inaccessible to the whitelist instead of everyone else. Someone they've whitelisted is someone they trust, but the player can't change the locks back to not-limited to make them accessible.
This change makes an exception to not automatically limit the two locks when on the Owner/Lover/whitelist permission only, so that their whitelisted players can use them.


Ben987:
There's an extra lock in that group now, can you resolve the conflict and test again?  Thanks.

Nina-1474:
Resolved so that the new lock is included, and retested.

Nina-1474:
I've also moved things a bit so that the list of password locks only has to be defined in one place (MainHall.js) instead of two. I retested both features to ensure they still work.

Ben987:
Sorry, there are new conflicts with other pull requests, double you resolve once again and tests?  Thanks.

Ben987:
Thanks a lot!  Sorry you had to double-check twice.  Cheers!

PR #1898, opened 2021-01-13
Feature/Beta Fix: Adds a "reset to default" button to the color picker

elliesec:
## Summary

This is something of a follow-up to #1882, as I feel that fully resetting a character's expression to reset their expression color to default doesn't fully meet the needs of players. This does two things:
* Adds support for the special "Default" keyword to the color picker - typing "Default" into the input box will reset an item (or layer) to its default color
* Adds support for a new reset button to the color picker which will reset an item (or layer) to its default color (by auto-filling the input with "Default")

Below is an image of the updated color picker with the default button:

color-picker-default


Ben987:
That's a good idea, we might need a third beta if it triggers some bugs though.  We still have time for it, thanks :)

PR #1899, opened 2021-01-13
Ceiling Ropes and Chains

Ayesha-678:
two new items in the addon section of the body, ceiling ropes/chains they can be applied if you have a collar, arm item, torso item, or pelvis item. they can ALSO be used to suspend someone.


Ben987:
Absolutely!  I'm sure that players will love it.  Thanks :)

PR #1900, opened 2021-01-13
Blocked OOC Gagged Workaround Fix

Nina-1474:
When the 'Block OOC when gagged' setting is on, messages like "(OOC talk)" would be blocked but "something (OOC talk)" would not. Fixed the check to block the latter as well.


Ben987:
Thanks!

PR #1904, opened 2021-01-14
Notifications

Nina-1474:
## Feature
When a player misses a particular event, this feature can change the tab/window header from "Bondage Club" to "(1) Bondage Club" to notify the player of it. The number increases for each notification and is reset after an appropriate action.

## Preferences
There's a new Notifications preferences screen to configure which events a notification can be raised for. There's a Reset button just to cover any unforeseen cases where the intended process to clear notifications doesn't work, as a stopgap until they're fixed.

## Beeps
If the player is never focused on the screen for the duration that the beep message banner is displayed, a notification is raised. The notification is cleared when the player views the 'Beep history' screen.

## Chat Messages
If the player is either not focused on the screen when a new message is posted to a chatroom's chat log, or is not scrolled to the bottom, a notification is raised. This occurs for normal messages, emotes (text inside **) and whispers. They can optionally also be raised for messages prompted by item changes or arousal activites where the player is involved. 
The notifications are reset if the player refocuses on the window and/or scrolls to the bottom of the chat log. It also resets if the player loads the Chat Search screen, to cover cases where they're ejected from a room.


Ben987:
That could be useful.  Have you tested it on many browsers and on mobile devices?  

There's also a code conflict in chatroom.js, can you resolve it?  Thanks.

Nina-1474:
Conflict resolved.
I've tested this on several browsers. I don't have a way to test on mobile but I don't think this shouldn't cause any issues there as the only potential problems areas I can think of are the additions of `document.title` and `document.hasFocus()`, which are both supported on mobile. At worst the notifications just won't be easily visible.

Ben987:
Excellent!  I also don't test much on mobile, we'll see if it triggers problems when the beta version rolls in.  Thanks!

PR #1905, opened 2021-01-14
Beta Fix: Slave Collar

elliesec:
## Summary

Fixes incorrect function calls when an owner tries to add/remove a sub's slave collar.

## Steps to reproduce

Two characters:
* Character A
* Character B - owner of Character A

1. Character B attempts to add/remove Character A's slave collar through chatroom dialog
2. Observe that the collar is not added/removed


Ben987:
Good catch, thanks for that one.  Cheers :)

PR #1906, opened 2021-01-14
Beta Fix: Hides the new corset slot in the AllFours pose

elliesec:
## Summary

The `AllFours` pose previously hid bras, but when corsets got moved out of the bra slot, the new slot wasn't added to the `Hide` array for the pose, resulting in corsets appearing like this:


Ben987:
Pretty funny bug, thanks for the fix :)

PR #1907, opened 2021-01-14
Bugfix: Allows timer locks to trigger chatroom messages in other chatroom subscreens

elliesec:
## Summary

Not strictly a bug with the current beta, but when timer locks expire, they won't trigger chatroom messages if the player is in their wardrobe, information sheet, etc. This fixes that.


Nina-1474:
Maybe this should use the recently created `ServerPlayerIsInChatRoom()` which covers the full set of chatroom "subscreens".

elliesec:
> Maybe this should use the recently created `ServerPlayerIsInChatRoom()` which covers the full set of chatroom "subscreens".

I had no idea that was a thing! Thank you!

Ben987:
Thanks for the little fix!

PR #1908, opened 2021-01-14
Refactor: Wand Rope belt to extended vibe

SandrineBC:
# Summary
Refactors the wand rope belt to the extended vibrator script


elliesec (approved):
Looks good to me.

PR #1909, opened 2021-01-14
Refactor: Wand belt to extended vibrator script

SandrineBC:
# Summary
Refactor the wand belt to use the extended vibrator script.


elliesec (approved):
Looks good to me.

PR #1910, opened 2021-01-14
Refactor: Vulva vibrating egg to extended vibe

SandrineBC:
# Summary
Refactor the vulva's vibrating egg to use the extended vibrator script


elliesec (approved):
Looks good to me.

PR #1912, opened 2021-01-14
Suspension Pose fix

Nina-1474:
The Suspension pose doesn't actually affect the arms or legs, as that's always determined by another pose it's paired with (e.g. LegsClosed). So it doesn't need to belong to the BodyFull category which makes it block arms/leg pose changing.


Ben987:

PR #1916, opened 2021-01-14
Refactor: Vibrating butt plug to extended vibrator

SandrineBC:
# Summary
Refactors the vibrating butt plug to the use of the extended vibrator script


elliesec (approved):
Looks good to me.

PR #1917, opened 2021-01-14
Confirmation dialog for calling maids to leave the room

Ada18980:
Because players were surprised that they would lose 40 dom rep. Also fixes a bug that players won't leave on other players' screens.


Ben987:

PR #1920, opened 2021-01-15
Bugfix: Removes an unneeded console.log call

elliesec:
I just noticed that I'd left in a `console.log` statement that I was using to debug the recent color picker change which is no longer necessary.


Ben987:
Yes!  I did not see that one.  Thanks :)

PR #1921, opened 2021-01-15
College Entrance Uniform Colour fix

Nina-1474:
It's possible for the college uniform to have a Color value of "Default" instead of the more typical ["Default","#202020","Default"] which was not allowing the player inside the college despite being the right colours. Now fixed.


Ben987:

PR #1923, opened 2021-01-15
Default Colour Check

Nina-1474:
A function to check if an asset currently has the default colours. 
Made to replace the hardcoded colour checks at the College Entrance with something more generic/reusable.


Ben987:
This one creates a conflict with a previous pull request.  Can you double-check and test?  Thanks a lot.

Nina-1474:
Resolved and retested.

Ben987:
Awesome!  Thanks a lot.

PR #1924, opened 2021-01-15
Bugfix: Prevents the bed spread eagle hemp rope type from ignoring toe ties

elliesec:
## Summary

The new bed spread eagle pose for the feet hemp rope did not include a `LegsOpen` prerequisite. In order to add this I needed to add a validate function (similar to the web in the arms group), to prevent the rope from conflicting with itself.

## Steps to reproduce

1. Add a toe-binding item (e.g. hemp rope in `ItemBoots`)
2. Add hemp rope to the `ItemFeet` group
3. Select the bed spread eagle type
4. Note that the player is forced into the legs spread pose, but the toe tie is still applied


Ben987:
Great!  Thanks a lot for that fix.

PR #1925, opened 2021-01-16
Timer Password Padlock

Ada18980:
Adds a password lock that also has timer functionality based on the mistress padlock. Timer can only be changed by the person who set the lock.


BondageClub/Scripts/Inventory.js
```diff
@@ -765,6 +765,10 @@ function InventoryUnlock(C, Item) {
 		delete Item.Property.Password;
 		delete Item.Property.Hint;
 		delete Item.Property.LockMemberNumber;
+		delete Item.Property.Password;
+		delete Item.Property.CombinationNumber;
+		delete Item.Property.LockSet;
+		delete Item.Property.Hint;
```
Nina-1474, line 771: All but the CombinationNumber are actually already deleted.

Ada18980:
Thanks Nina!

Ben987:
Great!  Can't go wrong with more locks :)

PR #1926, opened 2021-01-16
Fix: Raise wooden sign priority

SandrineBC:
# Summary
Raises the wooden sign to priority 57 so that is displayed in front of the display frame.


Ben987:
Very well.  Thanks!

PR #1927, opened 2021-01-16
Pilot suit fix

Ada18980:
Stops the upper pilot suit from hiding the butt and vulva...


Ben987:

PR #1928, opened 2021-01-16
Extended Item Self-Blocking Validation

Nina-1474:
Web and hemp hope on feet both have a custom validation function to work around cases where the item is blocking itself from being changed by failing the prerequisite check of the new type.
This moves the remove-validate-reapply process into the general ExtendedItem script and adds a new property that calls it, and only to the item types that need it.

Also added the setting to hemp rope and chains for arms to fix that same issue for them.


Ben987:
Thanks for that fix!  I will try it soon.  Cheers :)

PR #1930, opened 2021-01-16
Fixed return to chatroom allowing people to create non-asylum rooms

Ada18980:


Ada18980:
Tested with 2 accounts, seems to work out.

Ben987:
Thanks for the fix!

PR #1931, opened 2021-01-17
Feature: Item Opacity/Wooden Box enhancements

elliesec:
## Summary

I'm afraid this is quite a big one. It adds a few new features:

* Adds the ability to set an item's opacity. By default, items have opacity changes disabled to prevent console griefing (e.g. turning items invisible that shouldn't be).
* Adds a new `ElementCreateRangeInput` function, for creating range (slider) inputs, and adds corresponding CSS for styling these ranges - I've tried to make sure that slider styles are compatible with all major browsers, and have tested on Chrome, Firefox, Chromium Edge and old Edge. I've not tested on Safari (as I don't have a Mac), but as Safari uses WebKit, it should behave in the same way as Chrome. If not, it will just fall back to the default native range input.
* Extends the Wooden Box item: the box now has a back and a front asset (the back renders behind the player). The extended item screen allows players to change the front asset's opacity (revealing the player inside), and also allows custom text to be written on the front of the box.

Aside from the box, there are quite a few items that could utilise the transparency functionality: other boxes, the fairy wings, and for example certain clothing items could be set with a minimum transparency of (for example) 0.8, to allow for semi-transparent t-shirts, blouses, etc.

Items and individual layers can be configured with a minimum and maximum opacity (these are all set to 1 unless explicitly defined, which should prevent console users from abusing the functionality on other players).

The gif below demonstrates the new functionality:

opacity-demo-3

## Implementation Notes

Below is a summary of the technical changes in this PR:
* New `Opacity`, `MinOpacity`, and `MaxOpacity` properties that can be set on assets. Unless specified otherwise in `Female3DCG.js`, all three properties are set to 1 (that is, fully opaque). These properties are checked both on incoming item updates, and at the drawing level to ensure that items can't be turned transparent unless the item allows it.
* A new `HasImage` property that can be set at the layer level. Setting it to `false` prevents the drawing code from drawing that layer, but dynamic drawing functions are still called. This removes the need for dummy empty images which are used for dynamic-drawing only layers
* New `ElementCreateRangeInput` function for creating slider inputs
* New `CommonCallFunctionByNameWarn` function, which behaves the same as `CommonCallFunctionByName`, except warns if the function doesn't exist. Calls to the dynamic drawing functions have been replaced with this, which should prevent crashing when contributors are adding new items with dynamic drawing functions, outputting a (hopefully) useful warning message instead.
* Updates to `CommonDraw.js`, `Drawing.js`, `GLDraw.js` and `Appearance.js` to handle the new opacity option
* Removes some visual artefacts in the Studded Blindfold's preview image (not apparent in the preview, but became apparent when using it as a slider thumb)


Ben987:
Let me review the smaller ones first and we can test that one after.  Thanks!

Ben987:
This one is huge!  Lets hope it won't create any bugs, it could be useful for lots of items though.  Thanks a lot!

PR #1933, opened 2021-01-18
Hotfix: Correct a naming error on the ceiling cuffs layers

SandrineBC:
# Summary
The name of the ceiling cuffs layers was starting with 'ItemLegs' instead of 'ItemArms'. This PR fixes that.


Ben987:
Thanks!  I will prepare a hotfix for it.  Cheers :)

PR #1934, opened 2021-01-18
Bugfix: Fix for chess square overflow in Firefox

elliesec:
## Summary

This fixes an issue where the squares of the chessboard overflow the rows at certain window widths in Firefox (see image below). This is likely due to some internal rounding error in the way Firefox calculates the widths/positions of floated elements. I've added a new `Chess.css` file, separate from the existing library file, which modifies the CSS on certain chessboard elements to make it a bit more robust. It also moves the file letterings over to the right of the squares so that they don't overlap with the rank numbers. Tested and working on Firefox, Chrome and Edge. I've not tested Safari, as I don't have a Mac, but there shouldn't be anything in the added CSS that should cause issues.

Before fix:

After fix:
image

### DIFF
--- a/BondageClub/CSS/Chess.css
+++ b/BondageClub/CSS/Chess.css
@@ -1,0 +1,14 @@
+/* Fixes for square overflow in Firefox (overwrites default styling) */
+.row-5277c {
+  display: flex;
+}
+
+.clearfix-7da63 {
+  display: none;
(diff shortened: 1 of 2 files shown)


Ben987:
Thanks!  I will prepare a hotfix for that one.  Cheers :)

PR #1935, opened 2021-01-19
Fix: Aligns the Up-Button with the left hand side ones

SandrineBC:
# Summary
Aligns the "Up" button's height to the other buttons of the left hand side (namely face and pose changes).
Before:
image
After:
image
This is solely UI-design related, so no extensive testing was done. The button still works as intended...


Nina-1474:
You've changed the Draw function but there needs to be a corresponding change in the Click.

Ben987:
Thanks!  Let me know when the click is also ready and I'll merge it.

SandrineBC:
Fixed this. Things like that are easy to miss.

Ben987:
Awesome!  Thanks a lot :)

PR #1936, opened 2021-01-19
Improvement: DynamicGroupName enhancements & duplicate asset removal

elliesec:
## Summary

Sorry, this is another really big one, but most of it is removing duplicated assets - there are very few actual code changes. I've discussed the changes with @Nina-1474, and we're pretty sure this is a safe change. This PR does two main things:

* Revises the `DynamicGroupName` property (which was previously only used by the handheld toys) on assets so that `CommonDraw.js` now uses `A.DynamicGroupName` instead of `AG.Name` to find an item's assets.
* Adds a new `OverrideAllowPose` property to assets which will override any poses in the `AllowPose` array defined at the group level

The first change means that assets that are duplicated across groups (for example gags, corsets, ears, etc.) now no longer need to be. By setting `DynamicGroupName`, you can redirect the game to the dynamic group when it fetches assets (for example, this PR modifies the corsets in `ItemTorso` to have `DynamicGroupName: "Corset"`, meaning corsets in `ItemTorso` now fetch their images from the `Corset` directory).

The second change means that it is no longer necessary to duplicate an item's assets across poses. For example, `ItemTorso` has (had) several entries in its `AllowPose` that most items in the group didn't have unique assets for, but they still had to provide duplicated assets to prevent 404s at runtime.

Between these two changes, we should be able to massively reduce the total number of assets in the game, which should have a substantial impact on both bandwidth usage, and the game's memory footprint (because fewer textures need to get stored in the `GLDraw` texture cache).

Because of the huge number of assets that we _could_ potentially remove with this change, I've limited this PR to only the `ItemTorso` and `Corset` groups. In the future, we could remove assets that are duplicated across poses from all groups, and many assets that are duplicated across groups (notably gags, ears, tails/butt plugs, shoes, and probably a few others), which would reduce the game's memory usage substantially.

I've also added one other property to assets: `HideForPose`, which takes a list of pose names. These assets will not be drawn when the character is in the named poses. This is largely because several items do not currently support the `Hogtied` or `AllFours` poses, and this prevents them from having to provide blank dummy assets.

## Testing

I've manually checked every asset that has been changed by this PR. Assets which have been removed have been checked to be identical to (using MD5 hashes), or visually indistinguishable from (manual check by eye), existing assets, and these have been deduplicated as appropriate. I've tested out every asset several times in game, in every supported pose, and am confident that this should create no visual differences to the game.


Ben987:
This is huge!  Let me check the smaller changes first and I'll get back to that one.  Thanks a lot.

Ben987:
This is HUGE!  I won't be able to test all of it, but having less assets will certainly help with loading times.  Thanks a lot.

PR #1937, opened 2021-01-19
Bugfix: Removes the medical restraints from the random pool

elliesec:
The medical restraints weren't removed from the random pool, which could lead to some strange issues.


Ben987:
Good catch!  I will prepare a hotfix :)

PR #1938, opened 2021-01-19
Bugfix: Fixes the layer names for the Futuristic arm cuffs

elliesec:
Fixes missing layer names for Futuristic Cuffs in `ItemArms`. Also removes 2 unused layer name entries.


Ben987:
Excellent.  I will also hotfix that one.

PR #1939, opened 2021-01-19
Bugfix: Fixes posture collar name orderings

elliesec:
## Summary

As part of #1794, the strict posture collar and the latex posture collar got moved in `Female3DCG.js`, but their names weren't moved inside `Female3DCG.csv`, causing some of the collar names to show up incorrectly in game. This moves the names in the CSV to match their order in `Female3DCG.js`.


Ben987:
Very good one also.  I will add it to the batch.  Thanks a lot.

PR #1940, opened 2021-01-19
Vacbed + Flippers edit

EmilyR42:
- Gave the vacbed options to hide/show hair and hats/hair accessories
- Vacbed graphic tweaked to remove latex on forehead beneath hair
- Flippers are now hidden with most full-body items


Ben987:
That's a pretty big one, thanks a lot.  I might not be able to test all the combos but we have betas for that.  I'm sure the players will like it :)

PR #1942, opened 2021-01-20
Bugfix: Fixes alpha mask issues with the open straitdress

elliesec:
The alpha masks on the open straitdress were being applied to all layers, resulting in visual issues with some items (see image). This changes the masks to target specific groups, and also revises the `Hide`/`HideItem` arrays. This should remove any bad combinations, for now.

image


Ben987:
Thanks for the fix!  I will hotfix the live version.

PR #1943, opened 2021-01-20
Tweaked tail heights

wildsj:
Tweaked the height many of the tails are drawn at. Some butt plugs tails looked they were coming out half way up the characters spine.

They are now adjusted so the tails on butt plugs should look like they are coming from the anus and the strap plugs so they are coming fron the coccyx, slightly higher up.

The images are unchanged, only individual Top: values added as needed.


Ben987:
Absolutely!  Thanks for reviewing the heights.

PR #1944, opened 2021-01-20
Chains Feet LegsClosed fix

Nina-1474:
The chains on feet options weren't changing the pose to close the legs like they should have.


Ben987:
Very good, I will prepare a hotfix.  Thanks a lot :)

PR #1945, opened 2021-01-20
Chatroom Cut-off Name

Nina-1474:
When a chat room has two rows, fall letters in names (g,j,p,q,y) get cut off at the bottom. This often makes them quite hard to read, for instance trying to distinguish v from y or i from j.
This change will move the names up enough to make the letters readable, tested across various fonts. It will only apply in this two-row-room situation, since in every other instance there's already space for the name.


Ben987:
Yes!  I've seen that one before.  Thanks for fixing it :)

PR #1946, opened 2021-01-21
Bugfix: Adds the Corset group to the Hide array for several items

elliesec:
A few existing items which previously hid the `Bra` slot were missed when the corsets got moved out into their own slot, meaning some previously-hidden corsets becoming visible, resulting in visual errors like below, as well as some clipping issues. I've manually tested each of the items that have been changed. I've also added the corsets with garters to the new `HideItemExclude` array for some of the leotards, so that the garters should still display from underneath the leotard.

image


Ben987:
Sure!  I haven't checked the other pull requests yet so I can slide another late hotfix.  Thanks :)

PR #1947, opened 2021-01-21
Hollow butt plug

wildsj:
Item added with no special effects.


Ben987:
Yes!  Another kinky item for our kinky gang.

PR #1948, opened 2021-01-21
Improvement: Adds retry logic to CommonGet

elliesec:
## Summary

This PR adds retry logic to the `CommonGet` function. Occasionally, GET requests (usually for CSVs) will fail, usually due to an intermittent error. That then results in textual errors like those below, which can only be resolved by exiting/re-entering the screen, or in some cases by refreshing the page completely. This change should prevent the vast majority of those cases. If a request succeeds first time, the current behaviour is preserved.

image

image

## Technical Details

Retries are backed off exponentially up to a limit of 60 seconds between retries, and by default a request is retried a total of 10 times. If no successful response is received after 10 retries, the function will stop retrying.

This is the console log output and network requests for a request that has failed 10 times:

retries-1

And this is the output for a request that succeeded after a few retries:

retries-2

### DIFF
--- a/BondageClub/Scripts/Common.js
+++ b/BondageClub/Scripts/Common.js
@@ -196,8 +196,8 @@ function CommonGet(Path, Callback, RetriesLeft) {
  * retries up to a maximum of 10 times.
  * @param {string} Path - The path of the resource to request
  * @param {function} Callback - Callback to execute once the resource is received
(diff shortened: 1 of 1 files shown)


Ben987:
Sure!  A late retry with a timer might help with players loading new versions.  Thanks :)

PR #1950, opened 2021-01-22
Bugfix: Adds a missing AllowBlock entry to the Pet Suit

elliesec:
The PetSuit was missing an `AllowBlock` entry for `ItemVulva`, meaning that the item never blocked the vulva zone, even when zipped.


Ben987:
The timing is great, I haven't had time to review the latest changes, I will hotfix it.  Thanks!

PR #1953, opened 2021-01-23
Improvement: Simple chatroom optimisations

elliesec:
## Summary

Last one for today! This PR makes some simple optimisations to the chatroom code to reduce the amount of work done when just idling in a chatroom. I selected the optimisations based on running a CPU profiler over the chatroom, and identifying optimisations to functions (outside of the draw functions) that took a notable amount of time. They aren't huge performance improvements, but they're a couple of simple "easy wins" to reduce the game's overhead in chatrooms, and might benefit those with slower devices.

The changes that were made are:

* Reduces duplicated use of `Player.GetGagLevel()` and `Player.IsSlow()` by calculating once and assigning to a variable
* Reduces the overhead of the `ChatRoomTarget()` function - this gets called every frame and was accounting for about 7% of CPU time when idling in a chatroom (presumably due to the text lookup and DOM manipulation). Chatroom target updates now go through a new `ChatRoomSetTarget()` function, which sets a dirty flag accordingly. The main body of `ChatRoomTarget()` then only runs if the dirty flag is set, reducing the function's CPU time when idling to < 0.1%


Ben987:
Very good!  We'll see if it helps a little with game performance :)

PR #1955, opened 2021-01-23
Dev feature: Asset check script

jomshir98:
Script for checking assets (to be exact types of their properties for now, I might extend the script in the future)
The script does not run as part of BondageClub, it is external tool
As discussed with @elliesec 

Fixes for errors found by the script are included in this PR


Ben987:
Sure!  A script to validate the assets file is a great idea.  Thanks!

PR #1957, opened 2021-01-24
Bugfix: Adjusts the yoked lock asset for the Futuristic Arm Cuffs

elliesec:
The lock asset for the Futuristic Arm Cuffs in the `Yoked` pose was in the wrong place (copy of the default pose lock asset). This adjusts the lock asset to correctly match the position of the cuffs.


Ben987:
Very good!  Thanks.

PR #1958, opened 2021-01-24
Improve (re)login and ForceDisconnect handling

jomshir98:
Makes login, re-login and password reset buttons gray and unclickable while not connected to server.

Improves handling of `"ForceDisconnect"` event, displaying the error message and either properly disconnecting or attempting to reconnect in case of `"ErrorDuplicatedLogin"`


Ben987:
Sure!  I'll also check the server side changes pretty soon.  Thanks a lot.

PR #1959, opened 2021-01-25
Bugfix: Prevents non-owners/lovers from adding owner/lover locks

elliesec:
## Summary

There is currently a loophole in the validation for appearance bundles which allows non-owners/non-lovers add owner/lover locks through the console. This PR addresses that. I've also refactored the code for deleting invalid locks into a single `ServerDeleteLock` function, as it was getting repeated a lot.

## Steps to reproduce

Two characters required: Character A & Character B. Character B should not be Character A's owner
1. Apply a lockable item to Character A
2. As Character B, run the following in the console (assuming `ChatRoomCharacter[1]` refers to Character A, the item is in the `ItemArms` group, and `2` is the member number of Character A's owner):
```js
InventoryLock(ChatRoomCharacter[1], "ItemArms", "OwnerPadlock", 2);
ChatRoomCharacterUpdate(ChatRoomCharacter[1])
```
3. Note that an owner lock has been added to Character A (and made it through validation)

The same steps can be repeated by a non-lover to apply lover locks.

### DIFF
--- a/BondageClub/Scripts/Server.js
+++ b/BondageClub/Scripts/Server.js
@@ -286,19 +286,7 @@ function ServerValidateProperties(C, Item, Validation) {
 				    // Check if a lock on the items sub type is allowed
 				    (Item.Property && Item.Asset.AllowLockType && !Item.Asset.AllowLockType.includes(Item.Property.Type))
 			    )) {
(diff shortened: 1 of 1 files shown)


Ben987:
Yes!  Thanks for that fix.

PR #1961, opened 2021-01-25
Futuristic harnesses now have same priority as other harnesses

Ada18980:
Should go on top of the futuristic bra now~


elliesec:
Just an FYI - this looks like it's going to have some pretty substantial conflicts with #1936, so whichever one goes in first, that's probably something to be aware of. If this gets merged first, I'll likely be consulting you when I resolve my merge conflicts.

PR #1962, opened 2021-01-25
Github action: Asset check

jomshir98:
Creates GitHub Action for checking asset definition, that is automatically run on all PRs and pushes to master
This is a followup automating #1955

Also contains fixes for new errors since the last PR

> GitHub Actions usage is free for both public repositories and self-hosted runners.


Ben987:
@jomshir98 , so how does that work?  The script will generate a PR if there's something wrong with the asset file?  What if someone wants to add a new property to the assets, will that create an unwanted pull request?

elliesec:
@Ben987 - this won't generate a PR. With this in, Github should automatically run the asset checking script over any PR that anyone raises to check for asset issues (and the results should appear in the "Checks" tab up at the top of the PR). It won't generate any new code or pull requests or change anything on its own - but it'll flag up issues with the PR, which would be on the person contributing to fix.

For example, if I created a pull request that added a new item with `Priority: "someString"` rather than a number, the script will run over that pull request, identify that error, and that will be displayed in the "Checks" tab. Then it would be on me (the PR author) to fix that issue (by changing the priority back to a number), and update my PR. After updating, the script would run again, confirm that there are no more errors, and the PR would be safe to merge.

Ben987:
That sounds good to me, it should help to prevent errors.  Thanks a lot!

PR #1963, opened 2021-01-25
Grill and spatula

Ada18980:
Adds a grill and spatula for picnics


EmilyR42:
Barbecue

It's not for picnics.

Ben987:
There's no controversy, thanks for the grill :)

PR #1964, opened 2021-01-26
Lock Picking RNG tweaks and UI changes

Ada18980:
Reduces the RNG on lock picking by varying pin height based on order. Depending on your skill (higher lockpicking skill means a more pronounced visual difference), pins that are earlier in the order move less (because they are binding harder) and pins earlier in the order move more. 

This also changes the difficulty of the minigame to adjust for the fat that it is now a memory game. Picking locks is going to be pretty challenging without the memory aspect, or having a high skill.

This also changes the game to no longer consume a try when getting a pin right, and reduces the number of tries accordingly.


jomshir98:
@Ada18980 Looks like you've forgotten now unused `IsPickable` in `AssetAdd`. Could you please remove it?

jomshir98:
Also `PickDifficulty` defaults to `false`, but on some places is compared with `null`, which fails! Would you replace `false` with `null`?

Ben987:
There's also a few conflicts with other pull request, please resolve them and test again.  Thanks a lot!

Ada18980:
@jomshir98 @Ben987 I made the requested changes

Ben987:
Thanks!  I will test it soon.

PR #1966, opened 2021-01-26
Bugfix: Fixes for hidden chat message handling

elliesec:
## Summary

This fixes a few issues with the handling of hidden chat messages, probably largely as a result of bad merge resolutions. This PR:
* Removes duplicated handling of `MaidDrinkPick`, `PayQuest` and `OwnerRule` messages
* Changes remaining `if` statements into `else if` statements for efficiency
* Removes the blanket `return` at the end of the hidden message handling block - this was causing owner & lover rules to not show up in the chat (their type gets modified when the hidden message is handled, and the type is no longer `Hidden` after modification)


Ben987:
Very good!  Thanks for the fixes.

PR #1956, opened 2021-01-23
Longer Bio & Chat Messages

Shiranui-Izayoi:
Lengthened the Chat Messages to 1000 characters so 4 times the original length because honestly 250 is way to short for more detailed RP.

Lengthed the max length of Bios to 10,000 characters as it is crazy hard to explain yourself in 1000 characters.  While most likely won't use it, it is nice for those that do.

If there is a reason that it is limited like this then please let me know. But to my knowledge there is no reason for it to be like this. Since it only hurts those that want to go into more detail. Like myself.

PR #1967, opened 2021-01-26
Longer messages & bio (with bio compression)

jomshir98:
Extension of #1956 by @Shiranui-Izayoi 

> Lengthened the Chat Messages to 1000 characters so 4 times the original length because honestly 250 is way to short for more detailed RP.
> 
 > Lengthed the max length of Bios to 10,000 characters as it is crazy hard to explain yourself in 1000 characters. While most likely won't use it, it is nice for those that do.
> 
> If there is a reason that it is limited like this then please let me know. But to my knowledge there is no reason for it to be like this. Since it only hurts those that want to go into more detail. Like myself.

Additionally the Bio is compressed (if it is efficient or required) on update and decompressed on character load.

**Breaking changes:**
Current version users will see beta user's description as `╬` + lots of random characters
Beta users will see current version user's description without problem
If current version user has description, that literally starts with `╬` It will be lost when they first try editing it in Beta/Next version.

Originally was prefixed with `--compressed-`, but changed to `╬` on Ben's request.

### DIFF
--- a/BondageClub/Screens/Character/Login/Login.js
+++ b/BondageClub/Screens/Character/Login/Login.js
@@ -361,7 +361,7 @@ function LoginResponse(C) {
 			if (CommonIsNumeric(C.Money)) Player.Money = C.Money;
 			Player.Owner = ((C.Owner == null) || (C.Owner == "undefined")) ? "" : C.Owner;
 			Player.Game = C.Game;
-			Player.Description = (C.Description == null) ? "" : C.Description.substr(0, 1000);
+			Player.Description = (C.Description == null) ? "" : C.Description.substr(0, 10000);
 			Player.Creation = C.Creation;
 			Player.Wardrobe = CharacterDecompressWardrobe(C.Wardrobe);
 			WardrobeFixLength();
--- a/BondageClub/Screens/Character/Login/Login.js
+++ b/BondageClub/Screens/Character/Login/Login.js
@@ -361,6 +361,9 @@ function LoginResponse(C) {
 			if (CommonIsNumeric(C.Money)) Player.Money = C.Money;
 			Player.Owner = ((C.Owner == null) || (C.Owner == "undefined")) ? "" : C.Owner;
 			Player.Game = C.Game;
+			if (typeof C.Description === "string" && C.Description.startsWith("--compressed-")) {
+				C.Description = LZString.decompressFromUTF16(C.Description.substr(13));
+			}
 			Player.Description = (C.Description == null) ? "" : C.Description.substr(0, 10000);
 			Player.Creation = C.Creation;
 			Player.Wardrobe = CharacterDecompressWardrobe(C.Wardrobe);
--- a/BondageClub/Screens/Character/Login/Login.js
+++ b/BondageClub/Screens/Character/Login/Login.js
@@ -361,8 +361,8 @@ function LoginResponse(C) {
 			if (CommonIsNumeric(C.Money)) Player.Money = C.Money;
 			Player.Owner = ((C.Owner == null) || (C.Owner == "undefined")) ? "" : C.Owner;
 			Player.Game = C.Game;
-			if (typeof C.Description === "string" && C.Description.startsWith("--compressed-")) {
-				C.Description = LZString.decompressFromUTF16(C.Description.substr(13));
+			if (typeof C.Description === "string" && C.Description.startsWith("╬")) {
+				C.Description = LZString.decompressFromUTF16(C.Description.substr(1));
 			}
 			Player.Description = (C.Description == null) ? "" : C.Description.substr(0, 10000);
 			Player.Creation = C.Creation;
--- a/BondageClub/Screens/Character/OnlineProfile/OnlineProfile.js
+++ b/BondageClub/Screens/Character/OnlineProfile/OnlineProfile.js
@@ -48,8 +48,8 @@ function OnlineProfileExit(Save) {
     if ((InformationSheetSelection.ID == 0) && (InformationSheetSelection.Description != ElementValue("DescriptionInput").trim()) && Save) {
         InformationSheetSelection.Description = ElementValue("DescriptionInput").trim().substr(0, 10000);
         let Description = InformationSheetSelection.Description;
-        const CompressedDescription = "--compressed-" + LZString.compressToUTF16(Description)
-        if (CompressedDescription.length < Description.length || Description.startsWith("--compressed-")) {
+        const CompressedDescription = "╬" + LZString.compressToUTF16(Description)
+        if (CompressedDescription.length < Description.length || Description.startsWith("╬")) {
             Description = CompressedDescription;
         }
         ServerSend("AccountUpdate", { Description });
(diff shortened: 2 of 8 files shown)


Ben987:
What will happen to players that currently have a description?  Will they lose it when we migrate from one version to another? Thanks!

Shiranui-Izayoi:
Holy crap you are much more meticuolus and note takey than I am. Nice job by the looks of it. Hope this gets in soon.

jomshir98:
> What will happen to players that currently have a description? Will they lose it when we migrate from one version to another? Thanks!

As mentioned, they will only loose it in the unlikely case if it currently starts with `--compressed-`. And only after they press confirm button in the edit window. If players from Beta than play normal, they will see their own description garbled, but unless they try editing it, there won't be a problem.

Ben987:
So if a player plays on the beta version, then goes back to the regular version for whatever reason, they won't lose their description?

jomshir98:
> So if a player plays on the beta version, then goes back to the regular version for whatever reason, they won't lose their description?

No, they won't loose anything, unless they follow exactly this; everything else is okay

Steps to break your description:
1) Play beta
2) Edit your description while in beta
3) Play regular version
*At this point your description won't be readable, but will be okay when you play beta again*
4) Attempt to edit your description (<- This would break it when you press green confirm)

Ben987:
It will probably happen to some players, so I will put a warning message in the beta description.  Instead of --compressed-, could we put something shorter to save even more space?

jomshir98:
> It will probably happen to some players, so I will put a warning message in the beta description. Instead of --compressed-, could we put something shorter to save even more space?

Yes, we could, but honestly 13 characters are never a problem, considering one can create new account whenever he wishes and that takes a LOT more space than 13 bytes. And this has added benefit of being self-descriptive

Ben987:
Excellent, make it as short as you can and we are good to go.  The leaner the better.

jomshir98:
> Excellent, make it as short as you can and we are good to go. The leaner the better.

Okay. Could you please suggest something else, that is shorter and still has following benefits?:
- Extremely low to none chance of currently being in anyone's description (as that would mean it will get lost)
- When human sees it (especially non-beta players, because they will see it exactly like this), they can easily recognize, that the description isn't corrupted, but only compressed and that is the reason it is not readable to them?

Example of how non-beta user would see compressed description currently:
```
--compressed-ᩢఫ䅌઀Ȩ㎬⇜ސ䰦ᠢ怺ැ䁙嬠㘢㲀ʅ'ᐡ泉øᠧ塓{㰍⿠Ҁଆ毾థઁ䄠琠倠⌍尡㡳堣ᦧ挥Ⱦȣᅄ♑楪䣥ၧ嗬ᷙ皐ᤅ涧保ውㄠ᫪ྮ䌦揙奐†❃৆嶲ၥo㋁㦉┚碁⑴Ṡೃ᝱堦ြア堨亜ḪЂ戒嗴愃መ⢘㵆Ȥউ╤ႄ慀⒲䩡㮳Ṡৎ熠䖐ᣀ┒汊ቓ䊆⡓窻䡢㈒㴩㲠࿦#ဤ䠢撎J㲸㨜䬝䫘䂺ァঀℷG≁塕䈭挑Ȩ͎窜ᴴਠӝヽ䝃惚㮧dᬓオ䋀泦㯿恥㜔ᙅ܈Ô䴤ઁ票灒ᜣ様硜䎕ゲ຾䐂ᦦ䳀⌣咨ᣞ゚͠⭞Ꭰ᥀す͓䰶ᡐ䭳୶䣁桦ᦦ朻 〠牖ၪ⏑ᠠ偰抁③♏᥺䝲⥮悔曖ᣋ˴ಘṉ⃱㒈妎aᲜࢧ穁磺Ф㏨Y>⻅ⅹ䣇⼳ǉ砤⏁䂨磈卢ᅂၓढ䟪繙௡砠੃Ǵ琻㒈ቍ欹瀠ከ⇫䃸㩤ŷ㦺ద盂㤆糥┽բ檕ᠣؐ㜧䔵䅔Რᤄ慘䠣౅放の⒤弣湌Ƣ䅙⧡૯ⶽ䶓☴ᢈไ爤慀喦ʮ爿Ȧ⟻䅤孰桞␱ø渾祰太㩮䷳挨漷怤瑣⦊ᯪǚጭڀ沧ዐ⤡㎺簽Ⴊ塏愠
```

Nina-1474:
Also down the line, maybe after another account purge and when we're sure everyone's profile is compressed, we could always remove the extra tag then.

Ben987:
Yes Nina, we could remove the tag in a few months, that's a good idea.

For now, can we only put one character?  A weird character that nobody would use in their current description?

Or could we try to decrypt the description and if it fails we put it "as is" and then encrypt it again, so we would not need any extra characters?

Thanks!

jomshir98:
> can we only put one character
Definitely doesn't have the second benefit, but if you give me such character, I will put it in.

> Or could we try to decrypt the description and if it fails we put it "as is" and then encrypt it again, so we would not need any extra characters?
No. There is no way at all to do this; because EVERYTHING is valid, it will only decompress into something, that doesn't make sense. Also there is no reasonable way to detect if decompressed value makes sense, because there are descriptions in different languages (not even using latin letters) and also descriptions containing emoticon and similar things

jomshir98:
> Also down the line, maybe after another account purge and when we're sure everyone's profile is compressed, we could always remove the extra tag then.

> Yes Nina, we could remove the tag in a few months, that's a good idea.

Not in current implementation, because the descriptions is only compressed if following things are met:
- Player edits description - so even if stale players are removed, lots of players would still have uncompressed description
- The compressed description is actually shorter (including the prefix!) than the non-compressed format.
  - Beware, that even empty description is 2 characters long (even without prefix) in compressed format
  -

Ben987:
Good catch, if there's no description can you leave the field blank and not compress it?  It will take even less space.

Yeah, maybe we won't have any choice but to keep that extra character forever, we'll see.

For a single character, you can use any weird char that should not begin a description, like: ¤ or ┼ or ╬ etc., up to you.

Thanks!

jomshir98:
Done, changed to `╬` @Ben987

elliesec:
I realise this has already been changed several times, but I feel like using a single printable character is pretty risky, even if it's a somewhat obscure one. People put all kinds of things in their bios, including odd symbolic characters for the sake of bio artwork or decorating it.

Personally, I'd be inclined to use at least a couple of characters - probably non-printable ones provided that doesn't present any issues.

Nina-1474:
I actually saw this bio just today. Granted the symbol isn't the first character in this case, but it does highlight the possibility.
image

elliesec:
Yes, this is exactly the kind of thing I'm thinking about - I've seen all sorts of strange characters in bios. And it's all great until someone uses the single character you're using as a "this is compressed" indicator. At the very least, using a few obscure characters would greatly reduce the chances, and I'd say moreso if they're non-printable.

Generally though, I do feel like something like "--compressed-" is fine - it's a handful of extra bytes in a profile that's probably a few kilobytes in size, and having the explicit "compressed" indicator at least makes it a bit more self-explanatory as to what's going on.

Ben987:
Yes, it's a good idea to check first.  I've ran a query in the database and there's no description that starts with ╬ expect one that's already compressed for Katie, I'm guessing this is your character Jomshir98.  Thanks a lot everyone!  The next compression should target the bigger packets we send every time a player is synchronized in a chat room.  Cheers.

PR #1968, opened 2021-01-26
New item: Floor Shackles

jomshir98:
A new simple item: Ceiling Shackles
Shackles holding Player's legs spread and to the floor

**Assets created by my wonderful in-club wife Claudia**


elliesec:
Failing check is because #1936 got merged before I could update it. #1969 fixes that.

Ben987:
Good idea!  Thanks you two :)

PR #1969, opened 2021-01-26
Bugfix: Adds type definitions for the new asset properties

elliesec:
This adds asset property definitions to the asset type check script for the new asset properties that were added as part of #1936 (I was going to update that PR, but it got merged before I could).


Ben987:
Thanks for the fix!

PR #1970, opened 2021-01-26
Appearance Dynamic Menu Buttons

Nina-1474:
This refactors the appearance/clothing code so that the row of buttons at the top are built dynamically, similar to the menu buttons in the inventory items screen.
This will mean better use of space, make it easier to add new buttons, and reduces the work in the Run function.
The behaviour is all the same as before except in the Wardrobe and Permissions subscreens, which now use the green tick/accept button to jump back one screen instead of having unique "dress manually" and "reset" exit buttons, to make them more consistent with the rest of the appearance subscreens.


Ben987:
Sure!  That sounds like a good idea.  It will be easier to add new buttons or change the order later.

PR #1972, opened 2021-01-26
 InventoryGroupIsBlocked(C, null, true) 

Ada18980:
Replaces InventoryGroupIsBlocked(C, null, false) with InventoryGroupIsBlocked(C, null, true)  in dialog.js in two places where it should have been. (merge error)


Ben987:
Thanks a lot for the fix!

PR #1973, opened 2021-01-27
Clothes Group Colour Button

Nina-1474:
This adds a colour-picker button into the top menu for the Cloth Appearance subscreen, i.e. the individual clothes group selections.


Ben987:
Sure!  This should be useful.

PR #1974, opened 2021-01-27
Bed Rope access fix

Nina-1474:
The ItemFeet ropes which tie the legs to the bed were blocking the character from interacting with anything, as if their arms were bound.


Ben987:
Yes!  Thanks for the fix.

PR #1975, opened 2021-01-27
Bugfix: Adds more items to the RemoveItemOnRemove entry for the bed

elliesec:
## Summary

Several items that require the bed weren't being removed when the bed was removed on login, leaving players stuck in situations like below. This adds all items with the `OnBed` prerequisite into the bed's `RemoveItemOnRemove` array.

image


Ben987:
Good catch!  Thanks a lot.

PR #1976, opened 2021-01-27
New item pose: Ceiling shackles - allow suspension

jomshir98:
Adds option to use "Ceiling shackles" during suspension - connecting player's arms to the ground instead

**Assets created by my wonderful in-club wife Claudia**


Ben987:
Sure!  This is a good idea.  Thanks :)

PR #1977, opened 2021-01-27
Bugfix: Infinite Replication of SpankingToys

tickler2000:
Fixed a bug were SpankingToys were replicated infinetely in the ItemHands slot, when this slot was blocked for items but not for activities.


Ben987:
I've never seen that one, but thanks for the fix!

PR #1979, opened 2021-01-28
Appearance Group button scrolling

Nina-1474:
This turns the single-click group buttons in the appearance screen into back/next buttons, but using an enhancement to the buttons which allows only the far left and far right to act as the back/next buttons. The middle of the button covers the current behaviour to jump into a new screen.
This way we have both the old scroll-style selection method where the user can quickly jump through every item in a group on one screen without having to click around, and the newer 3x3 grid method with preview boxes across multiple pages.

This also includes two small fixes to the recent appearance screen changes.

image
image
image


Ben987:
How could we help mobile players with these buttons?  Since they don't have hover it will get pretty confusing to know where to click.  Thanks!

Nina-1474:
The best I can think of is drawing a line for the divisions, something like the below examples of faint line or a slightly stronger dashed line. It would only appear on mobile.

image
image

Ben987:
Yes, maybe drawing a line to see clearly where to click would work.   Can you fit the text to squeeze it in the center button and maybe make the two side buttons a little smaller?  Thanks!

Nina-1474:
Something like this?:

image

Although given that this is already specifically for people on smaller screens, making either the text or clickable area even smaller both sound like signficant negatives to me. Is having semi-solid lines intersect the text that big a drawback?

Ben987:
It looks better yes, but the text might get too small to be readable on a phone.  Do we have some room on the left to enlarge the buttons?  To compensate for the lost space for the text.

Nina-1474:
Enlarging the buttons to the left will definitely overlap the character too much, and won't even help in that many cases since even now mobile users already have to contend with long text like "Hair Accessory (All): Wolf Ears One Piece)".

It's making the arrow region smaller that I consider to be the much bigger downside. The original size is the same size as the naked/strip button to the left of the group button, so I'd prefer not to go any smaller than that.

I really think the dividing lines need to intersect the text if we want them at all. (There's also an argument to be made for no lines and letting mobile users just figure the regions out by quick experimentation)

One last idea is using a fill instead of a divider, although I don't know what colour would look nicest.

image

Ben987:
Sure, graying the zones will probably help.  Maybe later we'll find a better solution to make it pretty.  Thanks a lot!

PR #1980, opened 2021-01-28
Persistent Lock Picking

Ada18980:
This PR changes the lock picking mechanic to be persistent through subsequent attempts. This has the side effect of any pickable lock being possible to eventually break through, with enough dedication, even if the chances of randomly picking pins would make it impossible.

However, most of the times you set a pin, the pin will be in 'false set'. These pins will eventually unset if you set the true pin that comes next. However, this serves to obfusticate the true order of the pins and make it into a sort of memory puzzle.

This also adds a temporary lockout (30 seconds) before you can attempt to lockpick again, if you fail too many times (about 1 in 4 tries, but requires that you set the wrong pin at least 5 times just to keep it from happening on the first fail), to simulate the player getting winded and encourage players to ask others for help~


Ben987:
Alright!  I'm curious to try it.  Thanks :)

PR #1981, opened 2021-01-28
Optimise DialogFind for Player + error message

jomshir98:
Creates new function `GetPlayerDialog`, that works in similar way to heavily used `DialogFind(Player, ...)`, but is lot faster and if text is not found, it returns error instead of last dialog.
Error is in format `MISSING PLAYER DIALOG: ${KeyWord}` similar to `TextGet`'s `MISSING VALUE FOR TAG: ${TextTag}`

Also replaces all appropriate calls with new function.

I didn't test every change, but this change will make errors easily visible.


Ben987:
If it's faster to process I'm good with that.  Can you rename the function so it's starts with "Dialog" instead of "Get"?  Everywhere in the code, the functions are named based on which file they are in.  So a Dialog.js function should start with Dialog.  It could be named DialogFindPlayer() or something similar.  Thanks!

jomshir98:
@Ben987 Done!

Ben987:
Awesome!  I will test it soon.  Cheers :)

PR #1983, opened 2021-01-28
Over panties crotch rope now doesn't visually sink into the panties

Ada18980:
I thought it looked weird before.


Ben987:
And how does it look when you're naked?  Does it still sinks in?  Thanks!

Ada18980:
It does not, I don't think there is a way to do it based on whether there is an item in the panties slot. But there are two options to tie the crotch rope, one is over the panties and one is under. I think if players want the original look while the sub is naked they can just use the original crotch rope type.

Ben987:
Alright then!  Thanks a lot :)

PR #1985, opened 2021-01-28
Messages when plugged or vibrated

Ada18980:
Plays messages occasionally when you kneel/stand up, enter a chat room, use the struggle action, or fail a struggling minigame and exit out. 
These messages appear only to the player and not to others. They add a bit of arousal as well.


Ben987:
Will there be an option to avoid these messages?  Not everyone will want to output automatic text.  Thanks!

Ada18980:
I'll add an option to the chat settings

Ben987:
Thanks!  Can you also make a preference so players will be able to skip that mechanism if they want to?

Ada18980:
There, I added an immersion setting @Ben987 The messages are still controlled by the sexual activities chat room message, but the actual feature is an immersion option as I plan to expand it later (getting 'distracted' while trying to do things...)

Ben987:
Alright!  I will test it soon.  Thanks :)

PR #1986, opened 2021-01-28
Intro text on intricate padlock

Ada18980:
Since the functionality of the lock is changed, the intro text has changed as well.


Ben987:
Great!  We might receive less questions that way.  Thanks.

PR #1987, opened 2021-01-29
More responsive vibrator visual effects

Ada18980:
The pink glowing meter is now more responsive, without changing overall speed of arousal gain while vibed


Ben987:
Alright!  Thanks :)

PR #1988, opened 2021-01-29
Lock picking fixes

Ada18980:
Change ExclusiveUnlock to false by default for assets
Replace MemberNumberList with MemberNumberListKeys to avoid issues with timer locks
Fixed issues with giving keys and tested so that giving keys works now with both intricate and high security locks
Added back button to security dialog menu
Remove intricate and high security padlock keys to avoid issues with players being given keys but not having the key in inventory


Ben987:
Yes, removing the intricate padlock key is probably best.  It will be less tricky for players.

Why did you rename the "Metal Padlock Key" to "Metal Padlock Key`" ?  Why the extra ` ?

Thanks!

Ben987:
Also, there's a conflict in HighSecurityPadlock.js, can you resolve it and test again?

Ada18980:
Fixed metal padlock key name and merge conflict

Ben987:
Alright!  Thanks for the adjustments, I will try it soon.

PR #1990, opened 2021-01-29
Two new struggle minigames

Ada18980:
Adds two new struggle minigames,  which can be selected when the player tries to struggle out of an item:
-Flexibility: Click the ropes before they hit the ground
-Dexterity: Click when the player icon is on top of the buckle as it moves around

The difficulty level for each is different. The intent is to encourage players to try out different ways of escaping items, in order to increase the immersion aspect of trying to figure out what works. Some items will now be escapable where they weren't before; it is the nature of things.

Generally, flexibility is favorable when you have few items restricting your movement. Locks and hand restraints restrict you considerably less, but restricting the arms and legs makes it a lot harder to slip out of items. 
Dexterity is favorable when items are unlocked and when the arms are restricted but not the hands. However, dexterity does poorly against locked items and when the mouth, hands, boots, and feet are all blocked.

Tested on other players (you can only brute force items on other players) and when applying/removing items. Note that this splits the struggle minigames off into a separate file; this WILL cause merge conflicts whenever someone tries to modify the lockpicking minigame--just setting a mental note on that.


Ben987:
Thanks a lot!  New struggles have been on my list for a long while.

For code structure.  Make sure to uppercase your Struggle.js file.  And can you change all functions and variables to start with "Struggle".  So DialogDrawStruggleProgress(C) could be StruggleDrawProgress(C)

That way we can quickly identify where the functions and variables belong.

Ada18980:
There, I fixed merge conflicts and change the convention on all the functions and variables
@Ben987

Ben987:
In game, how will player know that escaping is easier using method A, B or C ?  They need to try all 3 and learn from experience which method work better with each item combo?

Is it still always impossible to escape from the highest difficulty items like the slave collar?

If someone plays in roleplay difficulty and skips the struggle mini-game, what happens if they try to struggle in the other mini-games?

Thanks!

Ada18980:
@Ben987 Hi Ben,

1) Yes, the intent is that the player needs to try out all 3 methods.

2) Yes, the difficulty of the item is added to all 3 difficulty methods. So something with difficulty 50 like most collars is going to be impossible to escape.

3) The code for skipping the minigame is the same. If the challenge is 6 or lower then it will set the bar to automatically progress without any user input required.

Ada18980:
@Ben987 Fixed the latest merge conflicts

Ada18980:
@Ben987 I fixed the merge issues

PR #1991, opened 2021-01-29
Refactor: Unify Preference Init

jomshir98:
No functional change (but maybe fixes some edge-case bugs)

Moves things around in `PreferenceInit`, making it more clear and removes unnecessary or temporary parts.

Tested with both new and existing accounts.

~~Will create conflict with #1669 and #1985 ; I recommend merging those before this as my changes are easier to fix.~~

### DIFF
--- a/BondageClub/Screens/Character/Creation/Creation.js
+++ b/BondageClub/Screens/Character/Creation/Creation.js
@@ -100,7 +100,7 @@ function CreationResponse(data) {
 			ImportBondageCollege(Player);
 
 			// Calls the preference init to make sure the preferences are loaded correctly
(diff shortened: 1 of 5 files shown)


fleisch11:
this makes controlles not work

Ben987:
Alright!  Can you remove the console.log and throw calls?  We don't need to add stuff to the browser console, lets leave it for real crashes.

jomshir98:
@Ben987 Rebased and ready to be merged.

Ben987:
That's a pretty big one, let's hope it won't create any bugs.  It makes the code easier to read though, thanks a lot :)

PR #1992, opened 2021-01-29
Feature: Adds kneeling support to the wooden box

elliesec:
## Summary

This PR adds support for the kneeling pose to the wooden box, which makes it behave better in combination with items that force a kneeling pose. I've also updated the dynamic text drawing function to accommodate this. See the image below:

wooden-box-kneel-1


Ben987:

PR #1993, opened 2021-01-29
Pose-specific fixes (Paw Mittens, Padded Mittens, Reverse Bunny Suit Top)

EmilyR42:
- Improved BackCuffs pose graphics for Padded Mittens and Paw Mittens
- Removed feet and bodysuit bottom from Reverse Bunny Suit Top hogtie pose sprites


Ben987:
Many images are in conflict with another pull request.  Can you resolve the conflicts and test again?  Thanks a lot.

Codefreak64:
I might be wrong on this, but if the whole PR is fixing up images, then it should override the existing images, possibly causing conflicts? so I think in this case it has to conflict?
Correct me if im wrong

PR #1994, opened 2021-01-30
Blocked or Limited Checks refactor

Nina-1474:
I've refactored checks for whether an item (and sometimes type) is blocked or limited and inaccessible to the player into one function, as the same logic was repeated in a lot of places. I tested all changed areas and they work just as before.


Ben987:
Very good!  It makes the code a little easier to read.  Thanks!

PR #1995, opened 2021-01-30
Default Extended Option enhancement 

Nina-1474:
**This is dependent on PR #1994 and should be merged after that.**

---

This enhances the extended item functionality so that the initial option used on applying the item can be different to the "null" option corresponding to the base asset, by simply moving it to be the first option in the options array.

In particular, this has been applied to the HempRope and Chains items for ItemArms. Now the option that first appears will be the WristTie, as it's the simpler tie in comparison to the current BoxTie default.
This solution means that random restraints applied to NPCs will still be the more visually interesting BoxTie, as that's still the what the base asset is.

The ExtendedItem load function has also been cleaned up to remove comparisons between the default option and base asset that prompt an update, since these are no longer needed. Discussed and confirmed with @elliesec.


Ben987:
Yes!  I will test it soon.  Thanks a lot.

PR #1996, opened 2021-01-30
Dress Reset Icons

Nina-1474:
The Gambling, Shibari and Asylum Therapy rooms all have a button to reset the player's clothes back to a previous state. However since these are using the same 'Dress' icon that's used for opening the Appearance screen everywhere else, it can seem like a bug when the player clicks it and nothing happens. Changed these to use a new 'Dress Reset' icon to make it clearer what they do.


Ben987:
Sure!  If there was confusion it's better to resolve it.  Thanks.

PR #1998, opened 2021-01-30
Item Permissions on Blocked Zones

Nina-1474:
When an item zone is blocked by another item and the player clicks the item permissions toggle, nothing appears. While the item list should be empty in normal mode, there's no reason to block editing item permissions of the zone regardless of whether the zone itself is blocked. This is now possible.


Ben987:
That makes sense, thanks for that fix!

PR #1999, opened 2021-01-30
Bugfix: Fixes double clicks on the appearance screen

elliesec:
## Summary

This fixes an issue with the `AppearanceClick` function that would cause an extra click to appear to be registered when transitioning between the default mode and the cloth selection mode. It changes the handling of each appearance mode to be in its own `else if` block, so click handling should only happen for one mode per click. 

Previously, the click handler for the default `""` appearance mode contained this line:
```js
CharacterAppearanceMode = "Cloth";
```
However, the click handler would then proceed into the `if (CharacterAppearanceMode == "Cloth")` statement, causing the `"Cloth"` mode handling code to be run.

## Steps to reproduce

1. Enter the appearance screen
2. Click one of the appearance groups
3. This takes you into the item submenu as usual, but also seems to register a click on whichever item happened to be under the mouse pointer when I clicked on the group.

See the below gif for an example of the issue:

double-click


Ben987:

PR #1989, opened 2021-01-29
Refactor: DrawAssetPreview

elliesec:
## Summary

The game uses asset previews in several locations, and despite the existence of a `DrawItemPreview` function, it is only used in a couple of places. This made it a lot harder to chase down all of the preview images when reworking for #1936.

This PR reworks the `DrawItemPreview` function into two new functions: `DrawAssetPreview` and `DrawPreviewBox`, and changes all preview images across the core game to use them. It also breaks the horrible ternary operator in `Appearance.js` for determining the preview color into its own slightly more efficient and much more readable `AppearanceGetPreviewImageColor` function.

After this, the `DrawAssetPreview` and `DrawPreviewBox` functions could be used across most extended item screens, but I've not done that here to keep the scope of this PR down.

All refactors have been manually tested and confirmed to still be working as before.

PR #2004, opened 2021-01-30
Refactor: Preview images part 2

elliesec:
## Summary

Sorry, another huge one! This is a follow on from #1989, which changes most extended items to use the `DrawAssetPreview` and `DrawPreviewBox` functions in their extended item screens. This should mean that almost every preview image in the game is now routed through the same functions, which reduces the amount of code, and should help with consistency. The PR also contains numerous other bug fixes that I discovered along the way with various items, as well as in the general codebase.

## Notable changes

* Modifies the `DrawAssetPreview` and `DrawPreviewBox` functions to accept new `Hover` and `Disabled` options to allow them to have button-like behaviour
* Adds a new `MouseHovering` function to `Mouse.js` this can be used in places where we do a hover check (`MouseIn(x, y, w, h) && !CommonIsMobile` - which is used in many places throughout the code)
* Adds two new functions to `Asset.js`: `AssetGetPreviewPath` and `AssetGetInventoryPath`, which take an asset and return the paths to its preview image directory and inventory directory respectively
* A _lot_ of other small fixes and removal of duplicated code in extended item scripts

## Testing

I have manually tested _every_ item affected by this PR (which took quite a long time), and verified that they are all still working as before (and in some cases, better than before).


Ben987:
So many huge changes this month, I'm starting to worry for the next beta.  Let's hope there won't be too many bugs.

You're on fire, do you ever sleep?  Thanks a lot :)

PR #2005, opened 2021-01-31
Bugfix: Fixes a click handler ordering issue in the appearance menu

elliesec:
Follow-up to #1999 - the addition of `else if`s in that PR causes an issue with the `ItemColor` click handler not being called for clicks in the top menu. This increases the priority of the click handling for color mode, which fixes the issue.


Ben987:
Thanks for the fix!  Cheers.

PR #2009, opened 2021-01-31
Assets: Adds copies of pantyhose and stockings to SuitLower

elliesec:
This PR adds copies of the pantyhose and stockings items to the `SuitLower` group using `DynamicGroupName` (no new assets required), based on a suggestion from Discord. It modifies the hogtied pose to remove the 30px x-offset for the `SuitLower` group, and adjusts `SuitLower` hogtied assets accordingly, and move positions are now calculated based on an asset's `DynamicGroupName` when drawing an asset.


Ben987:

PR #2010, opened 2021-01-31
Bugfix: Rebuilds appearance inventory when selecting an item

elliesec:
## Summary

I haven't quite traced back to exactly what caused this, but it was a while ago (before Nina's appearance menu changes at least). When the appearance screen is in the `Cloth` subscreen, clicking on an item no longer displays that item in pink as the first item in the list. This adds a call to `DialogInventoryBuild` after selecting an item, which restores that functionality.


Ben987:
Alright!  Thanks for another fix :)

PR #2011, opened 2021-01-31
Appearance Group buttons - mobile

Nina-1474:
The final suggestion from #1979 for button fills on mobile screens hadn't actually been commited.


Ben987:
Thanks!

PR #2012, opened 2021-01-31
Bugfix: Fixed limited permission check for individual spanking toys

elliesec:
## Summary

This fixes an issue with limited permission checks against handheld toys. the `InventoryCheckLimitedPermission` function wasn't using the dynamic asset/group names when running the check for whether an item's permission was limited.

## Steps to reproduce

1. On character A, set an individual spanking toy's item permission to limited (e.g. the cane)
2. Set character A's global item permission to "Everyone, no exceptions", "Everyone, except blacklist", or "Owner, Lover, whitelist & Dominants"
3. On character B (who should not be on character A's whitelist, or their lover/owner), equip the spanking toy selected in step 1
4. Note that character B can still use the spanking toy on character A


Ben987:
Awesome!  Thanks for the fix :)

PR #2014, opened 2021-02-01
Chat Room Camera

Nina-1474:
**In-Game Screenshots**
The feature in the Photographic room to capture a portion of the canvas in an image in a new tab has been tweaked and made available for re-use elsewhere.

**Chat Room Photos**
A new 'Camera' button has been added to the chatroom button menu. Clicking this will screencap the full room of characters. 
The dialog options for another character will now include 'Have a photo taken' which will screencap just the player and other character.

**Photo Mode**
Certain visual elements will be controlled during the screen capture for the sake of better photos. In particular, the drawing of the character blinking will be skipped, the pink tint for high arousal will be excluded, partial blindness will be excluded (full blindness will not so it can't be used as an exploit), and AFK emoticons will be excluded.

**Darkness Refactor**
The calculation to determine how dark the screen/background should get based on blindness level has been refactored into one function. This also fixes one error where light blindess could actually darken the screen more than heavy blindness.


Ben987:
This is great idea!  Another pull request that will be helpful for me.  Thanks a lot :)

PR #2015, opened 2021-02-01
Wardrobe Compresses & Decompresses Properties (FIX)

Shiranui-Izayoi:
Added code so that the properties of the saved outfits in the wardrobe are compressed and decompress.

Currently when saving an outfit the properties of an outfit will be saved such as if a Latex Corset has garters or not. But upon logging out and logging back in, as the properties are not compressed or decompressed when loading that same outfit it will default to it's default state.

I believe that this is unintended and just slipt through the cracks due to the low amount of clothing items with properties. I did do some testing and it did seem to work fine with no adverse effects.

### DIFF
--- a/BondageClub/Scripts/Character.js
+++ b/BondageClub/Scripts/Character.js
@@ -1066,7 +1066,7 @@ function CharacterCompressWardrobe(Wardrobe) {
 			var Arr = [];
 			if (Wardrobe[W] != null)
 				for (let A = 0; A < Wardrobe[W].length; A++)
-					Arr.push([Wardrobe[W][A].Name, Wardrobe[W][A].Group, Wardrobe[W][A].Color]);
+					Arr.push([Wardrobe[W][A].Name, Wardrobe[W][A].Group, Wardrobe[W][A].Color, Wardrobe[W][A].Property]);
 			CompressedWardrobe.push(Arr);
 		}
 		return LZString.compressToUTF16(JSON.stringify(CompressedWardrobe));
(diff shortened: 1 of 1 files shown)


Ben987:
Are you sure it won't cause any problem with time when asset properties will change?  Should we validate these properties when we reload the wardrobe?

Shiranui-Izayoi:
> 
> 
> Are you sure it won't cause any problem with time when asset properties will change? Should we validate these properties when we reload the wardrobe?

What do you mean "when asset properties will change", I also don't think there is a reason to validate, since even if someone does something silly it shouldn't be able to execute code anywhere. 

As for problems I can't think of any. Since properties already get saved by the wardrobe and will last until you log out. Then because properties aren't compressed they can't be retrieved when decompressing.

Also properties are compressed along with everything else when sent to the server to be stored. That should prevent anything bad if I am thinking about it correctly.

Ben987:
@elliesec , you've played a lot with character drawing lately.  Do you see any possible issues with that one?  Thanks

elliesec:
At a glance, I can't see any reasons for it to cause any issues, but there are a few edge cases I'd like to test, just to be sure. I'll pull it down and test a few things out and let you know.

As far as validation on decompress goes, I think it's reasonably safe to ignore validation for now, provided we trust what we're saving.

elliesec:
@Ben987 - I've pulled this down and done some testing, and also had a slightly more in-depth look at the code around the wardrobe, and I'm pretty confident that it won't cause any issues.

On the validation front, the only real potential issue as far as I can see is if someone were to set an undesired property on a piece of clothing, and for that property to be saved to the wardrobe. But ultimately, that's not really any different from saving the character's appearance - it's already possible to do that right now (although arguably it shouldn't be possible to set arbitrary properties on items in any case). 

If anything, I'd say that more in-depth property validation is probably a separate piece of work that should probably be applied to both clothes _and_ items, and in the case of the wardrobe, should happen when items get saved, rather than when they get loaded.

Shiranui-Izayoi:
So then I assume it can be merged then? Since everything seems fine?

Shiranui-Izayoi:
Cool. Glad this got in quickly since it is really fixing a base game bug. Looking forward to the next Beta/Release.

PR #2016, opened 2021-02-01
Lock menu icon

Ada18980:
Adds a different icon for the lock menu that opens up the new lock menu, to prevent confusion with the "inspect lock" Icon.

Also adds a new icon for "inspect lock" so it's different from "use a lock"


Ben987:
Sure!  Thanks for that one.

PR #2017, opened 2021-02-02
Some futuristic item changes

Ada18980:
3 changes:
1 - fixed crash with futuristic chastity belt shocks
2 - Made futuristic chastity belt have a low cooldown for the shocks for struggling. I don't see a need to limit the shock time. However, the orgasm timer is still there.
3 - Added the copy colors feature to the futuristic harness, per request from players. Lots of people already wear collars, so the feature is pretty useful to have on a non-collar item.


Ben987:
Great!  Thanks for the changes :)

PR #2018, opened 2021-02-02
Futuristic Earphones

Ada18980:
Adds some techy new futuristic earphones which can be locked (imagine it as a metal band which locks in place and can't be stretched apart to let the plugs out of the ear)

In addition to the normal modes, they have a noise cancelling mode which applies DeafTotal


Ben987:
Sure!  In the future maybe we'll have locking headphones :)

PR #2019, opened 2021-02-02
Adds Noise Cancellation to all earphones, and changes name garbling

Ada18980:
When names are garbled, it will now say Someone instead of "mmmmm," to avoid giving away the number of letters in the speaker's name.

Also adds noise cancelling options to the other headphones besides the futuristic ones, allowing DeafTotal.


Ben987:
Sure!  I'm curious to try it.

PR #2020, opened 2021-02-02
Sensory deprivation messages

Ada18980:
Adds a new immersion setting to hide messages from others such as adding restraints and activities.

Also fixes a bug (in the same block of code) with stimulation messages where the sender would stimulate herself instead of the target. Now both the sender and the recipient will be stimulated~


Ben987:
Very good!  Thanks again :)

PR #2021, opened 2021-02-02
Futuristic Vibrator

Ada18980:
Adds a futuristic vibrator, which has voice commands! Each command can be set to a 12 character trigger word. Since it is a futuristic item, it can also be locked and if the item is locked then others cannot modify the setting unless allowed. 

The settings are:
-Increase: Increases by 1 level
-Decrease: Decreases by 1 level
-Disable: Turns off
-Tease: Default vibrator feature
-Deny: Default vibrator feature
-Edge: Default vibrator feature
-Random: Default vibrator feature


Ben987:
Sounds fun!  Thanks a lot.

PR #2023, opened 2021-02-03
Futuristic Ball Gag

Ada18980:
Adds a new ball gag, a ball gag variant of the futuristic one. It is identical in functionality except it does not have a padded mode, so it will always apply at least GagLight.


Ben987:
Can't go wrong with another ballgag :)

PR #2024, opened 2021-02-03
Futuristic Collar Permission

Ada18980:
Makes it so you need to be able to unlock the collar in order to toggle whether or not permissions are open on the collar. This is to prevent trolling where someone changes your settings and leaves you stuck with them until your owner helps you.


Ben987:
Very good!  Thanks.

PR #2025, opened 2021-02-03
Improvement: Adds eyes closed expression support to the blackout lenses

elliesec:
## Summary

> ⚠️ Will conflict with #2031 - I'd recommend merging that first

This PR adds properties that enable assets (or layers) to copy the expression of another asset when being drawn. This enables the blackout lenses to follow manual eye expression changes (specifically winking or blinking) by splitting them into right & left layers. In the future this could enable items like cosplay lenses to be implemented without looking strange.

A quick summary of what's changed:

* Enables `AllowExpression` to be set on a per-asset basis
* Adds a new `MirrorExpression` property to assets and layers. This can be set to the name of an asset group to copy the expression changes of the asset in that group.
* Splits the Blackout Lenses out into a right and left layer - this enables each layer to copy the expression of their corresponding eyes, and also means that each lens can be colored separately.


Ben987:
Yep, you're correct about the conflict.  Can you resolve it and test again?  Thanks!

elliesec:
Conflicts have been resolved, and everything seems to be working as expected.

Ben987:

PR #2026, opened 2021-02-04
Feature: Updates the Transport Box with opacity and custom text options

elliesec:
## Summary

This applies the same changes to the Transport Box as have been made to the Wooden Box, to support custom text, and variable opacity. Changes were discussed and agreed with @Ayesha-678. I've also added a kneeling pose to the Transport Box, and added some missing locks for the Wooden Box.


Ben987:
Awesome!  Thanks a lot.

PR #2027, opened 2021-02-04
Crib Update

mrbakucamp:
# Changes
1. Splits the crib into multiple layers
2. Add leashing to the crib (meant to convey pushing the crib around)
3. Makes it so the crib can be filled with plushies without closing it
4. Fixes the issue where a player needs to be bound for the increased difficulty of have the crib closed to come into effect


Ben987:
Great!  Thanks for the update.

PR #2028, opened 2021-02-04
Kigurumi Mask

Ada18980:
Adds a mask which covers the head and can be configured to have up to 256 combinations of facial expressions!
The scripts are copied almost entirely from the High Security Straitjacket.

Much thanks to Ellie & Jomshir for help and code, and Miku for giving the the idea :)


elliesec:
FYI @Ada18980 - I've just raised PR #2031, which once merged should allow you to replace your 256 lock assets with a simple `DrawLocks: false` flag on the asset.

Ben987:
Sure!  Can you try the new property and remove all the blank lock assets?  Thanks!

Ada18980:
@Ben987 Done!

Thanks @elliesec !

Ben987:
Thanks!  I've never seen a mask like that before but I'm sure that players will like it :)

PR #2029, opened 2021-02-04
Image with Alpha Fix

Nina-1474:
A recent change to one of the drawing functions left an undeclared variable. This meant that on the non-WebGL drawing method certain assets with alpha masks would freeze the game.


elliesec:
Hehe, my bad. Good catch.

NoneNoname:
#1507 fixes this too if that is getting merged no need for this (this will just create conflict in that case)

Nina-1474:
There's a small difference between this fix and the #1507 one, commented over there. If my thinking is not incorrect (which is more than possible!) and the change is made there, I'll close this off.

NoneNoname:
Yes the comment was correct, I made the change

PR #2030, opened 2021-02-04
Improvement: Don't draw empty text

jomshir98:
This change seems very minor, but in places with lots of buttons without text (only using images, like MainHall) I measured almost 20% improvement in performance


Ben987:
Good idea!  Thanks.

PR #2031, opened 2021-02-04
Improvement: Optional lock assets & custom lock layers

elliesec:
## Summary

This PR adds two new properties to assist with easing some of the restrictions of the current lock asset system:
* `Asset.DrawLocks` - a boolean flag which can turn off the game's default lock drawing altogether at an asset level (useful for lockable items that don't need any lock assets - for example #2028)
* `Layer.LockLayer` - a boolean flag that indicates that this layer should only be drawn when the item is locked

Between these two new properties, the lock drawing system should be a lot more flexible (for example, I've modified the High Security Straitjacket so that locks are only drawn for the types that need lock assets).


Ben987:
Sure!  We'll be able to remove some blank assets.  Thanks!

PR #2032, opened 2021-02-04
Improvement: Draw empty background

jomshir98:
Draws a grey background if background image isn't ready, preventing things like this from happening:


elliesec:
Nice! This was on my list too - should also prevent the out-of-control halo glow issue when switching to a new room for the first time.

Ben987:
Sure!  That's a good idea for slow connections.

PR #2033, opened 2021-02-04
Bugfix: Ensures that the DrawRun function always gets refreshed properly

elliesec:
## Summary

In #1507, the `DrawRun` function would not get properly refreshed if a screen change happened at the same time as a `CharacterSetCurrent` call, resulting in a stale `DrawRun` function being used - this was only an issue for minigames where leaving the minigame would return the player into dialogue.


Ben987:
Thanks for the fix!  I will test it soon.

PR #2034, opened 2021-02-04
fix: chess with css

NoneNoname:
fixes: Uncaught TypeError: Cannot read property 'toString' of undefined at MiniGameChessResize (chess.js:338)


Ben987:
Thanks for the fix!  I will test it soon.

PR #2035, opened 2021-02-04
Improvement: Chess

jomshir98:
Refactors chess to run asynchronously, not blocking main thread, so the connection to server doesn't die during longer thinking.
Also removes unused parts of code


Ben987:

PR #2036, opened 2021-02-04
TightJacket Message Fix

Nina-1474:
Corrects chatroom messages like 'adds all addons to DestinationCharacter tightjacket'.


Ben987:

PR #2037, opened 2021-02-05
Futuristic Straitjacket

Ada18980:
Adds a new extended asset called the futuristic straitjacket, modeled after the latex strait leotard.


Shiranui-Izayoi:
Hey I think it is quite close to perfect. I would say the only issue I can see in terms of look is that the shading on the chest part, could be a little more uniform/symetrical since, the off center shaded part looks a little strange to me at least. Maybe also have some shading on the top of the bust/upper torso? Since in my case I can't really tell when the bust is meant to start to protrude/form. 

Looking at the other items that look busty, the shading is always around the bust with with the frontmost part being brighter. Might that work? So like for example with the shading on the original leotard is like that, also for the former baby dresses are like that as well, the lockdown suit and even things like the long chinese dress have that same thing.

Ben987:
Can't go wrong with more straitjackets, thanks :)

Shiranui-Izayoi:
Damn I was hoping the shading could have been updated a bit before it was accepted. Mostly to add some shading to the upper torso/chest area. And then also some around the edges. Since the shading looks a little uneven and random when I checked it in game.

Ada18980:
> Damn I was hoping the shading could have been updated a bit before it was accepted. Mostly to add some shading to the upper torso/chest area. And then also some around the edges. Since the shading looks a little uneven and random when I checked it in game.

Check again--the small is deliberately pretty flat. The medium and above should have smoother shading on the bust

PR #2038, opened 2021-02-05
Wooden Sign Upside-Down

Nina-1474:
Flipping the wooden sign to hang downwards when suspended upside-down.


Ben987:
Good idea, thanks!

PR #2039, opened 2021-02-05
Garble & Lisp switched to Switch Statements

Shiranui-Izayoi:
Ok so had this in my back pocket for a while, forgot to publish this. What this does is switch the Garble & Lisp speak from if & else if statements to switch statements.

This should run much better for players since switch statements are faster than else if. While it does make the functions longer it should also make them more optimised.

Not sure if this will get in but I feel I should submit it just in case. I have tested it in general use thanks to the console, and didn't notice any issues.

Biggest change is that I added capital letter replacements to most parts mostly for asthetic reasons, and that with lower gag levels I believe that you would be able to use them.

By the way the worst offender was Lisping/baby talk speak which was only if statements. Also added an replace to it to remove certain issues with words like thanks turning to thhanks which looks strange since the th sound is already there. So it wouldn't need to be lisped.

### DIFF
--- a/BondageClub/Scripts/Speech.js
+++ b/BondageClub/Scripts/Speech.js
@@ -95,9 +95,17 @@ function SpeechGarble(C, CD) {
 			if (H == "(") Par = true;
 			if (Par) NS = NS + CD.charAt(L);
 			else {
-				if (H == " " || H == "." || H == "?" || H == "!" || H == "~" || H == "-") NS = NS + H;
-				else if (H == "z" || H == "q" || H == "j" || H == "x" || H == "k" || H == "v" || H == "b" || H == "y" || H == "w" || H == "g" || H == "p" || H == "f" || H == "u" || H == "c" || H == "d" || H == "l" || H == "h" || H == "r") NS = NS + " ";
-				else NS = NS + "m";
+				switch(H) {
+					case " ": case ".": case "?": case "!": case "~": case "-": 
+						NS = NS + H;
+						break;
+					case "z": case "q": case "j": case "x": case "k": case "v": case "b": case "y": case "w": case "g": case "p": case "f": case "u": case "c": case "d": case "l": case "h": case "r": 
+						NS = NS + " ";
+						break;
+					default: 
+						NS = NS + "m";
+						break;
+				}
 			}
 
 			if (H == ")") Par = false;
@@ -114,9 +122,17 @@ function SpeechGarble(C, CD) {
 			if (H == "(") Par = true;
 			if (Par) NS = NS + CD.charAt(L);
 			else {
-				if (H == " " || H == "." || H == "?" || H == "!" || H == "~" || H == "-") NS = NS + H;
-				else if (H == "z" || H == "q" || H == "j" || H == "x" || H == "k" || H == "v" || H == "b" || H == "y" || H == "w" || H == "g" || H == "p" || H == "f") NS = NS + " ";
-				else NS = NS + "m";
+				switch(H) {
+					case " ": case ".": case "?": case "!": case "~": case "-": 
+						NS = NS + H;
+						break;
+					case "z": case "q": case "j": case "x": case "k": case "v": case "b": case "y": case "w": case "g": case "p": case "f": 
+						NS = NS + " ";
+						break;
+					default: 
+						NS = NS + "m";
+						break;
+				}
 			}
 
 			if (H == ")") Par = false;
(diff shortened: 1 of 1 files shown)


Ada18980:
Does removing the toLowerCase cause speaking in all caps to not be garbed?

elliesec:
> Does removing the toLowerCase cause speaking in all caps to not be garbed?

Without trying it out, it certainly looks that way.

Shiranui-Izayoi:
Actually no it doesn't. I included the capital letter versions of each letter for that very reason in fact. Meaning that if you speak in caps you get garbled caps.

elliesec (commented):
Performance-wise, I like the principle, but we're also extending a function that was already ~400 lines long into one that's 1000+ lines - which makes things a lot less digestible. Personally, I'd take this further before considering it to be good to go - just because it's a performance improvement, I don't think we should be sacrificing readability in what's already a really big function. What I'd do additionally is:
* Refactor the `switch` statement at each level of garbling into its own function that depends only on the character being transformed (letter character, not player character)
* Refactor the looping logic into its own function (taking a character garbling callback), which would handle the OOC logic too
* Pull out a function that takes a gag effect level & a deafness level, and returns a garbling callback
* At that point it should be pretty apparent that the whole function can be boiled down to something pretty simple:
```js
// The usual preamble - calculate the gag effect level, and the character's deafness level once

// This is a function which transforms a single character - one of the `switch` statements
const garblingFunction = SpeechGetGarblingFunction(GagEffect, DeafLevel);

// This is encapsulates the `for` loop logic - it runs over each character and runs the garbling function on it
let NS = SpeechGarbleLine(CD, garblingFunction);
NS = SpeechStutter(C, NS);
NS = SpeechBabyTalk(C, NS);
return NS;
```
Breaking these down into modular functions will also have the benefit of letting us actually run tests on the individual garbling functions, which is something we may well want to do further down the line.

Shiranui-Izayoi:
> 
> 
> Performance-wise, I like the principle, but we're also extending a function that was already ~400 lines long into one that's 1000+ lines - which makes things a lot less digestible. Personally, I'd take this further before considering it to be good to go - just because it's a performance improvement, I don't think we should be sacrificing readability in what's already a really big function. What I'd do additionally is:
> 
>     * Refactor the `switch` statement at each level of garbling into its own function that depends only on the character being transformed (letter character, not player character)
> 
>     * Refactor the looping logic into its own function (taking a character garbling callback), which would handle the OOC logic too
> 
>     * Pull out a function that takes a gag effect level & a deafness level, and returns a garbling callback
> 
>     * At that point it should be pretty apparent that the whole function can be boiled down to something pretty simple:
> 
> 
> ```js
> // The usual preamble - calculate the gag effect level, and the character's deafness level once
> 
> // This is a function which transforms a single character - one of the `switch` statements
> const garblingFunction = SpeechGetGarblingFunction(GagEffect, DeafLevel);
> 
> // This is encapsulates the `for` loop logic - it runs over each character and runs the garbling function on it
> let NS = SpeechGarbleLine(CD, garblingFunction);
> NS = SpeechStutter(C, NS);
> NS = SpeechBabyTalk(C, NS);
> return NS;
> ```
> 
> Breaking these down into modular functions will also have the benefit of letting us actually run tests on the individual garbling functions, which is something we may well want to do further down the line.

In that case I will likely have to close this pull request as it is as that is a lot of stuff to do. And I will likely need time to work out the best way to do each part. I understand the general premise that being to make each gag and deafness level it's own function with the switch statements incorpeated. 

Not sure about the second thing.

With the 3rd I assume that you mean to have the inital deafness/garble level be caculated by it's own function and then return it to a second function that would the launch the appropriate garble level for the message.

So yer with the amount of changes I would need to pull it down and work on it in my own time. To test things out. I do see what you mean though. While it is performance wize a good thing it does make it a hellva lot longer. Thanks for the list. As said I will be closing it for now.

elliesec:
Yeah - I appreciate it's a lot of changes (sorry about that) - I just feel like it would be a shame to improve the performance in this way without going all the way and refactoring it into something a bit more modular, rather than sacrificing readability for performance when we could have the best of both worlds. Like I say though, in principle I think it's definitely a good change.

As far as each of the points go, it sounds like you've got it. On the second point - once you've pulled out all of the `switch` statements into their own functions, you should notice that you end up with a function consisting of a bunch of `for` loops which all look the same - which means that you can pull that common code out into its own function (which is in essence garbling an entire string) - hopefully that should be fairly apparent once you get there.

Shiranui-Izayoi:
> 
> 
> Yeah - I appreciate it's a lot of changes (sorry about that) - I just feel like it would be a shame to improve the performance in this way without going all the way and refactoring it into something a bit more modular, rather than sacrificing readability for performance when we could have the best of both worlds. Like I say though, in principle I think it's definitely a good change.
> 
> As far as each of the points go, it sounds like you've got it. On the second point - once you've pulled out all of the `switch` statements into their own functions, you should notice that you end up with a function consisting of a bunch of `for` loops which all look the same - which means that you can pull that common code out into its own function (which is in essence garbling an entire string) - hopefully that should be fairly apparent once you get there.

Yes I guess that makes sense. I will do my best. It might need a little going over when it is done most likely it will be for the actual way it is garbled. Like what gets garbled to what. Since my general idea was to allow number and symbol useage since a lot of people use the symbols for expression.

And I am sure others that aren't me will have a better idea of what could use caps, what can't what symbols make sense etc. But once it is readable and modular that should be easy. Anyway I will start to work on it when I can/have the motivation to. Thanks for the suggestions for improvement. 

Honestly though changing it switch was fairly easy. Mostly just copy and paste for the capital letter versions, with Notepad++ convert to caps. After that I could honestly do a lot of find and replace. Anyway I am off to sleep now it is late.

PR #2042, opened 2021-02-06
Increased sensory deprivation level

Ada18980:
With "Hide others messages" and "Sensory Deprivation: Total" enabled in the immersion options, being totally blind and deaf will now outright hide all chat messages and emotes from other players, unless the emote has the player's name in it (not case sensitive).

This means that unless someone physically interacts with you or whispers to you, you have no way of knowing that someone is in the room with you.


Ben987:
This is intense!  But I'm sure that some players will like it, thanks :)

PR #2043, opened 2021-02-06
In Sensory deprivation, player names in emotes get replaced with Someone

Ada18980:
E.g.

"Ada laughs at Bob"

becomes "Someone laughs at Someone"

This does not replace the player's name, however, so "Charlie tickles Ada" becomes "Someone tickles Ada," assuming I am Ada.

This may be an issue if someone's name is a common word, like slave. For now, it only works if the first letter of the name is capitalized, so if someone types "Charlie tickles slave" and slave is in the room, then slave will not get replaced.


Ben987:
Alright!  I'm curious to try it.

PR #2044, opened 2021-02-06
Bugfix: Reinstates missing asset for the vibrating dildo

elliesec:
It seems that the dildo layer of the Vibrating dildo accidentally got deleted as part of #2021. This PR reinstates the missing asset.


Ben987:
Thanks for the fix!

PR #2045, opened 2021-02-06
Feature/Refactor: Modular item generalisation and extended item archetypes

elliesec:
## Summary

Another relatively big change I'm afraid, but the impact should be minimal (the only item affected by this change right now is the High Security Straitjacket).

This PR does two main things:
* Refactors the extended item script for the High Security Straitjacket into a new `ModularItem.js` script, which allows items to be created with nested menus, like the high security straitjacket (this could also be used by the new Kigurumi Mask in #2028 so that it no longer requires its own script)
* Adds a new `Female3DCGExtended.js` file, where extended items (currently only items that use the modular item script) can be registered and configured. Items registered in this file have their extended item `Load`, `Draw` and `Click` functions _automatically_ generated at asset build time, removing the need for the items to have their own extended item script.

The PR also adds a new `AllowModuleTypes` property, that should make it much easier to set an individual layer's `AllowTypes` array for modular items by specifying single module values (e.g. `AllowModuleTypes: ["a1"]` - again, will probably be very useful for #2028).

Eventually, we could move most items making use of the current `ExtendedItem` and `VibratorMode` scripts to use this system, which would allow us to completely remove dozens of extended item scripts, and have them all registered within `Female3DCGExtended.js`, which should make the creation of most generic extended items much simpler, requiring only an asset definition in `Female3DCG.js` and extended item configuration in `Female3DCGExtended.js`, and the addition of relevant assets and dialogue entries.

When/if this change is merged, I'll likely extend the current `AssetCheck` script to also cover `Female3DCGExtended.js`, so that extended item configurations can also be validated at PR time.

In the future, it should be possible to extend the idea even further, and allow submodules that can take a different extended archetype (for example, and vibration module, along with standard type-based modules), to allow for complex composite items with types and extra functionality in a uniform interface. For now though, this should be a good step in the right direction, without being too much of a complete system overhaul.

## Testing

The only item that is currently affected by this change is the High Security Straitjacket, and I've fully tested it to ensure that it still works exactly as before.


Ben987:
That's another big one, you're on fire.  Since it only targets one item it should be pretty safe to include it in the next beta.  Thanks :)

PR #2048, opened 2021-02-07
small controller fix for appearance menu

fleisch11:


Ben987:

PR #2049, opened 2021-02-07
Notifications Enhancements

Nina-1474:
Changes:
- A new option has been added to raise a notification when someone enters the chat room while the player is not focused on the window, which can be restricted to owners and/or lovers and/or friends only.
- An option to trigger the 'beep' sound on receiving a notification has been added. This setting is available in the Audio preferences screen as well. The sound will only occur for the first instance of a notification type. For example one chat message and one player entrance will prompt two beeps, whereas two chat messages will only prompt a beep on the first message. This won't occur for Beep notifications since player beeps already have a dedicated sound setting.
- The notifications code has been moved into its own file instead of cluttering up Common.
- The code to create a grey box to simulate a disabled checkbox has been generalised into DrawCheckboxDisabled for re-use.
- Minor fixes and improvements to notifications in general.


Ben987:
Yes!  This should be useful for many players.  Thanks for that one :)

PR #2050, opened 2021-02-07
Photo Blink Fix

Nina-1474:
Restoring a change from #2014 that got lost in a merge.


Ben987:
Sure!  Thanks for the fix :)

PR #2040, opened 2021-02-05
Refactor: ChatRoomDrawCharacter

jomshir98:
Refactors `ChatRoomDrawCharacter` function to be more readable and more easily extendable.

Functional changes: Player can now always click self or current whisper target to stop whispering

Technical changes:
Separates character overlay icons drawing into `ChatRoomDrawCharacterOverlay`
Separates character click handling into `ChatRoomClickCharacter`

All clicking tested carefully

PR #2052, opened 2021-02-07
Feature: Version number

jomshir98:
~~**Depends on (and includes): #2040**~~

Adds a `GameVersion` variable, set in index. @elliesec agreed to update it with each changelog update

The version is always saved in `OnlineSharedSettings.GameVersion`
When the version increases, maid will meet the player, telling them the Club updated and option to open changelog.
When another player has different game version, warning icon is shown above their head (this will only be visible during betas and is aimed to warn players to expect more problems when interacting with others; they can also confirm both players are on same version before reporting bug)

The version has format: `R[0-9]+(Beta[0-9]+)?`
For example (ordered ascending)
- `R64`
- `R65Beta1`
- `R65`


Ben987:
This one has a conflict with another pull request.  Can you resolve it and test again?  Thanks!

jomshir98:
@Ben987 Rebased and resolved!

Ben987:
That should be helpful for some.  Thanks!

PR #2053, opened 2021-02-08
Bugfix: Prevents expression reset from removing lip color

elliesec:
## Summary

With the introduction of the color picker to the expression menu, and the modification to the reset button, the reset button now removes players' lip color, which was something of an oversight - it seems logical to have the reset button only reset colors on groups that aren't body parts. This PR prevents the button from resetting a player's mouth color.


Ben987:
Sure!  Thanks for the fix.

PR #2054, opened 2021-02-08
Kigurumi Mask Refactor

Ada18980:
Refactors the Kigurumi Mask to use the same code as the high security straitjacket for cleaner code--thanks Ellie!


Ben987:
Sure, thanks!

PR #2056, opened 2021-02-08
Bugfix: Several lock-related fixes

elliesec:
## Summary

This makes several small fixes to certain lock functionality. ~~I'd be grateful if @Ada18980 could cast an eye over this to ensure that I've not inadvertently broken anything about lockpicking or the high security padlock (although I've tested it, and it seems to work fine)~~ (Ada has confirmed that this shouldn't cause any issues).

Fixes:
* Fixes an issue where `MemberNumberList` was being incorrectly set to a string in `InventoryLock`, rather than an array, which broke use of `MemberNumberList` in timer padlocks
* Fixes an issue where people were unable to add/remove timer from owner timer padlocks
* Fixes a problem where the "Append owner" feature of the high security padlock was appending the owner's name rather than their member number
* Fixes an issue where the "Append lovers" feature of the high security padlock would crash if the player had an NPC lover (as NPCs have no member number)


Ben987:
Thanks!  Right on time for the beta version :)

PR #2057, opened 2021-02-08
Bugfix: DialogHasKey error fix

elliesec:
##

This fixes an error when viewing an equipped item via the dialog screen which intrinsically has the `Lock` effect (e.g. the metal cuffs or the four limbs shackles). The code was erroring because `InventoryGetLock` was returning `null`.


Ben987:
Alright!  Thanks for fix.  Cheers :)

PR #2059, opened 2021-02-09
Collar Ropes

Codefreak64:
Adds a rope variant of the collar chains in the neck restraints.
-all can be bought at once
-works exactly like the collar chains
thank you ayesha for helping with the preview images.


Ben987:
Sure!  Thanks a lot you two :)

PR #2061, opened 2021-02-10
Chatroom Photo Changes

Nina-1474:
Changes to the new photo feature based on feedback:

- Adding a chatroom "Photographs" block category which disables the full room photo button and character photo dialog options.
- Since the available chatroom menu buttons can change, I've refactored them to be built/drawn/clicked dynamically in a list, similar to the recent Appearance menu button refactor.
- A dialog option for the player's character to take a self-portrait has been added.
- Changing all browsers to use the Chrome opening method to avoid lag.


Ben987:
Sure!  Thanks for all of these changes :)

PR #2062, opened 2021-02-10
Improvement: AllowModularTypes enhancement

elliesec:
## Summary

This PR is an enhancement to the modular item script created in #2045, which makes the `AllowModularTypes` property on layers more flexible. It is now possible to specify multiple module options in an `AllowModularTypes` entry to cause the layer to only be drawn if all modular types are met.

The new behaviour is that a layer is drawn if the item's current type matches _any_ of the entries in `AllowModularTypes`. A type is considered a match for an entry if it matches all modular options specified in that entry.

For example, in a modular item with 3 modules: `a`, `b` and `c`, where each module has 3 options (0, 1 and 2), the entry `AllowModularTypes: ["a1c0"]` would expand to:
```js
// All types containing both a1 and c0
["a1b0c0", "a1b1c0", "a1b2c0"]
```
And `AllowModularTypes: ["a1c0", "b2"]` would expand to:
```js
// All types containing either b2, or both a1 and c0
["a1b0c0", "a1b1c0", "a1b2c0", "a0b2c0", "a0b2c1", "a0b2c2", "a1b2c1", "a1b2c2", "a2b2c0", "a2b2c1", "a2b2c2"]
```
Currently this isn't used for the two existing modular items, but @Ayesha-678 has an item on the way that would need this level of granularity to succinctly describe layer types.


Ben987:

PR #2063, opened 2021-02-10
Clothes Back/Next Fix

Nina-1474:
The hover text for the back/next arrows for clothes groups in the Appearance screen were the wrong way around.
Thanks to @elliesec for spotting it!


Ben987:
Thanks for the fix!

PR #2065, opened 2021-02-10
Bugfix: Ensures an extended item's exit function is called on extended item type selection

elliesec:
This ensures that any cleanup on items that have extra features (e.g. input boxes) in their extended menus gets executed in single player areas.


Ben987:
Sure. Thanks for the fix.

PR #2066, opened 2021-02-10
Bugfix: Fix for disappearing lover/owner locks on relog

elliesec:
## Summary

This fixes disappearing owner/lover locks on relog by changing the `LoginResponse` function to report that the appearance bundle originated from the player themselves. I'm not exactly sure why this wasn't already an issue, but this should fix it.


Ben987:
Yes!  Thanks for that fix.

PR #2067, opened 2021-02-10
Transport Box Exit fix

Nina-1474:
Linking the exit function so controls are properly removed.


Ben987:
Sure!  Thanks for the fix.

PR #2068, opened 2021-02-10
Fix: Mobile landscape MouseX

NoneNoname:


Ben987:
Lots of players will be happy if it fixes the mobile bugs.  Thanks!

PR #2069, opened 2021-02-10
Wardrobe Cancel Button

Nina-1474:
After the menu button refactor, the wardrobe only had an Accept button which jumped back up to the main appearance screen, maintaining the 'Dress manually' behaviour from before. Having to then Cancel from that screen in order to not save the wardrobe change was counterintuitive and was reported as an issue by users.
Now the wardrobe screen has a Cancel button of its own that restores the appearance from before entering the wardrobe, and jumps back up to the main appearance screen to stay consistent with other subscreen accept/cancel button behaviour.


Ben987:
Sure, thanks!

PR #2070, opened 2021-02-10
Futuristic, stimulation, and lock fixes

Ada18980:
Fixes the following
-Crash on clicking a locked-out futuristic vibrator
-Picking a lock that's impossible will allow you to set all pins without unlocking it
-Futuristic harness ball gag does not gag when initially put on
-Lock layers on all futuristic items appear all the time
-Stimulation messages for vibe plug front/back are fixed
-Futuristic straitjacket default color
-Futuristic straitjacket now hides bra and panties in the chastity-hiding versions.
-Stops stimulation messages from causing orgasms while edged
-Reduced spam from stimulation messages happening when they werent supposed to


Ben987:
Nice!  Thanks for all the fixes.

PR #2071, opened 2021-02-10
Bunny Tail Vibe Plug fix

Nina-1474:
The BunnyTailVibePlug item was in a different position in the 'js' and 'csv' files resulting in the wrong preview appearing.


Ben987:
Yep!  Thanks for fixing it :)

PR #2073, opened 2021-02-10
Bugfix: Adds missing kneeling assets for the Transport Jacket

elliesec:
## Summary

The Transport Jacket was missing support for the kneeling pose. This adds `Kneel` to the item's `AllowPose` array and adds the missing assets.


Ben987:
Very good!  Thanks for adding that.

PR #2074, opened 2021-02-11
Bugfix: Fixes the click handler for the clit & dildo vibrator belt

elliesec:
Back in #2004, I removed some duplicated click handler logic for some vibrator items which had invisible exit "buttons" independent of their actual exit buttons. In the case of the Clit & Dildo Vibrator Belt, I accidentally deleted the wrong click handler, so you can only exit the extended item screen by clicking 200px lower than the actual button. This fixes that.


Ben987:
Thanks for the fix!

PR #2076, opened 2021-02-11
Bugfix: Ensures lovers can unlock intricate/high security locks

elliesec:
Lovers couldn't unlock high security or intricate padlocks, which seems incorrect given that game has always followed the general rule of owner > lovers > everyone else.


Ben987:
Yes!  Absolutely.  Thanks a lot.

PR #2077, opened 2021-02-11
Futuristic chastity belt dialog

Ada18980:
Changes the button names on the futuristic chastity belt to avoid confusion. 

They are now:

Open Front
Open Back
Closed


Ben987:
Thanks for the change!

PR #2078, opened 2021-02-11
Lock picking fixes

Ada18980:
Fixes 2 things with lock picking:
-People getting tired a lot less than they should be due 
-Chat room messages not being published properly


Ben987:
Thanks for the quick fixes.

PR #2079, opened 2021-02-11
Beta Fix: Pet Bed clipping

Codefreak64:
When the item is in use a white line appears through some tails and some wings. 
this fix changes what parts of the bed are layered (since the line was the asset in front of the tail)
-Visual fix for ItemDevicesPetBed
-clipping fix with wings and tails


Ben987:
Alright!  Thanks for the fixes.

PR #2083, opened 2021-02-12
Fix: Render warning

jomshir98:
Fixes following warning in console: `RENDER WARNING: texture bound to texture unit 0 is not renderable.`
The reason for this warning is attempt to render not yet loaded texture.
The fix works by assigning empty, 1px transparent texture, that is replaced when image loads


Ben987:
Sure!  I've seen that one before.  Thanks for the fix!

PR #2084, opened 2021-02-12
Fixed Position Assets

Nina-1474:
This creates a new `FixedPosition` setting for assets. When true, the asset will always appear in the same position in the character's space regardless of how much the character is moved around due to poses or height modifiers.
It does this by calculating where to position the asset in the canvas in order to 'cancel out' all height modifier movement and shrinking for short character calculations that occur later on. 180 degree rotations are applied when inverted so the asset is right-way up again after inverting later.
There is one small unavoidable graphical oddity: when the character is drawn full size and inverted, the asset will move out of line a little. This force full-size drawing only happens in the character appearance screen (where being inverted is already a bit broken), the arousal preferences screen, and the maid and dojo mini-games, so it's unlikely to even be noticed.

When the setting is off, there's no effect on the current drawing process. I haven't enabled the feature for any assets in this PR because there are some beta issues for new items that this resolves (in other dependent PRs) which would be useful to get into this release.


Ben987:
That's a pretty big beta fix!  Thanks a lot :)

PR #2085, opened 2021-02-12
Pet Bed Fixes

Nina-1474:

Also it would be preferable to merge #2086 before this.

---

This fixes issues of the Pet Bed appearing in the wrong position when in all-fours or suspended, as well as clipping issues. This is done by setting the new FixedPosition setting on and applying a single alpha mask. This removes the need for pose-specific images and alpha masks. Discussed and agreed with @Codefreak64.


Ben987:
This one creates a conflict with the other pull request.  Can you resolve them and test again?  Thanks a lot.

Nina-1474:
Resolved and retested.

PR #2086, opened 2021-02-12
Item Tweak: Pet Bed

Codefreak64:
A few changes to my pet bed item, these reflect and are meant to compliment PR #2085 with more fixes unrelated to those relied in PR #2084 

It does not need either of the PRs listed above to work and function as intended, it just refines code from the base item such as removing Item hides and getting the effects right


Ben987:

PR #2087, opened 2021-02-12
Ceiling Rope/Chain changes

Nina-1474:
At the moment a player can hang from a single rope/chain to the ceiling connected to a collar item, which has rather morbid connotations. Putting in a change to remove ItemNeck, leaving the arms/torso/pelvis as the possible points to connect to, and hide neck rope in particular. Discussed with @Ayesha-678.
Also added RemoveItemOnRemove to avoid having the ceiling rope left behind unable to be removed.


Ben987:
Sure, let's try it right now :)

PR #2088, opened 2021-02-12
Wooden Box Fixes

Nina-1474:

---

This fixes issues of clipping out of the top and bottom of the Wooden Box and Transport Box, and simplifies the kneeling support, by enabled the FixedPosition setting and placing alpha masks around the outside of the boxes.
Also while the character appeared behind the left and right borders of the front panel, this was not true for the top and bottom borders. An extra layer accomplishes this to make the character appear inside the box more.


Ben987:
Yes!  Absolutely.

PR #2089, opened 2021-02-12
BBQ Fixes

Nina-1474:

---

This fixes the issue of the BBQ Grill sticking to the ceiling when suspended by the legs, by enabling the FixedPosition setting. This also allows the duplicate images for different poses to be removed. The preview image has also been fixed up to not appear distorted. Discussed with @Ada18980.


Ben987:
You're on fire!  Thanks a lot :)

PR #2090, opened 2021-02-12
Fix: Canvas and element positioning

jomshir98:
Fixes Canvas css position and replaces logic for positioning elements.
Tested on both mobile and desktop (both landscape and vertical)
There should be no change in look against earlier versions


Ben987:
Yes!  That's the most important fix for mobile players, lets hope it will fix all the issues.  Thanks a lot :)

PR #2091, opened 2021-02-12
Item Tweak: Pet Bed again

Codefreak64:
put in the wrong effect, put in freeze when tethered should have been in there. this PR makes sure that happens and you get clothing access while in the bed


Ben987:

PR #2092, opened 2021-02-12
Fixed Assets Up Button fix

Nina-1474:
If you're using the Up button to view a character with a fixed-position asset in a non-standing pose and someone else changes something to refresh that character, the asset gets moved to the wrong position as if they're standing.
This would get automatically corrected in the next character refresh, but could look wrong on your screen for a little while. Now the correcting refresh happens as soon as the "Up" mode is exited.


Ben987:
Yes!  Thanks for the fix, cheers :)

PR #2093, opened 2021-02-12
Bugfix: Modular item framework fixes

elliesec:
## Summary

This restores previous behaviour on the High Security Straitjacket where it was not possible to modify a locked straitjacket without being able to unlock it. It also fixes an error that occurs in the extended item script for the validation of extended item options that have no `Property` (which wasn't an issue until modular items).


Ben987:
Thanks for the fix, I will slide it in the next beta :)

PR #2094, opened 2021-02-12
Fix: Zoomed touch position on safari

jomshir98:


NoneNoname:
ColorPicker.js has some old event listeners that should be updated this one as well:
`function ColorPickerGetCoordinates(Event)`

jomshir98:
> ColorPicker.js has some old event listeners that should be updated this one as well:
> `function ColorPickerGetCoordinates(Event)`

Thank you @NoneNoname! I put a fix in and asked @elliesec to test it.

elliesec:
I've just tested this on iOS, and can confirm that the color picker also now seems to be working correctly when zoomed (barring the drag functionality, but I suspect that's always been the case).

PR #2095, opened 2021-02-13
Item Permissions Colour fix

Nina-1474:
For assets that had both Item and Clothes versions, the button background was not properly set to red in permissions-mode when the item is blocked.


elliesec:
Can't remember off the top of my head, but does end up causing issues with permissions mode for the spanking toys?

Nina-1474:
> 
> 
> Can't remember off the top of my head, but does this end up causing issues with permissions mode for the spanking toys (as the function is reused in `DialogDrawItemMenu`)?

I considered that too and checked, and it seems fine.

Ben987:
Alright!  Thanks for fixing the previous fix :)

PR #2096, opened 2021-02-13
Hands Colour Fix

Nina-1474:
Hands were removed as a separate appearance option relatively recently and will always take the colour Default which lets them inherit the BodyUpper colour. However saved wardrobe slots from before that time will have Hands saved with a non-Default colour. Trying to change the body colour after loading a slot will leave the hands at the old colour with no way to change them. This forces hand colours to always be Default.


Ben987:
Yes!  This makes perfect sense.  Thanks :)

PR #2097, opened 2021-02-13
Add High Security Lock to Strong Locks

Nina-1474:
Since the High-Security lock can only be unlocked by owners/lovers and the equipper (plus anyone the equipper delegates) but can be applied by anyone, someone without an owner/lover could be permanently locked into a corset or other non-restraint item by a stranger.
Adding the lock to the 'strong lock' list alongside others like the password locks to prevent this.

### DIFF
--- a/BondageClub/Screens/Room/MainHall/MainHall.js
+++ b/BondageClub/Screens/Room/MainHall/MainHall.js
@@ -14,7 +14,11 @@ var MainHallMaidWasCalledManually = false;
 
 var MainHallBeingPunished = false;
 var MainHallFirstFrame = false;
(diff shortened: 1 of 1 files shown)


Ada18980:
Hey @Nina-1474 and @Ben987 -- these locks are intended to be pickable by any player with lockpicks who is not restrained. Right now there is an unintended bug where extreme mode players are locked into preventing lock picking being done to them, which is not intended.

Since Mistress Locks are not considered strong locks, I don't think High Security padlocks should, since I have seen people picking them fairly reasonably often in the beta.

Ben987:
Yes, thanks you two :)

PR #2098, opened 2021-02-13
Bugfix: Stops long password lock hints from overflowing

elliesec:
Not exactly a beta fix - this just prevents the hints on password locks from overflowing for particularly long hints by using `DrawTextWrap` instead of `DrawText`.

Before:

After
image


Ben987:
Sure!  That's a good idea.

PR #2100, opened 2021-02-14
Fix "Disable picking locks on self" forced in hardcore

Ada18980:
This is not intentional, not sure why it was set this way.


Ben987:
Sure!  Thanks for changing that.

PR #2101, opened 2021-02-14
Owner and Lover locks are stronger

Ada18980:
Makes owner and lover locks more difficult to struggle out of.


Ben987:
No, having a different type of lock doesn't make it harder to struggle out of an item.  You don't pick the lock, you struggle out.

PR #2102, opened 2021-02-14
Self Lockpick Fix

Nina-1474:
Whenever anyone successfully picked a lock on themselves, the screen would freeze.


Ben987:
Yes!  Thanks for the fix :)

PR #2103, opened 2021-02-14
Updated lockpicking intro text

Ada18980:
Updated it to be more descriptive of the correct method.


Ben987:
Absolutely!  Thanks for clarification.

PR #2104, opened 2021-02-14
Main Hall Maid Release fix

Nina-1474:
When the player automatically enters a chatroom on logging in, then later talks to the main hall maid, the dialog functions don't work. In particular trying to get released doesn't work and returns an error of 
> Trying to launch invalid function: MainHallChatSearchMaidReleasePlayer

This is because by the time the dialog functions are being created, `CommonScreen` is "ChatSearch" instead of "MainHall" as required. This error predates this release.
Now the loading of the NPC is delayed until the player loads the main hall when not about to jump into a chatroom.


Ben987:
Yes!  Thanks for the fix :)

PR #2105, opened 2021-02-14
Bugfix: Adds a missing AllowType to the Halo in HairAccessory1

elliesec:
The `HairAccessory1` version of the halo was missing an `AllowType` array, causing selection of the "Broken" type not to persist properly.


Ben987:
Sure, thanks for the fix.

PR #2107, opened 2021-02-14
Bugfix: Ensures Futuristic Vibrator works with the default triggers on first equip

elliesec:
This causes the futuristic vibrator to use the default triggers when it is first equipped, without the need for someone to visit its extended item menu first. If no triggers have been set, the vibrator will respond to the default triggers in `ItemVulvaFuturisticVibratorTriggers`.


Ben987:
Thanks for the fix!

PR #2108, opened 2021-02-14
Bugfix: Removes incorrect LockLayer properties

elliesec:
This removes a few `LockLayer: true` properties from the Lock Necklace, Love Chastity belt and High Security Straitjacket that were accidentally added by #2070, which were causing the items' `Lock` layers not to display.


Ada18980:
Oh no! Sorry about that >.<

Ben987:
Yes!  Thanks for the fix :)

PR #2109, opened 2021-02-15
Bugfix: Fix for an infinite loop in the shop job

elliesec:
## Summary

This fixes an infinite loop in the shop job that can occurs if the player is wearing a gag that doesn't muffle their speech (e.g. the futuristic gags in padded mode). I've removed the `while (true)` altogether and replaced it with a safer alternative, and added dialogue lines to the shopkeeper to handle this case

## Steps to reproduce

1. Equip arm, ankle and leg cuffs, but don't connect them
2. Equip a futuristic gag and set it to padded mode
3. Start the shop job. The shopkeeper should give you a blindfold (the issue only occurs if the blindfold doesn't also gag the player)
4. After completing the job (either successfully or unsuccessfully), ask to try again
5. The game will freeze in an infinite loop, as all demo groups are occupied, so the `while` loop in `ShopJobRestrain` will never terminate


Ben987:
Never seen that one, but it makes sense.  Thanks for the fix :)

PR #2110, opened 2021-02-15
Remove AllowItem from security action prerequisites

Ada18980:
This will avoid mistresses having to whitelist their subs to allow subs to hand over their keys~


Ben987:
Yes!  Thanks for that one.

PR #2112, opened 2021-02-15
Fix: Lover items disappearing if player has no owner

jomshir98:


Ben987:
Alright!  Thanks for the fix, I will slide it in the next beta :)

PR #2113, opened 2021-02-15
Bugfix: Fixes a crash in LARP item selection

elliesec:
Fixes an issue which causes a crash in LARP when a player tries to select an item, caused by accidentally mistaking the `Item` variable for an item rather than an asset back when I made the preview image changes in #1989.


Ben987:

PR #2114, opened 2021-02-15
Shackles Over Head Fix

Nina-1474:
For characters with a short fringe hairstyle, the wrist shackles in the over-the-head option had the chain appearing between the head and the front hair. The overlap between the chain and the head has been removed to avoid this.


Ben987:
Thanks for the fix!

PR #2116, opened 2021-02-15
AllFours ItemTorso fixes

Nina-1474:
Some of the corsets in ItemTorso that were converted to copy the Corset versions were no longer being hidden in the AllFours pose, resulting in them hanging below in midair.


Ben987:

PR #2117, opened 2021-02-15
Kneel Button Fix

Nina-1474:
The chat room Kneel button would always make the player stand on the first press if they had any pose other than kneeling. Even if that pose was already standing in some form, e.g. legs closed or arms in any pose. 
Now it should always make the player kneel or stand as expected and will no longer remove the arms pose.


Ben987:
Great!  Thanks for that fix :)

PR #2118, opened 2021-02-16
Bugfix: Futuristic vibrator crash

elliesec:
## Summary

This fixes an issue where the Futuristic Vibrator's extended item menu would crash if the vulva zone was blocked. If the zone was blocked, the inputs were not being added, and subsequently calling `ElementPosition` in the draw function would crash. I've put in a fix to the `ElementPosition` to ensure it logs a warning in such situations, and I've modified the extended item screen for the vibrator the vibrator now functions properly regardless of whether or not the zone is blocked.


Ben987:
Yes!  Thanks for that fix, right on time for the next version :)

PR #2119, opened 2021-02-16
Refactor: Refactors the crib to use the modular item script

elliesec:
## Summary

This PR refactors the Crib to use the new modular item script, as it was using the same pattern as the High Security Straitjacket/Kigurumi Mask. I've tested and confirmed that there are no changes in functionality or behaviour. I've also notified @mrbakucamp regarding the change.


Ben987:
Of course!  Thanks a lot :)

PR #2121, opened 2021-02-17
Add Edged to AllowEffect of most vibrating toys

Ada18980:
This is to prevent cases where the Edge and Deny modes won't work with those items.


Ben987:
Thanks again for that new fix :)

PR #2122, opened 2021-02-17
Bugfix: Prevents the handheld remote from bypassing remote owner rules

elliesec:
## Summary

Because the new handheld vibrator remote is in the same buy group as the existing vibrator remotes, it can be purchased in the shop to bypass the "cannot buy remotes" owner rule, and is not confiscated at the same time as other remotes, causing them to appear in the player's inventory on their next login. This PR fixes these issues.


Ben987:
Absolutely!  Thanks :)

PR #2125, opened 2021-02-17
Missing SuitLower Names

Nina-1474:
Adding in the missing csv asset names for the newly added Suit Lower copies of Socks items.


Ben987:
Yes!  Thanks for that fix :)

PR #2127, opened 2021-02-17
Bugfix: Stops skill modifiers from overlapping skills in the information sheet

elliesec:
## Summary

With the addition of the lockpicking skill, the skill modifiers in the information sheet can now overlap with the skill list if the player has every skill in the game (see image). Luckily, there's still (just about) space to move the skill modifiers down a line.

image


Ben987:
We might be stuck with the next skill, but lets go with that for now :)

PR #2128, opened 2021-02-18
Safeword Padlock and Electronic Tag w/ Tamper Evidence Integration

Ada18980:
Adds a padlock that leverages the new server functionality in https://github.com/Ben987/Bondage-Club-Server/pull/67

This lock is like the password padlock, except:
-The user who locked it on, and the date of locking, is displayed
-The player who is locked can see the password

Also adds an Electronic Tag which is like the custom name tag, but it also tells you if all the locks on the player are yourrs or not. This is to help doms make sure that their subs have not swapped out their locks.


Ben987:
Lets put that one on hold for now.  A pretty big server change is coming for packet size and compression.

Ada18980:
@Ben987 Sure.

Would you prefer I make a version without the dependency on server changes? The futuristic tag and safeword padlocks both have features which work regardless of the server change.

PR #2129, opened 2021-02-18
Fix/Refactor: Asset description

jomshir98:
Changes how asset description is resolved, preventing the error of names being assigned to different items.

Tested, including translations

### DIFF
--- a/BondageClub/Assets/Female3DCG/Female3DCG.csv
+++ b/BondageClub/Assets/Female3DCG/Female3DCG.csv
@@ -236,7 +236,7 @@ Shoes,SandalsRS,Rhinestone Sandals
 Shoes,PawBoots,Paw Boots
 Shoes,WoollyBootsTall,Woolly Boots
 Shoes,ThighHighLatexHeels,Latex Heels
-Shoes,Heels3,Multi-Strap Heels,
+Shoes,Heels3,Multi-Strap Heels
 Shoes,BarefootSandals1,Barefoot Sandals
 Shoes,LatexAnkleShoes, Latex Ankle Shoes
 Shoes,Flippers,Flippers
--- a/BondageClub/Assets/Female3DCG/Female3DCG.csv
+++ b/BondageClub/Assets/Female3DCG/Female3DCG.csv
@@ -553,6 +553,10 @@ Emoticon,,Emoticon
 Emoticon,Emoticon,Emoticon
 Eyebrows,,Eyebrows
 Eyebrows,Eyebrows1,Eyebrows
+Hands,,Hands
+Hands,Default,Default
+Head,,Head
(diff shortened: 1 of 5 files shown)


Ben987:
Good idea!  And it doesn't slow down assets loading?

jomshir98:
> Good idea! And it doesn't slow down assets loading?

Yes, it does slow it down. To be exact it takes now 2.3ms instead of 1.4ms to load all assets

Ben987:
@elliesec , @Nina-1474 , with this change, Female3DCG.csv could become quite a mess and the game would still work with it.   Can you think of way we could still keep some structure in there?  Thanks!

elliesec:
I guess there are a few things you could do, but personally I'd say the best way would simply be by checking pull requests that change it. If you really wanted to, it wouldn't be too hard to write a script like the asset check one to check that the entries are suitably grouped, although I don't personally feel that should be necessary.

Other than that you could break the file up into multiple files, or use a more structured format than CSV, but I'd say those would be more trouble than they're worth (not to mention the fact that loading 65 CSV files would be a _much_ bigger performance hit than this change).

Ben987:
Yeah, lets not create multiple CSVs, it will cause more issues and more F5 for players.   And I don't want to run a script each time I pull a new asset change.

Let me know if you find a good solution to keep the CSV in order, we can keep that one on hold until then.

jomshir98:
@Ben987 I can (and planned to do this anyway) write tests in AssetCheck for this

Ben987:
@jomshir98 Awesome!  Let me know when it's ready.

jomshir98:
@Ben987 I've made the checks. Currently it is failing because:
> ERROR No description for asset "ItemHands:SpankingToysLongDuster"
Do you have a name for me to put there?

Ben987:
"Long Duster" is fine for the name.  Thanks!

jomshir98:
@Ben987 Done!

Ben987:
Great!  So how will it work?  We will receive in-game console errors if a description is missing or in the incorrect order?

jomshir98:
> Great! So how will it work? We will receive in-game console errors if a description is missing or in the incorrect order?

No, in game the order doesn't matter at all and there shouldn't be errors in in-game console. If an asset without name somehow makes it in, it will be seen as `MISSING ASSET DESCRIPTION: Group:Name`

The check already happens the same as assetcheck script - putting the cross or checkmark next to the commit, with ability to click on the cross and open details to see error (just like is visible with the "Add check in AssetCheck" commit above, which had error about the Long Duster. Ordering is also checked in the same place

Ben987:
Perfect!  If we can see the missing name or the wrong sort order before committing the pull request it's even better.   Thanks for that one.

PR #2130, opened 2021-02-18
SuitLower Order Fix

Nina-1474:
Correcting the SuitLower name order so they appear on the right assets.


Ben987:
Thanks!  I will prepare a hotfix later today :)

PR #2131, opened 2021-02-18
Bugfix: Compatibility fixes for the floor shackles

elliesec:
## Summary

This adds the floor shackles to the list of "leg spreader" items in the `NoFeetSpreader` prerequisite, as they have incompatibilities with several arm items that would usually force the legs closed (see pictures). It also adds the `CanKneel` prerequisite to the strict leather pet crawler in line with the pet suits.

image

image


Ben987:
Thanks!  I will prepare a hotfix later today :)

PR #2132, opened 2021-02-18
Lockpick Dealer

Ada18980:
Adds lockpicks to the list of items you can buy from the asylum patient in the meeting hall. Allows subs to get lockpicks even if their owner doesn't let them. The owner can still confiscate them though.

I think this is acceptable because you can already be given lockpicks by other players and it adds a bit of a puzzle to subs stuck in intricate locks.


Ben987:
Sure!  That's a fun idea :)

PR #2133, opened 2021-02-19
Bugfix: Adds some missing dialogue and audio entries

elliesec:
## Summary

The recent change to player dialogue (#1981) has highlighted a few places where there is missing dialogue in the game. This adds dialogue entries where appropriate, and also adds some missing audio for manually triggering the automatic collar shock unit via the shock unit remote.


Ben987:
Absolutely!  Thanks for that one, I will prepare a hotfix.

PR #2134, opened 2021-02-19
Giving keys is possible when wearing just collar/nonrestraint

Ada18980:
Removes the .IsRestraint check from the key giving functions.


Ben987:
Sure!  Thanks :)

PR #2135, opened 2021-02-19
Futuristic Vibrator dialog change 

Ada18980:
Changes the dialog when switching modes to something more neutral so it doesn't break immersion.


Ben987:
Absolutely, thanks again!

PR #2136, opened 2021-02-19
Gag stimulation messages

Ada18980:
Adds a stimulation message for gags with GagTotal and GagTotal2, since these are usually dildos and are hard to breath with.


Ben987:
Sure!  More kinky messages :)

PR #2138, opened 2021-02-20
Improvement: Item save format

jomshir98:
**Requires [Bondage-Club-Server#68](https://github.com/Ben987/Bondage-Club-Server/pull/68)**
Otherwise blocked and limited items are not correctly updated in some cases

**Includes #2137**
**Should only be merged the release after #2137 is fully deployed to stable version!**
Otherwise players would loose their inventory and blocked items as well as others would be unable to correctly interpret this player's!

Changes the format in which `Inventory`, `BlockItems` and `LimitedItems` are saved with focus on smallest size transferred over network, expecting per-message-deflate


Ben987:
It's time to test the final step.  Thanks again :)

PR #2139, opened 2021-02-20
Rejoin Failed Message

Nina-1474:
When a player 1) has the rejoin chatroom setting on, 2) does not have the room recreate setting on, and 3) tries to rejoin a room that no longer exists, then they're greeted with an empty search screen and the "No chat room found" message which doesn't really communicate what happened.
Now in such cases the "Room doesn't exist anymore" or "This room is already full" messages will appear in the bottom left as they normally would if the user clicked on a room button. The Search button will now reset the message as well.


Ben987:
Yes!  It will make things a little clearer.  Thanks.

PR #2140, opened 2021-02-20
Futuristic chastity bra no display lock icon

Ada18980:
Removes the lock icon from the nodisplay type


Ben987:
Thanks!  I will prepare a hotfix for it :)

PR #2141, opened 2021-02-20
Makes futuristic earphones hidden by most hoods

Ada18980:
This helps prevent problems where the earphones look ugly on top of the hood graphic


Ben987:
Absolutely!  Thanks :)

PR #2142, opened 2021-02-21
Exit Club Confirm

Nina-1474:
It's been requested a few times to make it harder to hit the Main Hall's exit button and accidentally log out. Now it shows a popup to confirm the exit and give a chance to cancel.


Ben987:
Sure!  That's not a bad idea :)

PR #2143, opened 2021-02-21
Fix for crash if character with invalid effect enters the chat room

TessaTech:
This adds an additional check to `ServerValidateProperties` to check if Item.Property.Effect is an array.

This missing check actually was abused by a player to purposefully crash other player's clients.
This very same bug could be abused to enter chat rooms invisible under other circumstances

### DIFF
--- a/BondageClub/Scripts/Server.js
+++ b/BondageClub/Scripts/Server.js
@@ -281,7 +281,7 @@ function ServerValidateProperties(C, Item, Validation) {
 	}
 
 	// For each effect on the item
-	if ((Item.Property != null) && (Item.Property.Effect != null) && Array.isArray(Item.Property.Effect) == true) {
+	if ((Item.Property != null) && Array.isArray(Item.Property.Effect)) {
 		for (let E = Item.Property.Effect.length - 1; E >= 0; E--) {
 
 			// Make sure the item or its subtype can be locked, remove any lock that's invalid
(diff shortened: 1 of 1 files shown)


BondageClub/Scripts/Server.js
```diff
@@ -281,7 +281,7 @@ function ServerValidateProperties(C, Item, Validation) {
 	}
 
 	// For each effect on the item
-	if ((Item.Property != null) && (Item.Property.Effect != null)) {
+	if ((Item.Property != null) && (Item.Property.Effect != null) && Array.isArray(Item.Property.Effect) == true) {
```
elliesec, line 284: For the record, you can skip the `== true` comparison - `Array.isArray()` always returns either `true` or `false`. Additionally, the `Array.isArray()` condition makes the `Item.Property.Effect != null` condition redundant. Any value such that `value == null` will also cause `Array.isArray(value)` to return `false`.

TessaTech: Well, that `== true` is a personal codestyle but you're right that it's not the project's code style so i'll change it up. That `Array.isArray()` makes the `!= null` check redundant is a good call, though. I'll change that as well.

Ben987:
Thanks for the fix!  I will prepare a hotfix for the current version :)

TessaTech:
@elliesec The github name is fine, but thanks for asking!

PR #2145, opened 2021-02-22
Collar no longer needed for futuristic items to be locked out

Ada18980:
Due to a bug, a collar on the wearer was required for futuristic item settings to be locked out if a player cannot unlock it. Now the player is locked out if they don't have a collar either.


Ben987:
Sure!  Thanks for the fix :)

PR #2146, opened 2021-02-22
Change: Modifies the difficulties of the cryo capsule & coffin

elliesec:
## Summary

This changes the behaviour of the Cryogenic Capsule and the Vampire Coffin so that they are easy to escape from when open, even if locked. Previously when locked, they would have an effective base difficulty of 48 when open, and 54 when closed. Now they have a base difficulty of -20 when open, and 30 when closed.


Ben987:
It makes perfect sense.  Thanks!

PR #2149, opened 2021-02-23
Futuristic panel gag and shock collar now trigger on activity and emote

Ada18980:
There is a list of keywords in emotes that trigger a shock or inflation, and the activities one can do while gagged also cause the items to trigger.


Ben987:
Does it make sense to use the "MakeSound" property for that?

Ada18980:
@Ben987 I added a new property for it to avoid special-casing it or using a keyword search (I could just use the "Moan" keyword for it). Does this add any sort of overhead?

Ben987:
"MakeSound" sounds a little strange, but we can go with it.  Thanks!

PR #2150, opened 2021-02-23
Flexibility Minigame Difficulty

Ada18980:
Tweaks the flexibility minigame to be more difficult, since it was easy to escape at challenge 6 compared to the button mashing game.


Ben987:
Absolutely!  Thanks a lot.

PR #2151, opened 2021-02-23
Safeword Padlock and Electronic Tag w/o serverside changes

Ada18980:
This adds the electronic tag and safeword padlock from (https://github.com/Ben987/Bondage-College/pull/2128) without needing any serverside changes.

The safeword padlock is like a password padlock, except that the locked player can always see the combination

The electronic tag is just a reskinned version of the custom name tag.


Ben987:
Yes!  Without any server change it's way easier.  Since we are reworking the packets, we don't want to alter too much stuff.  Thanks!

PR #2152, opened 2021-02-24
Fix for crash caused by invalid item special case

TessaTech:
I found another potential exploit. I have no knowledge of it being used by now but in theory it could be abused to crash the client again.
If the an item prepped with Property.Effect being an object, the inspecting player's client freezes, making it very hard for the wearing player to get rid of the item again.
This, in itself, already is troublesome. But joining an chat room with the prepped item will also crash everyone in the room weither they're inspecting the item or not. The wearing player included.


elliesec:
Just as an FYI, I'm very close to pushing a pretty substantial PR which should negate this being an issue anywhere in the game by fixing it at the source.

TessaTech:
I see. In that case i'll refrain from further crash testing until your change is merged.

elliesec:
Sounds good. With that change, all appearance updates to any character will be passing through the same validation functions. So it should be easy to fix something like this at the root by just changing the validation function to handle it (for example if we were to see similar issues with the `Block` array, although I've included sanitization of that too).

Ben987:
Lets wait for the big PR then.  Thanks for fix Tessa :)

PR #2154, opened 2021-02-25
New Hairstyles 1 Fix

EmilyR42:
- Added 2 new HairFronts, 1 new HairBack (not satisfied with the second one just yet)


Ben987:
Hello.  There's no "HairBack47" images or entry, why are you adding it to the CSV file?

Ben987:
Thanks for the quick fix, the new hair looks great!

PR #2155, opened 2021-02-26
Item fix: Rope collar

Codefreak64:
This fix is to lower the collar itself which lines it up to the leashes and neck restraints. it also gets rid of the 'Hideitem' in the ceiling ropes, making the item get seen as well

-Updated Image file to line up with 'ItemNeckRestraints'
-Makes the item not hidden when using 'CeilingRope'


Ben987:
The Hideitem is there so you cannot "hang" someone from the ceiling.  Some players freaked out about it.  Can you leave it there?

Ben987:
Thanks!  I'll see how it looks at re-add the property myself if needed.

PR #2156, opened 2021-02-26
Respirator/Gas Mask attachments update

EmilyR42:
- Added multicoloring to respirator mask
- Adjusted respirator mask's priority
- Removed gagging on respirator and gas mask tube A by popular demand


Ben987:
Sure!  Thanks for that upgrade :)

PR #2157, opened 2021-02-26
Revert: "Remove Extra Console Messages"

jomshir98:
This reverts commit 9f856b9bf5f742a6979cf1efe7dd44c0bd2e92c2.

Reasoning:
First of all the console messages don't have any negative effect at all.
Quite the contrary, they are very helpful whenever someone sends console screenshot:
- The "Loading BondageClub version: ..." message immediately tells us or the player if they are on latest beta/release
- "Server connection established" is helpful to see where connection to server was restored, useful when working around problems connected with connections
- "WebGL Drawing enabled: GLVersion " is REALLY important when working out any drawing problems, because it tells if the browser is using legacy drawing method or which GL version it uses, making it way easier to replicate problems

Also there is a reason opening changelog didn't exit the maid:
In my experience changing screen at the same time new window opens is bad for user, as it can easily make them disoriented, when they don't see the change happen

### DIFF
--- a/BondageClub/Screens/Room/MainHall/MainHall.js
+++ b/BondageClub/Screens/Room/MainHall/MainHall.js
@@ -432,7 +432,6 @@ function MainHallClick() {
  */
 function MainHallOpenChangelog() {
 	window.open("./changelog.html", "_blank");
-	DialogLeave();
 }
 
 /**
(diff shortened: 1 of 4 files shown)


Ben987:
We can always reactivate these messages when we code, but there's no need to have them all the time.

PR #2158, opened 2021-02-26
Feature: Send messages with beeps

jomshir98:
**Depends on: [Server#69](https://github.com/Ben987/Bondage-Club-Server/pull/69)**

Adds ability to send messages along with beeps.
If no message is written, works exactly like old beep
Character limit for beep is 1000 (Like chat message)

Preview:


Ben987:
This one will be very popular.  Will there be problems for players using the regular version beeping players using the beta version and vice-versa?  Thanks!

jomshir98:
> This one will be very popular. Will there be problems for players using the regular version beeping players using the beta version and vice-versa? Thanks!

No:
Normal->Beta: Beep without message works exactly the same
Beta->Normal: Message will be ignored, looking like normal beep (Maybe worth mentioning in release notes, that during beta players can't expect, that other party will see the message, because they might not be on beta)

Ben987:
Awesome!  That should be fine then.  Yep, I can put a little note on that in the beta release.  Thanks :)

PR #2159, opened 2021-02-27
Camisole

EmilyR42:
- Added new camisole item


Ben987:
Absolutely!  Thanks a lot, you could also make it free later if you want everyone to have it by default.  Cheers :)

PR #2160, opened 2021-02-28
Fix: Game version

jomshir98:
`R66Alpha` is not valid version.

The version **needs** to be in format `R[0-9]+(Beta[0-9]+)?`, otherwise code cannot tell which version is newer.
If alpha version is really desired, `R66Beta0` can also be used

### DIFF
--- a/BondageClub/Scripts/Common.js
+++ b/BondageClub/Scripts/Common.js
@@ -12,7 +12,7 @@ var CutsceneStage = 0;
 
 var CommonPhotoMode = false;
 var GameVersion = "R0";
-const GameVersionFormat = /^R([0-9]+)(?:Beta([0-9]+))?$/;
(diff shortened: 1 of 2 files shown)


Ben987:
Could you add "Alpha" to your format?  So we can identify the pre-beta versions.  Thanks!

jomshir98:
@Ben987 Done, the new rules for version comparison are:

Regex: `R([0-9]+)(?:(Alpha|Beta)([0-9]+)?)?`
Than the versions will be compared lexicographically:
First the group 1 - number after R
Than group 2 - Alpha/Beta/[nothing]; in this order
Than number after Alpha/Beta, defaulting to 0

Ben987:
Wonderful!  Thanks for adding the alpha versions, cheers :)

PR #2161, opened 2021-03-01
Asset description line break fix

Nina-1474:
Correcting the buiding of the asset description list, since it included line breaks resulting in chat messages like this:
> (Player uses a ball gag
 on her mouth
.)


Ben987:
Yes!  Thanks for the fix :)

PR #2163, opened 2021-03-02
VR headset

Ada18980:
Adds a VR headset into the game with options to either not blind, display a black screen, or show a custom background. Also adds the new custom background, and features to allow any blindfold or hood to have this kind of effect.


Ada18980:
The asset check fails because I added a new asset type. Is this expected, @jomshir98 ?

Nina-1474:
You need to add the new asset property into \Tools\Node\AssetCheck_Types.js.

Ada18980:
> You need to add the new asset property into \Tools\Node\AssetCheck_Types.js.

Thanks!

Ben987:
Sure!  That's a fun idea.  Too bad we have to hardcode the room names in Drawing.js but we can always find a better solution later.  Thanks!

PR #2164, opened 2021-03-02
Fix lockpicking impossible for locks placed in scripted rooms

Ada18980:
Some people have been making scripted rooms with very high bondage levels on applied stuff. This is making players stuck, so I am adding a cap for the difficulty in lockpicking so it can't be set to something ridiculous like 100.


Ben987:

PR #2165, opened 2021-03-02
Skips struggle minigame choice if the item will be easy to remove

Ada18980:
If the challenge level is 1 or 0 then it will be very easy to remove, and so the game goes directly to the brute force minigame.

Thanks @Nina-1474 for the suggestion!


Ben987:
Yes, the choice isn't necessary in that case.  Thanks!

PR #2166, opened 2021-03-02
Fix dark BG when talking to characters

Ada18980:
Fixed an issue with a variable not being initialized properly


Ben987:
Thanks for the fix!

PR #2167, opened 2021-03-02
Removes the struggle options when not restrained and the item is unlock

Ada18980:
This should solve issues of things like handheld toys bringing up the minigame choices.


Ben987:
Great!  Thanks for the extra fix.  Cheers :)

PR #2168, opened 2021-03-03
Audio Volume Fix

Nina-1474:
Certain audio playing cases (futuristic items and maid bell) were mistakenly treating a 0% volume setting as missing and ironically playing the sound at full volume instead. The function has been tweaked to fix such cases.
This also includes a fix for the futuristic chastity belt's audio trigger which wasn't checking whether the player was in a chat room properly.


Ben987:
Yes!  Thanks for the fix, I've never notice that one.

PR #2172, opened 2021-03-03
Stop the player from being slowed by skirts

Ada18980:
This prevents players from being slowed for wearing a skirt unless legs are bound or covered


Ben987:
Are you sure?  It's kind of fun that tight skirts slow you down.  And the slow effect is very short in-game.

Ada18980:
> Are you sure? It's kind of fun that tight skirts slow you down. And the slow effect is very short in-game.

Well, the thing is these are fashion items, not restraints, and players complained about having to switch to roleplay mode so that they don't get slowed down by just wearing a skirt. Maybe Ill run a poll.

Ruilov3:
As the one who made the pencil skirt - its intended behaviour is to not slow or impair player movement. 
When I was making the item there were certain pose limitations when it comes to slowdowns and kneeling, that we recently got the tech to circumvent.
If it was a long hobble skirt going down all the way to the ankles - yes, it'd be appropriate to slow, but this one at this length at best limits shaking your butt too fast

Thumbs up from me on the PR fixing the item to its intended behaviour

Ben987:
Alright!  Thanks you two.  Yes, we could apply the slow effect for some clothes later, I like it.  Cheers :)

PR #2175, opened 2021-03-03
Restriction setting to bypass NPC punishment

Ada18980:
On roleplay mode, players can now elect to bypass punishment for the following:
-Being prevented from wearing clothes after the maid punishes you (player still gets stripped)
-Being prevented from wearing clothes after losing mistress status
-Being punished for calling the maids to leave the room


Ben987:
This will only affect the extra maid code you've added or it can also affect something else from the base club?  It should not affect anything from the single-player side.

Ada18980:
> This will only affect the extra maid code you've added or it can also affect something else from the base club? It should not affect anything from the single-player side.

It affects these two cases:
Being prevented from wearing clothes after the maid punishes you (player still gets stripped)
Being prevented from wearing clothes after losing mistress status

In both cases it prevents the LogAdd entry that forces the player to be naked for an hour. It does not disable the rest of the punishment.

Ben987:
Can you only add a dialog option to skip the punishment sequence for when you call the maid in online chat rooms when playing on "Rolepay/East" ?  The difficulty level is only intended for online play, so if you simply change the dialog you've added it won't impact the single player experience.  Thanks!

Ada18980:
@Ben987 Done.

Ben987:
Awesome!  That was quick.  Thanks a lot :)

PR #2176, opened 2021-03-03
Password padlock tab order

Ada18980:
Fixes the tab order in password padlocks so that hint is before the password


Ben987:
Sure!  Thanks for the fix :)

PR #2177, opened 2021-03-03
Eyepatch/Metal Leg Spreader update

EmilyR42:
- Added option to change eye patch position
- Added option to change metal leg spreader length


Ben987:
Sure!  This is a good idea.  Thanks :)

PR #2178, opened 2021-03-04
Lock picking arousal fixes

Ada18980:
2 fixes
1) Fixes the pin being unset when arousal ticks being a real set pin
2) Fixes arousal-related effects during lock picking not being disabled when the meter is disabled


Ben987:
Nice!  Thanks for the fix :)

PR #2179, opened 2021-03-05
Bugfix: Fixes leather armbinder lock positioning issues

elliesec:
## Summary

This fixes an issue where the leather armbinder's locks are positioned incorrectly under certain circumstances. The issue was happening because the two strap layers had different image top/left positions to the main binder layer, meaning their corresponding lock layers would be drawn correctly when the layers were visible (since the lock would be drawn at the top/left coordinates of the last drawn layer), but would be drawn too low and far to the right if the layers were hidden (through the `HideAs` property) - see the image below.

I've accordingly removed the additional lock images and configured the item to draw the same lock layer for all types. I've also removed the `top`/`left` properties from the two strap layers, and trimmed out the unnecessary whitespace so that all layers are drawn at the same top/left coordinates.

## Steps to reproduce

1. Equip the leather armbinder, and lock it
2. Set the armbinder's type to either the ring straps or wrapping straps
3. Change the player's clothes to an item that hides the straps (e.g. the Admiral Top)
4. Observe that the locks are now drawn in the wrong position

image


Ben987:

PR #2182, opened 2021-03-07
Notification Popups and Rework

Nina-1474:
This is a fairly major rewrite and enhancement to the notifications feature from #1904 and #2049.

New Event Settings
---
- The notification when a character joins a room can now be filtered/enabled for characters owned by the player
- A notification can occur when the game disconnects from the server
- A notification can occur when the player's turn starts in a LARP game and they're away from the game.
- The preferences screen now has a button to generate a test notification

Popups
---
Individual event settings can now be toggled in the preferences screen to create a popup window at the bottom-right of the screen, instead of incrementing a number in the document title as it currently does.
The popup that appears will indicate the event that occurred, and can also display extra event-specific information:
- Beep / Chat Join / Chat Message: The name of character that prompted the event will appear in the title
- Beep: The room name can appear in the title and the beep message can appear as the main content
- Chat Message: The character's face can appear as the icon

Clicking on the popup will refocus to the browser and reset the notifications.

Popup Availability
---
The exact behaviour/features/permissions for popups varies by browser, but the code should just skip features wherever they aren't supported. Some browsers such as iOS Safari won't support it at all, but most browsers will to some degree, although often with certain prerequisites such as the user giving permissions, using a secure https link, not running in private. The preference screen's test button should help out in these matters.

Audio Settings
---
Individual settings can now have separate audio settings. Each can also have one of two different types of audio alerts:
1. A sound only plays the first time the particular event occurs. Further occurrences of that event will not prompt a sound, until the player returns to the game and resets it.
2. A sound will play every time the event occurs.

If the event is configured to use popup notifications and the browser's popups support sound, we'll activate the popup sound instead of using the in-game beep sound.

Preferences
---
The preferences screen has been mostly rewritten from scratch to support the alert and audio settings now being configurable separately per event. If the current browser does not support popups, this option won't appear in the buttons and settings currently set to it will be disabled. 
If the notifications setting in the audio preferences page is disabled, all audio settings in the notifications page will be disabled and conversely if any are enabled in the notifications page, the audio page's setting will be enabled. 
There's a Test button at the bottom of the page which the user can play around with to find out how the alert and audio settings work and which ones will be available for them.

Testing
---
I've tested this across several browsers, but can't test it on mobile. Any new functionality that isn't supported is just skipped and should be for mobile as well. In the worst case scenario the popup just won't appear. The existing functionality hasn't been affected.


Ben987:
This is huge!  Thanks a lot, right on time for the next beta :)

PR #2183, opened 2021-03-07
Fix having struggle options for things like the bed

Ada18980:
Anytime when DialogProgressAuto >= 0, it will use the strength minigame to save time from picking a struggle option.


Ben987:
Sure!  Thanks for the fix :)

PR #2184, opened 2021-03-08
Improvement/Cleanup: Remove +Dark images

jomshir98:
*This is continuation of #1528, done with author's approval*
**Co-Authored-By: Sekkmer**

Removes all Dark background images.
Currently these were only used in SP rooms and on chat room administration background preview.

I did fair amount of testing and load-testing and can't find any issue or slowdown caused by this.


Ben987:
Have you tested it with slower computers?   We added the Dark versions back then because some users with slower computer were having lots of lag when darkening the backgrounds.  It greatly improved performance for them.  Thanks!

jomshir98:
Yes, I tested it with throttling to simulate slow computers and as expected absolutely no change.

That is because this **doesn't** change how dark backgrounds work; That was done all the way back in #625

Ben987:
Time to test this one!  Lets hope nobody will get slowed down.  Thanks a lot you two, it should make the game a little lighter.  Cheers :)

PR #2185, opened 2021-03-08
Character in Preview Images

Nina-1474:
A clothing asset group can now have a `PreviewZone` setting which if defined will change how the selection screen's preview images appear. Each preview box will contain a copy of the character, focused in on the defined zone co-ordinates, wearing that item.
This has been enabled for HairFront and HairBack (which currently only has the previous/next button).

image


Ben987:
Very nice!  I love it.  Thanks a lot :)

PR #2186, opened 2021-03-08
Improvement/Refactor: Appearance update validation rework

elliesec:
## Summary

I'm afraid this one's pretty huge, but I've spent a _lot_ of time testing it, and am about as confident as I can be without automated tests that it now all works correctly.

This PR reworks the validation logic for item appearance updates. There are several motivations for doing this:
* Improved consistency between validation in the `ServerAppearanceLoadFromBundle` and `ChatRoomSyncItem`, which currently both implement a lot of their own logic around item updates
* Better validation and locking down of item properties, to prevent people from using the console to set bad properties on items (e.g. setting an item's `Effect` to a non-array value)
* Closing off some loopholes and errors in the current appearance validation code
* Adding validation for the more recent appearance rules (around cosplay items, and allowing players to modify each others' body parts)
* Making it easier to track down inconsistencies in item updates (e.g. locks disappearing, which we've seen a lot of recently)
* Splitting appearance update validation logic out into item addition, removal, modification and swap logic, to hopefully make it easier to understand

## Technical Details

This change almost completely overhauls the validation of character appearance updates on the client, and in most cases, makes it stricter and more robust. 

Validation of appearance updates is now handled asset group by asset group, diffing the previous and new items in each group, and validating the diff as the composition of 3 atomic operations:
- (A) Item addition (with no item properties)
- (R) Item removal
- (M) Item modification (modification of item color, difficulty or properties)

Diffs are then treated as either the above atomic operations, or the following composed operations (referring to the keys in brackets above):
- Item addition (with properties) - A + M
- Item swapping - R + A + M

I've probably forgotten a few things, but the main changes are as follow:

* Splits appearance update validation out into its own `Validation.js` file
* Modifies `Server.js` and `ChatRoom.js` so that both `ServerAppearanceLoadFromBundle` and `ChatRoomSyncItem` use the same appearance validation logic. This means that in the future, we can add/modify validation in one central location and have it apply to all appearance updates from other players
* Adds stricter validation for the `Effect`, `Hide` and `Block` properties on items, ensuring that they are arrays (or null/undefined), and that all values inside the arrays are strings
* Enables most validation checks to be run by _all_ players in a chatroom, meaning if someone sends a bad update, only the person sending the update should see the bad changes in most cases
* Adds improved defence against an issue where players could add owner/lover locks or other invalid appearance items to other players if the other player disconnected after an invalid appearance update due to the bad update being persisted in the database
* Ensures that appearance update validation now properly respects the player's rules around cosplay items, body modification, and any lover rules they may have (i.e. whether or not owners can use lover locks)
* Adds warning messages when an invalid appearance update is received, which should make it easier to track down any remaining issues with things like disappearing locks, etc.

## Testing

As this is such a big change, and needs to account for any possible appearance update, it's difficult to test every possible combination update, but I've spent a good couple of weeks testing and refining the logic, and believe it should cover most eventualities.

As there are so many possibilities, I've compiled the following [spreadsheet](https://docs.google.com/spreadsheets/d/1l9nyXFuWC83S6Cz_rW71Rvfo0WIVFlnbhIeugOOKnXM/edit?usp=sharing) that I've been using for testing purposes, which should outline the desired behaviour with regards to item addition, removal and modification (see the three spreadsheet tabs). I've validated the current code against (I think) all cases on that spreadsheet for testing. I anticipate that there may still be some small bugs, but hopefully the beta would help us to iron those out.


Ben987:
This one is huge!  Thanks a lot.  Lets wait until the next version though, I want to validate socket.io V3 on all clients first.

Ben987:
Time to test this big one!  There are so many validations that I won't be able to try all of them, that's why we have betas :)

PR #2187, opened 2021-03-09
Character in Previews - Performance Improvement

Nina-1474:
A followup change for #2185: the preview character canvases will only be rebuilt when necessary. For example they won't be when selecting one of the items, since that just shuffles the inventory order.


Ben987:
Absolutely!  It was a little slow on my side.  Thanks for the fix :)

PR #2188, opened 2021-03-09
larp typo fix

4i0:
Matron will no longer detrain her target. She will detain her instead as she should.


Ben987:
Sure!  Thanks for the typo fix :)

PR #2190, opened 2021-03-09
Online Dialog overflow fix

Nina-1474:
In the last release, the "(Security actions.)" and "(Have a photo taken.)" dialog options for other online players were both added. Both have the potential to push the displayed options over the limit of 8 and make the "(Leave this menu.)" option disappear.
Now both have been moved inside the "(Character actions.)" option's page.


Ben987:

PR #2191, opened 2021-03-09
Wardrobe Name Persistence

Nina-1474:
When the player loads a wardrobe slot, its name is now loaded into the name input text box. And when the player leaves the wardrobe screen using the Accept button, the text in the box is remembered and reloaded on returning to this screen (unless the appearance screens are exited altogether).
This makes it much easier to load a wardrobe slot and make changes to it, since the name no longer needs to be re-typed when saving back down.


Ben987:
Yes!  That will be useful.  Thanks :)

PR #2193, opened 2021-03-09
Chat Room VIP List

TessaTech:
Depends on Server PR: []

Adds a VIP list to the chat room data.
VIP players of a chat room are able to find/enter/leave locked chat rooms even when they are not admin themselves.

VIP players are indicated by a new symbol where usually the admin symbol is. Since admins have all the benefits VIPs have this symbol only is shown if a player is VIP but no admin.

The VIP list can be changed by room admins in the ChatAdmin screen. I also added buttons to quickly set all friends or whitelisted players as vips.

Screenshot:
Screenshot VIPs


Ben987:
Sorry, we'll need to put that one on hold until the server bugs are fixed.  We will try to optimize the server before adding new stuff.  Thanks!

TessaTech:
Makes sense. Is there a list of known bugs? I am in work of some server optimisations myself right now to cut chat room syncs into smaller chunks that only the required data. If there are bugs in that section i could take care of them as well in that process.

Ben987:
Cutting the chat room sync packets would certainly help.  The biggest gain would probably be to have a transaction that updates the full appearance of a character without syncing anything else.

Ada18980:
@Ben987 Are we going to start merging server changes or what o.o

PR #2194, opened 2021-03-09
Metal Cuffs in front of the body

TessaTech:
A rather small change i made to get a hang of working with items.
This turns Metal Cuffs into an extended item and adds the "In Front" Option as default. The old look of the item now is the "Behind Back" option.

Screenshot:
Screenshot - Cuffs in front


Ben987:
Thanks!  I will slide it in the beta version.  Cheers :)

PR #2195, opened 2021-03-09
Character in Previews - More Groups

Nina-1474:
This adds the character-in-preview-images setting from #2185 to many more clothes groups - most of them in fact!
There is now also a 'Character' button in the main appearance screen which toggles whether 3x3 grid screens will use the current static preview images or previews containing the character. The latter can still cause slowdown in some browsers so this button will be off by default.


Ben987:
Yes, since some players are freezing when using that feature, it's better to have it off by default until we can get better speed.  There's a conflict with another PR, can you resolve it and test again?  Thanks :)

Ben987:

PR #2196, opened 2021-03-10
Asset: Blanket for Pet Bed

Codefreak64:
Recently it was brought up that you can put on the Bed item, then swap over to the Pet Bed to preserve the blanket. People seemed to like that idea so with this PR it makes that concept a reality, by adding a blanket in via the selection boxes.

-Adds extended item to PetBed
-Minor fix to the PedBed as well (not fully hiding items with the alpha mask)
-Added multicoloring to both the PedBed and the Extension


Ben987:
Sure!  That's a fun idea, thanks :)

PR #2197, opened 2021-03-10
Sensory Deprivation Message Hiding fix

Nina-1474:
This rearranges the chat-hiding sensory deprivation change from #2042 so that it happens earlier and skips notifications for messages that won't appear, along with any other message-prompted steps we may add in future.


Ben987:
Absolutely!  Thanks a lot :)

PR #2198, opened 2021-03-10
Enclosing Item Device Priority fix

Nina-1474:
For enclosing box-like ItemDevices which cover the character's head there are two items that incorrectly appear outside/in front of them: the wooden sign (priority changed in #1926 in order to be above the ItemDevices display frame) and the snorkel's tube.
These ItemDevices have now been moved from the group's priority 56 up to priority 58. Their display ordering against all other items around that range will not be affected.

image


Ben987:
Yes!  Thanks for the fix :)

PR #2199, opened 2021-03-10
Pole Message Fix

Nina-1474:
The text of the two action chat messages for selecting a pole option were set the wrong way around.


Ben987:
Thanks for that quick fix :)

PR #2200, opened 2021-03-10
Metal Cuffs Behind Back option difficulty fix

TessaTech:
I accidentally added difficulty to to the behind back pose of the metal cuffs which causes them to be much harder to escape now. 

This PR fixes this by removing the additional difficulty completely.

Edit:
Also fixes the missing lock attribute in both variations


Ben987:
Absolutely!  Thanks :)

PR #2201, opened 2021-03-10
Chatroom Subscreen Kicked Fix

Nina-1474:
If a player is in a chatroom subscreen and then is forced out of the room for any reason (kicked, leashed, etc) then various bugs can occur depending on the screen. 
I've corrected a few that I found. Namely: missing elements when re-entering the preferences or appearance screens if kicked when viewing them, and the text getting stuck aligned left instead of center if kicked from certain preference screens.


Ben987:
Absolutely!  Thanks for that one, I've seen it before.

PR #2202, opened 2021-03-10
Block Arousal Activities in Room

Nina-1474:
Adding a new Block category for admins to set on their rooms which disables arousal activities and actions in that room.
Also tweaked some Block settings text a little to reflect that it's not only for blocking items now.


Ben987:
Sure!  Your room your rules.  Pure roleplayers will like that one.

PR #2203, opened 2021-03-11
Increase gag stimulation messages frequency

Ada18980:
Increases chance from 10% to 30%, since gagged people dont have much to say anyway.


Ben987:
Alright!  Thanks :)

PR #2204, opened 2021-03-11
Bugfix: Fixes incorrect synthwave background name casing

elliesec:
The casing of the background name on the VR headset didn't match the casing of the checked in background file, causing 404s.


Ben987:
Thanks for the case fix :)

PR #2205, opened 2021-03-11
Fix notifications happening in sensory deprivation

Ada18980:
Adds a check to prevent sending notifications if the player is in sensory deprivation


Nina-1474:
This has already been fixed by #2197.

Ben987:
Duplicated fix, I will close this one.

PR #2206, opened 2021-03-11
Increases the difficulty of the minigame to get up if you are kneeling

Ada18980:
Seems like it should be harder to get up than to get down


Ben987:
It makes perfect sense :)

PR #2208, opened 2021-03-11
Fixes futuristic collar not coloring the tag properly

Ada18980:


Ben987:
Thanks for the fix :)

PR #2209, opened 2021-03-11
Deluxe Mistress Boots alpha groups

Nina-1474:
Adding the Groups settings from MistressBoots into DeluxeBoots so that they don't cut rectangular chunks out of other items such as beds.

image


Ben987:
Absolutely!  Thanks a lot :)

PR #2210, opened 2021-03-11
Added cyrilic for detect words to futuristic gag.

DoberBit:
Because "futuristic gag" fully ignoring saying with cyrilic


Ben987:

PR #2211, opened 2021-03-11
Fix: Changelog on account creation

jomshir98:
Do not display changelog maid for new players


Ben987:
Yes!  Thanks for the fix :)

PR #2212, opened 2021-03-11
Refactor: Migrates the leather/ornate/futuristic leg cuffs to the extended item script

elliesec:
## Summary

This migrates the leather, ornate and futuristic leg cuffs over to use the extended item script, so that they now use types rather than the old `Restrain` property. All three items have been tested and I've verified that they work as before (including the password/validation step of the futuristic cuffs).


Ben987:

PR #2213, opened 2021-03-11
Bugfix: Incorrect variable reference in collar auto shock unit

elliesec:
## Summary

This fixes an incorrect reference in the collar auto shock unit where it was referencing `GagTriggerPunish` rather than `ShockTriggerPunish`, causing errors as below:

image


Ben987:
Yes!  Thanks for the fix :)

PR #2214, opened 2021-03-11
Notification Fixes

Nina-1474:
Some notifications were resetting and disappearing too quickly, before the user even refocused back on the game. 
This is because I assumed those functions that called the reset step only ran while the screen was active, but this is not true for all browsers.

Notifications broke after creating a new character rather than logging in. They will now be loaded properly. Also added safety checks to the raise/reset.


Ben987:
Alright!  Thanks for the fixes :)

PR #2215, opened 2021-03-11
Respirator gag fix

EmilyR42:
- Fixed respirator mask still gagging when applied in a certain slot


Ben987:
Sure!  Thanks for the fix :)

PR #2216, opened 2021-03-11
Refactor: Migrates the Butt Plug Lock to the extended item script

elliesec:
## Summary

Refactors the Butt Plug Lock to use the extended item script, and converts the previous custom validation function for the short variation into prerequisite checks on the type options.


Ben987:
Absolutely!  Thanks :)

PR #2217, opened 2021-03-11
Futuristic collar permissions fix

Ada18980:
Fixes the remote blocking being changeable on the futuristic collar when locked out, even if the open permissions setting is blocked.

Also adds a text message to make it clear to the user that they can't change the setting because the collar is locked.


Ben987:
Nice!  It should be clearer.  Thanks :)

PR #2218, opened 2021-03-12
Fixes crash when applying prelocked extended item to unowned lover

TessaTech:
I encountered a rather weird crash with the handcuffs in front pose.
It turned out that this was a unhandled special case when adding an extended item with default "Lock" effect to an unowned lover which never occured until now since all default locked extended items until now were owner only.
However, this PR fixes this.


Ben987:
Very well!  Thanks for the fix :)

PR #2219, opened 2021-03-12
Fix: Prevent sending friendlist autorefresh multiple times

jomshir98:


Ben987:
Lets do a button to refresh that list in a future version.  Lets try it without the auto-refresh for now.

PR #2220, opened 2021-03-12
Spread vs Hogtie fix

Nina-1474:
Rearranging AllowPose for socks and shoes so that when a player has both the Hogtied and Spread poses applied (now possible due to the SpreaderMetal Wide option), the Hogtied image is used.


Ben987:
Yes!  Thanks for that fix :)

PR #2223, opened 2021-03-13
Leg items now have the slow effect

Ada18980:
Simplifies the IsSlow function so that items have to have the slow effect in order to slow.

This should eliminate unexpected cases of items slowing, and also catch all leg items except those which have


Ben987:
Yes!  Thanks for the fix :)

PR #2224, opened 2021-03-14
Chatroom Background Drawing Refactors

Nina-1474:
This refactors the steps that draw the chatroom background into a separate function so it can be reused for the top and bottom rows, to help prevent mistakes like the recent case of the custom background being missing for the bottom row.

This also enhances the blindless level determination so that it can check for eye openness only. This fixes the issue of the VR headset's custom background always appearing, even when the character closes their eyes.


jomshir98:
This will cause conflicts with #2184

Nina-1474:
It will. If this gets in first, I can resolve the conflicts in that PR if you'd prefer.

Ben987:
Sure!  Thanks for the fix, it should prevent the bug we had with the SynthWave background.

PR #2225, opened 2021-03-14
Suspension Kneel Fix

Nina-1474:
When suspended by the feet, the kneel button is available which can result in kneeling in mid-air.
This adds empty `AllowActivePose` to the options so that they override the assets' `AllowActivePose: ["Kneel"]` and prevents kneeling.


Ben987:
Yes!  Thanks for that one :)

PR #2226, opened 2021-03-14
Chess Progress Change

Nina-1474:
The process to calculate the total major and minor values on both sides (`CollegeChessGameProgress`) was running every frame, but it miscalculated and undervalued black's value at some point during their move. This means when their move ended and the value went back up, it thought white lost pieces and punished the player even though no pieces had been taken.
This PR changes this calculation to only occur after a move has completed via an event instead. This avoids unnecessary recalculations and seems to have corrected the miscalculations.


Ben987:
I've never seen that bug happen, but thanks for the fix!

PR #2227, opened 2021-03-14
Improvement/Fix: Moves chatroom rejoin functionality into ChatroomSync

elliesec:
## Summary

This moves the chatroom rejoin functionality, as well as updating of the last chatroom on the player's account, into the `ChatRoomSync` function instead of the `ChatRoomRun` function, so that the logic is not being run every frame, but only when there's a chatroom update. It also fixes an issue where an `AccountUpdate` message was being sent unnecessarily every time someone entered/left a chatroom due to a bad equality check (`!=` was being used to compare the admin arrays, which would always evaluate to `true`, resulting in a needless `AccountUpdate`).

This should result in a minor performance improvement in chatrooms, and also prevent the unneeded messages being sent to the server in this particular case.

~~Note that this will probably conflict with #2186 in `Common.js` as I've pulled the same `CommonDeepEqual` function that I used in there into this branch in order to compare the arrays properly.~~

## Bug: Steps to reproduce

1. Enter a chatroom (it's easiest to see the unnecessary message by viewing the messages being sent down the websocket)
2. Note that an `AccountUpdate` message is sent containing the chatroom data (i.e. name, description, admin list, etc.) - this is expected
3. Have another character enter the same chatroom
4. Note that an `AccountUpdate` message is sent containing exactly the same data as that sent in step 2 - this message is unnecessary


TessaTech:
This will create a conflict with #2193.
I suggest merging this one first since it's a structural change. I'll resolve #2193 once this is merged.

Ben987:
Yes, yes, yes!  Thanks for that one, it makes perfect sense.  Cheers :)

PR #2228, opened 2021-03-14
Fix: InventoryItemHasEffect

jomshir98:
Improves `InventoryItemHasEffect` to properly detect empty effects and makes it check properties by default (because that is 99% wanted use case)


Ben987:
Sure!  Thanks for that fix :)

PR #2230, opened 2021-03-15
Chess Simultaneous Move Fix

Nina-1474:
It's possible to move your own chess pieces during the opponent's turn, which freezes the game.


Ben987:
Thanks!  I've never seen that one also but it's probably because the computer plays too quickly.  Cheers :)

PR #2231, opened 2021-03-15
Chess - Play as Black

Nina-1474:
In the chess minigame there's now a 50% chance that the player will play as the black pieces instead of the white pieces.


Ben987:
That's a fun idea!  Thanks :)

PR #2232, opened 2021-03-15
Bugfix: Fixes feet items clipping with the long skirt

elliesec:
## Summary

Because of the length of the long skirt, items in the `ItemFeet` slot would clip over it. Unfortunately, as there are no gaps in the item priorities, it's currently impossible to get the skirt to cover feet items whilst still displaying underneath the hands. I've therefore split the item into two layers, one behind the hands and one above them, which resolves the visual issues.


Ben987:

PR #2233, opened 2021-03-15
Fix/Improvement: Add friend list refresh button

jomshir98:
Adds a refresh button to manually refresh friend list, as ~50% of people said that without showing that the list needs to be refreshed it would cause confusion


Ben987:
Sure!  A refresh button is fine.  Thanks!

PR #2234, opened 2021-03-16
Background Darkening Fix

Nina-1474:
Correction to the background darkening which was causing the top row of a two-row chatroom to be darkened twice instead of once.


Ben987:
Yes!  Thanks for the fix :)

PR #2235, opened 2021-03-16
Pig Nose Hook

Codefreak64:
adds another nose hook item, but this time its 3 straps and not 1, this new pignose hook might be repetitive, but at least its another for the ItemNose slot

-adds Pig Nose Hook when you buy the nose hook


Ben987:

PR #2236, opened 2021-03-16
Bugfix: Collar cuffs now allow neck activities

elliesec:
Not strictly a beta issue, but it's been noted that the collar cuffs block activities on the neck (which I think has been the case for several months at least). This adds an `AllowActivityOn` entry to the item to allow neck activities.


Ben987:
Very good, thanks for that little fix :)

PR #2237, opened 2021-03-16
Futuristic Collar does not override light color

Ada18980:


Ben987:

PR #2239, opened 2021-03-17
Fix: Possible crashes when text items have bad properties

jomshir98:
Items tested to work and no longer crash


Ben987:

PR #2240, opened 2021-03-18
Sci-Fi Pleasure Panties

Shiranui-Izayoi:
Assets Made by Ada18980 and used with permission.

The item is meant to be a futuristic/sci fi like belt that looks like panties, but works as a chastity device. It also is capable of both pleasuring and shocking the wearer at the same time. While being able to open or close access to the characters crotch or ass.

Also added in a unique Denial Mode that when turned on completely prevents the character from cumming regardless. This works similar to Edged except that it doesn't require all items to have this effect. If the effect is anywhere on the character then orgasms are prevented.

No advanced modes were added due in part to a lack of space. But considering that Denial Mode effectively replaces Edge and Deny advanced modes I don't think this is an issue.

I don't know how to tie it into Stimulation Messages yet so I haven't done so currently, but may do so in the future. It is also fine for anyone to tie them in for me if they so wish.

This has been tested and in what I could think to do it works fine with no issue. It is part of the "FuturisticChastityBelt" buy group so will be added automatically to anyone that owns that.

If anymore infomation is required or anything wants to be asked feel free to do so.


Ben987:
Sure!  That's a fun concept, I'm curious to try it.  Thanks :)

Shiranui-Izayoi:
> 
> 
> Sure! That's a fun concept, I'm curious to try it. Thanks :)
Glad you like it and I hope it lives up to your expectations. XD

PR #2241, opened 2021-03-18
Extended clothes on others fix

Nina-1474:
Fix to prevent crashing on changing the option of extended clothes on other characters.


Ben987:
Yes!  I will prepare a hotfix.  Thanks a lot :)

PR #2242, opened 2021-03-18
Incorrect Clothes Pose fix

Nina-1474:
Removing unneeded allow pose settings for the Teacher and Admiral outfits as these are already covered in the group's allow pose setting. This prevents an error where a character could have both OverTheHead and BackBoxTie poses from Ceiling Chains and Pole respectively but was incorrectly using the BackBoxTie clothes image since the asset settings take preference.

Also removing Horse and KneelingSpread from the Admiral top as there are no separate images for those poses and there don't need to be.


Ben987:
Absolutely!  I will hotfix it, thanks again :)

PR #2243, opened 2021-03-18
Bugfix: Corrects extended item arguments for the eyepatch

elliesec:
## Summary

This is somewhat related to #2241 (and also fixes the same issue for the eyepatch, although it makes sense to include both, as they fix slightly different things) - this adds the `IsCloth` argument to the extended item function calls in the eyepatch's extended item script. This ensures that the options are drawn in the correct place, and prevents the game from sending an appearance update every time the player modifies the eyepatch whilst inside their wardrobe.


Ben987:
Thanks for that one!  I will hotfix the live version.

PR #2244, opened 2021-03-18
Fix: Bundle fix for extended items

jomshir98:
This PR fixes many small problems or inconsistencies with extended items.

Where the behaviour of item changed, authors were asked for intended one.


Ben987:
This is huge!  Thanks a lot.  I won't be able to test all of this, but that's why we have betas.   Cheers :)

PR #2245, opened 2021-03-19
Sell Items in the Shop

Nina-1474:
This allows the player to sell items they own back to the shop for half of the purchase price (rounded up).

When selling, the dialog option for each item category will only appear when the player actually has an item in it to sell. Certain items will not be sellable if they can be obtained in the club for free. E.g. nylon ropes from the introduction room.

The dialog text for the item categories has been simplified so it makes sense for both purchases and sales.
The general page structure of the shopkeeper's dialog has been rearranged a bit for ease of navigation. Before, the first page was a list of the top-level item categories plus the "I'd like to ask you some questions." line linking to a second page with all the other questions/activities. Now the first page will instead cover all the main activities (buy/sell/tie/job etc) with one option to start the buying process and one to start selling.


Ben987:
Sure!  That's a fun idea.  I'm not sure if your local sex shop would buy back your old butt plug but why not :)

PR #2246, opened 2021-03-20
HairFront17/17b Position Adjustment

ItsJustMeVerity:
Have changed the assets for HairFront17 and HairFront17b because they didn't quite line up as I'd intended when designing them. Rather than mess about with the code, I just tweaked the image to make the adjustment.

I apologise for any mistakes I may have made. Never used GitHub so I'm surprised I even got this far!
If I have done any part of this pull request process wrong, please do let me know. I'd like to do it right next time :stuck_out_tongue_closed_eyes:


Ben987:
Very good!  Thanks for that change, cheers :)

PR #2247, opened 2021-03-20
Metal Cuffs Default

Nina-1474:
The recent change to have Metal Cuffs lock the player's hands in front by default means that when the cuffs are applied randomly by an NPC, it's very easy to escape.
This has been changed to use the feature from #1995 so that the default null type is the previous behind-the-back option, so that NPCs will use this. But when a player is the one applying the cuffs, it defaults to hands-in-front instead.


Ben987:
Absolutely!  Thanks for that fix, I've seen the problem before :)

PR #2248, opened 2021-03-20
No Items Message

Nina-1474:
When selecting a zone that the player has no items in, it will now say "You have no items in this category" instead of the "You cannot access your items" message that appears when you don't have access.


Ben987:
Yes!  Thanks for that one :)

PR #2249, opened 2021-03-20
Appearance Esc Fix

Nina-1474:
Opening an appearance category in the appearance menu then exiting using the Escape key instead of the Accept/Cancel buttons was not setting the character's FocusGroup to null. This meant that when clicking on the character later, it immediately opened the inventory screen showing clothes options in a 3x4 grid instead of starting the expected dialog.


Ben987:
Sure!  Thanks for the fix :)

PR #2250, opened 2021-03-24
Bunny Girl Accessories Update

EmilyR42:
- Added multicoloring support
- Added two additional styles hiding the collar or cuffs


Ben987:
That's a lot of files!  Thanks a lot, I will try it soon :)

PR #2251, opened 2021-03-26
Bugfix: Removes duplicate properties from Futuristic Harness layers

elliesec:
## Summary

Just a little fix to remove some duplicate properties from two layers of the Futuristic Harness


Ben987:
Sure!  Thanks for the quick fix :)

PR #2252, opened 2021-03-26
Refactor: Removes handling of the unused Restrain property

elliesec:
## Summary

All items that were actually using the `Restrain` property have now been moved to use the extended item script. I've therefore removed handling of `Restrain` from the validation code, and have removed the property from the padded mittens and paw mittens, where it was not being used.


Ben987:
Absolutely!  Thanks :)

PR #2254, opened 2021-03-27
assets cleanup/improvement

fleisch11:
-removes a few grey pixels of the inflatable bodybag in the lower right (it's most visible in the photographic room)
-hides the upper body for the prison lockdown suit (that way the body isn't visible through the restraint)
-widens the legbinder and seamless legbinder a little bit (that way the legs aren't visible through the restraint)


Ben987:
Alright!  Thanks for the asset fixes :)

PR #2255, opened 2021-03-27
Bugfix: Fixes issues with new validation script when run on private room NPCs

elliesec:
## Summary

This PR fixes errors that occur in the new appearance validation script with private room NPCs. Once I'd fixed the issue, it turned out that there was a minor issue with the definition for the `Nipples` asset group which means that NPCs are actually never populated with nipple assets, due to the group having `Default: false` set. I've accordingly fixed that issue, and added some additional logic into `PrivateLoadCharacter` to update the appearance of any existing private room NPCs whose appearance is invalid. I also added some JSDoc to `Server.js` that I intended to add in the original PR, but forgot about.


Ben987:
Thanks for the fixes!  I will test it right now :)

PR #2257, opened 2021-03-28
Feature: Lock interactions - Combination Padlock

elliesec:
## Summary

This PR adds some new functionality to the combination padlock to make interactions with it more interesting when restrained or blindfolded. Only the combination padlock is affected by this PR, but I intend on adding similar lock interactions to other locks, one at a time. The idea is to allow players to interact with locks whilst blind, and also to simulate the effect that certain restraints might have on lock input (e.g. you shouldn't be able to perfectly manipulate a combination padlock whilst wearing mittens).

What this does:

* Adds a new `Character.GetClumsiness` function to characters, which returns a "clumsiness" rating representing their ability to interact with locks, ranging from 0 (completely unimpeded) to 5. Clumsiness is determined by whether or not the character can interact (is wearing a blocking item), and whether or not they are wearing hand restraints (i.e. mittens). Non-restraint hand items (e.g. spanking toys) do not affect clumsiness.
* Modifies `DialogMenuButtonBuild` so that combination padlocks can be inspected when blind
* Modifies the combination padlock so that the input is hidden when the player is blind, and a consistent random offset is added to the player's inputs
* Modifies the combination padlock so that numbers typed into the combination inputs are modified based on the player's clumsiness:
    * At clumsiness 0 (unrestrained), the lock behaves as before
    * As clumsiness increases, the actual number will be randomly offset from the what the player tried to input - the variation of this random offset will increase with player clumsiness
* Adds messages to the combination padlock that indicate to players why the inputs might not be behaving as expected

### Player is blind, but not restrained

Inputs will be hidden, and digits entered will be offset by a consistent random amount. For example, if the player types "1234", the input might be populated with "4567" or "8901", depending on the offset (but the player will not be able to see what was actually input).

### Player is clumsy, but not blind

The player will be able to see the input digits, but digit entry will be inaccurate (the level of inaccuracy will depend on player clumsiness). For example, if the player types "1234", the input might be populated with "0253" - each digit is "close" to the digit the player tried to enter. This means that if they know the code, they can eventually enter it by typing and deleting until they get lucky.

### Player is both blind _and_ clumsy

The two mechanics above are combined, making it very difficult to correctly input the padlock combination.

Below is an example of the new lock interface when the player is both blind and clumsy:

image


Ben987:
I like it!  Let's hope the players won't be too confused in-game when it's introduced.  Thanks!

PR #2259, opened 2021-03-29
Feature: Display login queue position

jomshir98:
Improvement for: [Server#78](https://github.com/Ben987/Bondage-Club-Server/pull/78)

Small change to display on which position the player is at the time the login was pressed. The position doesn't update and is meant mainly to notify players, that something is going on while they wait - that they are in queue.


Ben987:

PR #2260, opened 2021-03-29
NPC Missing Arousal Settings fix

Nina-1474:
This moves the preferences-initialisation step that constructs the Zone, Activity and Fetish lists in ArousalSettings from `PreferenceInitPlayer` which is only run for players, into `PreferenceInit` which is run for both players and NPCs.

### DIFF
--- a/BondageClub/Screens/Character/Preference/Preference.js
+++ b/BondageClub/Screens/Character/Preference/Preference.js
@@ -222,6 +222,49 @@ function PreferenceInit(C) {
 	if (!Array.isArray(C.ArousalSettings.Activity)) C.ArousalSettings.Activity = [];
 	if (!Array.isArray(C.ArousalSettings.Zone)) C.ArousalSettings.Zone = [];
 	if (!Array.isArray(C.ArousalSettings.Fetish)) C.ArousalSettings.Fetish = [];
+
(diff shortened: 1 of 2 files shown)


elliesec:
One thought on this - looking at the implementation of `NPCArousal`, it looks like this won't fix private room NPCs whose arousal is already broken and has been persisted, as they will have their arousal settings defined, but empty? Happy to be corrected on that though.

Nina-1474:
> One thought on this - looking at the implementation of `NPCArousal`, it looks like this won't fix private room NPCs whose arousal is already broken and has been persisted, as they will have their arousal settings defined, but empty? Happy to be corrected on that though.

That is correct. I was planning on seeing if there was a straightforward fix for that separately later. You're quite welcome to give it a go if you'd like.

Ben987:
@elliesec , do you want to create a different fix for NPCs that are currently broken?  Should I merge this PR?  Thanks you two.

Nina-1474:
Yes regardless on when/whether a fix for existing broken NPCs is made, this can still be merged to fix all future NPCs.

Nina-1474:
Actually I've now made and tested a quick fix to correct existing NPCs with broken arousal settings too.

Ben987:
Great!  Thanks Nina :)

PR #2263, opened 2021-03-30
Feature/Refactor: Extended Archetypes - typed items

elliesec:
## Summary

This PR is a follow-up to #2045, and extends the extended item archetype framework to include an additional archetype: typed items. These are items that currently use the extended item script. Extended items can now have their configurations registered in `Female3DCGExtended.js`, and their load/draw/click functions will be automatically generated, removing the need for the item to have its own script file, which should make it easier for people to make extended typed items.

In order to keep the scope small and test the waters, this change only affects three items: the latex sleeveless/boxtie/butterfly leotards, which I've migrated over to the new framework - they should behave exactly as they did before. However, eventually we should be able to move all items that currently use the extended item script over to the new archetype, which will allow us to remove a lot of script files, and will also make it possible for us to have stricter item validation. We could potentially in the future also change typed items so that only the new type needs to be sent in appearance updates and stored in the database, which would reduce message sizes and storage usage. In addition, this will make it easier to dynamically modify item types through the game code, which could have benefits when coding single-player content (i.e. it would make it easier for NPCs to dynamically use item types).

Main Changes:
* Adds a new `TypedItem.js` file with configuration functions for items with the `TYPED` archetype in `Female3DCGExtended.js`.
* Modifies `Asset.js` to handle automatic configuration of extended items with the `TYPED` archetype
* Adds a new `CommonChatTags` enum to `Common.js` - used to encapsulate common chat message dictionary tags (e.g. `SourceCharacter`, `DestinationCharacter`, etc.)
* Migrates the Latex Sleeveless/Boxtie/Butterfly Leotards to the new system and removes their script files
* Extends the asset check script to cover some of the extended item configuration

## Testing

The only items that are currently affected by this change are the latex sleeveless/boxtie/butterfly leotards, and I've fully tested them to ensure that they behave exactly as before. Providing there are no issues once the new code has been through player testing, I'll then start to add more features to the script and migrate other extended items to use it.


Ben987:
That's another huge one!  Thanks a lot, since it doesn't target many items it should be pretty safe.  Cheers :)

PR #2264, opened 2021-03-30
Bugfix: Validation typo fixes

elliesec:
## Summary

I noticed that #2255 introduced an issue with validation where asset validation would not get properly run on appearance updates due to a refactoring error. This also fixes a bug with the validation of lockpick seeds for locks that define the `LockPickSeed` property.


Ben987:
Sure!  Thanks for the fix :)

PR #2265, opened 2021-03-30
Feature: facial expressions can be saved/loaded

fleisch11:


BondageClub/Screens/Character/Login/Login.js
```diff
@@ -447,6 +447,17 @@ function LoginResponse(C) {
 			Player.OnlineSharedSettings = C.OnlineSharedSettings;
 			Player.GraphicsSettings = C.GraphicsSettings;
 			Player.NotificationSettings = C.NotificationSettings;
+			Player.SavedExpressions = C.SavedExpressions;
+			if (Player.SavedExpressions === undefined) {
+				Player.SavedExpressions = [];
+			}
```
TessaTech, line 453: It might be safer to use `Array.isArray` here to make sure other special cases or bug related illegal entries that might occure are taken care of as well

BondageClub/Screens/Character/Login/Login.js
```diff
@@ -447,6 +447,17 @@ function LoginResponse(C) {
...
+			if (Player.SavedExpressions.length < 5) {
+				var x = Player.SavedExpressions.length;
+				while (x < 5) {
+					Player.SavedExpressions.push(null);
+					x += 1;
+				}
```
TessaTech, line 459: If you're iterating over a known number of elements, a `for` loop would be a less error prone aproach. It is usually more readable and less error prone. If someone changes this section in the future, they will have a easier time with a ´for´ loop.

In the same go, i'd suggest replacing your ´var´ with ´let´. Because ´let´ is scoped, it can't be accidentally be influence or influence itself variables outside of the scope with the same name.
Example:
```
var x=5;
for(int i = 0; i <= 10; i++) {
  var x=i;
  //Do something with local x
}
console.log(x);
```
The log entry will say `10` here. If ´x´ was ´let´, it would still output ´5´.

Therefore maybe use something like `for(let x = Player.SavedExpressions.length; x<5; x++)` here and also for the other, simillar loops in `Dialog.js`.

TessaTech (changes requested):

BondageClub/Screens/Character/Login/Login.js
fleisch11: alright, but i still need the === undefined part, cause .isArray doesnt work if it's undefined

BondageClub/Screens/Character/Login/Login.js
fleisch11: allright

fleisch11:
allright both changes are done

TessaTech:
Oh, i actually was refering to the `while` loops in `Dialog.js` as well. It would be nice if they also were `for` loops.
The `if`-clause in `Login.js` L454 also can be left away since the `for`-loop won't be executed in case it's initial loop condition already is wrong

BondageClub/Screens/Character/Login/Login.js
TessaTech: Oh, i actually was refering to the `while` loops in `Dialog.js` as well. It would be nice if they also were `for` loops.
The `if`-clause in `Login.js` L454 also can be left away since the `for`-loop won't be executed in case it's initial loop condition already is wrong

fleisch11: yeah, i noticed that

fleisch11:
both changes requested by tessatech are now done

fleisch11:
all changes requested by tessatech are done now

BondageClub/Screens/Character/Login/Login.js
```diff
@@ -447,6 +447,15 @@ function LoginResponse(C) {
 			Player.OnlineSharedSettings = C.OnlineSharedSettings;
 			Player.GraphicsSettings = C.GraphicsSettings;
 			Player.NotificationSettings = C.NotificationSettings;
+			Player.SavedExpressions = C.SavedExpressions;
+			if (Player.SavedExpressions === undefined || Player.SavedExpressions.isArray == false) {
```
elliesec, line 451: I don't think this is quite what @TessaTech had in mind. You can use `Array.isArray` (`isArray` being a static method of the `Array` constructor):
```js
if (!Array.isArray(Player.SavedExpressions)) {
    Player.SavedExpressions = [];
}
```
That catches the case when `Player.SavedExpressions` is `undefined` (because `Array.isArray(undefined)` evaluates to `false`) and any other case where `Player.SavedExpressions` is somehow not an array.

Nina-1474:
Minor bug: clicking Load for a number before saving something for that number will throw a "Cannot read property" error.

BondageClub/Scripts/Dialog.js
```diff
@@ -826,6 +832,57 @@ function DialogFacialExpressionsBuild() {
...
+/**draws the savedexpressions menu */
+function DialogDrawSavedExpressionsMenu() {
+	DrawText(DialogFindPlayer("SavedExpressions"), 195, 25, "White", "Black");
+	DrawText(DialogFindPlayer("SavedExpressionsSave"), 140, 180, "White", "Black");
+	DrawText(DialogFindPlayer("SavedExpressionsLoad"), 260, 180, "White", "Black");
+	for (let x = 0; x < 5; x++) {
+		DrawButton(100, 200 + (x * 100), 80, 80, x + 1, "White");
+	}
+	for (let x = 0; x < 5; x++) {
+		DrawButton(220, 200 + (x * 100), 80, 80, x + 1, "White");
```
elliesec, line 865: It's probably worth combining these two loops into one:
```js
for (let x = 0; x < 5; x++) {
    DrawButton(100, 200 + (x * 100), 80, 80, x + 1, "White");
    DrawButton(220, 200 + (x * 100), 80, 80, x + 1, "White");
}
```

fleisch11: allright, i'll do that

BondageClub/Screens/Character/Login/Login.js
fleisch11: allright, i'll test that

fleisch11:
> Minor bug: clicking Load for a number before saving something for that number will throw a "Cannot read property" error.

while it does throw an error, it doesnt break anything, so i left it as it was.
i guess, i'll fix that

fleisch11:
allright, all changes requested by ellisec and nina-1474 are now done

Ben987:
Thanks for that one!  It's a good idea, I will test it very soon.  Cheers :)

PR #2258, opened 2021-03-28
Greyscale + Multi-Coloring: Leather cuffs

Ayesha-678:
Greyscaled and multi-colored the leather ankle, leg, and arm cuffs

PR #2267, opened 2021-03-31
Bugfix: Removes duplicated DefaultColor property

elliesec:
Removes a duplicated `DefaultColor` property from `Female3DCG.js` that was probably missed when #2258 went in. Also adjusts the type of the `Bonus` property in the asset check script so that the script no longer fails, as noted by @jomshir98.


Ben987:
Absolutely.  Thanks again :)

PR #2269, opened 2021-04-01
Padded Blindfold

Ada18980:
Adds a new padded blindfold which has a colorizable trim, and also has BlindHeavy (to allow for gag/blindfold combinations that blind fully while having a profile similar to the leather blindfold)


Ben987:
Very nice!  Thanks for that new item :)

PR #2274, opened 2021-04-02
ChatRoomSync-Split: Enable ChatRoomSyncMemberJoin

TessaTech:
This PR enables the use of function `ChatRoomSyncMemberJoin` added by PR #2272 on which it depends. It requires Server PR [#81](https://github.com/Ben987/Bondage-Club-Server/pull/81) to work.


Ben987:
Hi there.  If I merge the MemberJoin & MemberLeave PR for both client & server, and then update the live server, are you able to test it from your side before the 10th?  I'd like to include at least one of these transactions before the next beta to start the testing process with players.  Thanks a lot :)

TessaTech:
Yeah, shouldn't be an issue. I already tested it a lot on a local server, repeating these tests on the live server shouldn't be an issue.

PR #2281, opened 2021-04-03
Devious Dungeon Only happens in Arcade

Ada18980:
Restricts Devious Dungeon Challenge to only occur inside the arcade, to prevent weird situations with players getting bound in chat rooms.


Ben987:
Very good!  Thanks for that fix :)

PR #2283, opened 2021-04-04
Virtual world upgrade

Ada18980:
Upgrades the Virtual World feature of the VR headset to allow players to see other players in the virtual world, and also allows players to whisper to other players in the virtual world, bypassing the usual whisper restriction settings from Immersion options


Ben987:
Alright!  Thanks for that one.  I will test it soon :)

PR #2285, opened 2021-04-05
Enhanced chat search

Ada18980:
The room filter now searches through descriptions as well as titles. So if someone has 'rp' in their room description but not their title, their room will still show up in the search.

### DIFF
--- a/BondageClub/Screens/Online/ChatSearch/ChatSearch.js
+++ b/BondageClub/Screens/Online/ChatSearch/ChatSearch.js
@@ -16,6 +16,8 @@ var ChatSearchPreviousActivePose = null;
 var ChatSearchIgnoredRooms = [];
 var ChatSearchMode = "";
 var ChatRoomJoinLeash = ""
+var ChatSearchQueryAdd = 0
+var ChatSearchQueryAddTimeout = 5000 // 5000 second timeout
 
 /**
  * Loads the chat search screen properties, creates the inputs and loads up the first 24 rooms.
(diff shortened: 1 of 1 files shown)


Ben987:
Instead of doing two calls on the client side?  Should we do one call and modify how the server returns the result?  Cheers!

TessaTech:
Correct me if i'm frong, but the way i understand it you do two queries. One without search phrase and one with. Then you use the full result to filter for descriptions on client side and join them with the server's name filter results.

NoneNoname:
the only reason for the server to get the search, in this case, is to add the private room which may exist and wouldn't be added otherwise

so we want the server to have minimal effort to complete the query just do not check against include in Name nor Description, just add a room if it private and have the exact name as the query, and always add all rooms, and let the client filter, sort however it wants it

TessaTech:
It depends if we want to cut the calculation power or bandwidth. Calculating things on the server might result in a higher cpu usage for the filtering but reduces the transaction's load since only the filtered room names are transmitted.
Always sending all room names increases the packet size but saves some CPU ressources in return.

Since filtering happens on server side anyways right now, the removal of that from the server would require a server change anyways. If we leave the server like this and let the client handle filtering now, we'll only get higher packet sizes without performance gain for the server.

Ada18980:
My main justification for the double query is that there are a lot of server changes lately and I don't want to add additional complicating factors. But it should be a simple change on the serverside as well.

PR #2286, opened 2021-04-06
Dev Feature: ESLint

jomshir98:
Adds config for ESLint and GitHub action to enforce linting, similar to assetcheck script.

ESLint is a tool for checking JavaScript without running it and has over 13M weekly downloads.
It is highly customisable and the config was discussed with Ellie and other developers.
Things marked as errors either in official recommended or our config will make GitHub action error.

As there are currently many problems found within BondageClub, the config was made quite permissive (There are currently 8045 warnings). The config will be made more restrictive as things are fixed/improved.

This PR also contains fixes for all current 286 found errors

### DIFF
--- a/.github/eslint-matcher.json
+++ b/.github/eslint-matcher.json
@@ -1,0 +1,21 @@
+{
+    "problemMatcher": [
+        {
+            "owner": "eslint-stylish",
+            "pattern": [
+                {
+                    "regexp": "^([^\\s].*)$",
+                    "file": 1
+                },
+                {
+                    "regexp": "^\\s+(\\d+):(\\d+)\\s+(error|warning|info)\\s+(.*)$",
+                    "line": 1,
+                    "column": 2,
+                    "severity": 3,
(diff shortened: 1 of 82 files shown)


elliesec (approved):
Looks good, and I think the rules are a decent selection based on where the code is at the moment - a bunch of nice little fixes in here as well!

@Ben987 - I'd definitely consider merging this when we're out of the hotfix period. It'll run passively as a check on pull requests, and can help catch coding errors before they become a problem. The way results are formatted is also informative for people that don't know how to run it themselves, and people shouldn't need to know how to run it in order to be able to fix linting errors.

jomshir98:
Added badge similar to #2323

Ben987:
Ok, time to test this huge one!  Big changes like these are always scary for me because you never know which new bugs it will trigger.  I will try to go in all the screens and test most mini-games, but the beta period will be best to test that one out.  Thanks a lot for all this hard work.  Cheers!

PR #2287, opened 2021-04-06
Fix: exposed tape missing dialog

fleisch11:
forgot to add chatroom dialog for the exposed tape


Ben987:
Yes!  Thanks for the fix :)

PR #2288, opened 2021-04-06
Bugfix: Missing lock assets for HarnessPanties2

elliesec:
## Summary

The `HarnessPanties2` asset is missing a lock asset for the legs spread pose, causing 404 errors. Since all lock assets for the item are blank assets, I've removed all lock assets and made use of the new(-ish) `DrawLocks` attribute to prevent lock layers from being drawn for the item.


Ben987:

PR #2292, opened 2021-04-07
Removed multiples of ChatRoomSync-Split functions

TessaTech:
These functions somehow got merged multiple times. This PR deletes the copies.


Ben987:
Yes!  Thanks for the fix :)

PR #2293, opened 2021-04-07
Custom backgrounds updates

Ada18980:
Changes items with custom backgrounds (for now just the VR headset) to still replace the background in Light sensory deprivation mode. This also allows items that do not have BlindHeavy to use a custom background.

The current behavior is to override DarkFactor only when DarkFactor is 0 (pure black). In the future this can be changed once there is some clarity in how blindfolds should interact with custom backgrounds from something other than a blindfold (which do not exist yet)


Ben987:
Alright!  Thanks for that one, cheers :)

PR #2294, opened 2021-04-08
Devious Dungeon 1.1

Ada18980:
Adds:
-Several new spells
-System for finding new spells in chests
-System for opening random pages in the spellbook when bound
-A random armbinder trap in chests


Ben987:
Very good!  Thanks for that upgrade.  Right on time for the next beta :)

PR #2296, opened 2021-04-09
Fix: chastity belt response for private characters

fleisch11:
makes it so the custom characters from the private room react to one adding/removing a chastity belt


Ben987:
Yes!  Thanks a lot for that one, it was overdue :)

PR #2297, opened 2021-04-09
queue and await AccountUpdate, reduce emit calls

NoneNoname:
only a draft for now since this won't get pulled before the next beta

Example of a user buying items one after another:
image

With this change only a single update is sent to the server:
image


NoneNoname:
this doesn't require server-side change it only reduces the amount of AccountUpdate messages

NoneNoname:
@Ben987 is this still on hold, if not I will resolve the conflicts

Ben987:
@NoneNoname , I'd like to test it for the next version, the more air we can give to the server the better.  But I've never messed with classes yet so far, so it's kind of tricky for me.  Could you change the PR so it doesn't use classes?  Maybe use a regular function that calls itself every 10 seconds to see if there's something to sync on the account.  Thanks a lot.

jomshir98:
@Ben987 

I would strongly recommend keeping it as a class. Classes are quite simple (if you don't use inheritance, which this doesn't) and very powerful for organising things!

If you already know how to work with objects, classes should take you ~10 minutes to understand. I strongly recommend following site, which will give you everything you need to know.
[https://www.w3schools.com/js/js_classes.asp](https://www.w3schools.com/js/js_classes.asp)

NoneNoname:
we already have classes, none of these prs rise any problems from you, so I don't see why this change should be made at all, and `10 seconds` is a bad idea itself, that is not the same concept with this pr, that will not be a code improvement

pr #2364 `class NotificationEventHandler` accepted by you @Ben987 
pr #1368 `class TextCache`

Ben987:
@elliesec , @Nina-1474 , are you able to support and debug this pull request if we get trouble during the beta?  Anything that touches the client/server communication is always touchy.  Thanks!

Ben987:
It's time to test this one!  Lets hope it will help with the server spikes.  Thanks a lot :)

PR #2298, opened 2021-04-09
Change: Allow Asylum chat access without being committed

jomshir98:
This changes asylum chatroom access for patients.
You no longer need to be committed patient to access the rooms, having proper clothing is enough.
( But you must have non-zero reputation, so you must have commited yourself in the past )
Also adds dialog option to get clothes from nurse without committing yourself.

Reason:
Currently the Asylum is almost dead. If attempting to enter it didn't lock you up for 15 minutes, maybe more people would try to do so.


Ben987:
No, we do need a solution to help the Asylum, but it's made to be different. immersive & special.  Maybe I'll start a survey on that topic soon.

PR #2299, opened 2021-04-09
Improvements: Friendlist

jomshir98:
- adds "Reply" button when reading beep message, opening send beep dialog
- If friendlist was opened from chat search screen, the names of rooms on list of friends and on beeps will be clickable, exiting the friendlist and searching for said room (only if friend is in same space as player)


Ben987:
Sure!  That's a good idea.  Thanks :)

PR #2302, opened 2021-04-10
Bugfix: Fixes casing for asset paths in the Kinky Dungeon minigame

elliesec:
## Summary

The Kinky Dungeon was using `Minigame` instead of `MiniGame` in asset paths, resulting in 404 errors when trying to load assets.


Ben987:

PR #2303, opened 2021-04-10
Fix `WhiteList` loading, add `BlackList` loading

jomshir98:


Ben987:
What does it fix in-game exactly?   How can I test it?

PR #2304, opened 2021-04-10
Bugfix: Fixes incorrect appearance update rollbacks

elliesec:
## Summary

There was an error in `Validation.js` where item colors were being compared incorrectly, resulting in rollbacks for valid appearance updates (see screenshot), especially when cosplay/body items were multi-colored. This uses a more complete comparison for item colors which should prevent these false alarms.

Validation reports


Ben987:
Thanks for the fix!

PR #2305, opened 2021-04-10
Custom Tag Regex Fix

Nina-1474:
When the text of a Custom Collar Tag or Electronic Tag was checked for invalid characters, the lastIndex property was updated to the end of the string since the global flag was unnecessarily included. This made the next test fail and revert the text to the default "Tag" even though the custom text was unchanged and still valid.


jomshir98:
I'm not sure it should be **m**ultiline either.

elliesec:
Beat me to it! I was about to raise exactly this (incidentally, I agree with @jomshir98 - I'd removed the multiline flag on my branch).

Nina-1474:
> I'm not sure it should be **m**ultiline either.

True. Removed that too now.

Ben987:

PR #2306, opened 2021-04-11
Sci-Fi Pleasure Panties Fixes

Nina-1474:
- An error message appears for missing "AssetsItemPelvisSciFiPleasurePantiesScriptDraw" when the SciFi Pleasure Panties are worn since the configuration had DynamicScriptDraw: true despite not using a dynamic draw function.
- Corrected missing text for the Lock layer name.
- Added missing character name in message on clicking Trigger Shock.
- Adjusted the default colour and preview image.
- Swapped the two plug layer names.

Credit shared with **Aeren** for a few of the fixes which I checked in on their behalf.


Shiranui-Izayoi:
Ahhhh sorry my bad I must have missed that I set that to true. Meant to set it to false. Thanks for catching it. I was looking up the issue just now.

It never came up in my testing since the function that warns of it wasn't in yet. Thank you Nina.

Nina-1474:
> That said can you change Lock to Lock Icon. Since that was the inital idea.

I somehow didn't notice that original line just below it! Change now made.

Shiranui-Izayoi:
No problem I managed to not notice that the lock layer didn't work so both are equal. Honestly I am having to make a few changes to the files anyway because I missed a couple small things mostly just a colour issue. So will upload those soon.

Shiranui-Izayoi:
Ow one last thing I noticed I got the Clit Toy and Pussy Toy the wrong way round.

They are currently like this
> ItemPelvisSciFiPleasurePantiesPlug2,Clit Toy
> ItemPelvisSciFiPleasurePantiesPlug1,Pussy Toy

When they should be like this
> ItemPelvisSciFiPleasurePantiesPlug2,Pussy Toy
> ItemPelvisSciFiPleasurePantiesPlug1,Clit Toy

Ben987:
Yes!  Thanks a lot for these fixes.

PR #2307, opened 2021-04-11
Lock Timer Validation Fix

Nina-1474:
Timer locks were dropping the "Time left" property and displaying null/NaN due to an error in the lock validation.
I also changed the max-time-exceeded correction to apply the lock's max time instead of its default time, since normal non-consoled time additions managed to occasionally exceed this max time.


Ben987:
Absolutely!  Thanks :)

PR #2308, opened 2021-04-11
Hair Accessory Cleanup

Nina-1474:
A number of changes to the three HairAccessory groups, making use of recent developments in order to remove duplicate images and neaten code.
- Now allowing an Asset Group to set a DynamicGroupName which will be passed to all Assets within to avoid needing to individually setting the same value on every asset in the group.
- Changed HairAccessory2 and HairAccessory3 to dynamically reference HairAccessory1, since these two groups are strict subsets of the 1 group. All assets within them were duplicates, and have been dropped.
- The Suspension pose variants for the 3 groups has been dropped since the Suspension versions of the assets are identical to the base versions and don't need to be there.
- The asset-level BodyCosplay settings on every HairAccessory2 asset have been removed as this is covered by the group-level setting.
- The Halo in HairAccessory3 was the "base" version before, with the HairAccessory1 version dynamically referencing it. This has been flipped so that 1 version is now the main version, for consistency with how groups 1 and 3 are used elsewhere and to allow all HairAccessory3 assets to be dropped.
- A few unneeded ElfEars files have been removed.

This also resolves an error where FoxEars3 would disappear when suspended upside-down.


Ben987:
That's a great idea, thanks.  Less images to upload with each version :)

PR #2309, opened 2021-04-11
Item Color Layer Order Fix

Nina-1474:
When opening the color menu for an item with multiple colorable layers, the order the layers appeared in was inconsistent depending on the browser. In Chrome they were the expected order, while on Firefox they were in reverse.


Ben987:
Great!  Thanks for that one.

PR #2310, opened 2021-04-11
Kidnap Dress Up Fix

Nina-1474:
The kidnap game's function to process a selected upper move wasn't updated to account for the new move list. The only move that was actually affected was UndoCloth (Dress Back Up) which was doing nothing when selected.


Ben987:

PR #2311, opened 2021-04-12
Re-added support for the Restrain Property

Shiranui-Izayoi:
Re-added support for the Restrain Property. As while it is not used by items anymore. Having it allows those that use console to enhance their experience, while also minimising risks to other people.

As it currently is everything that could be done before can still be done, but doing so is much more risky and likely to break something. While yes this only effects those that use console, it allows greater diversity in RP oppotunities. Such as being able to hide things, set poses etc. Once again I would like to say that all of this is still possible, but doing so is now MUCH more likely to cause negative effects/outcomes.

I also removed the new validation of the lock member number since non number locks are used in the game currently. Off the top of my head the Nursery uses text instead of member numbers. And to my knowledge allowing stings in the lock member number doesn't have any negative consequence.

### DIFF
--- a/BondageClub/Scripts/Validation.js
+++ b/BondageClub/Scripts/Validation.js
@@ -475,8 +475,8 @@ function ValidationSanitizeProperties(C, item) {
 		}
 	}
 
-	// Remove invalid properties from non-typed items
-	if (property.Type == null) {
+	// Remove invalid properties from non-typed & non-restrain items
+	if (property.Type == null && property.Restrain == null) {
 		["SetPose", "Difficulty", "SelfUnlock", "Hide"].forEach(P => {
 			if (property[P] != null) {
 				console.warn("Removing invalid property:", P);
@@ -549,11 +549,6 @@ function ValidationSanitizeLock(C, item) {
 
 	// Remove any invalid lock member number
 	const lockNumber = property.LockMemberNumber;
-	if (lockNumber != null && typeof lockNumber !== "number") {
-		console.warn("Removing invalid lock member number:", lockNumber);
-		delete property.LockMemberNumber;
-		changed = true;
-	}
 
 	// The character's member number is always valid on a lock
 	if (lockNumber !== C.MemberNumber) {


NoneNoname:
since `property.Restrain` has no validation and no limitation this change makes that whole validation for those specific effects pointless

I feel a much better approach could be adding user specific settings in `OnlineSharedSettings` that may allow certain validations to be turned off on that character

jomshir98:
I'm very much against this. The whole point of this validation is to prevent console from griefing other players. There also were some bugs with non-number `LockMemberNumber` and I see no harm to any script, if it just uses bot's membernumber instead of arbitrary text.

I think that support for adding arbitrary effects using console/script would be very nice and could enhance some scripted room very nicely, but it shouldn't be done as a hack, so I have following proposal:
How about adding item, that cannnot be blocked from being removed, that would literally allow any effect to be put on it using console? This item could even be limited (whitelist-only) by default, so you would have to whitelist bot for it to be used on you. I think adding such item would only benefit things as it would provide non-hacky **opt-in** way to do scripts, instead of scripts "forcing" things on others as some scripts are doing now.

elliesec:
I'm strongly against this change. The `Restrain` property allows console users to impose effects and graphical issues that are not intended in the game upon other players without their consent. Like @NoneNoname has said, a much better approach would be to allow players to disable specific validation checks with an "at your own risk" warning. That way, you can still add arbitrary effects and properties to an item, but you're not imposing them on players that don't want anything to do with it.

Shiranui-Izayoi:
> 
> 
> I'm very much against this. The whole point of this validation is to prevent console from griefing other players. There also were some bugs with non-number `LockMemberNumber` and I see no harm to any script, if it just uses bot's membernumber instead of arbitrary text.
> 
> I think that support for adding arbitrary effects using console/script would be very nice and could enhance some scripted room very nicely, but it shouldn't be done as a hack, so I have following proposal:
> How about adding item, that cannnot be blocked from being removed, that would literally allow any effect to be put on it using console? This item could even be limited (whitelist-only) by default, so you would have to whitelist bot for it to be used on you. I think adding such item would only benefit things as it would provide non-hacky **opt-in** way to do scripts, instead of scripts "forcing" things on others as some scripts are doing now.

Ok so the issue with an item is first where to put it? Would it be in the wardrobe or in an "item" slot. If so it would likely need to be a new slot to prevent it getting in the way of existing items. Then after that some effects are only possible when multiple items can be used to add properties to at the same time.

I do agree it would be nice to have since, personally I am very much against constantly restricting console as it heavily limits what can be done in the game. Since not everything is possible to make into a feature in game as it is simply to complex for the general player base. Since some things require you to know of the assets in game in order to use them. Such as pose names or slot names etc.

In my case multiple times I have tried to come up with a way to allow people to hide items in game to replicate what I use console for, or to allow them to do other things like say have a corset appear above an outfit. But the issue always comes back to 1 where to put it that it is easily accessable. And 2 how can I make it as easy to use and "idiot proof" as possible to prevent people accidentally creating an abomination.

For example once when I proposed the idea on the discord, multiple people wanted others to have the ability to hide their items, and for themselves to not be able to change it while bound. Which to say would result in more issues would be an understatement.

Also in this case I am not talking about bots. I know a few of them have been going around recently, but most of them seem to use very basic console. I would also very much like if you could tell me what bugs occured from non number member numbers. Considering that the Nursery has used and still does use non numbers in the member number field.

> 
> 
> I'm strongly against this change. The `Restrain` property allows console users to impose effects and graphical issues that are not intended in the game upon other players without their consent. Like @NoneNoname has said, a much better approach would be to allow players to disable specific validation checks with an "at your own risk" warning. That way, you can still add arbitrary effects and properties to an item, but you're not imposing them on players that don't want anything to do with it.

If I remember correctly we actually discussed about this before and I was for it the issue is implementation. I would not be opposed to trying to work it up. Because as said before I believe creative use of the console enhances RP and can even bring new people to contribute. I would not have ever contributed if console didn't exist as I wouldn't have wanted to learn the games scripting, and likely would have dropped the game. As I use a script that effectively was RP difficulty before RP difficulty was a thing.

> 
> 
> since `property.Restrain` has no validation and no limitation this change makes that whole validation for those specific effects pointless
> 
>     * yes item with types can have any of these effects since as of now type is not validated, which may change over time
> 
>     * items without type are usually clothes and it makes it harder to find which item needs to be removed/changed (ex: eye) to undo the hacked effect if it was applied without the user knowledge
> 
> 
> I feel a much better approach could be adding user specific settings in `OnlineSharedSettings` that may allow certain validations to be turned off on that character
> 
> note: if it is a private room where you RP, all side can have the same override for validation thereby allowing you to do more than what normally allowed

I will admit that it is hard to find a hacked item especially without console to track it down. Though admittedly most console users to my knowledge that use these kind of console codes will tell people or undo them. The vast majority of griefers would never look enough into the code to do it. And in the case of non (bondage) items loading an outfit from the wardrobe would reset it completely.

I also agree that a shared setting would be nice. Since as said I don't agree with constantly removing console from the game. And believe that while the general player should be protected it shouldn't be at the expense of the people that use it to enhance their experiance. I will look into making a setting of it that can be toggled. Likely with a few different levels, one that disables all item validations, one that disables the most basic etc.



Honestly I knew from the outset that this likely wouldn't get accepted, but I did feel I had to bring it up. Very often it is just accepted that this happens and should happen, But I believe that there should be a balance between things and not just pushing this community away.

Also I wouldn't say it makes the validation pointless as since `Restrain` has only been used a few times, it is VERY unlikely that most would even know what it does.

I will as said look into making some `OnlineSharedSettings` that will disable various validations and restrictions. Likely with a warning before hand.

NoneNoname:
so I thought more about this:
- `OnlineSharedSettings` to fine control what is allowed by the user
- new `<Item>.Property.Scripted = ['Hide', 'LayerPriority', 'SetPose', ...]` which specify which Property shouldn't be validated on an item
- Level 0: `Property.Scripted` is deleted
- Level 1: Extra group with a single item removable by dialog, only properties specified in `Scripted` array can exist on it
- Level 2: `Scripted` array can exist on normal items, these items can have highlighted and should be a way to remove the Scripted tags and revalidate the item if the user wants it
- `<Item>.Property.LayerPriority = [{ Name: 'Layer1', Priority: 3 }, ...]` allow scripted items to have different layers

each Property can be opt-in for level 1 or 2

Shiranui-Izayoi:
> 
> 
> so I thought more about this:
> 
>     * `OnlineSharedSettings` to fine control what is allowed by the user
> 
>     * new `<Item>.Property.Scripted = ['Hide', 'LayerPriority', 'SetPose', ...]` which specify which Property shouldn't be validated on an item
> 
>     * Level 0: `Property.Scripted` is deleted
> 
>     * Level 1: Extra group with a single item removable by dialog, only properties specified in `Scripted` array can exist on it
> 
>     * Level 2: `Scripted` array can exist on normal items, these items can be highlighted and should be a way to remove the Scripted tags and revalidate the item if the user wants it
> 
>     * `<Item>.Property.LayerPriority = [{ Name: 'Layer1', Priority: 3 }, ...]` allow scripted items to have different layers
> 
> 
> each Property can be opt-in for level 1 or 2

That could work the main issue I see is that a lot of stuff would need to be rewritten to make it work. Since you would need to add `.Scripted` to be checked by `Hide`, `SetPose` etc. 

I think a better idea would be having the levels as said. Either as a single setting which you can move up and down, or multiple checkboxes. Then alter the validations, to make it so if the appropriate settings are enabled it will bypass the check on the person. Meaning that the moment you turn the settings off it would re-enable the checks and thus delete the consoled variables.

So for example a setting could be to allow any effect on any item regardless (One of the very high level settings if it existed). If you had that checked then the game would bypass the check to see if that effect was allowed so you could say have a gag that vibrates. Then if you turned the setting off it would then notice "Hey that setting isn't allowed" and remove it.

This approch minimises changes to the code, doesn't require a new item and is still easy to remove since you can just turn off the setting. It may also be possible to add something so the item will be coloured differently if it has properties that would normally be not removed by validation, but that would be saved for later.

How does that sound?

TessaTech:
> That could work the main issue I see is that a lot of stuff would need to be rewritten to make it work. Since you would need to add `.Scripted` to be checked by `Hide`, `SetPose` etc.

Actually, only the validation would have to consider `.Scripted` to either delete `Hide`, `SetPose`, etc. or not. The rest of the game can still work as it does right now.

> I think a better idea would be having the levels as said. Either as a single setting which you can move up and down, or multiple checkboxes. Then alter the validations, to make it so if the appropriate settings are enabled it will bypass the check on the person. Meaning that the moment you turn the settings off it would re-enable the checks and thus delete the consoled variables.

This would still require a revalidation of all currently worn items since validation only is performed once right now and that is when the item is applied/changed/removed. A remove button would defeat the purpose of revalidating all items and only remove the one in question. It also would allow you to keep your current scripted items without allowing new ones to be added or your current ones from being modified.

But unrelated to that:
I also want to throw in that scripted items also can affect other players in the room, so an option to show/hide scripted items on others should also be included. Keep in mind, however, that this could cause asynchronous player states and therefore at least confusion among players (For example, a character has different poses for different players, be gagged for some players and others not, etc.). Worst case, this could result in undefined behaviour within custom scripts or even the game itself.

Shiranui-Izayoi:
> 
> 
> > That could work the main issue I see is that a lot of stuff would need to be rewritten to make it work. Since you would need to add `.Scripted` to be checked by `Hide`, `SetPose` etc.
> 
> Actually, only the validation would have to consider `.Scripted` to either delete `Hide`, `SetPose`, etc. or not. The rest of the game can still work as it does right now.
> 
> > I think a better idea would be having the levels as said. Either as a single setting which you can move up and down, or multiple checkboxes. Then alter the validations, to make it so if the appropriate settings are enabled it will bypass the check on the person. Meaning that the moment you turn the settings off it would re-enable the checks and thus delete the consoled variables.
> 
> This would still require a revalidation of all currently worn items since validation only is performed once right now and that is when the item is applied/changed/removed. A remove button would defeat the purpose of revalidating all items and only remove the one in question. It also would allow you to keep your current scripted items without allowing new ones to be added or your current ones from being modified.
> 
> But unrelated to that:
> I also want to throw in that scripted items also can affect other players in the room, so an option to show/hide scripted items on others should also be included.
> This, however, can result in other issues, starting with asynchronous player states resulting in at least confusion among players (For example, a character has different poses for different players, be gagged for some players and others not, etc.). Worst case, this could result in undefined behaviour within custom scripts or even the game itself if players without scripts allowed on others update players with scripted items.

Ok just wanted to address two quick things first that what I meant is that the game currently when looking for a pose would not check `Property.Scripted.SetPose`. It would only look at `Property.SetPose`. Second it would not cause asynchronous player states.

This is because by using `OnlineSharedSettings`, when a player goes to interact with someone those are well shared with someone. So their game knows that this person is allowing scripts on themselves and thus would not change anything. In addition all states will be the same for all in the room. If I have an item that has `Property.SetPose` and it equals `SpreadKneel` then it would put my character in that pose and the games of everyone in the room would see that "Ow this item is setting this characters pose as that" and thus render that person in that pose.

Then while not as important but worth bringing up the game doesn't validate once. It validates EVERYTHING every time you enter an online room or someone interacts with you in most cases. It also would not affect others anymore than a normal item. For example how much does using a hogtie on someone effect others. That would be the same way an item with the variables would.

What you are likely thinking of is loops or more complex console scripts. The term scripted item is very misleading in this case. What they mean is a item that has been modified outside of what would be possible by the game normally. For example having a collar that vibrates, or a ball and chain that is invisible. And other such things. All things that are possible in the base game just not for that item. Hope that makes sense.

elliesec:
> Ok just wanted to address two quick things first that what I meant is that the game currently when looking for a pose would not check Property.Scripted.SetPose. It would only look at Property.SetPose.

@Shiranui-Izayoi - you're misinterpreting the proposal. `Property.Scripted` is an array of strings which indicate that certain properties shouldn't be passed through normal validation. For example, if an item has `Property.Scripted = ["Effect"]`, it's indicating that the item may have effects that aren't in the item's `AllowEffect` array. So the above wouldn't be an issue - we wouldn't need to modify drawing code to make it happen - the only place that changes would need to be made is in the validation code (and since that all goes through one central channel now, that's a relatively simple change).

> This is because by using OnlineSharedSettings, when a player goes to interact with someone those are well shared with someone. So their game knows that this person is allowing scripts on themselves and thus would not change anything. In addition all states will be the same for all in the room.

You're misunderstanding @TessaTech's point. The point was that scripted items have the potential (for example) to make people invisible, and that gets imposed on others whether they like it or not. Tessa was proposing a preference to allow people to ignore scripted properties on _other_ players (which again, I believe could fairly simply be done at the validation level), meaning that a player with that preference set would perceive the scripted player differently to players that don't. I'm of the opinion that that's largely acceptable - we already have something similar in the form of the item visibility preferences (and we've discussed adding an icon to players with scripted items, which would indicate that, much like the item visibility preferences we already have).

Finally, to your earlier points: yes, we discussed this, and a system quite like what @NoneNoname described was actually in my plans. Contrary to what you may believe, I don't hate console, nor am I trying to stamp out its use in the game - in fact I am of the opinion that it can enhance the game massively. However, I _am_ of the opinion that console users should _not_ be able to impose graphical changes or issues on other players if those players do not consent to it. I'm also well aware of other workarounds that allow you to achieve similar things to what `Restrain` previously allowed, and I _am_ intending to add validation for those things too.

However, what I eventually want is a _controlled_ system which replaces that, which gives players the power to choose whether they want to experience the game in its vanilla format, free from any potential console abuse, or for them to give their informed consent to experiencing what scripts are capable of, whilst acknowledging that that comes with some element of risk. That, in my opinion, is the best of both worlds. It's not going to happen overnight, but the new validation system was a first step in that direction, and this invalidates a lot of the point of validation right now. We all know that there is still a workaround, which is no more dangerous than `Restrain` was, but its existence by no means invalidates the removal of `Restrain`.

Ben987:
Let's close this for now.  We don't want hackers to abuse that property.

PR #2314, opened 2021-04-13
Nose access change for blindfolds

Manilla32:
Altered access to the nose when a blindfold is on, so it makes more sense visually.
This will allow the changing of items on the nose when most blindfolds are on. This does not change the blindfolds that visually cover the nose.


Ben987:
Sure!  Thanks for that little tweak.

PR #2315, opened 2021-04-13
Kicked during Kinky Dungeon fix

Nina-1474:
If a player was playing the Kinky Dungeon minigame in a chat room and got kicked by an admin, on finishing their game they would still be in the room from their viewpoint but couldn't interact with anything. 
They will now be interrupted and immediately exit to the chat search screen as normal.


Ben987:
Yes!  Thanks for that fix :)

PR #2316, opened 2021-04-14
Urgent Pleasure Panties js update

Shiranui-Izayoi:
I messed up the click part for the show chat message box. By both including two versions of it AND also putting in the wrong XY values. Just found it out and fixed them. Needs to get in before next update.


Ben987:
Sure!  Thanks for that fix.

PR #2317, opened 2021-04-14
Bugfix: Properly enables partial appearance rollbacks

elliesec:
## Summary

This was something I'd intended to implement in the original validation change, but it's come to my attention that there was an issue with the way it was implemented. For the most part, this issue should only affect those using console, but there may be some edge cases where it causes unexpected behaviour in the standard game.

Currently, when an invalid appearance bundle is received, the character's full appearance is rolled back, rather than just the items that were invalid. Additionally, I'd intended for partial rollbacks to work on a per-item basis - for example, if an item is added with a lock that is blocked on the player in a single update, I'd intended to only roll back the lock part of that update, and permit the item to be equipped without a lock, but this isn't what happens right now. The final appearance bundle was being resolved correctly, but wouldn't get applied until _after_ the rollback update, which invalidated the creation of the validated bundle, and could result in appearance desynchronization between players in a chatroom. This PR addresses that by ensuring that the sanitized appearance bundle is applied _before_ the rollback happens.

### DIFF
--- a/BondageClub/Screens/Character/Login/Login.js
+++ b/BondageClub/Screens/Character/Login/Login.js
@@ -485,7 +485,7 @@ function LoginResponse(C) {
 			LoginDifficulty();
 
 			// Loads the player character model and data
-			Player.Appearance = ServerAppearanceLoadFromBundle(Player, C.AssetFamily, C.Appearance, C.MemberNumber).appearance;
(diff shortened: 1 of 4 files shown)


Ben987:
Absolutely, thanks for that one.

PR #2318, opened 2021-04-14
Bugfix: Improves character clumsiness calculation

elliesec:
## Summary

This makes a few refinements to clumsiness calculation for the combination lock. In particular, it now ensures that hand items have the `Block` effect, so that the Futuristic Mittens in glove mode don't trigger clumsiness.


Ben987:
Alright, thanks for that little tweak.

PR #2320, opened 2021-04-14
Bugfix: Chatroom messages for the pet bed

elliesec:
## Summary

The pet bed would previously not display chatroom messages when changing type, despite messages existing in `Dialog_Player.csv`. This was due to a typo in the function name. I also made some slight edits to the chat messages themselves to ensure that the tags worked correctly, and to make the wording flow a little better.


Ben987:
Yes!  Absolutely.  Thanks.

PR #2324, opened 2021-04-15
Account creation maid multiplayer questions

Ada18980:
Adds a question from the account creation maid asking the player what they would like to set their permissions to. The options are:
-No rules
-Block all padlocks (except safeword lock, since that one's always possible to escape)
-Set permission to whitelist only

Still looking for input as to whether the safeword lock should be blacklisted or not, since it's always possible to escape.

### DIFF
--- a/BondageClub/Screens/Room/MainHall/Dialog_NPC_MainHall_Maid.csv
+++ b/BondageClub/Screens/Room/MainHall/Dialog_NPC_MainHall_Maid.csv
@@ -157,7 +157,11 @@ ClubUpdated,,,Welcome back! The Club went through some renovation since you last
 1010,,Tell me about the reputation system.,"Your reputations will increase or decrease based on your actions and answers.  It affects your ranking, the activities you can do and how others perceive you.",,
 1010,,I want to learn about skills.,"There are different skills you can learn and get better with time and practice.  The maximum level is 10, the percentage is your current progress toward the next level.",,
 1010,1002,What are the most important rules again?,The first rule is that the club is only for adults.  This place isn't for minors.  Are you an adult in your country?,,
-1010,,(Thank her and leave her.),,DialogLeave(),
+1010,1011,That is all.,One more thing... would you like to set some rules for multiplayer chat rooms?,,
+1011,1012,I do not wish to set any rules.,Fair enough. I hope you enjoy your stay.,,
+1011,1012,"I am fine with being tied up, but not with padlocks",I have updated your blocked items to exclude locks.,BlockAllLocks(),
+1011,1012,I only want to be tied up by trusted players.,I have updated your permissions to whitelisted players only.,SetPermissionsToWhitelist(),
+1012,,(Thank her and leave her.),,DialogLeave(),
 1100,,,"(A pair of maids cut you loose and bring you to the main hall.)  You poor thing...",,Player.IsNaked()
 1100,,,"(A pair of maids cut you loose and drag you to the main hall.)  Oh how the tables have turned, head maid DialogPlayerName...",,IsHeadMaid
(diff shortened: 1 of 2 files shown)


tetris245:
I don't understand the proposed system of permissions, that has less options than the general system of permissions.

 To illustrate that: I don't use whitelist, only blacklist and friendlist, and I play in 'everybody except blacklist' mode. I have also disabled the safeword, because I dislike that it changes my global permissions when used.

About my approach: All people in my friendlist are people that I can touch and they are also allowed to touch me. In other words, there is for me no difference between friendlist and whitelist (I know many people make a difference between both lists, but not me). As some evolution of the game still refers more to whitelist to use some items, I really  would appreciate the addition of a selectable option allowing automatic adding of all friends to the whitelist (in the same way that ghosted people are automatically added to the blacklist). I ask that because I really have a big amount of friends, adding them one manually one by one would be a nightmare as it would ask so much time and patience.

TessaTech:
@tetris245 - I think this PR is more about making new players aware about permission settings and giving them some basic options to start with, other then the rather unintuitive default "Owner, Lovers, Whitelist **and Dominants**".

I don't know what your intent behind telling your own usage of whitelists is, but if you want to make a feature request or suggestion, i would ask you to visit the discord for this section is to discuss the concrete pull request.

If you want to add your complete friend list to your whitelist, this can also easily be done via console. But this also is not to discuss here and rather for the discord.

tetris245:
Thanks for the explanation.  Concerning my suggestion, I've posted it on discord.

Ada18980:
@tetris245 Tessa is right. This is more about adding "default" options. Nothing here is permanent or persistent, its simply meant to give new players an opportunity to set the defaults on some settings they might not know about.

For example, a player mentioned in the discord that they didn't know they could block items. For new players, the existing features to combat trolling (permissions, blocked items) aren't known to the player, so it ought to help to give a couple of defaults without having to explain the whole system.

elliesec:
One thought on this: it's fine as it is, but it might be nice for the dialogue to cycle back to the rules prompt (stage 1011) after selecting one of the rules (in combination with a `DialogRemove()` call), to allow the player to select both options if they wish before continuing.

As a natural extension to that, I could see the options at stage 1011 leading quite nicely into sub-dialogues which allow the player to effectively have an initial setup experience (i.e. "I'd like to configure my permissions", leading into a dialogue stage where the player can choose from multiple permission options). Obviously, that's something that could be added later - this is definitely a good starting point.

As far as locks go, I'd be inclined to say that it should block all locks (even the safeword lock). Two reasons:
1. It allows the player to explore what each lock does in their own time, without a panic moment of "I thought I blocked all locks!" when a safeword lock gets equipped on them, particularly if they don't yet know how it functions
2. It allows us to future-proof `MainHallLocks` by generating the list - just pull out all assets with an `IsLock` property, so that if a lock gets added in the future, it's automatically built into the list for any new players.

Ada18980:
@elliesec Good point, since it wont be till next beta when this has a chance of getting in, theres plenty of time for me to incorporate some of those

Ben987:
Lets try to think as a brand new player who never tried the game.  Is it too early to ask this question when you don't know anything about the game?  Can someone who doesn't know the game really answer this question right after creating her account?  Maybe we should let the players discover the game before we ask this.

Ada18980:
Hi @Ben987,

The motivation for this came up when a newer player complained that some locks were very difficult to escape, and ended up blocking the item.

This PR basically just gives a choice of 3 default settings. But on the other hand you are right, this is an adult game and not all players know what they like/don't like. Still, I do see some value in informing the player about things that can happen online.

Ben987:
Let's close this one for now and maybe figure out another solution if this becomes a huge problem.  I doubt that a new player will be able to answer that question properly, when she doesn't know anything about the game.

PR #2325, opened 2021-04-15
Extended Bunny Mask 1

Shiranui-Izayoi:
Basically the title. The edit isn't perfect and if anyone decides to put the earless version above the head it can look rather strange but no current hairstyles allow the user to see it.

Also preserved the original version aka it comes with the ears by default.

The reason being is that like some of the other delicate looking masks they can look really nice under the hair, but with the bunny one the "ears" are to long and stick out. If this is liked I will likely extend the other ones. There shouldn't be any bugs as I basically just copied the LatexCorset1 code.


Ben987:
Yes!  Thanks for that upgrade.

PR #2327, opened 2021-04-15
Asset: blanket hood

fleisch11:
asset done by ayesha
code done by me
since i already added the blanket, might as well
(though i guess, it'll have to wait until the next game version, won't it?)


Ben987:
Yes!  Thanks for that one you two.  Cheers :)

PR #2332, opened 2021-04-16
Bugfix: Ensures expression remove timers get removed when the expression is

elliesec:
## Summary

This is something that gets flagged by the new validation. If an expression is set with a `RemoveTimer`, but the player then manually changes their expression back to default, the remove timer isn't removed, which causes validation errors.

This PR modifies the `CharacterSetFacialExpression` function so that whenever an expression is changed, any existing `RemoveTimer` is deleted before changing the expression (and if a new `RemoveTimer` should be set, that still happens). This should prevent these validation warnings, and also fix a bug where manually setting an expression whilst a `RemoveTimer` is in effect causes the new expression to reset when the `RemoveTimer` expires.


Ben987:

PR #2333, opened 2021-04-16
Improvement: Automatically generates AllowEffect & AllowBlock for modular items

elliesec:
## Summary

This changes the modular item script to automatically generate the `AllowEffect` and `AllowBlock` properties on modular assets. Since all valid effects/blocks are defined in the extended item configuration, explicitly defining them in the asset definition is now information duplication and prone to human error. This only affects the high security straitjacket right now, but means there's less to think about with any future modular items.


Ben987:
Very good, thanks :)

PR #2335, opened 2021-04-16
Beta-fix: Remove Friendlist sorting

jomshir98:
In #2299 I fixed friendlist sorting, so it works as it was originally intended. This change removes the sort altogether.

Players however seem to prefer old way the friendlist was sorted (in order it was created on server - lovers&submissives first, everyone else in order they were added into friendlist)

This is temporary fix, until more stable sorting using actual friend types will be done.


tetris245:
About the friendlist, I think it should be in the following order: owner - lovers - subs - other friends and each sub-category (except owner ofc) should be alphabetically ordered.

TessaTech:
Actually, making the sorting order configurable would be a good note for the future

Ben987:
Sure, we can always add a custom sorting later.  Thanks :)

PR #2336, opened 2021-04-16
Pleasure Panties Password Fail fix

Nina-1474:
Interacting with the Sci-Fi Pleasure Panties when they're locked with an unopenable lock shows the usual futuristic item's "Enter Password" screen. However clicking 'Log in' and exiting this screen was not removing the input field.


Ben987:
Yes!  Thanks for the fix :)

PR #2337, opened 2021-04-17
Feature: Typed item archetype enhancements & Latex Corset refactor

elliesec:
## Summary

This PR makes several enhancements to the typed item archetype introduced in #2263, and migrates the `Corset` and `ItemTorso` versions of the Latex Corset to use the new framework, removing the need for individual asset scripts.

A full list of changes:
* Fixes and improvements to JSDoc in `TypedItem.js`
* Adds support to `TypedItem.js` for the `ShowImages` argument that can be provided to the extended item script functions, and updates the asset check script to permit the new configuration property
* Removes the `IsCloth` argument from extended item script functions. This is now instead read from `DialogFocusItem.Asset.Group.Clothing`
* Allows the existence of `Config` and `CopyConfig` simultaneously on item configuration in `Female3DCGExtended.js`. Configuration in the `Config` property will override any configuration obtained by the `CopyConfig` property. This uses a shallow object assignment right now - I may extend this functionality to enable deep copying later
* Modifies `ExtendedItemDraw` to use an asset's `DynamicGroupName` when drawing a typed item's type preview images to remove the need for duplicated assets
* Migrates the Latex Corset in both the `Corset` and `ItemTorso` groups to the archetype framework and removes the redundant scripts and assets, and adds chat messages to the `ItemTorso` version of the item when its type is changed


Ben987:
That's a big one, I won't be able to test all the items, but we have betas for that.   Thanks :)

PR #2340, opened 2021-04-18
Movie Studio Corset fix

Nina-1474:
The "Wear the latex corset" option in the movie wasn't working because that item moved groups from Bra to Corset.


Ben987:
Yes!  Thanks for the fix :)

PR #2341, opened 2021-04-18
Bugfix: Lock validation fixes

elliesec:
## Summary

This fixes a couple of things:
* There was an error in lock modification validation where the `RemoveItem` and `ShowTimer` lock properties were being reverted even if the player had permission to modify the lock. This resulted in some of the timer lock buttons not working properly
* When locks were removed via `InventoryUnlock` and `TimerInventoryRemove`, the `EnableRandomInput`, `RemoveItem` and `ShowTimer` properties weren't being removed, which resulted in the validation code flagging the update as invalid. These two functions now make calls to `ValidationDeleteLock`, to ensure that all lock properties are being properly deleted.


Ben987:
Thanks for the fix!  I will prepare a quick hotfix.

PR #2342, opened 2021-04-20
Canvas Draw Hooks + Virtual Avatars for VR Headset

Ada18980:
Adds a new option for the VR Headset: a virtual avatar mode. In this mode, the player can dress herself (if not unable to from other sources), and all restraint items are hidden. This required some modification of appearance functions, but now there are new hooks that are general use and can be used for other items:

BeforeSortLayers: Happens before layer sorted. Used to modify the appearance array before rendering, good if you want to make an item not appear when rendering

AfterLoadCanvas: Runs after sorting the canvas. Good for hiding individual layers, but does not alter the effects of HideItem, etc.


Ben987:
Is there another way to do that beside changing the core appearance functions?  I'm not a big fan of adding more complexity to the base code for such a niche change.  I've received a few complaints in the last months from previous asset makers telling me that the code is now too complex for them to make new stuff.  Thanks!

elliesec:
The suggestion of using hooks in the core game functions was largely intended to prevent special-casing in those functions to cater for niche use cases. I'm less sure about pulling apart `C` and `Appearance` in the appearance functions - how necessary is this? The idea with hooks was so that we wouldn't need to separate the two in this way. Perhaps the hooks idea in general needs a bit more thought around its design and implementation (or maybe just its usage) in order to keep core changes minimal.

On a broader point, I do think that what Ben mentioned is important though. Little has really changed about the core of creating new assets, but a lot of features have been added that should, _in theory_, make it easier, and also provide asset creators with powerful tools to do more than was possible before. However, there are a lot more attributes that you can put on an asset nowadays, but little explanation of how they all work. I think practically, the best solution is to introduce some solid documentation on asset creation (which is something I've been intending on adding to the wiki for a while now) which explains in detail how to use all of the options. All of those features are there, and are easy to use for creators that know how, but right now you pretty much need to be able to understand the code in order know how to use them.

Ada18980:
As @elliesec said, the whole point of the way the hooks were made was to do this in a way that minimizes special casing and allows for things to be done in the future with the same code. For example, 


The reason I needed to add TempAppearance was because having a gag in real life would csuse the mouth to be hidden in VR. I could either use a temp appearance array, or I could modify the core draw function to somehow not count the Hide property, but I wanted to do it in a way that minimally affects the drawing function. So thats why I went with TempAppearance.

Other things that will be possible with the hooks system include:
-Boxes that hide the player to other players but allow the player to see herself
-Appearance modification based on conditions in the future, like adding rope marks without adding a new asset for rope marks

I can probably avoid using TempAppearance if I hijack the system for hiding items, but I feel like that will be more of a special case for this item.

This will also allow cutscenes in the future to modify the player as appearance without changing their clothes

Ada18980:
@Ben987 if it helps, I can add comments detailing how to use a hook and how it works

Ada18980:
@Ben987 @elliesec I edited it to align the use of DrawPose and appearance modification so they both use the same method (variable stored in C which is modified by hooks).

The changes in the drawing functions are now minimal as they only change the variable being pointed to (DrawAppearance instead of Appearance) which is updated every time the canvas is reloaded.

BondageClub/Scripts/Character.js
```diff
@@ -689,9 +722,28 @@ function CharacterLoadCanvas(C) {
 	// Reset the property that tracks if wearing a hidden item
 	C.HasHiddenItems = false;
 
+	// We add a temporary appearance and pose here so that it can be modified by hooks.  We copy the arrays so no hooks can alter the reference accidentally
+	C.DrawAppearance = AppearanceItemParse( CharacterAppearanceStringify(C))
+	C.DrawPose = [...C.Pose] // Deep copy of pose array
```
elliesec, line 727: Just so you're aware, we've had browser compatibility issues reported previously with using spread syntax in arrays (some players have _very_ outdated browsers). It would probably be preferable to use something like `C.Pose.slice()` or a plain old loop to copy.

jomshir98: I don't think compatibility is problem with this exact feature. It is supported on every browser with at least 0.5% usage since 2017, except Internet Explorer:
[Can I use: Spread in array literals](https://caniuse.com/mdn-javascript_operators_spread_spread_in_arrays)

elliesec: I realise that. Like I said, I included array spread in a PR maybe about 6-7 months back, and it caused issues for some users.

elliesec: Ah, actually I'm misremembering. It was array destructuring that was causing the issues. Spread is a few browser versions before, so might be alright.

Ben987:
It's an interesting debate.  By complexifying the code, we might be alienating asset makers and story tellers that wants to contribute to the game but don't have lots of programming knowledge.  I myself would probably not contribute to a project if I cannot understand the code easily.  Maybe there is some kind of line we should draw so we don't lose too many contributors.  I'll start a little discussion on that one.

Ada18980:
@Ben987 judging from the discussion it seems the consensus is that the code isn't too complex, the problem is the lack of documentation. If I add documentation to this particular PR, will you give it the green light or will it be a waste of my time? Thanks.

Ben987:
Alright!  If you both say it works great, let's test this one then :)

Yes, if you could add some documentation to the code it would be great.  Seeing new unknown technologies added to code is always scary for an old dinosaur like me.  Cheers!

PR #2343, opened 2021-04-20
Background: Hotel Room

Codefreak64:
With all the work that everyone has put in, it would be nice to take time away in a hotel.
Now you can with this background

-Adds new Indoor background
-Not sure how else to hype this up, its a background that sets up a 'luxury' mood


Ben987:
Absolutely!  It's a nice idea.  Thanks :)

PR #2344, opened 2021-04-21
Added stepping sexual activity

Manilla32:
Target has to be kneeling for it to work.


Ben987:
Sure!  That's a fun idea :)

PR #2345, opened 2021-04-21
Bugfix: Fix for tongue strap gag lock asset colour

elliesec:
## Summary

This fixes an issue with the tongue strap gag where adding a lock would remove any asset colour from the gag. This happened because the lock asset contained both the full gag, and the lock. So locking it would draw the non-coloured gag asset (which was part of the lock asset) over the coloured one. This PR edits the lock asset for the item to cut out the gag parts, leaving only the actual lock graphics, which should fix the issue.

See [this Discord link](https://discord.com/channels/554377975714414605/554378725916147722/834480545525661727) for an example.


Ben987:
Very well!  Thanks for that fix :)

PR #2346, opened 2021-04-22
Fix: Crash during beep log rendering

jomshir98:
Fixes crash, that happens when you open beep log from chat search screen, while you have received beep from someone, who isn't in room.


Ben987:
Yes!  Thanks for the fix :)

PR #811, opened 2020-04-30
Expression fix

ADRCorp:
Expression no longer get reset upon WardrobeFastLoad

PR #2347, opened 2021-04-23
Bugfix: Prevents expressions from being saved to the wardrobe

elliesec:
## Summary

As of #2015, item properties are saved to the player's wardrobe. However, this overlooked the fact that player expressions would also be saved to the wardrobe (as they are stored in an item property). This contradicts the functionality introduced in #811 whereby player expressions would be preserved on loading from the wardrobe. The result is that any player expressions that were explicitly set (to a string, i.e. not `null` or `undefined`) would be preserved on appearance load, but any player expressions that weren't would take the value saved in the wardrobe, which is something of the worst of both worlds.

To address this, I've taken the approach of not saving expressions to the player wardrobe (and ignoring any existing saved expressions when loading appearance from the wardrobe). This means that all item expressions on the player will be the same after loading an appearance as they were before loading, rather than a hybrid of the player's expression before loading and the expression saved in the wardrobe. I feel this is probably the behaviour that most players want (expressions are naturally more transient than clothing, and it doesn't seem logical to save an expression as part of an outfit, as the primary purpose of the wardrobe is for storing clothing outfits).


Ben987:
Absolutely!  It doesn't make much sense to save the expressions.

PR #2348, opened 2021-04-23
Lock Key-List change fix

Nina-1474:
MemberNumberListKeys (used for High Security Padlocks) was listed as a basic lock property meaning it couldn't be changed after being initially set, which should actually be possible.


elliesec:
Only issue with this is that the property won't get removed from the item when the lock is removed. It's kind of annoying, but for now, it's probably best to add a `.concat(["MemberNumberListKeys"])` onto the end of the `ValidationAllLockProperties` definition. On the whole, the handling of locks here is something I'd like to improve in the future. Ultimately, I'd probably like to get to a place where a lock can declare its own validation function, and the validation code can delegate out to that.

Ben987:
Sure!  Thanks for that fix.  Cheers :)

PR #2349, opened 2021-04-23
Elf Ear Position Options

Nina-1474:
Extending the elf ears using the Typed archetype so that they can appear behind HairFront instead of in front of it.


Ben987:
Sure!  That's a good idea, thanks a lot :)

PR #2351, opened 2021-04-24
Bugfix/Improvement: Menu button permission clarifications

elliesec:
## Summary

This is a follow-up bugfix to #2186. It changes appearance/dialog menu buttons so that they render differently depending on the permission level of the item, and whether or not the player can modify/interact it. When a button is disabled, it will be greyed out, and have a tooltip explaining that this is due to player permissions. This also fixes some issues where existing menu buttons would either not work due to limited item permissions, or would work but cause a validation rollback. It also includes some bugfixes to some of the validation logic around item & lock modification.

Menu buttons modified:
* **Use** button - players may only use (i.e. extend) an item if they would be able to add it. In particular, this means that players cannot use (for example) shock functionality or change the types on an extended item if they do not have permissions.
* **Remote** button - same applies as above
* **Inspect lock** button - Again, the same applies as above. Eventually, I'd like to allow locks to have a "read-only" view that permits them to be extended but not modified by players that do not have permission
* **Color picker** button - same as above. Modifying a limited item's colour (without having limited item access) would previously cause a validation rollback.
* **Lock** button - same as above. Adding a lock to a limited item is only permitted if you have limited item access on the player

I've discussed the above behaviours with @Nina-1474 and @jomshir98, and we feel this is the best approach to item permissions for now. Note that in particular, it is still possible to _remove_ limited items and locks, but in general, not to add, modify or otherwise interact with them.


Ben987:
This one creates a conflict in Dialog.js   Can you resolve it and test again?  Thanks.

elliesec:
Conflicts have been resolved.

Ben987:
Thanks!  It should make the menus a little easier to use.  Cheers :)

PR #2352, opened 2021-04-25
Bugfix: Fixes a minor graphical issue with the pearl necklace

elliesec:
## Summary

There was a small graphical artifact in the pearl necklace that hadn't been erased properly which shows up on a character's chin (see image from Discord report below).

image


Ben987:
Sure!  Thanks for that fix :)

PR #2355, opened 2021-04-25
Bugfix: Fixes an owner/lover lock validation issue

elliesec:
## Summary

Fixes an issue which allowed players to incorrectly remove owner/lover locks from others. I'm not quite sure how this slipped through the net when I was testing previously, as I'm pretty sure I had a test case for this situation, but apparently it did.


Ben987:

PR #2357, opened 2021-04-27
Locking vibrating butt plug

Manilla32:
In short, a locking butt plug that can vibrate.


Ada18980:
Why is there a change to ballet heels? Also I think you might have carried over some changes from the giggle PR. Generally those should be kept separate unless there is a need to (I keep doing it accidentally myself)

Manilla32:
> Why is there a change to ballet heels? Also I think you might have carried over some changes from the giggle PR. Generally those should be kept separate unless there is a need to (I keep doing it accidentally myself)

I must have accidentally done something to the ballet heels file without noticing. I'll make a new PR for this soon.

PR #2360, opened 2021-04-28
clit ring

Manilla32:
Added a clit ring that can have a leash attached to it.


Ben987:
Alright!  Thanks for that one.  Cheers :)

PR #2361, opened 2021-04-28
Fix: CharacterLoadCanvas Lint

jomshir98:
Format function `CharacterLoadCanvas` changed by #2342 to fix linting error


Ben987:
Thanks for that fix!

PR #2362, opened 2021-04-28
Locking vibrating Butt plug

Manilla32:
Adding a vibrating butt plug that can be locked.


jomshir98:
Do not mind the failing Lint check, @Manilla32 
The errors it found aren't in your change and will be fixed by #2361 .
As far as I can see, your change looks good!

Ben987:

PR #2363, opened 2021-04-28
Immersion: Chat room names and descriptions are garbled while blind

Ada18980:
Adds a new immersion option, allowing chat room names and descriptions (plus the names of friends in them) to be garbled.

This required a refactoring of SpeechGarble. I have split SpeechGarble into two functions, one containing the base functionality based on a level of GagEffect, and one that assigns GagEffect based on either deafness (if the character is another player) or gagged level of the player.

### DIFF
--- a/BondageClub/Screens/Character/FriendList/FriendList.js
+++ b/BondageClub/Screens/Character/FriendList/FriendList.js
@@ -192,6 +192,9 @@ function FriendListChatSearch(room) {
 	FriendListExit();
 	ElementValue("InputSearch", room);
 	ChatSearchQuery();
+	// Change the text box so the player still cant read it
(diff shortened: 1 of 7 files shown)


Ben987:
It might get confusing for players but it's a fun idea.  You don't force that immersion setting for extreme or hardcore players?

Ada18980:
> It might get confusing for players but it's a fun idea. You don't force that immersion setting for extreme or hardcore players?

It's an immersion setting, so it is forced on for Extreme but the default is off.

It probably will be confusing for extreme players at first, but it won't be forced on hardcore players unless they switch it on.

Ben987:
It's very extreme as a setting, but I guess Extreme players want that kind of challenge and immersion.  We'll see how it goes.  Thanks :)

PR #2364, opened 2021-04-28
Notifications Improvements

Nina-1474:
- In the chat messages notification settings players can now separately select Normal Messages, Whispers and Items/Activities to be included or excluded.
- If a popup notification is expected but isn't available for some reason, the "(1) in title" notification will now occur in its place. This covers situations like a user switching from a device that supports popups to one that doesn't.
- If the user was prompted by their browser to allow or block notifications, the particular popup that triggered this request would not appear. Now it will appear after the permissions request is completed (if the user selected Allow).
- Added steps to the popup's onclick to ensure that it refocuses on the game on all OS's.


Ben987:
That's a big one!  Thanks a lot, cheers :)

PR #2366, opened 2021-04-29
Cleanup: Scripts formatting (Part 1 - little edits)

jomshir98:
- Changes missed `var` to `let` in `SpeechGarbleByGagLevel`

To keep this cleanup small, this part only focuses on files with less than 20 edits each.


Ben987:

PR #2367, opened 2021-04-29
Cleanup: Scripts formatting (Part 2 - `Scripts/Animation.js`)

jomshir98:

To keep this cleanup small, this part only focuses on file `Scripts/Animation.js`


Ben987:

PR #2368, opened 2021-04-29
Cleanup: Scripts formatting (Part 3 - `Scripts/Character.js`)

jomshir98:
- Formats functions inside `CharacterReset` to be more readable without changing them

To keep this cleanup small, this part only focuses on file `Scripts/Character.js`


Ben987:

PR #2372, opened 2021-04-29
Cleanup: Scripts formatting (Part 7 - `Screens/Inventory/*` - little changes)

jomshir98:

To keep this cleanup small, this part only focuses on files in `Screens/Inventory/` and only on files with at most 5 edits each


Ben987:

PR #2374, opened 2021-04-29
Cleanup: Scripts formatting (Part 9 - `Screens/Inventory/*` - rest)

jomshir98:

To keep this cleanup small, this part only focuses on files in `Screens/Inventory/` and only on files not handled before


Ben987:

PR #2375, opened 2021-04-29
New item: Short pencil skirt

SandrineBC:
# Description
Anna (graphics) and I added a new shorter version of the pencil skirt.
image
# Testing
Testing was done on my local instance with the various body sizes


Ben987:
It looks great.  Thanks a lot you two :)

PR #2376, opened 2021-04-29
Bug fix: Adding missing tag

SandrineBC:
# Description
Just added a missing tag for ZipArmsSetZipWrist


Ben987:
Yes!  Thanks for the fix :)

PR #2378, opened 2021-04-29
Cleanup: Scripts formatting (Part 10 - Scripts & Assets indentation)

jomshir98:
- Unifies indentation across files to use the most used one - 1 tab per indent

Note that 99% of changes in this PR is 4 spaces replaced by 1 tab
To keep this cleanup small, this part ignores files in `Screens`


Ben987:
Sure!  Thanks for that.

PR #2379, opened 2021-04-29
Refactor: Extended Vibrator Script for the Sybian

SandrineBC:
# Description
Does exactly that. The Sybian now uses the extended item script. Don't know, why I missed it the last time

# Testing
I tested that on my local version of the Club and everything worked fine.


Ben987:
Absolutely!  Thanks :)

PR #2381, opened 2021-04-30
Fix: automaticly remove invalid skills and reputations

fleisch11:
it works the second time you load the informationsheet.
i am not sure, whether or not i should make this a button, the player can use manually.


Nina-1474:
I don't think we can depend on finding text entries for this check. Sometimes text just doesn't load for players, particularly after a new release. Deleting their skills/reputations if that happens would not be good.

elliesec:
I agree with Nina. Sometimes the missing text entries are caused by players being on the wrong version link, or just an internet hiccup meaning that a CSV doesn't load.

I think the best solution for this would be to refactor the valid skills/reputations into enums and loop through those to remove invalid skills/reps (and also add some validation to the skill/reputation update functions to prevent invalid values being added in the first place).

PR #2384, opened 2021-04-30
Icon Notification Option

Nina-1474:
This adds a new alert option for notifications: "(1) in icon"
It draws a red circle onto the icon in the document/window header containing the incrimenting number of notifications, similar to the "(1) in title" option.


Ben987:
Sure!  That should be useful.  Thanks :)

PR #2385, opened 2021-04-30
Feature: BlindFlash

fleisch11:
makes it hard to see after removing a blindfold.
this can be turned off in the preferences.

### DIFF
--- a/BondageClub/Screens/Character/Preference/Text_Preference.csv
+++ b/BondageClub/Screens/Character/Preference/Text_Preference.csv
@@ -120,7 +120,7 @@ SensDepLight,Light
 SensDepNames,Hide names
 SensDepTotal,Heavy
 SensDepExtreme,Total
-DoBlindFlash,Do BlindFlash
+DoBlindFlash,Flash after blindness
 AllowPlayerLeashing,"Players can drag you to rooms when leashed"
 DisableAutoMaid,Maids come automatically when restrained
 DisablePickingLocksOnSelf,Locks on you can't be picked
--- a/BondageClub/Screens/Character/Preference/Text_Preference.csv
+++ b/BondageClub/Screens/Character/Preference/Text_Preference.csv
@@ -120,6 +120,7 @@ SensDepLight,Light
 SensDepNames,Hide names
 SensDepTotal,Heavy
 SensDepExtreme,Total
+DoBlindFlash,Do BlindFlash
 AllowPlayerLeashing,"Players can drag you to rooms when leashed"
 DisableAutoMaid,Maids come automatically when restrained
 DisablePickingLocksOnSelf,Locks on you can't be picked
--- a/BondageClub/Scripts/Inventory.js
+++ b/BondageClub/Scripts/Inventory.js
@@ -507,7 +507,7 @@ function InventoryRemove(C, AssetGroup, Refresh) {
 	}
 	catch (e) {
 		if (e instanceof TypeError) {
-			console.log("idk how to do nothing")
+			console.log("i have to put something here because of eslint");
 		}
 	}
 
--- a/BondageClub/Scripts/Inventory.js
+++ b/BondageClub/Scripts/Inventory.js
@@ -499,16 +499,10 @@ function InventoryGetRandom(C, GroupName, AllowedAssets) {
 * @param {false} [Refresh] - do not call CharacterRefresh if false
 */
 function InventoryRemove(C, AssetGroup, Refresh) {
-	try {
-		if (Player.GraphicsSettings.DoBlindFlash) {
-			var lastblindlevel = Player.GetBlindLevel();
-			lastdarkfactor = CharacterGetDarkFactor(Player);
-		}
-	}
-	catch (e) {
-		if (e instanceof TypeError) {
-			console.log("i have to put something here because of lint");
-		}
+	
+	if (Player.GraphicsSettings && Player.GraphicsSettings.DoBlindFlash) {
+		var lastblindlevel = Player.GetBlindLevel();
+		lastdarkfactor = CharacterGetDarkFactor(Player);
 	}
 
 
@@ -543,16 +537,10 @@ function InventoryRemove(C, AssetGroup, Refresh) {
 		if (C.Appearance[E].Asset.Group.Name == AssetGroup) {
 			C.Appearance.splice(E, 1);
 			if (Refresh || Refresh == null) CharacterRefresh(C);
-			try {
-				if (Player.GraphicsSettings.DoBlindFlash) {
-					if (Player.GetBlindLevel() < lastblindlevel && Player.GetBlindLevel() == 0) {
-						DrawBlindFlash(lastblindlevel);
-					}
-				}
-			}
-			catch (e) {
-				if (e instanceof TypeError) {
-					return;
+
+			if (Player.GraphicsSettings && Player.GraphicsSettings.DoBlindFlash) {
+				if (Player.GetBlindLevel() < lastblindlevel && Player.GetBlindLevel() == 0) {
+					DrawBlindFlash(lastblindlevel);
 				}
 			}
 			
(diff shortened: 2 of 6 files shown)


BondageClub/Scripts/Inventory.js
```diff
@@ -498,6 +500,12 @@ function InventoryGetRandom(C, GroupName, AllowedAssets) {
 */
 function InventoryRemove(C, AssetGroup, Refresh) {
 
+	if (Player.ImmersionSettings.DoBlindFlash == true && Player.ImmersionSettings.DoBlindFlash !== undefined) {
```
Nina-1474, line 503: This crashes the game on login. Whenever you're changing a function that provides a `C` Character parameter you will most of the time want to use that in place of `Player`. And if the changes should only ever happen for the player, including `C.ID == 0` in the `if` check will accomplish that.

Nina-1474 (changes requested):

BondageClub/Scripts/Inventory.js
Nina-1474: Also the check for true and not undefined can be simplified: `if (something && C.ImmersionSettings.DoBlindFlash) { ... }`

BondageClub/Screens/Character/Preference/Text_Preference.csv
```diff
@@ -120,6 +120,7 @@ SensDepLight,Light
 SensDepNames,Hide names
 SensDepTotal,Heavy
 SensDepExtreme,Total
+DoBlindFlash,Do BlindFlash
```
Nina-1474, line 123: "Do BlindFlash" doesn't explain the setting very clearly. Maybe something more like "Flash after blindness" could be clearer. I appreciate that space is limited at the moment. I'm going to implement a second page for immersion settings after this (or before if you'd prefer to wait).

Ada18980:
I think this ought to respect epilepsy by being toggleable for all players in the graphics settings. It's not right to lock extreme mode for disabled users. And it's not really that graphical or immersive so Im not sure it can afford a space in the immersion settings page.

Also I think this would be better implemented as a stimulation message (the world feels blurred to you as the blindfold comes off) with a slight change to the stimulation message system to allow different colors of screen flash

fleisch11:
about Nina's comments:
allright, i'll go ahead and fix it. it crashed for me too, but i though, it was because my local files didn't update correctly, since it worked the second try, but that obviously is not the case.
also i'll rename the settings thing.

about Ada's comment:
serious question; i'm no expert, but wouldn't the pink arousal flash do the same thing?
anyway, i'll definetly change it, since the main reason i did it this way anyway, is because i just copied code that behaved that way, so i had no reason for that anyway.

fleisch11:
i am aware, that my latest commit tdidnt fix nina's request. working on it

fleisch11:
allright, nina's requests are done now.
i will now work on ada's requests

fleisch11:
allright, all requested changes are done

BondageClub/Scripts/Inventory.js
```diff
@@ -497,6 +499,19 @@ function InventoryGetRandom(C, GroupName, AllowedAssets) {
...
+			lastdarkfactor = CharacterGetDarkFactor(Player);
+		}
+	}
+	catch (e) {
+		if (e instanceof TypeError) {
+			console.log("i have to put something here because of lint");
+		}
+	}
+
+
```
jomshir98, line 514: You **don't** want to be catching `null` property access with try/catch!!! EVER!!
You instead want to check that something isn't `null` in the first place!

And `console.log("i have to put something here because of lint");` Really?...
If you really need an empty block somewhere, putting a comment into it is enough to prevent the check from triggering!
But there is a reason the check is there, because it 99% of the time shows, that this is something that doesn't have to be there at all!
For example here, if you remove the `if`, there would be no change to the functionality at all, because no matter if `e` is or isn't `TypeError` the catch wouldn't do anything with it.

In case you wanted to not catch any error except `TypeError` you would do this:
```js
if (!e instanceof TypeError) {
	throw e;
}
```

BondageClub/Scripts/Inventory.js
```diff
@@ -528,6 +543,19 @@ function InventoryRemove(C, AssetGroup, Refresh) {
...
+						DrawBlindFlash(lastblindlevel);
+					}
+				}
+			}
+			catch (e) {
+				if (e instanceof TypeError) {
+					return;
+				}
+			}
+
```
jomshir98, line 558: Same as above! You should never be using try/catch to avoid `null` property access error!

BondageClub/Scripts/Timer.js
```diff
@@ -241,7 +241,13 @@ function TimerProcess(Timestamp) {
...
 	}
 
-	// Launches the main again for the next frame
+	if (BlindFlash == true && CurrentTime < DrawingBlindFlashTimer) {
+		if (Player.GetBlindLevel() == 0) {
+			DrawRect(0, 0, 2000, 1000, "rgba(255,255,255," + (1 - lastdarkfactor - 0.1) + ")");
+		}
+	}
+
+    // Launches the main again for the next frame
```
jomshir98, line 250:
```suggestion
	// Launches the main again for the next frame
```

jomshir98 (changes requested):
Do not use try/catch for `null` access errors! Instead check that the thing isn't `null`!

BondageClub/Scripts/Inventory.js
fleisch11: allright, working on both

BondageClub/Scripts/Inventory.js
fleisch11: ok, i didn't know that. also what i was trying to do, was to not do anything, if it results in an error

jomshir98:
Also I apologise for sounding so aggressive, @fleisch11 .
I had very bad experience in past with try/catch masking some nasty bugs...

The way you do want to do it is:
```js
if (Player.GraphicsSettings && Player.GraphicsSettings.DoBlindFlash)
```

fleisch11:
@jomshir98  don't worry about it and thanks for the help

fleisch11:
allright, jomshir's requests are done now

BondageClub/Scripts/Inventory.js
```diff
@@ -497,6 +499,13 @@ function InventoryGetRandom(C, GroupName, AllowedAssets) {
 * @param {false} [Refresh] - do not call CharacterRefresh if false
 */
 function InventoryRemove(C, AssetGroup, Refresh) {
+	
+	if (Player.GraphicsSettings && Player.GraphicsSettings.DoBlindFlash) {
+		var lastblindlevel = Player.GetBlindLevel();
```
NoneNoname, line 504: using `var` to leak out variable out of scope is bad practice, create the variable before the if and give it value inside the if, though it should be fine without any if since it will be tested later:
`const lastblindlevel  = Player.GetBlindLevel();`

note: `let` & `const` won't let you leak out of scope thereby you can easily spot this issue if you use them

NoneNoname (changes requested):

fleisch11:
@NoneNoname allright, done

BondageClub/Scripts/Inventory.js
```diff
@@ -528,6 +533,13 @@ function InventoryRemove(C, AssetGroup, Refresh) {
 		if (C.Appearance[E].Asset.Group.Name == AssetGroup) {
 			C.Appearance.splice(E, 1);
 			if (Refresh || Refresh == null) CharacterRefresh(C);
+
+			if (Player.GraphicsSettings && Player.GraphicsSettings.DoBlindFlash) {
+				if (Player.GetBlindLevel() < lastblindlevel && Player.GetBlindLevel() == 0) {
```
elliesec, line 538:
```suggestion
				if (lastblindlevel > 0 && Player.GetBlindLevel() === 0) {
```
Just a minor one here, but `Player.GetBlindLevel` actually encapsulates a fair bit of logic, so it's probably worth just calling it once here.

elliesec (approved):
Looks good to me. Checked out and tested locally.

Nina-1474 (approved):

Ben987:
Can it be unhealthy for some players to have the screen turn white suddenly?  If anyone thinks it could trigger epilepsy, we should disable this option by default.

PR #2386, opened 2021-05-01
Fix: allows one to login with invalid skills/reps

fleisch11:
it also makes invalid skills not show up at all, **but they never get deleted**

### DIFF
--- a/BondageClub/Scripts/Reputation.js
+++ b/BondageClub/Scripts/Reputation.js
@@ -41,9 +41,10 @@ function ReputationChange(RepType, RepValue, Push) {
 	}
 
 	else {
-		throw new Error("trying to change invalid reputation")
+		if (CurrentScreen != "Login") {
+			throw new Error("trying to change invalid reputation")
+		}
 	}
-
 }
 
 /**
--- a/BondageClub/Scripts/Reputation.js
+++ b/BondageClub/Scripts/Reputation.js
@@ -41,9 +41,7 @@ function ReputationChange(RepType, RepValue, Push) {
 	}
 
 	else {
-		if (CurrentScreen != "Login") {
-			console.warn(`Invalid reputation type "${RepType}"`);
-		}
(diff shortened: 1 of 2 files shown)


BondageClub/Scripts/Reputation.js
```diff
@@ -41,9 +41,10 @@ function ReputationChange(RepType, RepValue, Push) {
 	}
 
 	else {
-		throw new Error("trying to change invalid reputation")
+		if (CurrentScreen != "Login") {
+			throw new Error("trying to change invalid reputation")
```
elliesec, line 45:
```suggestion
			console.warn(`Invalid reputation type "${RepType}"`);
```
I didn't have a proper look at the previous PR before it was merged, so didn't notice this then, but I feel like throwing errors here is perhaps a little harsh. I don't feel that using an invalid reputation type should be game-breaking, but we should definitely be warning about it. 

I'd suggest maybe switching them to use a `console.warn`. Including the invalid type in the message might also be useful information for future devs, particularly if they want to add new reputation/skill types.

BondageClub/Scripts/Skill.js
```diff
@@ -45,7 +45,9 @@ function SkillChange(SkillType, SkillLevel, SkillProgress, Push) {
 	}
 
 	else {
-		throw new Error("trying to change invalid skill");
+		if (CurrentScreen != "Login") {
+			throw new Error("trying to change invalid skill");
```
elliesec, line 49:
```suggestion
			console.warn(`Invalid skill type "${SkillType}"`);
```
Same comment here as in `Reputation.js`.

fleisch11:
@elliesec allright, i will remove the login stuff too, once the repostery has downloaded

fleisch11:
@elliesec alright, everything is done now

elliesec (approved):

Ben987:
Absolutely!  Thanks.

PR #2387, opened 2021-05-01
Optimised Garble with switch statements + split functions.

Shiranui-Izayoi:
Did any tests I can think of and all seems to be working right. Hope this can get in before the beta comes out.

I swapped over the if statements to switch statements to make it more efficient. Also split each level into it's own function for easier editing. Doing the same for the baby talk function. I was considering adding a way to manually trigger baby talk without the item but felt that would need it's own pull request.

I THINK I managed to include everything that was initally talked about. When I tried submitting this before. Sorry for the long wait. I started it and had most of it written up, but then just bombed at the testing phase. 

Should work as it normally does but better. Also made it so that caps aren't instantly deleted when you are gagged. As I did before, though the higher levels those being 8+ aka the total gag levels do not have caps as it doesn't fit at that level.

@elliesec can you have a look over it to make sure I didn't miss anything that you mentioned last time?

**EDIT:** Ow yer also added some extra stuff to the baby talk to remove cases of thh shh etc since they don't really make sense. Due to already having that sound there. Adding the extra h makes it harder to read and look strange.

**EDIT2:** It seems that GitHub doesn't like the use of multiple case variables. But I am sure that this is the correct way to do this meaning if it is any of them it does it. So it is likely that these checks should be ignored.

### DIFF
--- a/BondageClub/Scripts/Speech.js
+++ b/BondageClub/Scripts/Speech.js
@@ -434,7 +434,7 @@ function SpeechGarble6(H) {
 		case "ж": case "л": case "р":
 			H = "а";
 			break;
-		case "з": case "с": case "й": 
+		case "з": case "й": 
 			H = "г";
 			break;
 		case "б": case "в": case "ы": 
@@ -452,7 +452,7 @@ function SpeechGarble6(H) {
 		case "Ж": case "Л": case "Р":
 			H = "А";
 			break;
-		case "З": case "С": case "Й": 
+		case "З": case "Й": 
 			H = "Г";
 			break;
 		case "Б": case "В": case "Ы": 
@@ -679,20 +679,14 @@ function SpeechGarble4(H) {
 		case "г": case "к": case "х": 
 			H = "к";
 			break;
-		case "в": case "у": case "ж": case "л": case "р": 
+		case "у": case "ж": case "л": case "р": 
 			H = "а";
 			break;
 		case "с": case "я": 
 			H = "х";
 			break;
-		case "д": case "ф": 
+		case "д":
 			H = "м";
-			break;
-		case "р": 
-			H = "ф";
-			break;
-		case "г": 
-			H = "н";
 			break;
 		case "В": case "Ф": case "Б": case "П": 
 			H = "ФЫ";
@@ -706,14 +700,8 @@ function SpeechGarble4(H) {
 		case "С": case "Я": 
 			H = "Х";
 			break;
-		case "Д": case "Ф": 
+		case "Д":
 			H = "М";
-			break;
-		case "Р": 
-			H = "Ф";
-			break;
-		case "Г": 
-			H = "Н";
 			break;
 	}
 	return H;
@@ -824,20 +812,14 @@ function SpeechGarble3(H) {
 		case "г": case "к": case "х": 
 			H = "к";
 			break;
-		case "в": case "у": case "ж": case "л": case "р": 
+		case "у": case "ж": case "л": case "р": 
 			H = "а";
 			break;
 		case "с": case "я": 
 			H = "х";
 			break;
-		case "д": case "ф": 
+		case "д":
 			H = "м";
-			break;
-		case "р": 
-			H = "ф";
-			break;
-		case "г": 
-			H = "н";
 			break;
 		case "В": case "Ф": case "Б": case "П": 
 			H = "ФЫ";
@@ -845,20 +827,14 @@ function SpeechGarble3(H) {
 		case "Г": case "К": case "Х": 
 			H = "К";
 			break;
-		case "В": case "У": case "Ж": case "Л": case "Р": 
+		case "У": case "Ж": case "Л": case "Р": 
 			H = "А";
 			break;
 		case "С": case "Я": 
 			H = "Х";
 			break;
-		case "Д": case "Ф": 
+		case "Д":
 			H = "М";
-			break;
-		case "Р": 
-			H = "Ф";
-			break;
-		case "Г": 
-			H = "Н";
 			break;
 	}
 	return H;
--- a/BondageClub/Scripts/Speech.js
+++ b/BondageClub/Scripts/Speech.js
@@ -106,11 +106,8 @@ function SpeechGarble(C, CD, NoDeaf) {
 	// The usual preamble - calculate the gag effect level, and the character's deafness level once
 	GagEffect = SpeechGetGagEffect(C, NoDeaf);
 
-	// If GagEffect is 0 then there is no need to run this part so, a simple if statement is used in order to check.
-	if (GagEffect > 0) {
+	NS = SpeechGarbleByGagLevel(GagEffect, CD);
 
-		NS = SpeechGarbleByGagLevel(GagEffect, CD);
-	}
 	NS = SpeechStutter(C, NS);
 	NS = SpeechBabyTalk(C, NS);
 
@@ -124,7 +121,7 @@ function SpeechGarble(C, CD, NoDeaf) {
  * @returns {string} - Returns the correct garbling function to run.
  */
 function SpeechGetGarblingFunction(GagEffect) {
-	
+
 	// Makes sure that the gag level is within acceptable parameters. As a GagEffect of more than 20 does nothing, but for this function above 20 will break it.
 	if (GagEffect > 20) GagEffect = 20;
 
@@ -169,6 +166,8 @@ function SpeechGetGarblingFunction(GagEffect) {
  * @param {string} CD - The character's dialog to alter
  */
 function SpeechGarbleByGagLevel(GagEffect, CD, IgnoreOOC) {
+
+	// If GagEffect is 0 then there is no need to run this part so, a simple if statement is used in order to check and prevent errors.
 	if (GagEffect > 0) {
 		const garblingFunction = SpeechGetGarblingFunction(GagEffect);
 		return SpeechGarbleLine(CD, garblingFunction, IgnoreOOC);
@@ -433,7 +432,7 @@ function SpeechGarble7(H) {
 			H = "";
 			break;
 
-		// Regular characters
+			// Regular characters
 		case "a":
 		case "e":
 		case "i":
@@ -627,7 +626,7 @@ function SpeechGarble7(H) {
  */
 function SpeechGarble6(H) {
 	switch (H) {
-		
+
 		case 9:
 		case 8:
 		case 7:
@@ -641,7 +640,7 @@ function SpeechGarble6(H) {
 			H = "";
 			break;
 
-		// Regular characters
+			// Regular characters
 		case "a":
 		case "e":
 		case "i":
@@ -847,7 +846,7 @@ function SpeechGarble6(H) {
  */
 function SpeechGarble5(H) {
 	switch (H) {
-		
+
 		case 9:
 		case 8:
 		case 7:
@@ -861,7 +860,7 @@ function SpeechGarble5(H) {
 			H = "";
 			break;
 
-		// Regular characters
+			// Regular characters
 		case "e":
 		case "i":
 		case "o":
@@ -1077,7 +1076,7 @@ function SpeechGarble4(H) {
 			H = "";
 			break;
 
-		// Regular characters
+			// Regular characters
 		case "v":
 		case "b":
 		case "c":
@@ -1285,7 +1284,7 @@ function SpeechGarble3(H) {
 			H = "";
 			break;
 
-		// Regular characters
+			// Regular characters
 		case "v":
 		case "b":
 		case "c":
@@ -1491,7 +1490,7 @@ function SpeechGarble2(H) {
 			H = "";
 			break;
 
-		// Regular characters
+			// Regular characters
 		case "c":
 		case "t":
 			H = "e";
@@ -1679,7 +1678,7 @@ function SpeechGarble1(H) {
 			H = "";
 			break;
 
-		// Regular characters
+			// Regular characters
 		case "t":
 			H = "e";
 			break;
--- a/BondageClub/Scripts/Speech.js
+++ b/BondageClub/Scripts/Speech.js
@@ -47,7 +47,7 @@ function SpeechGetEffectGagLevel(Effect) {
  * @param {string} AssetGroup - The name of the asset group to look through
  * @returns {number} - Returns the total gag effect of the character's assets
  */
- function SpeechGetGagLevel(C, AssetGroup) {
+function SpeechGetGagLevel(C, AssetGroup) {
 	var GagEffect = 0;
 	for (let i = 0; i < C.Appearance.length; i++) {
 		var item = C.Appearance[i];
@@ -69,7 +69,7 @@ function SpeechGetEffectGagLevel(Effect) {
 function SpeechGetGagEffect(C, NoDeaf) {
 
 	let GagEffect = 0;
-	
+
 	GagEffect += SpeechGetGagLevel(C, "ItemMouth");
 	GagEffect += SpeechGetGagLevel(C, "ItemMouth2");
 	GagEffect += SpeechGetGagLevel(C, "ItemMouth3");
@@ -78,7 +78,7 @@ function SpeechGetGagEffect(C, NoDeaf) {
 	GagEffect += SpeechGetGagLevel(C, "ItemNeck");
 	GagEffect += SpeechGetGagLevel(C, "ItemDevices");
 	GagEffect += SpeechGetGagLevel(C, "ItemHoodAddon");
-	
+
 	if (C.ID != 0 && !NoDeaf) {
 		if (Player.GetDeafLevel() >= 7) GagEffect = Math.max(GagEffect, 20);
 		else if (Player.GetDeafLevel() >= 6) GagEffect = Math.max(GagEffect, 16);
@@ -105,12 +105,12 @@ function SpeechGarble(C, CD, NoDeaf) {
 
 	// The usual preamble - calculate the gag effect level, and the character's deafness level once
 	GagEffect = SpeechGetGagEffect(C, NoDeaf);
-	
+
 	// Makes sure that the gag level is within acceptable parameters. As a GagEffect of more than 20 does nothing. So reduces it if above 20.
 	if (GagEffect > 20) GagEffect = 20;
-	
+
 	// If GagEffect is 0 then there is no need to run this part so, a simple if statement is used in order to check.
-	if (GagEffect > 0)  {
+	if (GagEffect > 0) {
 
 		// This is a function which transforms a single character - one of the `switch` statements
 		const garblingFunction = SpeechGetGarblingFunction(GagEffect);
@@ -120,7 +120,7 @@ function SpeechGarble(C, CD, NoDeaf) {
 	}
 	NS = SpeechStutter(C, NS);
 	NS = SpeechBabyTalk(C, NS);
-	
+
 	return NS;
 
 }
@@ -133,18 +133,27 @@ function SpeechGarble(C, CD, NoDeaf) {
 function SpeechGetGarblingFunction(GagEffect) {
 
 	let CorrectFunction;
-	
-	switch(GagEffect) {
+
+	switch (GagEffect) {
 		case 20:
 			CorrectFunction = SpeechGarble20;
 			break;
-		case 19: case 18: case 17: case 16:
+		case 19:
+		case 18:
+		case 17:
+		case 16:
 			CorrectFunction = SpeechGarble16;
 			break;
-		case 15: case 14: case 13: case 12:
+		case 15:
+		case 14:
+		case 13:
+		case 12:
 			CorrectFunction = SpeechGarble12;
 			break;
-		case 11: case 10: case 9: case 8:
+		case 11:
+		case 10:
+		case 9:
+		case 8:
 			CorrectFunction = SpeechGarble8;
 			break;
 		case 7:
@@ -169,7 +178,7 @@ function SpeechGetGarblingFunction(GagEffect) {
 			CorrectFunction = SpeechGarble1;
 			break;
 	}
-	
+
 	return CorrectFunction;
 }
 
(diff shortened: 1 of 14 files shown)


Nina-1474:
The lint error is due to the same values appearing in the cases more than once. For example the Cyrillic 'у' appears at both lines 321 and 327.
This was actually an error in the old method too but there was nothing to check it. It can't be ignored though, as every PR after this would then also fail the checks.

Shiranui-Izayoi:
> 
> 
> The lint error is due to the same values appearing in the cases more than once. For example the Cyrillic 'у' appears at both lines 321 and 327.
> This was actually an error in the old method too but there was nothing to check it. It can't be ignored though, as every PR after this would then also fail the checks.

Ahhh makes sense. I will go through and try to delete the duplicates. It may take a little bit since I am not sure which have capital versions and which don't. I guess that Cyrillic normally doesn't.

Thanks Nina for letting me know.

Shiranui-Izayoi:
Ok fixed all the check errors I HOPE.

May I also add that I now LOATH Cyrillic characters... Also vague checks those as well.

Hopefully I have got them all. And that it doesn't effect how things work to bad. Also removed some pointless things like, a character converting into another character which if it had been typed as the converted character would have been converted into something else.

Shiranui-Izayoi:
Removed un-needed case statements that just resulted the character translating into itself. AKA just a break. 

Since anything that isn't in the case list just gets returned as is.

Also sorry for like the +600ish lines was a result of me "beautifying" the code. As without it trying to fix the case erros would have been much much much more annoying. It won't effect any of the code when run. Just the amount of lines. Also causes all the code to be in the same style.

elliesec:
> Also removed some pointless things like, a character converting into another character which if it had been typed as the converted character would have been converted into something else.
I've not had a proper look through yet, but doesn't this modify the behaviour of the garbling functions? The fact that the functions aren't idempotent doesn't necessarily mean that behaviour like that is pointless.

@Ben987 - I'd like to pull this down and run some testing of my own before this goes through, as it's going to be difficult to determine the behavioural changes just from inspecting the code changes in this case. I'll post back on it when I've had a full look.

Ada18980:
Isn't removing SpeechGarbleByGagLevel going to crash the game when garbling the chat room titles?

Shiranui-Izayoi:
> 
> 
> Isn't removing SpeechGarbleByGagLevel going to crash the game when garbling the chat room titles?
> 
> I would prefer that garbling a line by a certain gag level be doable by calling a single function without using functions as parameters.

Unless it calls that function directly then no. That said I wasn't aware of that interaction. Though I reckon it will call SpeechGarble that will then caculate level which then would garble. In which case it will work as it currently does.

Also that function needs to be removed since it is what I was trying to replace as it uses all if statements and isn't very efficent.

Correct me if I am wrong but this is how the current system works. SpeechGarble gets called, it then calls the get gag level function which returns the value. And then calls the garble by gag level sending along the gag level it just got from the previous function. Once garbling is done that is returned to SpeechGarble which then walks through the rest and then sends it.

If that is correct then the way this works is similar. It calls SpeechGarble, which then gets gag level. (I have to admit the way that deaf is caculated is really nice.) After that if gag level is more than 0 it will run a function to find the right gag level function and return that. Then passes that along to the function that loops it through the correct function. And returns when done. To be stuttered etc and is done.

The goal was to break it down into multiple functions for easy editting.

elliesec:
`SpeechGarbleByGagLevel` is directly called in the `ChatSearchMuffle` function. Regardless, I think it would be sensible to have a single function that does what `SpeechGarbleByGagLevel` previously did. It makes the interface simple and easy to use, which is important for testing (which I'm planning on implementing for this file), and it makes life easier for any scripters or contributors that want to make use of the garbling functions in their own code (e.g. new rooms/features, or just custom scripts). `SpeechGarbleLine` is fine for internal use (and it reduces repetition), but it doesn't provide an easy to use interface.

Shiranui-Izayoi:
> 
> 
> `SpeechGarbleByGagLevel` is directly called in the `ChatSearchMuffle` function. Regardless, I think it would be sensible to have a single function that does what `SpeechGarbleByGagLevel` previously did. It makes the interface simple and easy to use, which is important for testing (which I'm planning on implementing for this file), and it makes life easier for any scripters or contributors that want to make use of the garbling functions in their own code (e.g. new rooms/features, or just custom scripts). `SpeechGarbleLine` is fine for internal use (and it reduces repetition), but it doesn't provide an easy to use interface.

Ahhh I was not aware this was a thing. I will update my PR and look through. I will also see if I can simplify the function to give a way to call a certain function directly. I think I will something to garble line that lets you manually set it easier. Though not sure how. I will look how `SpeechGarbleByGagLevel` worked and see if I can implement that.

Shiranui-Izayoi:
I think I can actually make it compatible. I think if I have the input for `garblingFunction` be something that doesn't start with `SpeechGarble` then I could potentially define it and be able to work it out. Sorta like

if (!garblingFunction.startsWith("SpeechGarble")) {
	garblingFunction = SpeechGetGarblingFunction(garblingFunction);
}
And then the text would work fine. Since you can include Ignore OOC in the function already. I will put that in and make the changes and then someone can test it. Since I don't know the default behaviour well so I can't really judge.

Thanks for letting me know about this issue.

elliesec:
You should be able to basically re-implement the old functionality with a couple of lines, based on your current changes:
```js
function SpeechGarbleByGagLevel(GagEffect, CD, IgnoreOOC) {
    const garblingFunction = SpeechGetGarblingFunction(GagEffect);
    return SpeechGarbleLine(CD, garblingFunction, IgnoreOOC);
}
```
Something along those lines.

Shiranui-Izayoi:
> ```js
> function SpeechGarbleByGagLevel(GagEffect, CD, IgnoreOOC) {
>     const garblingFunction = SpeechGetGarblingFunction(GagEffect);
>     return SpeechGarbleLine(CD, garblingFunction, IgnoreOOC);
> }
> ```

True that would work. I was planning to just change the thing. But yes that should work if I just insert that then everything can stay the same. Thanks Ellie. Will upload the update in a few minutes.

Shiranui-Izayoi:
Ok uploaded let me know if you find/think of any other issues.

Shiranui-Izayoi:
I moved the validation for GagEffect to the get garble function. Since a GagEffect of over 20 would break it. And with it in it's original spot it could be bypassed if this function was called directly.

BondageClub/Scripts/Speech.js
```diff
@@ -94,360 +89,1649 @@ function SpeechGarble(C, CD, NoDeaf) {
...
  */
 function SpeechGarbleByGagLevel(GagEffect, CD, IgnoreOOC) {
+    const garblingFunction = SpeechGetGarblingFunction(GagEffect);
+    return SpeechGarbleLine(CD, garblingFunction, IgnoreOOC);
+}
+
+/**
+ * This runs the for loop that actually garbles the text and returns NS when done to SpeechGarble.
+ * @param {CD} The message itself.
+ * @param {garblingFunction} The string that tells which function to run.
```
elliesec, line 198: As per the comment above, the format of both `@param` declarations is incorrect here. For the `garblingFunction` parameter, you can either use the `function` type, or you can define a [callback function type](https://jsdoc.app/tags-callback.html) - see the top of `CommonDraw.js` for examples of some callback function type declarations.

BondageClub/Scripts/Speech.js
```diff
@@ -94,360 +89,1649 @@ function SpeechGarble(C, CD, NoDeaf) {
...
+
+	// The usual preamble - calculate the gag effect level, and the character's deafness level once
+	GagEffect = SpeechGetGagEffect(C, NoDeaf);
+
+	// If GagEffect is 0 then there is no need to run this part so, a simple if statement is used in order to check.
+	if (GagEffect > 0) {
 
-	// No gag effect, we return the regular text
+		// This is a function which transforms a single character - one of the `switch` statements
+		const garblingFunction = SpeechGetGarblingFunction(GagEffect);
```
elliesec, line 113: Now that `SpeechGarbleByGagLevel` has been reinstated, we might as well use it here instead of these two lines.

elliesec (dismissed):
Okay, I've had a look through the code, and run some tests on it. I've run ~10,000 lines of text through the function in a variety of languages whilst testing.

@Ben987 - I've outlined my test results below, both in terms of performance and functionality differences to the current behaviour. I think it's up to you whether you think the behavioural changes are desirable or not.

## Performance

I've used the `SpeechGarbleByGagLevel` function to test the garbling output, as the function is common across both the old and the new versions of the file. In each case, I've run the function over each piece of test data with each of the 21 available gag effect levels (0-20), and for each piece of test data, I've run this 10 times, for a total of 2,205,420 runs of the function, with the following results on my machine:

**Old:**
```
Processed 2205420 messages in 33.449s
65934 messages/second
```

**New:**
```
Processed 2205420 messages in 29.933s
73679 messages/second
```
This shows that the new method is roughly 12% faster than the old one.

## Functional Differences

I've also scripted a primitive diff check to identify any differences in behaviour between the new and old functions. Aside from the main difference already mentioned (the new garble respects letter casing), the other thing of note is that with the new functionality, any characters that don't map to something are simply appended as is, whereas with the old functionality, they'd get removed.

I'm in two minds as to whether or not this is a good thing. Firstly, in languages that aren't handled by the garbling functions, most text is simply not garbled, meaning that gags have next to no effect:
```
Original:	5	@harukyan2580 ぇッまじ？営業停止とかの話聞いてない？
Old garble:	5	haaekean 
New garble:	5	@haaekean2580 ぇッまじ？営業停止とかの話聞いてない？
```
Secondly, it results in players being able to use characters that they arguably shouldn't be able to, in general increasing the range of expressiveness available when gagged (notice that it also means that numbers remain ungarbled):
```
Original:	4	Shopping! (@ Kohl's) http://t.co/I8ZkQHT9
Old garble:	4	hhoffinn! (@ Kohl's) heefe.eoihkkhe
New garble:	4	Hhoffinn! (@ Kohl's) heef://e.eo/I8HkKHE9
```
Asides from these differences, I've not noticed any problems with the garbling functionality. Notably the following string (which encapsulates all of the non-symbol characters handled by the old garble function) garbles exactly the same using both methods:
```
usetricAlokpmanghfdvbyOjSGLT4203168VH7M5NEzwDRUFxIPCB9qZQJXKYWáâàéêèëíîìïóôòúûùüçñÁÂÀÉÊÈËÍÎÌÏÓÔÒÚÛÙÜÇÑиоуюляесйхкжразгбвыфднмИОУЮЛЯЕСЙХКЖРАЗГБВЫФДНМпПчцщЧЦЩё
```

BondageClub/Scripts/Speech.js
```diff
@@ -522,27 +1806,48 @@ function SpeechStutter(C, CD) {
...
+						break;
+					default:
+						NS = NS + H;
+						break;
+				}
 			} else NS = NS + CD.charAt(L);
 			if (H == ")") Par = false;
 		}
-		return NS;
+		return NS.replace(/thh/g, "th").replace(/Thh/g, "Th").replace(/shh/g, "sh").replace(/Shh/g, "Sh");
```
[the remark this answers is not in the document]
Shiranui-Izayoi: Huh never knew this was a thing I will edit that in in a bit.

BondageClub/Scripts/Speech.js
```diff
@@ -94,360 +89,1649 @@ function SpeechGarble(C, CD, NoDeaf) {
...
 	NS = SpeechStutter(C, NS);
 	NS = SpeechBabyTalk(C, NS);
 
 	return NS;
+
+}
+
+/**
+ * Chooses the correct function to run via a switch statement and returns it when found.
+ * @param {GagEffect} The already caculated value that is used to determine the correct function.
```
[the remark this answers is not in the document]
Shiranui-Izayoi: I am new to JSDoc stuff so it doesn't surprise me that I got it wrong here.

BondageClub/Scripts/Speech.js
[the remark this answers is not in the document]
elliesec: It's nothing too fancy, but it allows us to use a single `replace`, which is nice. The `$n` notation allows you to substitute in a group captured inside a parenthesis (`([TSts])` in this case) into the replacement string (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#specifying_a_string_as_a_parameter).

BondageClub/Scripts/Speech.js
Shiranui-Izayoi: Doesn't it seem a bit pointless to call a function to call a function? When all that calling `SpeechGarbleByGagLevel` does is then call `SpeechGetGarblingFunction` just to return the value. 

Or do you also mean to remove the line under it with `NS = SpeechGarbleLine(CD, garblingFunction);`?

BondageClub/Scripts/Speech.js
```diff
@@ -94,360 +89,1649 @@ function SpeechGarble(C, CD, NoDeaf) {
...
+	
+	// Makes sure that the gag level is within acceptable parameters. As a GagEffect of more than 20 does nothing, but for this function above 20 will break it.
+	if (GagEffect > 20) GagEffect = 20;
+
+	let CorrectFunction;
+
+	switch (GagEffect) {
+		case 20:
+			CorrectFunction = SpeechGarble20;
+			break;
```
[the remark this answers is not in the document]
Shiranui-Izayoi: Ok swapped them all out for return.

BondageClub/Scripts/Speech.js
Shiranui-Izayoi: Honestly I might have to just cut out this and the other one mentioned as I don't currently feel confident enough to do them justice.

BondageClub/Scripts/Speech.js
elliesec: Yep, I'm talking about replacing both this line and the one below with the function call. Sorry, should have made the comment multiline.

BondageClub/Scripts/Speech.js
elliesec: In which case I'll update my comments with suggestions, which should be good enough for the moment.

Shiranui-Izayoi:
> 
> This shows that the new method is roughly 12% faster than the old one.
> ## Functional Differences
> 
> I'm in two minds as to whether or not this is a good thing. Firstly, in languages that aren't handled by the garbling functions, most text is simply not garbled, meaning that gags have next to no effect:
> 
> Secondly, it results in players being able to use characters that they arguably shouldn't be able to, in general increasing the range of expressiveness available when gagged (notice that it also means that numbers remain ungarbled):


Interesting to know that it is only 12% better overall. Though I guess that over multiple people the time would add up which is nice. If anything at least it makes it easier to edit and track things down. Instead of scrolling through one long function.

Also personally I have found that losing access to some of the previously non translated symbols can be annoying at times. And honestly if someone wants to bypass the gag stuff they can in a multitude of ways. So I would consider it a non-issue.

Also on the other hand languages that aren't garbled currently can't even speak in most cases when gagged which must be annoying honestly. So I think in the interest of allowing more languages there isn't much harm in there not being much effect. I think it is preferable that there is little to no effect rather than nearly or completely muting them.

BondageClub/Scripts/Speech.js
```diff
@@ -94,360 +89,1649 @@ function SpeechGarble(C, CD, NoDeaf) {
...
  * @param {string} CD - The character's dialog to alter
  */
 function SpeechGarbleByGagLevel(GagEffect, CD, IgnoreOOC) {
+    const garblingFunction = SpeechGetGarblingFunction(GagEffect);
+    return SpeechGarbleLine(CD, garblingFunction, IgnoreOOC);
+}
+
+/**
+ * This runs the for loop that actually garbles the text and returns NS when done to SpeechGarble.
+ * @param {CD} The message itself.
```
elliesec, line 197:
```suggestion
 * @param {string} CD - The string to garble.
```

BondageClub/Scripts/Speech.js
Shiranui-Izayoi: NP putting that in now. Honestly should have been obvious really. My bad.

Shiranui-Izayoi:
Ok think I have made all the needed changes @elliesec take a look to see if they are as you suggested and if possible resolve them when done. It should all be right.

elliesec:
> Also personally I have found that losing access to some of the previously non translated symbols can be annoying at times. And honestly if someone wants to bypass the gag stuff they can in a multitude of ways. So I would consider it a non-issue.
> 
> Also on the other hand languages that aren't garbled currently can't even speak in most cases when gagged which must be annoying honestly. So I think in the interest of allowing more languages there isn't much harm in there not being much effect. I think it is preferable that there is little to no effect rather than nearly or completely muting them.

I think this is an opinion thing. Personally, I like the fact that you lose access to a lot of characters when gagged. What's the purpose of a gag if not to restrict your range of expression? The fact that this (for example) allows you to straight up say a number clearly when you wouldn't realistically be able to is a loss in my opinion. And I don't think that the fact that there are a multitude of ways to bypass gag talk means we should make it less restrictive - we could take that argument to its logical conclusion and just remove garbling completely.

On the topic of languages, I'm in two minds. On the one hand, most people in the club converse in languages with Latin or Cyrillic alphabets. But for character sets like Chinese/Japanese, there are kind of two cases. On the one hand, the new behaviour means that gags basically do nothing when conversing in those languages, which is annoying in the other direction - gags simply don't stop you from talking. On the other hand, the old behaviour means that all gags effectively mute the player. Given the choice between the two behaviours, I suspect this is the sort of thing that is going to be down to player preference as to which they prefer (for example, I'd probably prefer the old behaviour because it means the gags actually do something).

Shiranui-Izayoi:
> 
> 
> > Also personally I have found that losing access to some of the previously non translated symbols can be annoying at times. And honestly if someone wants to bypass the gag stuff they can in a multitude of ways. So I would consider it a non-issue.
> > Also on the other hand languages that aren't garbled currently can't even speak in most cases when gagged which must be annoying honestly. So I think in the interest of allowing more languages there isn't much harm in there not being much effect. I think it is preferable that there is little to no effect rather than nearly or completely muting them.
> 
> I think this is an opinion thing. Personally, I like the fact that you lose access to a lot of characters when gagged. What's the purpose of a gag if not to restrict your range of expression? The fact that this (for example) allows you to straight up say a number clearly when you wouldn't realistically be able to is a loss in my opinion. And I don't think that the fact that there are a multitude of ways to bypass gag talk means we should make it less restrictive - we could take that argument to its logical conclusion and just remove garbling completely.
> 
> On the topic of languages, I'm in two minds. On the one hand, most people in the club converse in languages with Latin or Cyrillic alphabets. But for character sets like Chinese/Japanese, there are kind of two cases. On the one hand, the new behaviour means that gags basically do nothing when conversing in those languages, which is annoying in the other direction - gags simply don't stop you from talking. On the other hand, the old behaviour means that all gags effectively mute the player. Given the choice between the two behaviours, I suspect this is the sort of thing that is going to be down to player preference as to which they prefer (for example, I'd probably prefer the old behaviour because it means the gags actually do something).
> 
> Ultimately, I think it's very much on Ben to have the final say on whether the new or old behaviour is preferable in this regard.

True true. I can add a set of lines to the various gag levels that would remove numbers if that is an issue. Though honestly as said most seem to be of the idea I am gagged, I am not going to bypass it unless I have to. At least from what I have seen, this is the case.

BondageClub/Scripts/Speech.js
```diff
@@ -94,360 +89,1630 @@ function SpeechGarble(C, CD, NoDeaf) {
...
 
 /**
- * The core of the speech garble function, usable without being tied to a specific character
+ * Garbles a line of text based on the provided gag effect level.
  * @param {Int} GagEffect - The gag level of the speech
  * @param {string} CD - The character's dialog to alter
  */
 function SpeechGarbleByGagLevel(GagEffect, CD, IgnoreOOC) {
+    const garblingFunction = SpeechGetGarblingFunction(GagEffect);
+    return SpeechGarbleLine(CD, garblingFunction, IgnoreOOC);
```
elliesec, line 173:
```suggestion
	if (GagEffect > 0) {
		const garblingFunction = SpeechGetGarblingFunction(GagEffect);
		return SpeechGarbleLine(CD, garblingFunction, IgnoreOOC);
	} else {
		return CD;
	}
```
One final suggestion. It's not an issue now, but since `SpeechGetGarblingFunction` will return `undefined` for a gag effect of 0, calling `SpeechGarbleByGagLevel(0, someText)` would throw an error. This would make the function safer and a bit more flexible for consumers with no real downside (it also means you'd be able to remove the `if` clause on line 110).

Shiranui-Izayoi: Nice catch Ellie I forgot about that. I remembered it when writing up the inital function though so :)

Shiranui-Izayoi:
Ok so I stuck in a case statement to each gag level to remove direct numbers as I do agree with you on the number front.

I also added a couple symbols to the allowed list for the total gag levels those being +, \ and / as I have recently witnessed some people using them as sentence continuers and enders during RP. Such as putting a + at the end when they plan to continue on a new message. For example switching from talking to an action. And then putting a \ or / at the end when they are done.

I feel this change is harmless as it doesn't do much but does allow people using a system like this to do so even when total gagged.

elliesec:
@Ben987 - Code-wise, everything has now been addressed as far as I'm concerned. See [my earlier comment](https://github.com/Ben987/Bondage-College/pull/2387#pullrequestreview-650709734) summarising my testing of the changes, and [@Shiranui-Izayoi's comment](https://github.com/Ben987/Bondage-College/pull/2387#issuecomment-831584047) detailing subsequent changes that have been made. At this point, I think the rest is down to whether or not you think the behavioural changes are desirable.

Ben987:
Wow, this is probably the most commented pull request ever.  Let's not include that one if it barely speeds up the code and changes the way how garbles works.  Players are used to the current system.  Thanks for all your efforts everyone.

Shiranui-Izayoi:
Well... That is quite annoying. And it is the most commented because it overhauls a lot... I can also include a thing which would make it so that anything not on the list is removed. As it normally is if that is an issue.

PR #2389, opened 2021-05-01
Transport jacket usage

Manilla32:
Makes the transport jacket able to leash people.


elliesec:
I'm not sure about this one. Whilst it does have "transport" in its name, I'm not sure that adding the leashing mechanic to the item makes that much sense. Other items with the leash mechanic are either leashes, or were explicitly created with the intention of being able to move the person (i.e. the transport box), whereas this doesn't have anything that makes the wearer inherently "transportable". I'll discuss this one with Cecilia, who should probably get the final say as the original asset creator.

Manilla32:
> I'm not sure about this one. Whilst it does have "transport" in its name, I'm not sure that adding the leashing mechanic to the item makes that much sense. Other items with the leash mechanic are either leashes, or were explicitly created with the intention of being able to move the person (i.e. the transport box), whereas this doesn't have anything that makes the wearer inherently "transportable". I'll discuss this one with Cecilia, who should probably get the final say as the original asset creator.

What about the straps that are on it? It looks like someone could grab onto one of those.

elliesec:
Technically yes, but the same could probably be said of other items. My main concern with this is that it might put people off the item. The other items have leashing as their primary function - generally, people put them on with the explicit intention of leashing people around, whereas the transport jacket is primarily intended as a bondage item, and adding leashing to it might put off people that don't like playing with the leashing functionality (of which there are definitely a good few), but otherwise like the jacket for the aesthetic.

elliesec:
I'm not saying that it definitely can't go in. I just think that this is a discussion worth having, because it's making a gameplay change to an existing item, and that might change the way players view the item.

PR #2390, opened 2021-05-02
Kinky Dungeon 1.31

Ada18980:
-Some bugfixes and balancing
-Adds a rope snake, an enemy which sacrifices itself to place a rope restraint on you. The restraint is tougher than light charms but weaker than mummification
-Adds a new spell, Electrify
-Adds a new enemy, the Shock Witch, which hunts for you and launches Electrify spells at you!


Ben987:
Another big update!  Thanks a lot :)

PR #2394, opened 2021-05-02
Tennis ball

Manilla32:
Added a tennis ball with stripes


Ben987:
Hahaha, sure, why not.  Thanks :)

PR #2395, opened 2021-05-02
Refactor: VibratingDildoBar to extended vibrator script

SandrineBC:
# Description
Does exactly that. Uses the (not so) new vibrator options for the vibrating dildo bar

# Testing
Tested that on my local installation and all looked fine.


Ben987:
Sure!  Thanks for that one :)

PR #2396, opened 2021-05-03
Chat Setting: Shrink non-dialogue

Ada18980:
Sometimes players complain that the text from binding someone takes over the whole chat log. This setting allows users to shrink anything that isn't chat or an emote in order to focus on RP if they wish to do so.


Ben987:
Sure!  That's a good option, thanks :)

PR #2397, opened 2021-05-03
Sci Fi Pleasure Panties assets

Ada18980:
Added small, normal, large, and xlarge assets for the sci fi pleasure panties.

I made the original asset for these and checked with the creator to make sure it's OK.


Shiranui-Izayoi:
THANK YOU ADA! 

I am sure all the small/thin bodied girls will enjoy their ~~not at all devilishly evil and locked on~~ panties are now tighter and fit better. ~~Which definitly means they can no longer slip a finger under them *grins and giggles devilishly*~~

Ben987:
Sure!  Thanks a lot :)

PR #2398, opened 2021-05-03
Plastic wrap

Codefreak64:
ever want to keep your subs fresh? and no im not talking about the sandwich. A new shop order has come in, plasticwrap, and its got more uses than the sub wrapped up in it.

-Plasticwrap item added in the shop ($100)

-New buygroup (plasticwrap)

-Adds the new item in 6 slots

Im sure isses will pop up eventually and will be fixed durring beta, I would like this to be implemented in the next beta
From now on, in the change log please put me down as Natsuki and NOT Z-Freak. thankyou


Codefreak64:
omg i just realised i forgot alpha masks, i'll test in my version and add them into here manually via the github editor

Codefreak64:
just trying to pass all the chacks

BondageClub/Assets/Female3DCG/Female3DCG.js
```diff
@@ -2455,6 +2457,7 @@ var AssetFemale3DCG = [
 					{ Name: "StrapsLegs", CopyLayerColor: "Straps", ParentGroup: "BodyLower", AllowTypes: ["ShortsAndStraps"], HasType: false, HideForPose: ["Horse", "KneelingSpread", "Spread"] },
 				],
 			},
+			{ Name: "PlasticWrap", Value: 100, Difficulty: 7, SelfBondage: 3, Time: 30, RemoveTime: 25, BuyGroup: "PlasticWrap", SetPose: ["BackElbowTouch"], Block:["ItemBreast","ItemNipples", "ItemNipplesPiercings", "ItemTorso"], Hide: ["Cloth", "ClothAccessory"]},
```
elliesec, line 2460: Should this have the `Block` and `Prone` effects? It strikes me as odd that you'd be able to use hands & change clothes with the arm item on.

BondageClub/Assets/Female3DCG/Female3DCG.js
```diff
@@ -3150,6 +3153,7 @@ var AssetFemale3DCG = [
 					{ Name: "Funnel", Priority: 55, AllowColorize: false, AllowTypes: ["Funnel"] }
 				]
 			},
+			{ Name: "PlasticWrap", Value: 100, Difficulty: 4, Time: 30, RemoveTime: 25, BuyGroup: "PlasticWrap", Effect: ["BlockMouth", "GagLight"], Block:["ItemMouth2", "ItemMouth3"]},
```
elliesec, line 3156: Is there any need for this to block `ItemMouth2` and `ItemMouth3`? Other similar items like the duct tape allow for further gags to be layered on top.

BondageClub/Assets/Female3DCG/Female3DCG.js
```diff
@@ -3593,6 +3598,7 @@ var AssetFemale3DCG = [
 					{ Name: "Muzzle" },
 				]
 			},
+			{ Name: "PlasticWrap", Value: 100, Difficulty: 4, Time: 30, RemoveTime: 25, BuyGroup: "PlasticWrap", Effect: ["BlockMouth", "GagLight"], Block:["ItemMouth1", "ItemMouth2"]},
```
elliesec, line 3601: The `Block` entry shouldn't be needed here, as it's implicitly added already by the group.

BondageClub/Assets/Female3DCG/Female3DCG.js
```diff
@@ -3369,6 +3373,7 @@ var AssetFemale3DCG = [
 					{ Name: "Strap" },
 				]
 			},
+			{ Name: "PlasticWrap", Value: 100, Difficulty: 4, Time: 30, RemoveTime: 25, BuyGroup: "PlasticWrap", Effect: ["BlockMouth", "GagLight"], Block:["ItemMouth1", "ItemMouth3"]},
```
elliesec, line 3376: Same comment as above - any reason that this needs to block `ItemMouth3`?

elliesec: You could also add `DynamicGroupName: "ItemMouth"` here (and in the `ItemMouth3` definition), which would remove the need for the duplicate `ItemMouth2` and `ItemMouth3` assets (`DynamicGroupName: "ItemMouth"` essentially redirects the game to the `ItemMouth` directory when it's looking for item assets).

BondageClub/Assets/Female3DCG/Female3DCG.js
Codefreak64: forgot about those, yes it should have them like that

BondageClub/Assets/Female3DCG/Female3DCG.js
Codefreak64: I now realise that oversight, the blocking can be removed for the `ItemMouth1`

BondageClub/Assets/Female3DCG/Female3DCG.js
Codefreak64: so wait, we have a system that does that automatically? i was not aware of that if we do have one

BondageClub/Assets/Female3DCG/Female3DCG.js
Codefreak64: this is the first im hearing of `DynamicGroupName`. I'll refine this with that property once beta comes around if thats alright, since i'm planning on making multiple bug fixes for this item in 1 PR, just in case i missed something, if that makes sense

BondageClub/Assets/Female3DCG/Female3DCG.js
elliesec: Yes. If you have a look at the `ItemMouth2`/`ItemMouth3` groups, you'll see that `Block` is already defined at the group level. If the asset itself doesn't override that, it'll automatically inherit the `Block` entry from its group.

BondageClub/Assets/Female3DCG/Female3DCG.js
elliesec: Not a problem - it's relatively new (well, it's been around for a while, but it recently got updated so it can be used remove duplicated assets). I'm happy for this to go in without that change though, as I'm about to do a sweep of the gags anyway and add `DynamicGroupName` to the ones which duplicate something in the `ItemMouth` group. Functionally, it'll behave the same, it just means fewer assets in the game in total, which ultimately helps reduce the memory load of the game (as well as the game generally just being somewhat more lightweight).

elliesec (approved):

Ben987:
This is nice, I'm sure it will be popular.  Thanks a lot :)

PR #2401, opened 2021-05-03
Refactor: Ceiling shackles to new archetypes

SandrineBC:
# Description
Refactors the ceiling shackles to the new archetypes

# Testing
Was done on my local installation of the club


jomshir98 (changes requested):
The reason I didn't yet convert my item into new archetype is this line:
```js
var msg = "CeilingShacklesRestrain" + (C.Pose.includes("Suspension") ? "Suspension" : "") + Option.Name;
```
Your change would break "Suspension" behaviour of the shackles.

SandrineBC:
> The reason I didn't yet convert my item into new archetype is this line:
> 
> ```js
> var msg = "CeilingShacklesRestrain" + (C.Pose.includes("Suspension") ? "Suspension" : "") + Option.Name;
> ```
> 
> Your change would break "Suspension" behaviour of the shackles.

Good point.

elliesec:
For that behaviour, you might want to grab the latest master. You can now pass a function into `Config.Dialog.ChatPrefix` that would allow you to replicate the current behaviour (although you'll need to modify `TypedItem.js` slightly to pass the reference to `C` into the function when it gets called (see `TypedItemCreatePublishFunction`).

Ben987:
Sure!  Thanks again :)

PR #2402, opened 2021-05-03
Feature: Auto-generates `AllowEffect` and `AllowBlock` on typed items

elliesec:
## Summary

This modifies the typed item script slightly to automatically generate an asset's `AllowBlock` and `AllowEffect` entries based on its type configuration (very similar to what was done with modular items in #2333). This should mean that asset creators no longer need to worry about adding `AllowEffect` and `AllowBlock` to their items if using the archetype framework.

I've also tidied up the configuration for the ball and wiffle gags, as their configuration contained some unneeded `Effect` entries (which were just duplicating the item's base `Effect`).

This also modifies the chatroom publish functionality of the typed item script slightly to pass the character reference into `Config.Dialog.ChatPrefix` functions when they are called, which should enable conversion of the ceiling shackles as per @jomshir98's comment in #2401.

All affected items have been tested and are working exactly as before.


Ben987:
Any change that simplify assets making is always welcome.  Thanks!

PR #2403, opened 2021-05-03
"Hide others' Messages" applies to Hide Names sensory deprivation option

Ada18980:
Figured there was no reason to block it out.

It still does not affect Light because in Light the player cannot enter total sensory deprivation due to the IsInSenseDep function.


Ben987:
Sure!  Thanks a lot :)

PR #2404, opened 2021-05-04
Kinky Dungeon 1.32

Ada18980:
-Added chain bolt, a simple spell that stuns an enemy without dealing damage. Decent for running away.
-Added a chain witch, a new miniboss who is the toughest enemy yet. She hurls chains at you which are some of the most punishing restraints in the game and can't be cut. If she is in melee range she will lock them in place! Do not go near her if you have an armbinder on or you will have to pray to the goddess of leather to get it off!


Ben987:
You're a machine!  Thanks a lot :)

PR #2406, opened 2021-05-04
Asset: Allows for second back hair

fleisch11:
not sure, if this is a good idea, but since a few hairstyles are combinations allready, why not let the players do the combining?


Ben987:
No.  The back hairs are already combined by itself.  What we would need to do is remove all the combos first and then create that new slot.   But it would break lots of wardrobe so I'm not sure if players would want that.   You could do a quick poll to see if players are ok with it first.  Thanks a lot.

PR #2407, opened 2021-05-04
Bugfix: Fixes linting errors

elliesec:
Fixes linting errors introduced in #2404 & fd09289b5baf8e7e5ae64b87ca43ffc1b0b7a141.

PR #2408, opened 2021-05-04
Bugfix: Re-enables `DarkFactor` inside minigames

elliesec:
# Summary

#2163 modified `Drawing.js` so that `DarkFactor` was no longer applied inside minigames. This resulted in a few visual issues in some minigames, for example the Asylum Therapy minigame, where the instruction text is shown against an almost-white background, making it almost impossible to read (see image below). This removes the minigame check, which fixes the issue for the Asylum Therapy and any other minigames that were trying to set `DarkFactor` to something other than 1.

I assume that the minigame check was added to support the Kinky Dungeon minigame (although that hadn't yet been fully introduced at the time of the above PR). That said, I've checked this change inside the Kinky Dungeon minigame, and have not seen any issues as a result. @Ada18980 - do you foresee this causing any problems with the Kinky Dungeon minigame?

Asylum

### DIFF
--- a/BondageClub/Screens/MiniGame/KinkyDungeon/KinkyDungeon.js
+++ b/BondageClub/Screens/MiniGame/KinkyDungeon/KinkyDungeon.js
@@ -24,6 +24,7 @@ var KinkyDungeonRootDirectory = "Screens/MiniGame/KinkyDungeon/";
  */
 function KinkyDungeonLoad() {
 
+	CurrentDarkFactor = 0;
 	if (!KinkyDungeonGameRunning) {
 		if (!KinkyDungeonPlayer)
       KinkyDungeonPlayer = CharacterLoadNPC("NPC_Avatar");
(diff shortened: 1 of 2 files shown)


elliesec:
**Note:** Lint errors are those fixed by #2407, and shouldn't be an issue once that's been merged.

Ada18980:
Hi,

I believe when the headset is off and playing the kinky dungeon game, the background will be black. This is why I made the change. I didnt realize some minigames actually depended on the player being blind!

elliesec:
> Hi,
> 
> I believe when the headset is off and playing the kinky dungeon game, the background will be black. This is why I made the change. I didnt realize some minigames actually depended on the player being blind!

Thanks, I'll give it a test and see if I can come up with something that works in all cases.

elliesec:
Okay, I've opted to move the background handling into the Kinky Dungeon minigame in this case, as that's something of a special case right now. I set `DarkFactor` to 0 inside the minigame to prevent the background from potentially being drawn twice, and then explicitly draw the background in the `KinkyDungeonDraw` function.

I think the other option is to move the minigame special-casing into `DrawGetCustomBackground`, so that it returns `""` if inside a minigame, but I'm not sure what the intended behaviour of items with custom backgrounds is when inside another minigame (i.e. whether all minigames should override custom backgrounds, or whether just Kinky Dungeon should). This seems like a reasonable approach for now.

Ada18980:
Sounds good; Ill incorporate this into my branch so I can make sure it works as intended.

Ben987:
Alright!  Thanks you two :)

PR #2410, opened 2021-05-04
Fix: Ceiling Shackles

jomshir98:
Fixes suspension behaviour of Ceiling Shackles, which was broken by #2401


elliesec (approved):

Ben987:
Sure!  Thanks for that fix :)

PR #2411, opened 2021-05-04
Fix latex posture collar gag

Tyrsen:
Currently it does not apply any gagging effect whatsoever. This should fix that?


Ben987:
Sure!  Thanks for that one, I will try it soon :)

PR #2413, opened 2021-05-04
Feature: Modular item `Effect` support

elliesec:
## Summary

This PR does a couple of things:
* Adds support to modular items for the `Effect` property (which wasn't supported until now, since no items needed it). This supports a request from @Ada18980 to extend the Kigurumi Mask's functionality
* Adds an `Archetype` property to assets, which is automatically populated for items which have an extended archetype - this is primarily a convenience property, but may be useful to scripters, or to support Ada's Kinky Dungeon functionality
* Fixes an issue with the modular item script where module changes were unintentionally getting assigned directly to the underlying asset's `Block`, `Hide` and `HideItem` arrays


Ben987:
Absolutely!  Thanks for that upgrade.

PR #2388, opened 2021-05-01
Refactor: Migrates extended gags to the archetype system

elliesec:
## Summary

Another big one I'm afraid, but this one should be pretty safe. This PR takes advantage of some of the improvements that have been made recently to the extended item and drawing systems to remove redundant scripts and assets across extended gag items. What this does:
* Migrates all extended gags to the archetype system (using the typed archetype introduced in #2263), removing the need for individual extended item scripts
* Adds `DynamicGroupName` entries to all extended gags where assets are duplicated across the three mouth groups and removes the unneeded assets in `ItemMouth2` and `ItemMouth3` (as per #1936)
* Makes a minor change to the `TypedItem` script to allow an item's `Dialog.ChatPrefix` configuration entry to be a function as well as a string, to allow additional flexibility in the generation of chatroom messages for typed items

Affected items:
* Cloth Gag
* Wiffle Gag
* Ball Gag
* Harness Ball Gag
* Duct Tape
* Cupholder Gag
* Pump Gag
* Plug Gag
* Dildo Plug Gag
* Milk Bottle
* Funnel Gag

All items have been manually fully tested to verify that they behave exactly as they did before. I'm intending on following this PR up with a second PR which adds `DynamicGroupName` to the remaining `ItemMouth2` and `ItemMouth3` items in order to remove the remaining unneeded assets that are duplicated across all three mouth groups.

PR #2415, opened 2021-05-04
Refactor: Removes duplicated assets for `ItemMouth2` and `ItemMouth3`

elliesec:
## Summary

This is a follow-up to #2388, and uses `DynamicGroupName` to remove all assets in the `ItemMouth2` and `ItemMouth3` groups which were duplicating assets in `ItemMouth`. It also adds a missing `Audio` entry for the rope gag in `ItemMouth3` that I noticed whilst testing.

I've manually tested all affected items to ensure that they all behave exactly as they did before.


Nina-1474:
You could set DynamicGroupName at group-level for ItemMouth2 to simplify it. 
And I'd argue that the two muzzles that are only in ItemMouth3 should be included in lower mouth slots blocking the higher ones, giving the same effective functionality and allowing the same change for ItemMouth3.

elliesec:
> You could set DynamicGroupName at group-level for ItemMouth2 to simplify it.
> And I'd argue that the two muzzles that are only in ItemMouth3 should be included in lower mouth slots blocking the higher ones, giving the same effective functionality and allowing the same change for ItemMouth3.

I absolutely could move `DynamicGroupName` up to the group level - and that was actually something that I'd planned for a subsequent pull request. I don't disagree about the muzzles either, but this PR was purely about deleting things - I wasn't intending on changing anything.

elliesec:
I've moved `DynamicGroupName` up to group level for both `ItemMouth2` and `ItemMouth3`, but have not moved the two muzzles as I'd rather not increase the scope of this PR beyond `Female3DCG.js`. I'll make that change in a follow-up PR.

Ben987:
This one creates a conflict with another pull request.  Can you double-check?  Thanks :)

elliesec:
Conflicts have been resolved.

Ben987:
That's a big cleanup, I love it.  Thanks a lot :)

PR #2416, opened 2021-05-05
Feature: Futuristic & Cage Muzzles in all mouth groups

elliesec:
## Summary

> ⚠️ Depends on and includes #2415

This makes the Futuristic Muzzle and Cage Muzzle available in the `ItemMouth` and `ItemMouth2` groups, as suggested by @Nina-1474 in #2415, and uses `DynamicGroupName` to share assets between the groups. The version of the Cage Muzzle in `ItemMouth` and `ItemMouth2` blocks the mouth groups above it, and I've added a `BuyGroup` to the Cage Muzzle (the Futuristic Muzzle already shares a `BuyGroup` with the other futuristic gags).


Ben987:
This one also has conflicts.  Can you resolve and test again?  Thanks.

elliesec:
Conflicts resolved & re-tested.

Ada18980:
Im not sure that the futuristic muzzle needs to block the mouth layers above. I see it as identical to the muzzle gag in terms of blocking, so IMO it should be layerable.

elliesec:
> Im not sure that the futuristic muzzle needs to block the mouth layers above. I see it as identical to the muzzle gag in terms of being a flat surface (and it's basically the same as the futuristic gags on ItemMouth), so IMO it should be layerable.
> 
> Fully agree with the cage muzzle blocking the upper layers though.

That's fair enough - I did kind of wonder, but assumed it was intended as an "on-top-only" item due to it being placed in `ItemMouth3` only. I'll remove those blocks.

Ben987:
Thanks a lot!  I'll try it later today :)

PR #2421, opened 2021-05-06
Futuristic Assets

Ada18980:
Adds some new futuristic heels and also adds shiny options for the futuristic bra. These are asset changes with no special features so I lumped them together.

Also, I added HideForPose: ["Spread"] to all the boots that don't have "Spread" assets. This is because the new future heels have spread assets, and it will be needed once HideForPose is fixed to allow enabling ItemBoots options to exist for the spread pose. Right now there is an issue where HideForPose does not remove alpha masks or hidden assets, so the legs appear cut off in the spread pose with some boot assets.


Ben987:
Yes!  I love the look of these new heels.  Thanks :)

PR #2422, opened 2021-05-07
Bugfix/Improvement: Better compatibility for `HideForPose`

elliesec:
## Summary

This PR makes some improvements to the `HideForPose` property when used in conjunction with other asset properties. It moves the handling of `HideForPose` into the build stage of drawing, when sorting a character's appearance layers. 

* This means that it now works in conjunction with alpha masks (i.e. alpha masks won't be used if an item is hidden via `HideForPose`)
* This also fixes a bug where the pet bowl's dynamic text would still get drawn despite the bowl being hidden whilst the character was suspended
* Also includes a few improvements to the type definitions for asset layers and characters


Ben987:
Very good!  Thanks for that upgrade :)

PR #2425, opened 2021-05-07
Kinky Dungeon 1.37

Ada18980:
+Shield and Barrier, two shield spells for blocking projectiles
+Mistcall and Leap, two teleportation spells
+Shroud, a spell to protect yourself from melee
Vibrators now are charged by magical energy from spells, but eventually stop vibrating.
Slowlevels have now been reduced by 1 in severity. Also a bug was fixed that caused stamina to drain extra fast while slowed. Being slowed is no longer as punishing since the game is now harder.
You can now hear chat from within the game!

I had to add a few extra data hooks in ChatRoomChatLog. I also fixed an error in ChatRoomChatLog which would cause the futuristic vibrator to not work properly while deafened.


Ben987:
Thanks for that upgrade, I'm sure players will like it.  Cheers :)

PR #2426, opened 2021-05-07
Prevent players that are leashed from leaving

Ada18980:
Since the current behavior is that if you are leashed, you keep getting dragged to whatever room the leashing player takes you to, this prevents the player from manually breaking the leash connection except by removing the leash. 

Tested to make sure leashing itself wasn't broken, and also verified the functionality and that it can be escaped by no longer being leashable.


Ben987:
That makes perfect sense.  Thanks!

PR #2427, opened 2021-05-07
Prevents softlock from return to chat room when hiding full rooms

Ada18980:
This should prevent players from being stuck in ChatSearch when the room they are forced to rejoin is full but they ignore full rooms.


Ben987:

PR #2428, opened 2021-05-08
Kinky Dungeon 1.4

Ada18980:
Adds in multiplayer streaming! You can click on a player while they are playing in a chat room and watch them play. You can't interact with them yet, but can still hear the latest chat message and they can hear the latest chat message if they are not busy in the game.

The updates occur at the following times:
-At most once every 500ms, at the moment the player clicks or updates the game
-Once per 5s, if the user hasn't clicked

The update is sent to all players who have a) Requested kinky dungeon streaming and b) are in the chat room. The messages are not processed outside of kinky dungeon and do not affect people while they are playing their own games.

The updates are ~1k characters (string format) when inventory and map are sent, but inventory and map are only sent when the map is changed or a restraint is added/removed. They are 300-500 otherwise depending on bullet count and enemy count and the last time stats were sent.  Stats are sent every 5s.

This feature might be buggy on release. Players should watch for Unexpected JSON characters in the console, they will not crash the game but will prevent the game from updating correctly.


Ada18980:
I am aware of linting issues, will fix them in a few hours when I properly wake up

Ben987:
Thanks for that one!  But since it adds more client/server traffic, let's put it on hold until Tessa and Sekkmer changes are tested and we can monitor the lag spikes decrease.  Cheers :)

Ada18980:
> Thanks for that one! But since it adds more client/server traffic, let's put it on hold until Tessa and Sekkmer changes are tested and we can monitor the lag spikes decrease. Cheers :)

Hi Ben,

Will these changes be getting tested in the next beta?

I want to base the next beta's kinky dungeon branch off of this one to avoid merge issues.

Ada18980:
> Thanks for that one! But since it adds more client/server traffic, let's put it on hold until Tessa and Sekkmer changes are tested and we can monitor the lag spikes decrease. Cheers :)

Hi Ben,

Can we take this one off hold now that the hotfix period is over? Or do you want to wait further? I've been basing my latest changes on this commit so if you don't want to include these changes I can disable the feature if needed.

Ben987:
Hi @Ada18980, are you able to evaluate the server side traffic that will be generated by this pull request?  How much data one client will send on average every minutes?  Thanks a lot.

Ada18980:
@Ben987 The PR description states how many characters are sent. I'm not familiar with how to measure server traffic. @jomshir98 Is there any special tool I can use?

jomshir98:
@Ada18980 
I recommend hooking into `ServerSend` like this:
```js
function ServerSend(Message, Data) {
	const encoder = new TextEncoder();
	console.log(`Send ${Message}: ${encoder.encode(JSON.stringify(Data)).byteLength}`, Data);
	ServerSocket.emit(Message, Data);
}
```
But it will return larger numbers, than what actually goes over network, as this is before the data is compressed!

I tried to play around a bit and these are my results:
No matter how many people watches (but only if someone watches) - so this is the best possible option, the game sends ~300 bytes every 5 seconds.
For comparison this is equal to any item activity chat message and ~4 times less than any appearance update.

Also note, that size isn't nearly the most important thing for server as server is CPU limited~
The way this is done means, that server wouldn't do any processing of the message at all (it just relays it to everyone else), so this shouldn't have any noticeable impact on server at all.

Ben987:
300 bytes every 5 seconds is perfectly fine for me.  Thanks you two.

@Ada18980 , there is lots and lots of stuff to test in that one.  Could you pull the master branch and do a full round of test on your side to see if everything works as you want it?  Cheers!

PR #2429, opened 2021-05-08
Bugfix: `HideForPose`/`HideAs` compatibility issue

elliesec:
## Summmary

This fixes an issue introduced by the change to the `HideForPose` property in #2422 where items using the `HideAs` property would result in an error being thrown as the `ItemHidden` group no longer exists.

## Steps to reproduce

1. Equip an armbinder
2. Select one of the strap variations from the extended menu
3. Note that the menu does not exit properly, and an error is thrown in the console


Ben987:
Yes!  Thanks for that fix :)

PR #2430, opened 2021-05-09
Bugfix: Better null safety on combination padlock extended menu

elliesec:
## Summary

I'm not entirely sure on what caused the original problem (possibly a badly consoled-on combination lock), but if a combination padlocked item has no `LockMemberNumber` property, then it will cause an error in the combination padlock's extended item menu. This PR adds better null-safety to the extended item menu to catch this.


Ben987:
I've never seen that one but thanks for the fix :)

PR #2431, opened 2021-05-09
Locker Opacity

Nina-1474:
The Locker and Ventless Locker items have been combined into one Locker with two options to switch between the two.
The old Seethrough and Opaque types have been replaced with a slider to allow the player to fine-tune the transparency.
A throttle function wrapper similar to the debounce wrapper has been created to make the slider's effect more immediate.
The image has been split into layers to give it a frame and a back wall for that 'contained' feel.
The item is now "fixed" to allow kneeling down inside it.
The same changes have been made for the Small Locker and Small Ventless Locker.
Predefined alpha masks have been created for the extra canvas space above and below the main 500x1000 space so they're easier to set on items.


Ben987:
Great idea!  Thanks :)

PR #2432, opened 2021-05-09
Pandora Button Hover

Nina-1474:
A small change so that when the N/S/E/W arrow buttons are darkened and unavailable, they are set to disabled so that they aren't highlighted cyan when the cursor hovers over them.
The code snippet that determines this was wrapped up in a function just to make it a little more readable.
This also has a tweak to one of the NPC lines.


Ben987:

PR #2433, opened 2021-05-10
Blind Flash improvement

Ada18980:
Improves the blind flash so that it fades out rather than being constant. Based on user feedback during the beta.

I incorporated this in a way to try and reuse code from stimulation messages so we don't have completely different mechanisms for the same thing. Hopefully this is OK.


fleisch11:
@Ada18980  I don't know what you mean. i am not going to stop anyone from improving my work, if that is what you mean.

Ben987:
Alright!  Thanks you two :)

Ada18980:
> @Ada18980 I don't know what you mean. i am not going to stop anyone from improving my work, if that is what you mean.

People asked for it to be more gradual, so I just wanted to know what your original intent was. If you consider it an improvement, that is good!

PR #2436, opened 2021-05-10
Notification Popup fix

Nina-1474:
A fix to ensure the screen doesn't freeze if a popup doesn't exist when the reset occurs.


Ben987:
Absolutely!  Thanks :)

PR #2437, opened 2021-05-10
kinky dungeon fixes

Ada18980:
-fix locks not appearing on player
-fix missing text
-fix devious dungeon challenge not locking locks to the player
-fix harness grabbing not working when player is harnessed
-fixed harness grabbing not working when the player doesnt have any evasion buffs
-fix doors slowing down the game by the prompt coming up too often


Ben987:

PR #2439, opened 2021-05-10
Added sound to futuristic vibrator shock

Leilatwentytwo:
heh, forgot to trigger the sound when the shock happens. fixed.


Ben987:
Absolutely!  Thanks :)

PR #2440, opened 2021-05-11
Beta fix for clit ring

Manilla32:
Prevents you from being able to put the clit ring on while wearing clothing items that should block it.


Ben987:
Yes!  Thanks for that one :)

PR #2442, opened 2021-05-11
Character Down Button

Nina-1474:
It's possible for a character's height modifier to be high enough that the top asset zones are offscreen, e.g. when on the Saddle Stand and wearing heels.
Now a "Down" button will appear just like the Up button which centres the character to make all zones accessible.


Ben987:
Absolutely.  Thanks for that little fix :)

PR #2443, opened 2021-05-11
Removed Floating Pixels on Swimsuits

Shiranui-Izayoi:
Likely missed during initial creation nearly impossible to see, but can be seen sometimes such as wearing certain items in very dark or light colours. Like the Catsuit in a full white or black.

My guess is they are left over from a body import or something.

If you want to see them look near the bottom right corner of them most visible on dark backgrounds.

These were the only ones I know of. So are the only ones I fixed.


Ben987:
You have a keen eye.  Thanks!

PR #2444, opened 2021-05-11
new item: Ribbons binding

gatetrek:
adds ribbon binding items and some for in clothing
Screenshot_50
Screenshot_36


Ben987:
Wow!  This is huge!  Thanks a lot for that one, it will be appreciated.

PR #2445, opened 2021-05-11
Fixes for futuristic collar

Ada18980:
-Sci fi panties should be recolored by it
-Fixed crash when applying colors to a futuristic muzzle whose color is not default


Ben987:

PR #2449, opened 2021-05-12
If a leashing player disconnects, the leash is broken

Ada18980:
This makes it so that if the player receives a message that her leasher disconnected, the leash will break and the player will be able to leave the room again. This is to prevent cases where the player is stuck for no discernable reason.


Ben987:
Yes, that's a good one.  Thanks :)

PR #2450, opened 2021-05-12
Bugfix: WebGL alpha blending

elliesec:
## Summary

This fixes an issue with alpha blending where source and destination alpha channels were not being interpolated properly when drawing an image with semi-transparent pixels onto a canvas. This could result in visual artifacts being displayed on some setups (see before/after comparison for this change below).

| Before fix | After Fix |
| --- | --- |
| alpha-before | alpha-after |


Ben987:
Absolutely!  Thanks :)

PR #2453, opened 2021-05-13
Fixed Dungeoneer reputation

Ada18980:
Dungeoneer is supposed to give the max level youve been to, but it instead just increases each level currently. This update fixes that and also adds a bit of cheat prevention (if it's higher than the max level it resets). This is to reset people's scores that are currently at 100 without theoretically penalizing anyone in the future because it is impossible to go above a certain level.


Ben987:
Alright!  Thanks for the fixes.

PR #2456, opened 2021-05-13
MemberNumber = null for locks on player

Ada18980:
Makes it so the player cant just immediately unlock them after getting out of the game >.<


Ben987:
Absolutely!  Thanks.

PR #2458, opened 2021-05-14
Future vibe support for multiple commands at once

Leilatwentytwo:
Changed to allow the futuristic vibrator to receive multiple commands in one chat message.  Did this mainly to allow for both shocks and vibrator mode changes from one chat message.  Could be used for any triggers added in the future.


Ben987:
@Ada18980 , would you like to review this one?  Thanks a lot.

Ada18980:
This seems good; it doesn't cause multiple vibration modes to be set at once (which would be confusing and problematic) but it still allows a shock trigger and a vibe trigger, which is good.

Ben987:
It's time to test that one!  Thanks again :)

PR #2460, opened 2021-05-14
Feature: Serving Tray item

elliesec:
## Summary

As of the validation rework in #2186, the maid tray items can no longer be used on another player due to the `Enable: false` flag being enforced at the validation level. Despite the fact it was only accessible via console, it's become apparent that some players like having explicit access to the item for roleplay purposes.

With that in mind, I've added a new Serving Tray item, which is obtainable once the player joins the maid sorority. I've kept it as a separate item from the existing maid tray items so that they can still be used during maid jobs without other players interfering by removing the item. The Serving Tray is visually identical, but uses the type system to permit empty/full types, and does not have the `Enabled: false` flag, allowing it to be used on other players. I've also added a login check which adds/removes maid items from the player based on whether or not they've joined the maid sorority, similar to most other reputation-based items, which should make sure that anyone that was already a maid gets the item on login.


Ben987:
Sure!  It's a good idea.  Thanks :)

PR #2462, opened 2021-05-14
Fixes futuristic vibrator to work as it did before

Ada18980:
-Unaffected by deafness
-Gagged users have a hard time controlling it

Tested 3 cases: User saying Increase while gagged, someone else saying Increase, and someone else saying Increase while user is deafened.


Ben987:
Sure!  Thanks for that fix, cheers :)

PR #2463, opened 2021-05-14
Futuristic Collar Crash Prevention

Ada18980:
Adds in one last fix which should prevent crashes from the futuristic collar from happening

### DIFF
--- a/BondageClub/Screens/Inventory/ItemNeck/FuturisticCollar/FuturisticCollar.js
+++ b/BondageClub/Screens/Inventory/ItemNeck/FuturisticCollar/FuturisticCollar.js
@@ -244,6 +244,14 @@ function InventoryItemNeckFuturisticCollarUnlock(C) {
 function InventoryItemNeckFuturisticCollarColor(C, Item) {
 	for (let E = C.Appearance.length - 1; E >= 0; E--)
 		if (C.Appearance[E].Asset && C.Appearance[E].Asset.FuturisticRecolor && C.Appearance[E].Asset.Group.Name != "ItemNeck") {
+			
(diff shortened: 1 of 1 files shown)


BondageClub/Screens/Inventory/ItemNeck/FuturisticCollar/FuturisticCollar.js
```diff
@@ -244,6 +244,14 @@ function InventoryItemNeckFuturisticCollarUnlock(C) {
 function InventoryItemNeckFuturisticCollarColor(C, Item) {
 	for (let E = C.Appearance.length - 1; E >= 0; E--)
 		if (C.Appearance[E].Asset && C.Appearance[E].Asset.FuturisticRecolor && C.Appearance[E].Asset.Group.Name != "ItemNeck") {
+			
+			if (C.Appearance[E].Asset.Layer.length > 1 && typeof C.Appearance[E].Color === "string") {
+				let color = C.Appearance[E].Color;
```
elliesec, line 249: It's generally advisable to use `const` here rather than `let`, as the value is never modified. You _could_ also use `Array.prototype.map` below for brevity:
```js
C.Appearance[E].Color = C.Appearance[E].Asset.Layer.map(() => color);
```

elliesec: One other thing to note is that for future-proofing, it would probably be a better idea to use `C.Appearance[E].Asset.ColorableLayerCount` rather than `C.Appearance[E].Asset.Layer`, as not all layers need be colorable (even if the Futuristic items don't have non-colorable layers right now). Obviously that would remove the ability to use `Array.prototype.map` as above.

Ben987:
Thanks for that fix!

PR #2465, opened 2021-05-15
Fix for ribbon application

Manilla32:
Makes the ribbons in the breasts and pelvis slot not be able to be put on through clothes.


Ben987:
Absolutely!  Thanks :)

PR #2466, opened 2021-05-16
Refactor: Add layers to the tennis skirt

SandrineBC:
# Description
I added color layers to the tennis skirt, so cloth and trim can now be colored seperately


Ben987:
Absolutely!  Thanks again for all these assets :)

PR #2467, opened 2021-05-16
fix: metal leg spreader

gatetrek:
fixed for the metal leg spreader from being allowed to be applied when your are legs closed


Ben987:
Sure!  Thanks a lot for that fix!

PR #2471, opened 2021-05-17
Revert to Default Bugfix

Nina-1474:
There was a rare bug where if two players opened an extended item's options screen at the same time, then one of them selected a new option, it was incorrectly changed to the default option. The other player did not receive this update and was desynced. This only seemed to happen for items where the default/first option was not the null type (i.e. arm ropes, chains and cuffs).


Ben987:

PR #2472, opened 2021-05-17
Extended Self-Block Validation Fix

Nina-1474:
When a new extended item option could be blocked by other the current option, e.g. hemp rope on arms, a SelfBlockCheck setting temporarily removed then readded the item for validation. These changes are no longer pushed to the server unnecessarily, which also had the unwanted side-effect of resetting the memoization and rerunning the prerequisite checks constantly.
The steps to refresh character dialog during a character refresh have been moved into their own function and can now be skipped when not wanted.


Ben987:
Since that one is a little bigger, lets wait until the next release.  Thanks!

Ben987:
Time to try this one!  Thanks a lot :)

PR #2473, opened 2021-05-18
Chess Calculation Fix

Nina-1474:
Fix to an error that meant the NPC never makes any chess moves.


Ben987:
Yes!  I will prepare a hotfix.  Thanks a lot.

PR #2474, opened 2021-05-18
Bugfix: Ensures items get removed with timer locks

elliesec:
## Summary

The fix in #2341 meant that more properties were being removed from locked items on unlock than previously, including the `RemoveItem` property, which was later used in `Timer.js` to detect whether the locked item should be removed as well in the case of timer locks with a "Remove item when the lock timer runs out" checkbox, meaning that these locks would no longer remove the item upon expiry. This fixes that so that the redundant properties still get removed, but so does the item if the `RemoveItem` property was set.

## Steps to reproduce

1. Equip a lockable item and lock it with a Mistress, Lover or Owner Padlock
2. Inspect the lock and check the "Remove item when the lock timer runs out" checkbox
3. Wait for the lock timer to run out
4. Note that the lock gets removed, but the item does not


Ben987:
Yes!  Thanks for that one, it will be useful :)

PR #2475, opened 2021-05-18
Refactor: Add layers to the pajama

SandrineBC:
# Description
I separated the pajama pants and the stripes into different layers so they can now be coloured separately.

# Tests
Testing was done on my local installation and all looked good. For different body sizes as well as for different poses.


Ben987:
Yes, yes, yes, thanks!

PR #2476, opened 2021-05-18
Fixes issue where owner/lover lock on a leash makes it impossible to use

Ada18980:
Hotfix based on a beta feature that was added. The ChatRoomCanBeLeashed function was being used inappropriately, I instead modified a function to bypass the lock check if there is no membernumber (signaling that the check is a general one).


Ben987:
Thanks!  I will prepare a hotfix right now.

PR #2477, opened 2021-05-19
Refactor/Bugfix: Password lock refactors & UI improvements

elliesec:
## Summary

This was originally just going to be a slight UI fix to the password padlocks (password, safeword & timer password), but whilst I was in there, I noticed that there was a lot of code duplication and general inefficiencies in the code for the password padlocks, and ended up refactoring them. This PR does a bunch of things, but should not affect the behaviour of the padlocks themselves from a player perspective:

* Fixes an issue where the password inputs weren't wide enough to accommodate all possible passwords (passwords containing several wide letters like 'm' and 'w' could result in a letter or two getting cut off, which is particularly bothersome with the safeword padlock, as it prevents most players from being able to unlock it). This particular issue has now been fixed for the safeword padlock in #2478, although that hotfix didn't address the other two locks.
* Fixes an issue with the timer password padlock where clicking in certain areas could trigger chatroom item updates, despite nothing changing
* Fixes a handful of inaccurate click zones on the interfaces for all three locks
* Changes the password timer padlock so that the timer functionality is not available when the zone is blocked. I did this because when the zone is blocked, half of the padlock's interface becomes unavailable, but the other half doesn't, which I don't feel makes a lot of sense - I've therefore chosen to hide it in line with the standard password padlock
* Removes a duplicated lock member number display from the UI for the timer password padlock (duplicated when the lock has been set to allow random input and a player other than the wearer/lock-setter inspects the lock)
* Refactors a lot of common code around the three password padlocks into common functions that are shared between all three
* Reworks a lot of old code to remove inefficiencies and make use of newer functions like `MouseIn` and `CharacterGetCurrent`, and makes some general readability improvements
* Fixes the JSDoc in a handful of adjacent functions, where my IDE had flagged issues

I've tested all three locks pretty thoroughly, and as far as I can tell, they all behave as they did before.

### DIFF
--- a/BondageClub/Screens/Inventory/ItemMisc/PasswordPadlock/PasswordPadlock.js
+++ b/BondageClub/Screens/Inventory/ItemMisc/PasswordPadlock/PasswordPadlock.js
@@ -1,77 +1,98 @@
 "use strict";
 
+const InventoryItemMiscPasswordPadlockPasswordRegex = /^[A-Z]+$/;
+
 // Loads the item extension properties
 function InventoryItemMiscPasswordPadlockLoad() {
-	var C = CharacterGetCurrent();
-	if ((DialogFocusSourceItem != null) && (DialogFocusSourceItem.Property == null)) DialogFocusSourceItem.Property = {};
-	if ((DialogFocusSourceItem != null) && (DialogFocusSourceItem.Property != null) && (DialogFocusSourceItem.Property.Password == null)) DialogFocusSourceItem.Property.Password = "PASSWORD";
-	if ((DialogFocusSourceItem != null) && (DialogFocusSourceItem.Property != null) && (DialogFocusSourceItem.Property.Hint == null)) DialogFocusSourceItem.Property.Hint = "Take a guess...";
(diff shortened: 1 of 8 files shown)


elliesec:
@Ada18980 - As this is a reasonably wholesale change to the code for the password padlocks, it might be worth you having a look over it. For the most part, the actual logic shouldn't have changed, and from the player's perspective, all this change should really do is make the password box a bit wider on all three locks to accommodate for wide passwords. The majority of the changes are just refactors to remove code duplication and trim out dead code or make generally efficiency improvements.

Ada18980:
I will have to try it out. I've pushed PR #2478 to try to address a potential trolling issue ingame, but I think UI improvements are welcome too. Not sure if Ben will include these as a hotfix though.

elliesec:
> I will have to try it out. I've pushed PR #2478 to try to address a potential trolling issue ingame, but I think UI improvements are welcome too. Not sure if Ben will include these as a hotfix though.

Yep, on the surface, this doesn't do too much more than your PR (although it also does it for the password/timer password padlocks) - and I think that's still worth pushing through as a hotfix in the interim. I'm not expecting this one to go through as a hotfix this release, but it does also contains a lot of general code improvements and other small bugfixes to the locks.

Ada18980:
Yep. This also makes it much easier to implement the futuristic collar allowing you to lock/unlock combination/password locks

Ben987:
It's time to test that big one!  Thanks a lot :)

PR #2478, opened 2021-05-19
Increases size of text field for safeword padlock

Ada18980:
To prevent passwords that are too big to see the last letters of, inviting trolls.

Suggesting hotfix due to small size of change and troll potential.


Ben987:
Sure!  I can hotfix that one also :)

PR #2481, opened 2021-05-19
Avoid removing keys when picking lock in kinky dungeon

Ada18980:


Ben987:
Thanks for that one also, I'll hotfix it.

PR #2482, opened 2021-05-19
Refactor: Add layer to the admiral skirt

SandrineBC:
# Description
I added layers to the admiral skirt, so that stripe and cloth can now be coloured seperately.

# Test
I tested this on my local instance of BC and everything worked fine


Ben987:
Absolutely!  This is great, thanks :)

PR #2483, opened 2021-05-19
Variable Suspension Height

Nina-1474:
~~> Depends on #2472~~

---

### Extended Item Subscreens

The ExtendedItem code now supports the ability to use a custom screen for an individual item option. Selecting the option opens that screen instead of exiting the dialog. Entering the options list screen again shows the selected option highlighted green instead of grey to indicate it can be selected again.
Some cleanup has been made to ExtendedItem, in particular changing the button colour determination from a giant ternary to a readable function.

### Variable Height Suspended Hogtie

The Suspension Hogtie option for Hemp Rope on arms has been configured to use a new height-setting subscreen. The screen features a vertical slider, a numerical 1 to 100 input field, and a Confirm button. 
The character now starts at hogtie height (the floor) instead of the ceiling, with the suspension rope extending all the way up. Moving the slider or changing the number will move the character vertically, allowing the player to set the exact height. The character appearance changes on-screen but is only pushed to chatrooms on hitting Confirm, and is rolled back to the previous option/setting otherwise.
The "ItemHidden" configs for other items have been fixed up as some items were unnecessarily hiding the suspension rope.


BondageClub/CSS/Styles.css
```diff
@@ -331,7 +331,8 @@ input[type=range]::-webkit-slider-thumb {
 
 input[type=range][data-thumb=lock]::-webkit-slider-thumb,
 input[type=range][data-thumb=blindfold]::-webkit-slider-thumb,
-input[type=range][data-thumb=lightbulb]::-webkit-slider-thumb {
+input[type=range][data-thumb=lightbulb]::-webkit-slider-thumb,
+input[type=range][data-thumb=player]::-webkit-slider-thumb {
```
elliesec, line 335: I've been meaning to do this for a while, but now that we've got a bunch of these slider thumbs, it might just be worth checking for the presence of the `data-thumb` attribute, rather than explicitly checking for a bunch of specific values to keep things simple:
```css
input[type=range][data-thumb]::-webkit-slider-thumb {
    ...
}
```
I can do this in a separate PR if you'd prefer, but figured I'd make the suggestion while you're in the area (same to the other selectors further down).

Nina-1474: Done. (An attempt I made at building these settings dynamically in JS was clearly the wrong approach then!)

Ayesha-678:
Glad to see this finally done, great work.

Ben987:
This is impressive!  I love it.  Thanks a lot for that one :)

PR #2484, opened 2021-05-19
ExtendedItem Page Button

Nina-1474:
Changed the page changing buttons on the Extended Item option select screen from two arrow icons into a back/next button containing text "Page x / y".


Ben987:
Sure!  That's a good idea.  Thanks!

PR #2487, opened 2021-05-20
Bugfix/Refactor: Removes `AccessVulvaSuitZip` prerequisite & `Expose` on `SuitLower`

elliesec:
## Summary

PR #1475 effectively removed the use for the `AccessVulvaSuitZip` prerequisite on items as it added `Expose` entries to all `SuitLower` items. However, the prerequisite itself was never removed, so when the stockings were copied to the `SuitLower` group in #2009, this meant that they were the only items in the group which block vulva access, which didn't really make sense (with the exception of the pantyhose item). Rather than adding `Expose` entries to the stockings, I've entirely removed the `AccessVulvaSuitZip` prerequisite and all `Expose` entries on items in the `SuitLower` group, as the prerequisite wasn't actually being used, so it didn't make much sense to keep around.

This PR therefore fixes an issue where stockings in the `SuitLower` group block vulva access, and doesn't change the behaviour of any other items in the `SuitLower` group, as all other items had `Expose` entries which circumvented the prerequisite.

### DIFF
--- a/BondageClub/Assets/Female3DCG/Female3DCG.js
+++ b/BondageClub/Assets/Female3DCG/Female3DCG.js
@@ -375,14 +375,14 @@ var AssetFemale3DCG = [
 		PreviewZone: [0, 450, 500, 500],
 		Asset: [
 			{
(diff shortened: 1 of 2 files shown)


Ben987:
Sure!  Thanks for that fix, cheers :)

PR #2489, opened 2021-05-22
pair of shoes

gatetrek:
adding in this pull two pair of shoes.
added the breast binder to the breast zone and the bra 
added the straps of the armbinder to the clothing asset to let people customise more 
Screenshot_2


Ben987:
They look awesome!  Thanks a lot :)

PR #2490, opened 2021-05-23
Admiral Top Split for Multicolour

Nina-1474:
Split the Admiral Top clothing into three separately colourable layers.


SandrineBC:
_sighs and throws her pr into nirvana_

SandrineBC:
_sighs and throws her pr into nirvana_

Ben987:
Very nice!  Thanks a lot :)

PR #2494, opened 2021-05-23
Bugfix: `SetPose` property validation

elliesec:
## Summary

It is currently possible to use console to set the `SetPose` array on an item's `Property` to a value other than an array, which can cause another player's game to crash when they click on the character with the offending property (or if the property is set on them). This adds a sanitization check in `Validation.js` which ensures that `SetPose` entries on an item's property are valid string arrays, whose entries are all valid pose names.

## Steps to reproduce

1. Equip an extended item on a character
2. Using console, set `Property = {SetPose: "Broken"}` on the item
3. Note that the game will crash for the player on whom the property was set, and for anyone else, clicking on that character will crash their game, with the following stack trace:

```
Character.js:720 Uncaught TypeError: SetPose.find is not a function
    at CharacterItemsHavePoseAvailable (Character.js:720)
    at CharacterCanChangeToPose (Character.js:696)
    at Object.CanChangeToPose (Character.js:244)
    at CharacterCanKneel (Character.js:1421)
    at Object.CanKneel (Character.js:72)
    at ChatRoomMenuDraw (ChatRoom.js:1155)
    at ChatRoomRun (ChatRoom.js:1100)
    at DrawProcess (Drawing.js:1202)
    at MainRun ((index):385)
CharacterItemsHavePoseAvailable @ Character.js:720
CharacterCanChangeToPose @ Character.js:696
CanChangeToPose @ Character.js:244
CharacterCanKneel @ Character.js:1421
CanKneel @ Character.js:72
ChatRoomMenuDraw @ ChatRoom.js:1155
ChatRoomRun @ ChatRoom.js:1100
DrawProcess @ Drawing.js:1202
MainRun @ (index):385
requestAnimationFrame (async)
TimerProcess @ Timer.js:252
MainRun @ (index):386
```

### DIFF
--- a/BondageClub/Assets/Female3DCG/Female3DCG.js
+++ b/BondageClub/Assets/Female3DCG/Female3DCG.js
@@ -4431,6 +4431,12 @@ var PoseFemale3DCG = [
 	},
 ];
 
+/**
+ * List of all available pose names in the game
+ * @constant {string[]}
+ */
+var PoseFemale3DCGNames = PoseFemale3DCG.map(pose => pose.Name);
+
 // 3D Custom Girl based activities
 var ActivityFemale3DCG = [
 	{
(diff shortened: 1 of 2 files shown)


Ben987:
I've never seen that one, but thanks for the fix :)

PR #2495, opened 2021-05-24
[Asset] {Natsuki} Latching Pony Bit Gag

Codefreak64:
This makes it so that the HarnessPonyGagBits is extended, having both a latched and unlatched state. giving both speech and more RP actions for club members.

-makes item extended
-adds a new "state"
-more versatility with pony play RP

the only issue is that players will need to RE-collect the bit gag if they have it already but its minor and doesn't break anything.


Ben987:
Absolutely!  Thanks a lot :)

PR #2497, opened 2021-05-24
ribbons for the torso and lips.

gatetrek:
added two new types of ribbons assets for the torso and a gag
Screenshot_15
Screenshot_9


Ben987:
Yes, yes, yes!  That's very cute.  Thanks a lot :)

PR #2499, opened 2021-05-25
New asset: Dental gag

SandrineBC:
# Description
This PR adds a new gag that is wedged into the mouth of the target player and can force the target player to keep her mouth permanently shut or permanently open. Depending on the setting, different gag levels are applied. In general, the gag remains hidden.

# Testing
I tested this on my local version of BC with one and two players and everything works fine.

# Side note
To have this gag's effects work properly, `Effect: []` had to be added to the asset description. As the HarnessBitGag seemed to have the same problem, I added the empty array there as well.


Ben987:
This one creates a few conflicts with other pull requests.  Can you resolve them and test again?  Thanks a lot :)

Ben987:
That was quick, thanks a lot!  And thanks for the HarnessPonyBits fix :)

PR #2501, opened 2021-05-26
Modified Hand Mask

ItsJustMeVerity:
Made a quick and simple change to the masking on the Hands asset to remove the random section of arm it included to allow for assets with longer, tighter-fitting sleeves without people having to mess about with layer priorities. 

Tested with all current long-sleeve assets, and doesn't visibly impact those whatsoever.

HandPreview


Ben987:
Absolutely!  Thanks a lot :)

PR #2503, opened 2021-05-26
Refactor: Move sturdy leather belts to extended items config

SandrineBC:
# Description
Moves all instances of the SturdyLeatherBelts asset to the new extended item configuration. This relates to
* ItemArms
* ItemLegs
* ItemFeet

This also removes the need to undress a player to add more than one belt. The clipping issues, this prerequisite tried to avoid, happened only with the first of the ItemArms belt and thus didn't add anything to the game. Instead, this set of belts now hides all clothes with too bad clipping issues.

# Testing
I tested that on my local installation of BC with one and two players and everything worked fine.


Ben987:
You're on fire today!  I'll try to test all of these really soon.  Thanks a lot :)

PR #2505, opened 2021-05-26
Added transperancy to some hogtie poses

SandrineBC:
# Description
Some hogtie poses missed a transparency between the boobs. This PR fixes this
Prior to the pr
Normal_Asian
Now:
Normal_Asian


Nina-1474:
That spot isn't meant to be transparent; it's the white shine of the player's mid-section. It doesn't look amazing but personally I think it still looks better having the black outlines fade to nothing and skin sections without black outlines.
image

Ben987:
Correct, that spot isn't meant to be transparent.  But thanks for checking :)

PR #2508, opened 2021-05-27
Asset: Ceiling & Floor shackles update

jomshir98:
Updates Ceiling & Floor shackles to work better with new Seel Cuffs set.
Also bumps priority of Ceiling Shackles to 31, fixing some clipping with body-fitting clothes


Ben987:
Awesome!  Thanks for that update :)

PR #2509, opened 2021-05-27
Silk Straps (ItemPelvis Base)

Ayesha-678:
Added a new binding type: silk straps.
Currently only in the pelvis area but i'll be adding more groups soon!


Ben987:
It's great to see you back!  Thanks for this update :)

PR #2510, opened 2021-05-28
Feature: Archetyped item validation improvements

elliesec:
## Summary

This PR updates the modular and typed item archetypes to permit a few extra features around validation:
* Both archetypes now support a per-option `ChangeWhenLocked` property - if set to false, players will not be able to change the locked item's type away from the flagged type unless they are able to open the item's lock
* Both archetypes now support the `ChangeWhenLocked` at the top configuration level - if set to false, players will not be able to change the item's type at all unless they are able to open the item's lock
* The typed item archetype now supports a new `Validate` property, which allows a function to be passed into the configuration, much like the existing extended item validation functions

These changes should allow us to shift more items over to the archetype framework fairly easily, as the lock prerequisite is a relatively common usage of custom validation functions in existing items that use the extended item script. This should also unblock #2503.


Ben987:

PR #2514, opened 2021-05-30
[Asset fix] [Natsuki]: Over The Head Visual Fix

Codefreak64:
The right side of the pose 'OverTheHead' was misaligned and did not line up with the lower bodies, this fixes that misalignment by editing the images


Ben987:
I've never noticed that one before, thanks!

PR #2515, opened 2021-05-30
Silk strap bindings (item torso base)

Ayesha-678:
Silk bindings in the torso area


Ben987:
Yes, yes, yes!   Thanks for more of these :)

PR #2517, opened 2021-05-31
Refactor: Leather armbinder and straps

SandrineBC:
# Description
This PR moves the extended scripts of the leather armbinder and leather straps to the new config files.

# Testing
Both items were tested on my local instance with one and two players and everything looks fine.


Ben987:
You're on fire today!  I'll try to test all of these really soon.  Thanks a lot :)

PR #2518, opened 2021-05-31
Refactor: Leather and ornate arms cuffs

SandrineBC:
# Description
This moves the extended item scripts for the ItemArms leather and ornate cuffs to the configuration for extended items.
The steel cuffs, already using the extended item config were changed a bit so that they use the same configuration as the other two now.

# Testing
Was done on my local client with one and two players. All looks good on my side


Ben987:
Github isn't handling these multiple pull requests on the same file very well.  Is there another way you could make them so we don't get all the strange conflicts?  It's getting worse with each merge.  Thanks a lot.

SandrineBC:
> Github isn't handling these multiple pull requests on the same file very well. Is there another way you could make them so we don't get all the strange conflicts? It's getting worse with each merge. Thanks a lot.

Ugh... I see, what you mean. I'll merge all of my PRs into one big one and close the others.

SandrineBC:
Will be merged with the other PRs

PR #2520, opened 2021-05-31
Refactor: Various arms items to extended configuration

SandrineBC:
# Description
This PR changes the extended item script of various arms items to the extended item configuration:
* Wirst shackles
* Leather cuffs
* Ornate cuffs
* Steel cuffs
* Leather strait jacket
* Canvas strait jacket
* Collar cuffs

# Testing
Was done on my local version of the club and everything works fine.

PR #2521, opened 2021-05-31
Refactor: Leather and canvas straitjacket

SandrineBC:
# Description
This PR moves the extended item scripts for the leather and canvas strait jacket to the extended item configuration

# Testing
Was done on my local version of the client with one and two characters. All looked good for both items.


SandrineBC:
Part of #2520

PR #2523, opened 2021-05-31
Preserve scroll location in TextAreaChatLog on window resize

Leilatwentytwo:
When resizing the window, this prevents your location in the TextAreaChatLog, in chat rooms, from changing.  It keeps you at the exact end of the chat log if you were there at the beginning of the resize event, or otherwise keeps you at about the same position in the chat log.  Triggered by a resize event listener, which is added and removed when the chat log is added and removed.


Ben987:

PR #2526, opened 2021-05-31
Improvement/Fix: Asset check update for extended items

jomshir98:
Updates asset check to keep up with latest extended item archetype changes.


Ben987:
Once again, you'll have to test that one.  But as long as there's no bugs for players it's all good.  Thanks!

PR #2527, opened 2021-06-01
Change: small changes to alpha masks of two items

fleisch11:
•bolero straitjacket:
changes alpha mask, so that the neck is visible 
(reason for this is how the combination of (leather bolero + bolero straitjacket + latex posture collar gag) looks)

•prison lockdown suit
removes itemieck from hideitem
adds alpha mask for itemneck


Ben987:
Thanks for these fixes!  Cheers :)

PR #2528, opened 2021-06-01
Improvement: Drawing optimisations

jomshir98:
Creates function `DrawImageEx` unifying functionality of most image drawing functions in more optimised way.
Also improves how inversion (suspension) and blindness is drawn.

Performance effect: Character drawing while bindfolded or inverted is ~40% faster.

There is no visible change, everything looks the same.
Tested with various combinations of inverted characters and blindfolds.


Ben987:
The "..." code crashes some browsers, we've had trouble with it before.  Is there another structure you can use?  Thanks!

Ben987:
Thanks a lot!  Players with slow machines will appreciate :)

PR #2529, opened 2021-06-01
Improvement: Clientside GetAllowItem

jomshir98:
Implements clientside `ChatRoomGetAllowItem` from server to allow for instantaneous resolving of access to target character, removing delay after clicking.

Original behaviour is still preserved, including aking server and if there is mismatch, server has final say.


Ben987:
I like it!  Thanks a lot :)

PR #2530, opened 2021-06-02
Fix: Correct some wrongly hidden items on the short pencil skirt

SandrineBC:
# Description
This PR corrects a copy/paste error that hid too many items on the short pencil skirt


Ben987:
Yes!  Thanks for that fix :)

PR #2531, opened 2021-06-02
fix sync reset test

NoneNoname:
@jomshir98 suggested small changes to the timing and a fix so the timer won't always reset


jomshir98 (approved):

elliesec (approved):

Nina-1474 (approved):

Ben987:
Very well!  Thanks a lot for the fix :)

PR #2532, opened 2021-06-03
Animation Quality Preference

Nina-1474:
At the moment most functions that update the character appearance repeatedly are limited by debouncing (delaying updates until a set time has passed without change) with a wait interval of 100 ms.
This feature creates a new "Animation Quality" preference in the Graphics settings page to let users indirectly decide this limit themselves. The options are Low/Medium/High/Max corresponding to wait intervals of 200/100/50/0 milliseconds respectively. By default the setting will be Medium = 100 which is the current behaviour for most cases.
For Low and Medium these character refreshes will still be debounced, while for High and Max they'll be throttled instead (delaying updates until a set time has passed) which will provide a faster smoother response to appearance setting changes, provided the user's device can handle it. These refreshes all occur entirely client-side and won't affect the server.

At the moment this affects:
- The colour picker
- The hemp rope suspension height setting
- Opacity sliders for
	- Halo
	- Lockers
	- Wooden boxes 
- Text changes for
	- Wooden boxes
	- Transport jacket

While the term 'animation' isn't entirely accurate for current uses, I'm hoping to create an animated item implementing this setting and the impact will be very noticeable there.

### DIFF
--- a/BondageClub/Screens/Character/ItemColor/ItemColor.js
+++ b/BondageClub/Screens/Character/ItemColor/ItemColor.js
@@ -211,7 +211,7 @@ function ItemColorDrawDefault(x, y) {
  * indices
  * @const {function(): void}
  */
-const ItemColorOnPickerChange = CommonDebounce((color) => {
+const ItemColorOnPickerChange = CommonLimitFunction((color) => {
 	const newColors = ItemColorState.colors.slice();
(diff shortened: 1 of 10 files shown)


Ben987:
Thanks a lot!  I'm curious to test that one.  Cheers :)

PR #2533, opened 2021-06-03
fix credit over scroll

NoneNoname:


Ben987:
Yes!  I've seen that bug a few times before.  Thanks a lot :)

PR #2534, opened 2021-06-03
Added Rest head activity

Manilla32:
New activity to rest your head in someone's lap.


Ben987:
That's a fun idea, thanks a lot :)

PR #2535, opened 2021-06-03
New Binding Type: Thin Leather Straps (Item Arms Base)

Ayesha-678:
A new binding type, thin leather straps that lock, in the arm area, more areas coming soon!


Ben987:
Yes, yes, yes!  Absolutely.  Thanks for that one :)

PR #2536, opened 2021-06-04
Refactor: Zipties to extended item config

SandrineBC:
# Description
This moves the zip ties to the new extended item config and also adds the new option to trigger expressions for extended items. The former zip ties already had this options, but the extended item scripts never made use of this option.

# Testing
I tested this with one and two characters and several items on my local installation of BC and all worked fine.


Ben987:
This one creates a conflict with other pull requests.  Can you resolve it and test again?  Thank you very much.

Ben987:
Thanks a lot!  You're on fire.

PR #2537, opened 2021-06-04
Feature: Release submissive from ownership

Jomshir98:
**Depends on [Server#91](https://github.com/Ben987/Bondage-Club-Server/pull/91)**

- Reorganises relationship menu to fit new option and moves rarely used ones into submenu.
- Adds option to release submissive from the bond.
  - The option requires you to first remove any collar sub might be wearing
  - The option is intentionally *after* back button to prevent case, where the option and confirmation button could be on same position.
- Adds server messages for when this happens


Ben987:
Thanks a lot for that one.  But there is way too much stuff going on the server side.  I'll have to learn all that coding first, or maybe I'll code a simpler solution myself one day.   Let's put that one on hold for now.  Thanks again.

Ada18980:
@Ben987 Are there more server changes this version?

elliesec (approved):
@Ben987 - I think it's worth giving this one a try this release if at all possible. Server stability is looking pretty good right now, and we've not had any changes in a month or so. I've had a look through both this and the server-side component, and pulled them down for testing, and everything seems to work as expected. It's a feature a fair few players had asked for, and I feel like it's something that's been missing from the club for a long time.

The bulk of the changes on the server are actually just moving chunks of existing code to make the `AccountOwnership` function a bit cleaner (and refining them a little in some cases), and the only real new code in there is the handling of the new `Release` message (line 1519), which is fairly straightforward.

If you're unsure about it, you should be able to safely merge and deploy the server PR without needing this one and see how that goes. I don't anticipate any issues with it.

Ada18980 (approved):

PR #2538, opened 2021-06-04
Refactor: Collar ropes into layers

SandrineBC:
# Description
This divides the three collar rope types into layers, so that the plate does not her collared, with the rope

# Test
Works fine on my local installation of BC


Ben987:

PR #2539, opened 2021-06-04
fix and simplify LoginCreditsPosition check

NoneNoname:


Ben987:
Thanks for the fix!  I will try it right now.

PR #2540, opened 2021-06-04
Thin leather straps (item torso)

Ayesha-678:
Added thin leather straps in the torso section, along with a couple of fixes to the arm section


Ben987:
Very nice!  You're on fire.  Thanks a lot :)

PR #2545, opened 2021-06-07
Reconfiguring blackout lenses to not hide mask or glasses

kimei-nishimura:
First PR so hopefully didn't break anything.  This should be a nearly invisible change and would allow the blackout lenses to work in a more logical fashion with glasses and masks. Resubmitting after adding a priority change after discussion with Nina and Ace__ on Discord.


ace-1331 (approved):
tested with multiple gags, hoods, masks, glasses, blindfolds and everything looks good. Re-looked at the priorities and it makes sense as well

Ben987:
Thanks a lot for that first PR, I will test it very soon.  Cheers!

PR #2548, opened 2021-06-08
Feature: Hide item based on attribute

SandrineBC:
# Description
This is the PoC of a new asset feature: The ability to hide items based on certain attributes.
Within this PR all skirts in the ClothLower group got the attribute "Skirt" now and the regular strait jacket and the leather strait jacket now hide all clothes with that attribute.
This way, there is no need to check all `HideItem` entries on all assets, if a new item is introduced, it just have to have the correct attribute and will be hidden automatically, without changes to other assets.

# What it does
## Details
The main changes happened in the `Asset.js` and `Appearance.js`. 
In `Asset.js` the two new attributes `Attribute` and `HideItemAttribute` are introduced on asset level.
In `Appearance.js` a new check as added to `CharacterAppearanceVisible` where all items are hidden, if one of their attributes match with one of the entries in the hide attributes list

## Testing
I tested this with the local installation of my BC and everything looks good.


Ben987:
Sure!  This could be useful in many situations.  Thanks :)

PR #2551, opened 2021-06-08
Vacuum Bed Deluxe

TessaTech:
# Vacuum Bed Deluxe
This new vacuum bed for the club features a couple of new options along with an alternative design

The bed consists of multiple layers and makes it possible to colorize body, sheet and frame seperately. It utilizes the opacity property to allow for a freely chosable transparency. The oppacity is capped between 25% and 100%.

It also allows two differnt poses for arms and legs, resulting in a total of 4 different poses.

Vacbed2_20210606_4
*[Example image with 75% Opacity(Blue) and 60% Opacity(Purple)]*


Ben987:
Very nice!  But why are you logging your variables in the console?  Is that some debug information that was left over?

Ben987:
I'll merge it and remove the console.log myself.  I'd like to test everything before beta tomorrow.  Cheers :)

TessaTech:
Yes, the log output there was an oversight. Forgot to remove it after debugging.

PR #2552, opened 2021-06-08
[Asset] [Natsuki]: Clear Vacbed

Codefreak64:
Add in the Vacbed, but clear. 
now you can see through and really see whats inside.

-Adds Clear Vacbed Item
-Item located in Devices
-Comes with the Vacbed (the opaque one)


Ben987:

PR #2553, opened 2021-06-09
Futuristic Heels are now a clothes item as well.

Ada18980:
Adds a clothes version of the futuristic heels using DynamicGroupName and the new extended typed archetype.


Ben987:
Yes!  Thanks for that one :)

PR #2554, opened 2021-06-09
Bot Command

Ada18980:
Implements a /bot command which does nothing except send a hidden message to all other players. This message will get ignored on the default client, but a modded client might be able to interpret the bot command. This specifically enables bots that work even if the player can't whisper due to immersion settings.


Ben987:
Sure!  It could be useful for mods.

PR #2555, opened 2021-06-10
Missing preview image for vacbed deluxe

TessaTech:
Adds the missing preview image for the vacbed deluxe.
Also removes a dupilcate RemoveAtLogin key in the asset data


Ben987:
Great!  Thanks for that one.

PR #2556, opened 2021-06-10
Bugfix: Error thrown when inspecting the Bondage Bench

elliesec:
## Summary

Probably not strictly a beta fix, but this fixes an error that can occur when inspecting items using the `AllowLockType` property (currently only the Bondage Bench).

## Steps to reproduce

1. Player A equips a Bondage Bench on player B and immediately exits the extended menu
2. Player B clicks on themselves and selects the `ItemDevices` slot
3. The following error is thrown:

```
Dialog.js:733 Uncaught TypeError: Cannot read property 'Type' of undefined
    at DialogMenuButtonBuild (Dialog.js:733)
    at DialogInventoryBuild (Dialog.js:846)
    at DialogClick (Dialog.js:1300)
    at CommonClick (Common.js:258)
    at Click ((index):385)
    at HTMLCanvasElement.onclick ((index):467)
```


ace-1331 (approved):

Ben987:

PR #2542, opened 2021-06-06
Refactor: Various arm items

SandrineBC:
# Description
This PR moves four arms items to the extended item configuration:
* MermaidSuit
* TightJacket
* TightJacketCrotch
* WrappedBlanket
* Ribbons

# Testing
Was done on my local installation of BC with one and two characters. All looked good on my side

PR #2559, opened 2021-06-10
Bugfix: Reinstates accidentally removed extended script for web

elliesec:
## Summary

The script reference for the web arms item was accidentally removed in #2542, which means that the extended menu for the web no longer works. This PR reinstates the script.


ace-1331 (approved):

Ben987:
Yes!  Thanks for that one.

PR #2560, opened 2021-06-10
Bugfix: Prevents `HideItem` declarations from being short-circuited

elliesec:
## Summary

As part of #2548, some of the item-hiding logic was refactored to try to merge handling of `item.Property.Hide` and `item.Property.HideItem`. However, this would cause the `else if` block handling `item.Asset.HideItem` to be short-circuited incorrectly. This PR reverts that change, but maintains the new `HideItemAttribute` handling.

## Steps to reproduce

1. Equip a pair of Futuristic Earphones and exit the extended menu
2. Equip a gas mask (`OldGasMask`) on top
3. Note that the Futuristic Earphones are still visible, despite being listed in the gas mask's `HideItem` array


Ben987:
Absolutely!  You're a machine, thanks again :)

PR #2562, opened 2021-06-10
Bugfix: Reinstates accidentally deleted sturdy leather belt configuration

elliesec:
## Summary

#2520 accidentally deleted the extended item configuration for the sturdy leather belts, presumably as a result of a merge conflict, meaning that the extended menu for the belts no longer works. This reinstates the missing configuration.


Ben987:
Great!  Thanks for that extra fix :)

PR #2563, opened 2021-06-11
Adds a few keywords to the auto shock blacklist

Ada18980:
There is a blacklist of words that will always get the player shocked even if the player does an emote. For example, writing "/me shouts into her gag" will still shock the player even if they aren't sending a normal chat message. This just adds a few extra keywords to catch clever subs.


Ben987:
Good idea.  Thanks a lot :)

PR #2564, opened 2021-06-11
Revert "Modified Hand Mask"

Nina-1474:
Reverts Ben987/Bondage-College#2501

Reverting this on behalf of Verity. It wasn't needed for any current assets but causes issues with some skirts. Long-term another solution will be found for long sleeves.


elliesec (approved):

Ben987:
That was quick!  Thanks a lot :)

PR #2567, opened 2021-06-12
Bugfix: Prevents difficulty from increasing when sending single item updates

elliesec:
## Summary

This fixes an issue where single item updates could cause an item's difficulty to increase if the item's current difficulty is sent along with each update. I've also added some clarification to some of the JSDoc documentation regarding difficulty when adding items. Not strictly a beta fix, but would be good to include in this release.

## Steps to reproduce

1. Player A equips an extended item (with types) on player B
2. Player A leaves the room and re-enters (this ensures the difficulty value is synced properly between players, as it can go out of sync - this PR also fixes that)
3. Player A changes the type of the item on player B
4. Note that the difficulty of the item has increased by the base difficulty of the asset
5. Steps 2-4 can be repeated to increase the difficulty indefinitely


Ben987:
Yes, yes, yes!  This one has been annoying for a little while.  Thanks a lot :)

PR #2569, opened 2021-06-12
Allow permission setting for strong locks on Extreme difficulty

Nina-1474:
Locks deemed to be "strong" (namely CombinationPadlock, PasswordPadlock, TimerPasswordPadlock, and HighSecurityPadlock) are forced to Limited permissions on the Extreme difficulty since otherwise it's quite easy to be put into a situation that's not escapable by any reasonable means.
There have been objections and confusion around this, particularly since the lower difficulty of Hardcore doesn't have this limitation on the locks and therefore is in one sense more restrictive than Extreme.

This PR rectifies that by making the Item Permissions menu mode accessible on Extreme difficulty, but only allowing permission changes on those four locks, and not any other items. The player can only toggle the locks between limited/orange and open/green, and can't set them to blocked/red.

The validation that enforces Extreme settings has been relaxed to allow these new settings. There are some edge cases where the Limited permissions could be reapplied to the locks quietly but the player can at least re-open permissions again.
This change does introduce that possibility of being pseudo-permanently stuck, but the default lock setting is still Limited as before, so opening those permissions is a risk the player must accept. There's still the fallback of reducing difficulty at least.


Ben987:
Sounds good to me.  Thanks :)

PR #2570, opened 2021-06-13
Extended Item Type Permissions on Extreme Difficulty

Nina-1474:
On the Extreme difficulty, the button to switch to Permissions mode for extended items was accessible, allowing specific options to be blocked. These settings were at least reset on logging in again but could still be changed temporarily. Now the button will be hidden.


Ben987:

PR #2572, opened 2021-06-13
Character In Previews Setting

Nina-1474:
The appearance screen's "Show character in previews" button state is now saved to the player preferences so that it doesn't have to be re-selected after every login.


Ben987:
Yes!  That's a good idea.  Thanks a lot.

PR #2573, opened 2021-06-13
Bed Position

Nina-1474:
The Bed is being given an override height so that it isn't pushed offscreen in the petsuit, or moved around by things like heels.


Ben987:

PR #2574, opened 2021-06-13
Fix pink orgasm filter + refactor chat room drawing

Ada18980:
Refactors the custom drawing list in ChatRoom.js in order to support features besides the VR headset. This is to fix an issue with the orgasm filter that's in release (and also beta), but I did a little more work to ensure the fix is not a special case but supports future features (like a new immersion option I am working on)


Ben987:
It's time to test that one, thanks a lot!

PR #2575, opened 2021-06-13
Disable Examine While Blind only works outside of the virtual realm

Ada18980:
The virtual realm is supposed to allow players a bit more freedom with each other only. Also, this allows switches to edit their sub's settings while in the virtual realm, even if they have immersion options set, so now people with the option set who aren't bound can edit the restraints on someone who is bound.

Dependency on #2574


Ben987:
This one creates conflicts with other pull request.  Can you resolve them and test again?  Thanks a lot.

Ada18980:
Fixed

Ben987:
Thanks for the quick fix.  Cheers :)

PR #2576, opened 2021-06-14
Immersion refactor + only show adjacent players while partly blind

Ada18980:
Refactors the immersion page and rearranges some options so the sensory deprivation settings are clustered together at the top. Also adds the second room (Thanks so much @Nina-1474 for making it easy!!)

Since the second page needed an actual option to justify adding it, I've added a new option to hide adjacent players while blind. This makes it so if you are partially blind you can't see players that aren't right next to you, helping to create a restricted feeling without blocking out all players.

Dependent on #2574. Tested to make sure it works with examining. It does not apply while in the virtual realm.


Ben987:
This one creates conflicts with other pull request.  Can you resolve them and test again?  Thanks a lot.

Ada18980:
Fixed

Ben987:
Awesome!  Thanks again :)

PR #2577, opened 2021-06-14
tweaking fix

gatetrek:
some tweaking fixes to priority to stop clipping with the jeans
added a missing D


Ben987:
Yes!  Thanks a lot for these fixes :)

PR #2578, opened 2021-06-14
Bikini Updates

EmilyR42:
- Added multicoloring and closed-crotch variants for Cute Bikini and Demonique Bikini.


Ben987:
Great idea!  Thanks a lot, there are lots of assets :)

PR #2580, opened 2021-06-15
Fix: Dental gag cannot be changed when locked

SandrineBC:
# Description
Just adds the `CannotChangeWhenLocked` property to the dental gag.


Ben987:
Sure!  Thanks for that fix :)

PR #2581, opened 2021-06-15
Quick add owner and/or lovers to chat room admins list

kimei-nishimura:
Adds buttons to chat room admin to quickly add owner and/or lovers to the room admin list.
adminscreen-changes


Ben987:
That's a good idea!  Thanks a lot :)

PR #2582, opened 2021-06-15
re-doing the images of the prison gag

gatetrek:
with permission from the maker of the asset prison lockdown gag. I've just tweaked with some updates to it and also separated the belts off the mask as well


elliesec (approved):

ace-1331 (approved):

Ben987:

PR #2584, opened 2021-06-16
Bugfix: Adjusts alpha masks & `Hide` entries on the seamless open crotch strait dress

elliesec:
This adjusts the alpha masks and `Hide` entries for the Seamless Open Crotch Strait Dress to fix an issue where some items like skirts would clip through it (see image below).

image


Ben987:
Absolutely!  I'll slide it in beta 3.  Thanks :)

PR #2585, opened 2021-06-17
Strait Leotard Layer Priority Change

Shiranui-Izayoi:
A small but significant change to the Strait Leotard. I changed it so that the Strait Leotard has similar behaviour to the Futuristic Leotard.

That being that the suit itself is really low priority while the arms have a higher priority allowing items to be layered on top of the suit, but below the arms.

I chose to have them on layer 31 as this puts the arms over the top of most clothing items, giving a nice effect where they hug under the breasts.

Personally I think that some other strait jackets could have this behaviour. But I also think it might be a good idea to expand the `OverridePriority` system to allow manipulating of layers individually. Something like if it is in an array it applies each number to a layer but if it is a flat number then it applies to all. Either that or a `OverrideLayerPriority` variable to do that. Similar to `Hide` and `HideItem`

### DIFF
--- a/BondageClub/Assets/Female3DCG/Female3DCG.js
+++ b/BondageClub/Assets/Female3DCG/Female3DCG.js
@@ -2201,10 +2201,10 @@ var AssetFemale3DCG = [
 				]
 			},
 			{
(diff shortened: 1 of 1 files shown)


Ben987:

PR #2587, opened 2021-06-17
Strait Leotard Extended

Shiranui-Izayoi:
This builds on what I did in pull request [#2585](https://github.com/Ben987/Bondage-College/pull/2585).

This makes the Strait Leotard extended using the modular item framework. The default behavior of the leotard has not changed, but it is now possible to make any of the 4 things it previous hid visible. 

That means that clothes can be seen over the top of it, same for corsets etc if the appropriate option is clicked. That said normally Nipple and Vulva Piercings can't be seen due to them being drawn on a lower level. But a few can be. Still I felt that it was good to include the option just in case. 

I have tested it all in offline and it seems to work. I included all the needed dialog and all seems to be working. If this gets accepted then the previous one is not needed. But I felt that this was to big a change to include in the previous one. So I felt I should make it a seperate request.


BondageClub/Assets/Female3DCG/Female3DCGExtended.js
```diff
@@ -757,6 +757,31 @@ var AssetFemale3DCGExtended = {
 				],
 			}
 		}, // SturdyLeatherBelts
+		StraitLeotard: {
+			Archetype: ExtendedArchetype.MODULAR,
+			Config: {
+				ChatSetting: ModularItemChatSetting.PER_MODULE,
```
elliesec, line 763: I don't think you want to be setting this to `PER_MODULE` - that will mean it only looks for one chat message per module of the format "ItemArmsStraitLeotardSetCloth", but you've arranged your chat messages to follow the `PER_OPTION` pattern (which is the default - just removing this line should work).
```suggestion
```

Shiranui-Izayoi:
Ok thanks Ellie this was my first time using the Modular Item System so I copied off of the High Security Strait Jacket. Thanks for pointing that out it is commited.

jomshir98:
@Shiranui-Izayoi 
Instead of setting `HasType: false` for each layer, you can set it for whole item (that way it becomes default for each layer)

Shiranui-Izayoi:
> 
> 
> @Shiranui-Izayoi
> Instead of setting `HasType: false` for each layer, you can set it for whole item (that way it becomes default for each layer)

True thanks for reminding me. I had just noticed that the other Modular items didn't have that.

Ben987:
Time to test that item!

PR #2590, opened 2021-06-18
Advanced Denial Modes

Ada18980:
Adds a new feature to the Deny mode of most vibrators... ruined orgasms!

When the vibrator switches off, it will first edge the player and bring them to the orgasm screen. At this point there is nothing the player can do to orgasm, as the surrender button will be missing and the orgasm will cancel right before the timer runs out.

The mode is manually triggered by the vibrator. It will not be triggered in any other modes, so the change should be mostly safe.

This does not send a chat message to anyone but the player, in an effort to reduce chat spam in chat rooms. Deny mode will remain safe for use in rooms with 9 AFK dolls set to deny, for better or for worse.

Also adds it to the Sci Fi Pleasure Panties for more fun. I asked the creator for permission.


Shiranui-Izayoi:
Yes I approved this. It sounded like a fun alternative. Glad Ada managed to find space in my overcrowed item :)

I REALLY need to get on to making it have either more pages OR more menus similar to how the modular items work. Where I would have `Vibrator Modes` as one part then `Shock Level` then `Advanced Modes`. With the ability to lock and unlock at the top always. Since then I could add other things like voice commands or something. Hopefully this will kick me into gear due to the lack of space :)

Ada18980:
@Shiranui-Izayoi I am going to use Nina's extra pages in my futuristic training belt. Perhaps you can look into that.

Shiranui-Izayoi:
> 
> 
> @Shiranui-Izayoi I am going to use Nina's extra pages in my futuristic training belt. Perhaps you can look into that.

I will look  into it thanks.

Ben987:
That's a fun concept!  Thanks a lot you two :)

PR #2591, opened 2021-06-18
Add support for MU* style poses

tuxyq:
This adds : as a synonym for * and /me for MUD/MUSH users
It also avoids obvious smileys

Posting in this state to gauge if this is something you'd accept, one improvement I'm considering is making it a toggle in the settings as well.


TessaTech:
Yet another synonym for * and /me sounds a bit odd. I'm all for making things individualizable for users but if we add this now we might just as well add other emote formats that are used somewhere else.
I would suggest to at least make this configurable before merging it since most users are used to * or /me and using up a commonly used character like : all of a sudden could annoy some people.

Also there are much more emoticons that start with : and considering them all would making relying on : for general text emotes nothing one would want to get used to.

Like:
> :jumps jumps forward -> _*Tessa jumps forward*_
> :pounces forward ->**Tessa: :pounces forward**

BondageClub/Screens/Online/ChatRoom/ChatRoom.js
```diff
@@ -72,6 +72,25 @@ const ChatRoomArousalMsg_ChanceGagMod = {
...
+	'\'',
+	'[',
+	']',
+	'{',
+	'}',
+	'l',
+	'@',
+	'|',
+	's'
+];
```
TessaTech, line 93: Either this would have to turn all the lower care letters to upper case...

BondageClub/Screens/Online/ChatRoom/ChatRoom.js
```diff
@@ -1488,12 +1507,12 @@ function ChatRoomSendChat() {
 			Dictionary.push({ Tag: "CoinResult", TextToLookUp: Heads ? "Heads" : "Tails" });
 			if (msg != "") ServerSend("ChatRoomChat", { Content: msg, Type: "Action", Dictionary: Dictionary });
 
-		} else if ((m.indexOf("*") == 0) || (m.indexOf("/me ") == 0) || (m.indexOf("/action ") == 0)) {
+		} else if ((m.indexOf("*") == 0) || (m.indexOf("/me ") == 0) || (m.indexOf("/action ") == 0) || (m.indexOf(":") == 0 && (m.indexOf("'s") == 1 || !ChatMuPoseIgnoreCharacters.includes(m[1].toUpperCase())))) {
```
TessaTech, line 1510: ...or this would have to convert the tested letter to lower case
`!ChatMuPoseIgnoreCharacters.includes(m[1].toLowerCase())`

TessaTech (changes requested):
I think the conversion doesn't quite match up here.

tuxyq:
Good point regarding that last part.

I opened this PR to get feedback before I sink time into expanding it into being configurable, if this is a feature that would potentially be accepted, that would be something I'd be looking into.

Current state is the POC

BondageClub/Screens/Online/ChatRoom/ChatRoom.js
tuxyq: yeah, I swapped things around and neglected that I had lowercase in the array :)

TessaTech:
Also another few common smileys that wouldn't be catched by this:
:3, :o, :v, :x, :F
And i'm sure there could be found even more if one searches more thoroughly.

TessaTech:
I'd suggest asking around among the player base and on the discord in the "suggestions" or "work in progress" channels for oppinion.

tuxyq:
Updated how it decides if it parsing the emote, it's more naive, but also a faster method, I can't think of a case where you have a 3 letter emote, hence making it that naive.

tuxyq:
noticed that #1462 will conflict with my changes when/if it's merged, if/when that happens, Ill refactor this to match

BondageClub/Screens/Character/Player/Dialog_Player.csv
```diff
@@ -450,7 +450,7 @@ ChatRoomStimulationMessageVibePlugFront2,,,"The vibrator inside of you keeps you
...
-ChatRoomHelp,,,"Commands available : /help : show this help text, * or /me : write an emote, ** or /action : write an action, click on someone's name on the left to whisper, PgUp/PgDn keys to navigate chat input history, /coin : flip a coin, /dice Y or XdY : roll X dice with Y-sides (from 2 to 100 sides), /afk, /friendlistadd [ID], /friendlistremove [ID], /whitelistadd [ID], /whitelistremove [ID], /blacklistadd [ID], /blacklistremove [ID], /ghostadd [ID], /ghostremove [ID], /kick [ID], /ban [ID], /unban [ID], /promote [ID], /demote [ID]",,
+ChatRoomHelp,,,"Commands available : /help : show this help text, :(if enabled in settings), * or /me : write an emote, ** or /action : write an action, click on someone's name on the left to whisper, PgUp/PgDn keys to navigate chat input history, /coin : flip a coin, /dice Y or XdY : roll X dice with Y-sides (from 2 to 100 sides), /afk, /friendlistadd [ID], /friendlistremove [ID], /whitelistadd [ID], /whitelistremove [ID], /blacklistadd [ID], /blacklistremove [ID], /ghostadd [ID], /ghostremove [ID], /kick [ID], /ban [ID], /unban [ID], /promote [ID], /demote [ID]",,
```
Nina-1474, line 453:
```suggestion
ChatRoomHelp,,,"Commands available : /help : show this help text, * or /me (or : if enabled in settings) : write an emote, ** or /action : write an action, click on someone's name on the left to whisper, PgUp/PgDn keys to navigate chat input history, /coin : flip a coin, /dice Y or XdY : roll X dice with Y-sides (from 2 to 100 sides), /afk, /friendlistadd [ID], /friendlistremove [ID], /whitelistadd [ID], /whitelistremove [ID], /blacklistadd [ID], /blacklistremove [ID], /ghostadd [ID], /ghostremove [ID], /kick [ID], /ban [ID], /unban [ID], /promote [ID], /demote [ID]",,
```

Nina-1474 (commented):
Having : as both a divider in the help message and part of a comment makes it a little hard to read. Plus there's an extra comma in there. I'd suggest something like this which puts the extra option in parentheses after the normal options.

Ben987:
Sure!  I've never used : for emotes but if some players are used to that and there's a setting, it should be good.  Thanks :)

kimei-nishimura:
FYI: 3-char smilies that I have seen used would be like :-) or :-D or :^) (though recognizing these may be an indicator of how long I've been on the internet rather than common usage).

tuxyq:
That was considered with the last change for a more naïve approach, it only acts if the message is longer than 3 characters long, 4 character smileys are uncommon, and 3 letter and fewer emotes are rare, hence the naïve approach

PR #2593, opened 2021-06-18
Bugfix: Adds a null check to extended item `ChangeWhenLocked` checks

elliesec:
## Summary

Under some circumstances (that I've yet to reproduce), it appears to be possible to reach the extended item draw function in a state where the extended item script can't identify the currently selected extended option. This results in the following error being thrown:

error

I suspect this may happen when the item's `Property` ends up being `undefined` (generally due to the player leaving the extended menu immediately after equipping the item), although I've not managed to figure out how the game is hitting the `Draw` function without first hitting the load function immediately beforehand.

In any case, this adds a null check for `CurrentOption` inside the `ExtendedItemValidate`, which should fix this particular error, although I'm still unsure about the root cause.


Ben987:
Yes!  I will prepare the hotfix.  Thanks a lot :)

PR #2595, opened 2021-06-20
Bugfix: Prevents false lock modification warning messages in the console

elliesec:
## Summary

There was a minor issue with lock validation where full character updates on characters wearing limited locks (often combination locks etc.) would trigger validation warning messages about invalid lock modification in the console, despite the lock not having been modified. This would trigger a lock rollback, although as the lock hadn't actually been modified, the rollback wouldn't do anything. This fix prevents the warning messages and the unnecessary rollbacks that were occurring.

## Steps to reproduce

1. On player 1, equip an item
2. Lock the item with a lock whose permissions have been set to limited
3. As a player that _doesn't_ have access to limited items on player 1, trigger a full appearance update on player 1 (e.g. change their clothes and hit the "accept" button in the wardrobe)
4. Note that the following message will appear in player 1's console, and an appearance rollback will have occurred, despite the lock not having been modified

image

### DIFF
--- a/BondageClub/Scripts/Validation.js
+++ b/BondageClub/Scripts/Validation.js
@@ -164,46 +164,7 @@ function ValidationResolveModifyDiff(previousItem, newItem, params) {
 		return { item: previousItem, valid: false };
 	}
 
-	let valid = true;
-
-	const previousLock = InventoryGetLock(previousItem);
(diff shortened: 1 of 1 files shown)


Ben987:
Yes!  I've seen these extra messages before.  Thanks for the fix :)

PR #2596, opened 2021-06-20
Item Expression Trigger Preference

Nina-1474:
This adds a new "Items can affect your facial expression" shared preference in the General settings menu, true by default. When disabled, changes to facial expression triggered by assets or extended item options will no longer occur. It also disables expression changes from struggling. This option has been requested a few times.

Also:
- Fixes made to expression-affecting shocks which weren't applying the blush.
- Fixed the extended option expressions to allow several changes like assets (though the currently existing cases only ever change one).
- Tweaked some General settings to appear on Extreme difficulty but disabled instead of being hidden.


ace-1331 (approved):

Ben987:
Sure!  That setting could be used by a few players.  Thanks :)

PR #2597, opened 2021-06-20
Feature: Extended modular item property support

elliesec:
## Summary

This adds support for a few extra properties to modular items. These are things that are not currently used, but have been requested or discussed by other contributors. The newly supported properties are:

* `SetPose` - The `SetPose` array for each module will be merged into the final `SetPose` array. Asset contributors should take care not to create modular items where the `SetPose` properties might conflict or cause unpredictable results
* `OverridePriority` - The final merged `OverridePriority` will be the `OverridePriority` of the _last_ module that defines one. Again, asset contributors should take care not to create modular items where the `OverridePriority` properties might conflict or cause unpredictable results
* `HeightModifier` - The `HeightModifier` of each module will be summed to get the final `HeightModifier` value


Ben987:
Absolutely!  It will be useful.  Thanks :)

PR #2598, opened 2021-06-20
Bugfix: Fixes incorrect gas mask rebreather layer name

elliesec:
Fixes an incorrect layer name in the colour picker for the gas mask's lenses & rebreather item.


ace-1331 (approved):

Ben987:

PR #2599, opened 2021-06-21
Save input text when relogging in chat rooms.

Leilatwentytwo:
When relogging, the contents of InputChat are saved in the same manner that the chat log is saved.

This pull request is motivated by having, multiple times, lost a paragraph I was typing because there was a disconnect to the server and I had to retype the entire thing.  Now, the text that was being entered wont get cleared and wont have to be retyped.


Ben987:
Good idea!  Thanks for that one :)

PR #2602, opened 2021-06-21
Kigurumi Mask New Functionality and Options

Shiranui-Izayoi:
This PR is dependant on [2597](https://github.com/Ben987/Bondage-College/pull/2597) to work. As while I have tested all parts of it that don't rely on that PR. I am unable to test the `OverridePriority` part, though juging by how the rest of the systems work it should work fine.

This PR adds addition functionality to Kigurumi Mask with permission from Ada the creator of the item. The default functionality of the item is kept as per normal. The changes are listed below.

- Removed the Asset level `Block`, `Hide`, `HideItem` and `Effect`
- Moved them all to Property level via creating a new category in the extended menu the new `MaskStyle` menu.
- The `MaskStyle` menu has 2 options the default being the current mask the second being the Silicone/Hollywood mask mode.
- In the second mode the mask no longer blocks any zones, hides any items etc. Instead acting like face. Allowing the wearer to be gagged etc. This is to imitate how expensive Hollywood silicone masks work where they move with the face. Instead of being a solid mask like a normal Kigurumi.

This change allows more diverese uses for the mask, that previously would have only been possible with console. If these changes are liked I will likely further expand on it.


Ben987:
Lots of items to test!  Thanks a lot for these changes :)

PR #2603, opened 2021-06-21
added drawlocks to clitring

ace-1331:
Added drawlocks: false to clitring since it has no locks therefore it was missing the lock assets. 
There's still more but Jomshir is working on missing assets with a script, this one was the only one I could remember off the top of my head as it was bugging me due to its frequency.


elliesec (approved):

Ben987:
Yes!  Thanks for that one :)

PR #2606, opened 2021-06-21
Futuristic Item Update

Ada18980:
Adds a new item and some changes:

-Some futuristic items that had "AlwaysExtend" now use remotes instead when they are covered (added an Effect to support this on all items)
-Refactors a few futuristic items to make the code more extensible (mainly the gag)
-Merges the futuristic chastity belt into one item, and adds a third model
-Adds a brand new Futuristic Training Belt with a public modes feature and speech controls. It allows the owner to control who has access to the vibration modes (either public, Club Mistresses only, or keyholders only), and lets the user specify a word that the slave must say or else be punished (like "nyah" or "miss")

Depends on #2590


Ben987:
Wow!  This one is huge, lets hope it won't create too many bugs.  Thanks a lot :)

PR #2607, opened 2021-06-23
[Asset][Natsuki] Nylon Collar

Codefreak64:
Nylon is an interesting texture when you think about it, its strong and not that bad of a texture. we only have it for rope currently so whats stopping us from adding it to the rest of the body? starting with this.

-Adds a 'Nylon Collar' (45) into the game

-Collar is lockable but not added into the slave menu as of right now


Ben987:
Sure!  That's a good idea.  Thanks a lot :)

PR #2610, opened 2021-06-23
Advanced Vibrator Modes Disable

Ada18980:
Allows the player (if appearance is below Extreme) to block advanced vibrator modes in the arousal menu. If the player is extreme, it is forced off (because the player cannot block any items or types).

Also adds validation to prevent trolling.

### DIFF
--- a/BondageClub/Scripts/Validation.js
+++ b/BondageClub/Scripts/Validation.js
@@ -553,7 +553,7 @@ function ValidationSanitizeProperties(C, item) {
 	}
 	
 	// Block advanced vibrator modes if disabled
-	if (property && typeof property.Mode === "string" && Player.ArousalSettings && Player.ArousalSettings.DisableAdvancedVibes) {
+	if (typeof property.Mode === "string"&& Player.ArousalSettings && Player.ArousalSettings.DisableAdvancedVibes && !VibratorModeOptions[VibratorModeSet.STANDARD].includes(VibratorModeGetOption(property.Mode)) ) {
 		console.warn(`Removing invalid mode "${property.Mode}" from ${asset.Name}`);
 		property.Mode = VibratorModeOptions[VibratorModeSet.STANDARD][0];
 		changed = true;
--- a/BondageClub/Scripts/Validation.js
+++ b/BondageClub/Scripts/Validation.js
@@ -551,6 +551,13 @@ function ValidationSanitizeProperties(C, item) {
 			}
 		});
 	}
+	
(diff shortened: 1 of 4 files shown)


BondageClub/Scripts/Validation.js
```diff
@@ -551,6 +551,13 @@ function ValidationSanitizeProperties(C, item) {
 			}
 		});
 	}
+	
+	// Block advanced vibrator modes if disabled
+	if (property && typeof property.Mode === "string" && Player.ArousalSettings && Player.ArousalSettings.DisableAdvancedVibes) {
```
elliesec, line 556: The `property` check here shouldn't be needed - `property` is guaranteed to be an object due to the earlier checks.

As this function can be called on characters that aren't the player, we shouldn't be directly referencing `Player` here, we should be using `C` instead.

Also... as this is just checking whether or not `Property.Mode` is a string, won't this end up _always_ triggering a sanitization, even for basic vibrator modes?

BondageClub/Scripts/VibratorMode.js
```diff
@@ -198,7 +198,7 @@ function VibratorModeDrawControls(Options, Y) {
 		OptionGroup.forEach((Option, I) => {
 			var X = 1175 + (I % 3) * 225;
 			if (I % 3 === 0) Y += 75;
-			var Color = Property.Mode === Option.Property.Mode ? "#888" : "White";
+			var Color = Property.Mode === Option.Property.Mode ? "#888" : (!(OptionName == VibratorModeSet.ADVANCED && CurrentCharacter.ArousalSettings && CurrentCharacter.ArousalSettings.DisableAdvancedVibes) ? "White" : "Pink");
```
elliesec, line 201: Probably best to assign `CharacterGetCurrent()` to a variable here, and then use that instead of using `CurrentCharacter` directly. Similarly below.

elliesec (changes requested):

BondageClub/Scripts/Validation.js
Ada18980: Dang, I shouldn't be coding this late at night. Will submit a fix

BondageClub/Scripts/VibratorMode.js
Ada18980: Ill do that, thanks

elliesec (approved):

Ben987:
Alright!  Thanks for that new preference.

PR #2611, opened 2021-06-23
fix light/dark 2 mobile

ace-1331:
- Fixed the newer chat styles (dark/light 2) on mobile.
Font-size was insanely huge on mobile, removing these keeps it on par with other styles and doesn't change much other than making it usable on mobile/small windows.

(Had Ellie's permission for this fix)


elliesec (approved):
Hmmm, the one thing that does bug me about the removal of the font sizes is that the default chat window font size is huge on desktop (IMO), which is why I created these themes originally. I think removing the hard `!important` override is definitely the right thing to do though for responsiveness. What I might do after this (and what's probably the best solution in the long run) is to put in a chat preference for font scale (not specific font sizes, but more a generic small/large selection).

Ben987:
Thanks!  And we can always review the size later if we find a better setting.  Cheers :)

PR #2615, opened 2021-06-23
Fix floating cushion

ace-1331:
- fixed a bug where you could float with a cushion when using it with suspension or various other items that made kneeling look strange


elliesec (approved):

Ben987:
That's a funny bug, I don't think I've seen it before.  Thanks for the fix :)

PR #2618, opened 2021-06-24
Feature: Block cycle validation

elliesec:
## Summary

This is something I've been working towards for a while now, which should give players an extra level of protection against poor or malicious console usage, or problems in the game code that might lead to players being equipped with an impossible item combination.

The PR adds an extra step to the validation of appearance updates which checks for and removes block cycles (a series of items which block each other, making removal of any of them impossible by normal means. For example, the Floor Shackles block the `ItemDevices` group, and the Wooden Box, being an enclosing item, blocks the `ItemLegs` group. It's not usually possible to equip both items simultaneously through the game, but it is possible via console, which makes it impossible to remove either item, even by another player. The new validation will detect this block cycle and will remove (or roll back) one of the items. Which item gets removed is largely down to the circumstances. It will try to remove whichever item was included in the appearance update that triggered the validation (i.e. if the player was already wearing Floor Shackles, and an update which added a Wooden Box was sent, the Wooden Box would get removed).

At the moment, the fix only applies to items blocked via the `Block` property, or by the `Enclose` effect. It does not (currently) address items which block each other due to prerequisites, but this is something I will likely look to add in later.

This also adds a minor fix to the `ChatRoomCharacterItemUpdate` function which causes it to update the local `ChatRoomData`. This is because `ChatRoomCharacterUpdate` messages get sent back to the source player, causing `ChatRoomData` to be updated, whereas `ChatRoomCharacterItemUpdate` messages don't. This can cause sync issues between players when `CharacterLoadOnline` is called (when it tries to detect whether or not a a refresh should happen) due to stale chatroom data.

---

Of interest to @Ada18980 - the validation script now exposes a `ValidationFindBlockCycles(appearance)` function. You can pass this an appearance array, and it will return a list of all block cycles, formatted as arrays of group names. For example:
```js
ValidationFindBlockCycles(appearance);
// Returns:
[
	["ItemLegs", "ItemDevices", "ItemLegs"], // A block cycle where two items block each other
	["ItemArms", "ItemLegs", "ItemDevices", "ItemArms"] // A block cycle where three items block each other:
														// ItemArms blocks ItemLegs blocks ItemDevices blocks ItemArms
]
```


Ben987:
I have no idea what all that code does, but if you're able to debug it when it crashes and it helps solve console bugs, it should be good.  Thanks!

PR #95, opened 2019-02-05
Local

GIT4NICK:
Some Stable suff

New Skill Dressage
Some Characters, Minigame & Dialoge for the Stable
Dialoge Prison
Backgroundfix Photograph

PR #2619, opened 2021-06-24
Inventory favorites

ace-1331:
- added a new "favorite" list to mark certain items as favorites, it will send them to the beginning of the list.
- fixed a bug with permission toggles that made it so certain extended items did not behave as intended
- re-instated the hold blocking behavior as it was +/- a side-effect of other changes as well as the reason for the extended type blocking bugs. (We're discussing "grouped blocking" as a potential feature) 
- refactored the sort order variables into one enum

This has been a long requested feature of many, even myself since I've been around, was never sure how to do it till now, not sure why. 

It should help communicate preferences better, and helps to navigate around the menu by bringing frequently used items to the first page.

**Requires serverside [#95](https://github.com/Ben987/Bondage-Club-Server/pull/95)**


Ben987:
It's time to test that big one!  I like the concept, I'm sure players will like it also :)

PR #2624, opened 2021-06-25
broom item

ace-1331:
- added a broom in the handheld items

credits to Rui for the base asset ^^


elliesec (approved):

Ben987:
Yes!  I've been wanting an item like that for a while now.  Thanks you two :)

PR #2625, opened 2021-06-25
wardrobe info

ace-1331:
- added a disabled wardrobe button when the wardrobe is not owned

It's been noticed that with the increased complexity of the club, many people aren't aware they can have a wardrobe/save outfits. This adds the information that they can buy a wardrobe and this info exists in a place where they cannot miss it.


elliesec (approved):

Ben987:
Sure!  That's a good idea :)

PR #2626, opened 2021-06-25
Feature: Allows `DefaultColor` to be set in an item's `Property`

elliesec:
## Summary

This PR adds support for `DefaultColor` to be set in an item's `Property`. If the colour of a layer (or the whole item) is set (implicitly or explicitly) to "Default", then `Property.DefaultColor` will be used, if it exists. `Property.DefaultColor` can be an array of colour strings, or a single colour string:
* If `Property.DefaultColor` is an array, the individual array items will be used when their corresponding layer's colour is set to "Default"
* If `Property.DefaultColor` is a string, then it will be used for any colourable layer whose colour is currently set to "Default"

This will allow typed items to have different default colour palettes per type, which may be useful as we start to see more extended clothing in the game (and this is something that will support an item that @TheGnarp has on the way).

As an example:
```js
// Base asset
DefaultColor: ["#111", "Default", "#333"]

// Type definitions
{
	Name: "Type1",
	Property: { Type: "Type1", DefaultColor: "#222" },
},
{
	Name: "Type2",
	Property: { Type: "Type2", DefaultColor: "#444" },
}

// Evaluated colour when selecting Type1 from the extended item menu:
["#111", "#222", "#333"]
// Evaluated colour when selecting Type2 from the extended item menu:
["#111", "#444", "#333"]
```

I also removed instances of `var` in `CommonDraw.js` and replaced them with `const` and `let` as appropriate, as that had been bugging me for a while.

### DIFF
--- a/BondageClub/Scripts/CommonDraw.js
+++ b/BondageClub/Scripts/CommonDraw.js
@@ -83,7 +83,7 @@ function CommonDrawAppearanceBuild(C, {
 	drawImageColorize,
 	drawImageColorizeBlink,
 }) {
-	var LayerCounts = {};
+	const LayerCounts = {};
 
 	// Loop through all layers in the character appearance
 	C.AppearanceLayers.forEach((Layer) => {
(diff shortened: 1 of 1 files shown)


Ben987:
Absolutely!  It could be used for many items.  Thanks :)

PR #2627, opened 2021-06-26
Bugfix: Cafe item fixes

elliesec:
## Summary

Over the last release or two, several items have been migrated over to the archetype system. However, the Cafe room was making use of the options objects that were previously defined in the item scripts. This would cause the Cafe to crash on several of the Cafe maid's dialogue options. This updates the Cafe code to get the options objects out of the typed item data lookup, which restores the previous functionality.


ace-1331 (approved):

Ben987:
Yes!  I've heard of that bug.  Thanks a lot :)

PR #2628, opened 2021-06-26
cane/crop gags

ace-1331:
- added two "gags" which are a cane/crop being held by the mouth

credit to gatetrek for the cane asset and natsuki for the mouth.


elliesec (approved):

Ben987:
Nice idea.  Thanks again you two :)

PR #2629, opened 2021-06-26
Garters

ace-1331:
- added a need garters group, introduced with two different ones (more are being worked on)

credit to gatetrek for all the assets


Ben987:
That's a great idea.  Thanks you two :)

PR #2630, opened 2021-06-26
New Serving Tray options

EmilyR42:
Collab with Ruilove. Thanks to elliesec for assistance.

- Added Cookies, Cake, and Sex Toys options for Serving Tray
- Added multicoloring for each
- Updated tray preview icons


BondageClub/Assets/Female3DCG/Female3DCG.js
```diff
@@ -4104,7 +4104,12 @@ var AssetFemale3DCG = [
 					{ Name: "Text" },
 				]
 			},
-			{ Name: "ServingTray", Value: -1, Time: 5, Extended: true },
+			{ Name: "ServingTray", Value: -1, Time: 5, Extended: true, Layer: [
+					{ Name: "Tray", AllowColorize: false },
+					{ Name: "Objects" },
```
elliesec, line 4109: It's generally preferable to use `AllowTypes` on the layers to remove the need for empty placeholder assets. That makes it so that the layer is only rendered when the current type is in the `AllowTypes` array.
```suggestion
					{ Name: "Objects", AllowTypes: ["Drinks", "Cake", "Cookies", "Toys"] },
```
This means the "Objects" layer would not be rendered for the empty type, and the `ServingTray_Objects.png` asset could be removed. Similarly for the "Details" layer.

BondageClub/Assets/Female3DCG/Female3DCG.js
```diff
@@ -4104,7 +4104,12 @@ var AssetFemale3DCG = [
...
+					{ Name: "Details" },
```
elliesec, line 4110: Same as above.
```suggestion
					{ Name: "Details", AllowTypes: ["Drinks", "Cake", "Toys"] },
```

BondageClub/Assets/Female3DCG/Female3DCG.js
EmilyR42: Thanks a lot!

elliesec (approved):

ace-1331 (approved):

PR #2633, opened 2021-06-27
Hide Icon State Persistence

Nina-1474:
When the 'hide icon' button in the chatroom is changed, it should now only reset back to the default on entering a new chatroom, so that entering submenus such as wardrobes doesn't also reset the selection.


ace-1331 (approved):

Ben987:
Absolutely! Cheers :)

PR #2634, opened 2021-06-27
Added favorite colors to color picker

Leilatwentytwo:
Added the ability to save colors in the color picker for future use.  Will save up to 18 colors in HSV format and save the data on the server.

Displays the favorite colors in a palette at the bottom (as a result, the height of the large s/v panel is smaller), and moves the last color/current color palette to the right half to make room for a few buttons to save colors and move between pages of favorite colors.

Added French and German translations for the mouseover text for the new buttons.  Didn't find anyone for Russian.

Added a save icon, which was taken from Bondage College and resized.

Also fixed the spelling of palette throughout the ColorPicker.js file.  Was spelled in three different ways, all of them wrong 😆


Ben987:
Absolutely!  Thanks for that one, it's been on my list for a while :)

PR #2635, opened 2021-06-27
creation crash

ace-1331:
- fixed a crash that makes it impossible to sign up because of the recent addition of the preview setting


Ben987:
Sure!  Thanks for that fix :)

PR #2636, opened 2021-06-27
creation length validation

ace-1331:
- made the signup more robust, adding validation for length in case the maxlength is bypassed for some reason.

example of how this can happen: https://discord.com/channels/554377975714414605/554378725916147722/858561603230498826


Ben987:
Not sure if it's really needed, if someone hacks their DOM elements, they can hack the code also.  But why not.  Thanks :)

PR #2637, opened 2021-06-27
Feature: Typed item API improvements

elliesec:
## Summary

Afraid this is quite a large one, although most of the touched files are just simple method signature changes. It pulls some of the extended/typed item functionality out into standalone functions that can be more easily re-used elsewhere in the code, or by scripters to set types on items. The intent is to separate the UI-specific code from the core functionality (removing reliance on contextual globals like `DialogFocusItem`, etc.) in order to make these core functions more applicable elsewhere. It also adds a few utility functions to the typed item script, that make it easier to interact with typed items from a code point of view. This shouldn't change any existing functionality, but should open up the code to make it easier to use typed items for contributors that are writing single-player rooms, or for scripters.

Along with this, I'm also intending on setting up a page on the wiki that contributors and scripters can use as a reference for interacting with extended items programmatically.

In the future, this should open the way for NPCs to utilise item types without hard-coding (for example, we could update the random item functions to set a random type too).

## Technical details

* Pulls some of the item-setting and validation functionality out of `ExtendedItem.js` and into `TypedItem.js`, and removes dependence on `DialogFocusItem`
* Refactors extended item validation functions to also take an `Item` parameter, rather than relying on `DialogFocusItem`
* Adds several utility functions to `TypedItem.js`:
	* `TypedItemGetOptions` - gets the options array for a typed item
	* `TypedItemGetOptionNames` - gets a list of the available option names for a typed item
	* `TypedItemGetOption` - gets a named option for a typed item
	* `TypedItemSetOptionByName` - sets a typed item's type by option name
	* `TypedItemSetOption` - sets a typed item's option, given the options array and chosen option (can also be used by items using the extended item script)
	* `TypedItemSetRandomOption` - sets a random option for a typed item

## Examples

Here are a few code examples of how some of the new functionality can be used:
```js
if (item.Archetype === ExtendedArchetype.TYPED) {
    // If an item uses the typed archetype, use this to set it to a random type. Refreshes the character, but
    // doesn't persist changes (can set a third boolean "push" parameter to true if that's desired)
    TypedItemSetRandomOption(C, item);
}

InventoryWear(C, "DuctTape", "ItemMouth");
// Set a random type
TypedItemSetRandomOption(C, "ItemMouth");
// Set a named option - note this is the option name, not the type, so you'd use "Small" here rather than null
// to set it back to the default
TypedItemSetOptionByName(C, "ItemMouth", "Cover");
// Returns a list of the available option names - i.e. ["Small", "Crossed", "Full", "Double", "Cover"]
const optionNames = TypedItemGetOptionNames("ItemMouth", "DuctTape");
```


Ben987:
Another big one!  Thanks a lot.  Lets hope it won't create any bug in these items.  Cheers :)

PR #2638, opened 2021-06-27
Nurse Uniform OverTheHead pose fix

EmilyR42:
- Positioned Nurse Uniform to fit better on OverTheHead pose


Ben987:
Sure!  Thanks for that fix :)

PR #2639, opened 2021-06-28
Miscellaneous color picking fixes

Leilatwentytwo:
This pull request fixes a Crash when picking colors in the wardrobe on character creation.  it also fixes a few bugs I noticed while implementing favorite colors.  This fixes:

Crash in color selection on character creation since Player.SavedColors didn't exist and hadn't been initialized.

An error in color selection on character creation fixed with a suggested change from @ace-1331 to the file Common.js

Color picking menu height is slightly too large when picking your character's name's color, causing the lower part to get cut off.

The function DrawRGBToHex() returns the hex code with lower case letters.  Though, my changes here also make this function unused, it could still be used later at some point? Probably better to just leave it in.

When selecting character name color, if the name color is too dark, the hex code becomes illegible.

When selecting the slave collar color, the color is set by what is simply under the mouse cursor on a click, even if the color picker hasn't told it to change color yet.  This was fine before, if a bit weird (could select colors in the background image), but now that the color picker has been changed, it causes differences in how the UI functions.

Side-note:  The color picking code for the slave collar seems to be implemented differently than any other use of the color picker.  is the color picking code here outdated?  Unless there is a reason for this, I'm thinking the color picking code here should be made similar to how it works elsewhere in the code.


Ben987:
Sure!  Thanks for all these fixes, I will try it soon.

PR #2640, opened 2021-06-28
DrawTextFit Optimisation

ace-1331:
`MainCanvas.measureText(Text)` is resource heavy, and is called multiple times every frame whenever we call `DrawTextFit`, causing screens like `ChatSearch` to take a lot of unnecessary script time while idling or doing other things.

I extracted the logic that finds the size and text to render into its on memoized function so that it would only do the necessary calculations once. The difference here is astronomical (see pics below).

The decision to clear the memo cache when leaving chatsearch is to prevent the cache from growing too big if someone stays in the club for a long time. 

during ~26s of idling, before ~80% of script time was in `DrawTextFit` for a total of ~4.9s, after it was ~61% with ~0.9s which is ~80% less

Before:
before

After:
after


Ben987:
Sure!  Anything that can speed up the game a little is always welcome :)

PR #2643, opened 2021-06-28
Add Classic T-Shirt with prints

TheGnarp:
For the girls (and guys) who are looking for a more casual, classic style of t-shirt, I added this model with a couple of prints or logos to choose from if the plain version is just too boring.

Many thanks to @elliesec for addint the option to set default colors in an item's property (#2626)

My commit doesn't actually rely on this new feature, but it works a lot better with it!


Ben987:

PR #2644, opened 2021-06-29
auto shock unit fixes

ace-1331:
- fixed a few crashes with the automatic shock unit


Ben987:
Sure!  Thanks for that fix :)

PR #2645, opened 2021-06-29
Asset: Plain blouse

ItsJustMeVerity:
Adds a plain blouse upper clothing + undershirt version.

Having found myself being mildly annoyed that so many assets are very specific and can only really be used for one specific thing, I made just a regular blouse. Or shirt, depending on how you want to use it. For extra versatility, it can be used in both the Cloth and the fairly barren SuitUpper slot in order to go under clothes (Bodice, T-shirts, jackets, etc).

Very minor graphical issue with the suit version, which *would* have been fixed by my previous PR adjusting the hand layer alpha, if it hadn't also broken several skirts in the process. Should instead be fixed by Natsuki's upcoming changes to remove the hand cutout. The issue only affects the Suit version, and given that the version in that slot is intended to be used beneath other clothing, the issue should disappear under any assets that the previous alpha change broke.
Blouse Preview


Ben987:
Yes, yes, yes!  I love it.  Thanks a lot :)

PR #2646, opened 2021-06-29
missing dialogs

ace-1331:
- fixed missing cuffs & body bag dialogs


Ben987:
You're on fire!  Thanks for that fix :)

PR #2649, opened 2021-06-29
Feature: Adds extended item support to `InventoryWearRandom` (and NPCs!)

elliesec:
## Summary

This is a follow-up to #2637, and adds support for random item extension to the `InventoryWearRandom` function. If the item has the correct archetype (only typed items are supported for now), and the `Extend` parameter is passed to the function, `InventoryWearRandom` will now randomly select an item type and apply it. This means that anywhere which was previously using `InventoryWearRandom` to bind the player will now also be able to randomly select item types, as applicable. As we move more items over to the archetypes system, this feature should automatically support those items. Typed item options also now support a `Random` flag just like assets to indicate whether or not they can be randomly selected.

The PR also modifies some of the functions added by #2637 so that they now respect the player's blocked/limited types, and adds a dialogue message to the extended item screen for typed items and items using the extended item script, indicating that a given type is blocked/limited by player permissions.


Ben987:
That's a good idea, it will add more stuff that NPCs can do.  Thanks!

PR #2650, opened 2021-06-29
Improvement: Force reset the canvas after a context loss if it is not restored after 10 seconds

elliesec:
## Summary

In #1026, handling was added for when the WebGL context is lost and restored. Usually this is due to a lack of resources on the player's machine, but more often than not, it never seems to get restored when the issue occurs in real situations. This adds a timeout to the context lost listener so that if the context isn't successfully restored after 10 seconds, the game will forcefully remove the GLDraw canvas and recreate it, and rebuild the canvas of all characters currently on the screen, forcing them to be re-rendered.

If _another_ context loss then happens within a further 50 seconds (i.e. two context losses within a minute), the game will revert character rendering back to the original `CharacterAppearanceBuildCanvas` function, and no longer use `GLDraw` (thanks to @ace-1331 for that suggestion).

This shouldn't happen during normal gameplay, but will hopefully alleviate the "WebGL context lost" issue for players for whom it never automatically restores. This change was prompted by the latest WebGL context lost issue that was posted on the game's Discord (see image below). I've tested this with the `WEBGL_lose_context` plugin, and it seems to work well.

image

### DIFF
--- a/BondageClub/Scripts/Drawing.js
+++ b/BondageClub/Scripts/Drawing.js
@@ -28,6 +28,12 @@ let DrawCacheTotalImages = 0;
 var DrawLastDarkFactor = 0;
 
 /**
+ * A list of the characters that are drawn every frame
+ * @type {Character[]}
+ */
+var DrawLastCharacters = [];
+
+/**
  * Converts a hex color string to a RGB color
  * @param {string} color - Hex color to conver
  * @returns {{ r: number, g: number, b: number }} - RGB color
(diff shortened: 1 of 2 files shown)


ace-1331 (approved):

ace-1331 (approved):

Ben987:
Yes!  More and more players seems to be having these problems.  Lets hope it will help them.  Thanks :)

PR #2651, opened 2021-06-30
WebGL render options

ace-1331:
*builds on #2650*

While discussing the context issue, I started reading on WebGL context attributes more in detail, and there are a few options we can change to affect performances.

This PR adds device-specific graphic settings to provide options to lower/increase performances.

1- Antialiasing: It is already ON by default when the device supports it, now there is an option to force disable it which can help make the club run much faster on slower devices. This is supported by every browser that supports WebGL

2- Power consumption mode: The default mode is "default", but you can pass low/high to influence the browser/device's decision on how it will handle rendering. This feature is still non-normative, and the support is low to mid. I debated whether or not it should be in, but I think it's good to give  this option to people with modern browser/devices that support it as it can help save battery on laptops/phones (which the club is guilty of), or increase performances on PCs.  If support for this attribute is ever lost or change, this setting can easily be adapted, but won't cause any errors due to it's nature of just being an attribute passed to the creation of the context.


elliesec (approved):

Ben987:
This one creates conflicts with #2650, can you resolve them and test again?  Thanks.

ace-1331:
solved

Ben987:
Absolutely!  More and more players are having drawing issues lately, maybe it will help them.  Thanks :)

PR #2652, opened 2021-06-30
Extended item migration: ItemHead

elliesec:
## Summary

This migrates extended items in the `ItemHead` group over to the archetype framework, and removes their corresponding script files. The affected items are:

* Duct Tape
* Ribbons
* Web Blindfold

I've tested all three thoroughly and verified that they are all working as before.


Ben987:
Sure!  Thanks for that update :)

PR #2653, opened 2021-06-30
Bugfix/Improvement: `GagUnique` prerequisite improvements

elliesec:
## Summary

It came to my attention recently that ball gags (or, as it turns out any gags with the `GagUnique` prerequisite) were never being selected in the random item pool. The reason for this is that the `InventoryPrerequisiteMessage` function returns `null` rather than an empty string if `C.FocusGroup` isn't set, which meant it would fail this check in `InventoryAllow`:
```js
return (Msg == "");
```

This PR updates `InventoryPrerequisiteMessage` and `InventoryPrerequisiteConflictingGags` so that `C.FocusGroup` being set is no longer a hard requirement. However, the `GagUnique` prerequisite is slightly different from others in that it pulls some contextual information from the current focus group. Therefore, if no focus group is set, the prerequisite returns an error if _any_ of the mouth groups contains a conflicting gag. If a focus group is set, the prerequisite behaves as it did before. This should bring a lot more gag variety (back?) into the random item pool.


Ben987:
Thanks!  More random gags is always good :)

PR #2655, opened 2021-06-30
Update FoxyHarnessPanelGag to support all slots

Lara-BC:
All other panel gags support layering on top of the gag and, more importantly, they can be used in slots 2 and 3.
This pull requests updates the **Foxy Harness Gag** to allow it to be used in slots 2 and 3. I want to stuff my gags!

I believe I've done everything required to support multiple slots, including putting the gag into a `BuyGroup`.
I've tested it and it seems to work as I expect.

Note: I believe this update also lowers the gag level from the default `GagNormal` to `GagEasy`. I'm not 100% sure there as I am not certain how this works, but this also brings it in line with the other panel gags, as I made sure to follow their configuration.


ace-1331 (approved):

Ben987:
Sure!  Can't go wrong with more gag combos.  Thanks :)

PR #2657, opened 2021-06-30
New styles for Chinese Long Dress

EmilyR42:
- Added two new styles for Chinese Long Dress
- Removed stray pixels from Edges layer


Ben987:
Absolutely!  That's a good idea, thanks a lot :)

PR #2662, opened 2021-07-01
misc type related fixes

ace-1331:
Player noticeable changes:
- fixed `CommonConvertStringToArray` that had a condition that always evaluated to true (number === false) which was a mistake that caused empty strings to be converted to zeros ("1,,2" used to return [1,0,2])
-  fixed a crash that could occur when starting to pick a lock because `Math.Ceiling` was incorrectly used instead of `Math.ceil`
- fixed a few default online preferences never getting initialized properly (notably autoban and show full rooms)
- fixed the mistress cutscene maids not having the intended amount of restraints ("Lot" vs "LOT")

Type changes:
- fixed two instances of `DrawText `being called with two extra arguments.
- fixed Player.VisualSettings type definition missing some optional attributes.
- fixed some jsdoc in chatroom, larp, tennis, creation and chatsearch
- fixed two instances of extra arguments passed to `ChatSearchMuffle`
- fixed the LARP odds function returning a string instead of a number as expected everywhere else in the code (luckily, it did not cause any bugs before since there was no use of triple equals and such)
- removed an instance of an extra argument being passed to `PreferenceGetFetishFactor`
- removed two instances of an extra argument being passed to `Math.round`


BondageClub/Screens/Online/ChatRoom/ChatRoom.js
```diff
@@ -2747,15 +2747,15 @@ function ChatRoomStopLeave(){
 /**
  * Sends an administrative command to the server for the chat room from the character dialog.
  * @param {string} ActionType - Type of action performed.
- * @param {boolean} [Publish] - Whether or not the action should be published.
+ * @param {boolean | string} [Publish] - Whether or not the action should be published.
```
jomshir98, line 2750: I like to add this just to see what the default behaviour is without having to look through the code of the function.
```suggestion
 * @param {boolean | string} [Publish=true] - Whether or not the action should be published.
```

BondageClub/Screens/Character/Preference/Preference.js
```diff
@@ -399,11 +399,11 @@ function PreferenceInitPlayer() {
...
-	if (typeof C.OnlineSettings.AutoBanBlackList !== "boolean") C.ChatSettings.AutoBanBlackList = false;
-	if (typeof C.OnlineSettings.AutoBanGhostList !== "boolean") C.ChatSettings.AutoBanGhostList = true;
-	if (typeof C.OnlineSettings.DisableAnimations !== "boolean") C.ChatSettings.DisableAnimations = false;
-	if (typeof C.OnlineSettings.SearchShowsFullRooms !== "boolean") C.ChatSettings.SearchShowsFullRooms = true;
-	if (typeof C.OnlineSettings.SearchFriendsFirst !== "boolean") C.ChatSettings.SearchFriendsFirst = false;
+	if (typeof C.OnlineSettings.AutoBanBlackList !== "boolean") C.OnlineSettings.AutoBanBlackList = false;
+	if (typeof C.OnlineSettings.AutoBanGhostList !== "boolean") C.OnlineSettings.AutoBanGhostList = true;
+	if (typeof C.OnlineSettings.DisableAnimations !== "boolean") C.OnlineSettings.DisableAnimations = false;
+	if (typeof C.OnlineSettings.SearchShowsFullRooms !== "boolean") C.OnlineSettings.SearchShowsFullRooms = true;
+	if (typeof C.OnlineSettings.SearchFriendsFirst !== "boolean") C.OnlineSettings.SearchFriendsFirst = false;
```
jomshir98, line 406: Maybe worth adding migration/cleanup code for the wrong settings?

Ben987:
Thanks for all these fixes, it's strange that these situations were never reported before.  Lets hope it won't generate new bugs.  Cheers :)

PR #2665, opened 2021-07-01
Feature: New screen functions and optimisations

jomshir98:
Updates per-screen functions to following:
- `Run(time: number)` Called each frame [change: add `time` parameter]
- `Click(event: MouseEvent|TouchEvent)`  Called when user clicks on the canvas [change: pass event]
- `Load()` Called when screen is loaded using `CommonSetScreen` [no change]
- `Unload()` Called when this screen is being replaced [change: new function]
- `Resize(load: boolean)` Called when screen size or position changes or after screen load [change: new function]
- `KeyDown(event: KeyboardEvent)` Called when user presses any key [change: pass event]
- `Exit()` Called when user presses Esc [no change]

Also changes the way these functions are called, improving the performance
Tested and fully compatible with existing screens

### DIFF
--- a/BondageClub/Scripts/Common.js
+++ b/BondageClub/Scripts/Common.js
@@ -5,7 +5,10 @@ var Player;
 /** @type {number|string} */
 var KeyPress = "";
 var CurrentModule;
+/** @type {string} */
 var CurrentScreen;
+/** @type {ScreenFunctions} */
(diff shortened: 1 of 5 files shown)


Ben987:
Thanks!  Maybe these new events will be used soon :)

PR #2667, opened 2021-07-02
New item: Hitching Post

ace-1331:
- added a new hitching post item

I only did the code/layer splitting, all credits to Sybil for the base asset (and 3D model! >~>!)


Ben987:
Yes!  Thanks a lot you two :)

PR #2668, opened 2021-07-02
resize crash fix

ace-1331:
- fixed a crash occurring when leaving rooms after the recent addition of the resize manager.
- fixed chat resizes being applied while resizing when in a dialog
- improved the resize and element functions to be more robust
-  jomshir reworked it so DialogLeave no longer calls the resize function (instead it is kept track of inside the chatroom logic)


Ben987:
Yes, yes, yes.  Thanks for that fix :)

PR #2669, opened 2021-07-02
various color picker fixes

ace-1331:
- split the export/import color button logic to make sure they are both supported before enabling them (this prevents crashes when trying to import/export on browsers that only support one of these options, such as firefox)
- fixed a crash occurring after getting an error because of `console.err`
- modified return statement to avoid type errors


Ben987:
Thanks for these fixes, I will try them soon :)

PR #2670, opened 2021-07-02
Boxes/Cages/... fixes + New item: Kennel

ace-1331:
- Added a new item: Kennel (yes the first commits date back to october of last year.. that's not a bug with git)
- Added a new "OneWayEnclose" effect to enable one way interactions. (IE: the caged person cannot change or interact with others, but others can interact with the person)
- Fixed rendering of all boxes/cages and similar items to use the overflow alpha masks to prevent partial cuts and the new "NotLifted" prerequisites to block breaking combinations (such as overthehead ceiling shackles + ceiling rope + box)
- Fixed missing AllowEffect on the futuristic crate, as well as the alpha mask definition (they were not doing anything due to not being on a top layer)
- Removed unused property AllowBlock from the exposed boxes.

credits goes to Sybil for the assets (and 3D model! >~>), I just did the splitting&code

Side note: The original idea was to have an animation for the door opening/closing. I have working code for it, but it would require a bit of resizing of the prototype assets which might come in the future as a rework... If it can help someone make another item with an animation, feel free to message me about it.


Ben987:
Very nice, I'm sure the new Kennel will be popular.  Thanks :)

PR #2671, opened 2021-07-02
Removed unused files

ace-1331:
- removed unused .pdn files

confirmed that Ada still had the files first, they are recoverable through git anyway


Ben987:
Absolutely!  Thanks :)

PR #2675, opened 2021-07-03
Bra extended item refactor

ace-1331:
- refactored all bras to use the newer extended item system


Ben987:
You're on fire!  Thanks again :)

PR #2678, opened 2021-07-05
ItemPelvis extended item refactor

ace-1331:
- refactored the chastity belts to use the newer extended item system


Ben987:
Yeeeeees!  Thanks a lot.

PR #2680, opened 2021-07-05
Preview Icons

Nina-1474:
## Preview Image Icons
The small icons that display inside the preview boxes for items are now dynamically drawn using a single system. They're placed at the top-left of the box with additional icons added vertically below. Hovering the mouse over an icon shows a tooltip. This currently covers:
- Favorites: Replaces the star in the description text with a yellow star icon (with the exception of imageless text-only options). This also fixes a bug in extended item options where the player would still be seeing the target character's preferences in permissions mode.
- Handheld: Removed the hand embedded in the preview image files.
- Locked/Unlocked: Moved to the top-left corner as with the other icons.
- Owner/Lover Only: Newly added, appearing as a collar/heart.

## Dialog Inventory Refactor
Some parts of the inventory list logic have been refactored to support the feature, reduce code duplication and improve performance - mainly in the setting of certain properties (icons/hidden/vibrating) when building the list instead of on every draw.


Ben987:
What does it do on screen exactly?  Can you post a "before" and "after" image?  Thanks!

Nina-1474:
This is the "after" image at least:
image

Ben987:
That looks very nice, thanks a lot, I will test it soon :)

PR #2681, opened 2021-07-05
Extended item refactor

ace-1331:
- refactored 6 more extended items.

There are 131 items not included in this newer system yet, still a fair few that can without extending the current functionalities though.


Ben987:
Sure!  A little less code is always good :)

PR #2682, opened 2021-07-05
Kennel improvements

ace-1331:
- re-added the kennel animation after a little work on it

Managed to pull of a bit of salvage work to make the animation function properly with the modified assets! and it syncs with the audio. Hoping it will serve as an example/inspire others to push the limits of what they can do~

Also improves the following:
- fixed a few conflicts by adding "CanKneel"
- Improved the effects to allow a few more things while the door is open
- Added dynamic audio to play a different sound when struggling in the kennel
- Changed it so items can have dynamic audio play on struggle as well


Ben987:
Absolutely!  Thanks for these updates.

PR #2683, opened 2021-07-06
Updates and fixes to the FuturisticVibrator (and minor change to FuturisticTrainingBelt)

Lara-BC:
This pull requests does the following things:

**Update the FuturisticVibrator to escape user inputs rather than accepting raw regex**

This would confuse users, where characters like `.` would cause unintended side effects.
It is also a vector for attack, as it is possible to craft "evil regex" that would crash browsers.
Escaping avoids this issue.

**Update the FuturisticVibrator to accept \* wildcards**

As a result of the new changes (prior to this pull request), the belt now only triggers on full words.
This new change handles this by converting any `*` character into a wildcard that will match anything.

**Update the FuturisticVibrator to instantly respond to messages**

Previously, the vibrator would check the chat chat log for messages once every second.
This change has the script monitor the time of the last message,
and triggers the keyword searches the moment a newer message is seen.
This results in triggers occurring in the very next frame,
and a much snappier experience.

**Update the FuturisticTrainingBelt to not use FuturisticVibratorCheckChatTime**

As `FuturisticVibratorCheckChatTime` was no longer being used in the main vibrator,
it made sense to no longer use it in the training belt, and instead clear up the variable.
At the end of the day, the `FuturisticTrainingBelt` is still limited by its `UpdateTime`,
so this change has no practical effect on this item.

**Update FuturisticTrainingBelt to not trigger on OOC chat**

Add check for starting `(` in messages, and skip over messages that do this.
This is already included in the `FuturisticVibrator`, but missed in the `FuturisticTrainingBelt`.

**Minor extra clean-up**

Over the code there is some minor cleanup, such as improving loop ordering,
or taking checks outside loops. They should be obvious looking at the code changes.

### DIFF
--- a/BondageClub/Screens/Inventory/ItemPelvis/FuturisticTrainingBelt/FuturisticTrainingBelt.js
+++ b/BondageClub/Screens/Inventory/ItemPelvis/FuturisticTrainingBelt/FuturisticTrainingBelt.js
@@ -395,6 +395,9 @@ function InventoryFuturisticTrainingBeltCheckPunishSpeech(Item, LastTime) {
 		// Messages are in order, no need to keep looping
 		if (ChatRoomChatLog[CH].Time <= LastTime) break
 
+		// If the message is OOC, just return immediately
+		if (ChatRoomChatLog[CH].Chat.indexOf('(') == 0) return "";
+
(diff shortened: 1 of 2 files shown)


Ada18980:
I am fine with these changes.

Ben987:
Very good!  Thanks a lot for these changes.  Cheers :)

PR #2684, opened 2021-07-06
Up button improvements

ace-1331:
The up button could have strange behaviors in certain cases, some were accentuated by certain types of items, this aims to make it a bit more user friendly

- turned the up button into a toggle so it is possible to revert to normal height without losing focus on the character
- fixed the arrows being inverted while suspended
- fixed the character going back down on refreshes
- fixed the items with `FixedPosition: true` being rendered in a different position compared to the character, which can break most items that aren't just "decor"/ need to maintain consistency in height (like boxes, the kennel, etc.)


Ben987:
Thanks for that fix!  Yeah, that button is a little strange, lets hope it will help :)

PR #2685, opened 2021-07-06
validation lock check

ace-1331:
- fixed the validation code to understand cases where items were not locked

Expression triggers were causing false errors and refreshes because of the property RemoveTimer which is common on both locks and expressions.


Ben987:
Yes!  Thanks for that fix :)

PR #2687, opened 2021-07-06
Key accessory for collar

Manilla32:
Key accessory to go on the collar


Ben987:
Absolutely, this is a good idea.  Thanks :)

PR #2688, opened 2021-07-07
Slight adjustments before beta

ace-1331:
- split the garters for multicolor and left/right options
- adjusted the kennel's priority to work better with the wooden sign as well as making it easy to escape when the door is open


Ben987:
Sure, this will look nice.  Thanks a lot :)

PR #2689, opened 2021-07-07
minor fixes

ace-1331:
- corrected a line i improperly reverted in a previous pr
- added garters to the hide array to prevent them from floating when using the up button

Note: the check is not failing because of this PR, but because of one of the Pandora WIP commit from an hour ago


Ben987:

PR #2693, opened 2021-07-09
Recreate chat room fixes

Ada18980:
Fixes the following bugs:
1) Adds a small delay so that public rooms recreated have their admins changed properly - Tested as part of (2)
2) Changes the logic for new room creation to increment the number by 1 instead of setting it to a random number - Tested by creating a 3 person room, disconnecting one player, setting size to 2, then relogging as the 3rd character. Repeated twice to verify incrementation. Tested with 20 person room title.


Ben987:
Alright!  Thanks for these fixes :)

PR #2694, opened 2021-07-09
Update timing on auto shock collar and futuristic items

Ada18980:
Adds logic similar to the new instant logic for the futuristic vibrator to the following items:
Auto Shock Collar
Auto Shock Unit
Futuristic Panel/Ball Gags
Futuristic Training Belt


Ben987:
Sure!  Thanks for these updates.

PR #2695, opened 2021-07-09
Allows lovers and owners to trap players in private rooms

Ada18980:
The Auto-remake feature for chat rooms now looks at the player's items and determines if there are any owner-locked or lover-locked items that tether, enclose, or mount the player. Since these locks cannot be applied by non-owners or non-lovers, a troll cannot apply them to keep a player trapped. Also, since these items make it impossible to leash a player, a troll cannot leash a player who has these conditions into a private room.

This allows for owners to lock their subs in a private room as part of roleplay without the player being able to change the room properties (e.g. to make it public and plead for help)


Ben987:
What if the owner leaves for a week?  Will the sub be trapped for a week?  If that's the case we can't do it.

Ada18980:
> What if the owner leaves for a week? Will the sub be trapped for a week? If that's the case we can't do it.

Yes, same way a dub can get trapped in full sensory deprivation for a week due to owner leaving, unable to see, whisper, or speak to anyone. Regular difficulty and lower can still use safeword

PR #2696, opened 2021-07-09
Allow more tries with a lent lockpick

Ada18980:
This makes the game easier but the pick will still eventually break.


Ben987:
Thanks, many players seems to be having a hard time with lock-picking, this could help a little.  Cheers :)

PR #2699, opened 2021-07-10
Favorites for Extreme Difficulty & Worn Items

Nina-1474:
On the Extreme difficulty the inventory permissions mode is blocked for all items (except for strong locks as of this release), which means players can't use it to mark items as favorites.
The menu now shows all items regardless of difficulty and the function to cycle between permissions is what handles the difficulty check instead, skipping past the blocked and limited (except for strong locks) states on Extreme.

When an item is worn it previously ignored any clicks in permission mode. Now it allows clicks but only toggles between open and favorited, skipping the blocked and limited states.


Ben987:
Yes, yes, yes.  Thanks for that fix :)

PR #2700, opened 2021-07-10
Modular Item Validation Fix

Nina-1474:
A fix to prevent crashing for extended items with no Type in their config, such as the Kigurumi Mask.


ace-1331 (approved):

Ben987:
Of course, thanks for that fix!

PR #2701, opened 2021-07-10
Unusable Favorites Fix

Nina-1474:
When an item cannot be used due to failed prerequisites but is a favorite of the character, the box background was white when it should be grey. Now fixed.


ace-1331 (approved):

Ben987:
Sure!  Thanks for that one :)

PR #2703, opened 2021-07-10
Futuristic Crate Alpha Mask Fix

Nina-1474:
The Futuristic Crate's alpha masks were on the main asset, meaning they were applied while drawing the first layer on priority 1, before any other items were drawn. It's been moved to the frame layer so it masks items as intended, and additional masks have been added for above and below.


Ben987:
It makes perfect sense.  Cheers :)

PR #2704, opened 2021-07-10
minor fixes

ace-1331:
- fixed the futuristiccrate window type blind effect using an unexisting blindness level
- fixed the hitching post conflicting with handhelds 
- 5 color bracelet wasnt showing for XL/L bodies because they were lowercase
- fixed the locked icon label to properly display "Item is locked"


Ben987:
Yes!  Thanks for these fixes :)

PR #2705, opened 2021-07-10
Bugfix: Font size setting adjustments

elliesec:
Not really a fix so much as an adjustment based on feedback in the beta errors channel. This adjusts the font size settings so that:
* Medium is now the same size as the old default font size (for the dark 1/light 1 themes)
* Large should now allow for a larger size on mobile devices and smaller screens as it did before in the light 2/dark 2 themes (although not to quite the same extreme on mobiles)


ace-1331 (approved):

Ben987:
It makes sense.  Thanks a lot!

PR #2706, opened 2021-07-10
allfours fix for futuristic bra

ace-1331:
- hide the futuristic bra in allfours since it has no asset for it

Spoke with Ada for permission before submitting the fix


Ben987:
Sure!  Thanks a lot :)

PR #2708, opened 2021-07-10
Color Picker Fixes

Nina-1474:
When the colour in the colour picker screen was changed using certain methods, such as typing in a hex code or loading a saved colour using the new feature, the 'current colour' box (near the bottom beside an 'initial colour' box) didn't change.

When viewing another character but changing the colour for an object on yourself, some buttons such as Exit and Default did nothing.


Ben987:
Sure!  Thanks for that fix :)

PR #2709, opened 2021-07-11
WiredEgg missing AllowEffect

ace-1331:
- added missing AllowEffect on the wiredegg which caused the properties to be removed and cause unnecessary refreshes


Ben987:
Of course!  Thanks :)

PR #2710, opened 2021-07-11
kennel changes

ace-1331:
- made it so the door can always be opened, especially useful when paired with items that block the devices slot
- fixed which sound is played when others perform an action


Ben987:
Yes!  Thanks for that fix :)

PR #2711, opened 2021-07-11
garters shop

ace-1331:
- added a label for the garters group at the shop, under underwear
apparently more people than I do use this menu XD


Ben987:
Absolutely, thanks!

PR #2712, opened 2021-07-11
Shock unit: red light

ace-1331:
- updated the shock collar animation to trigger a red light when a shock is triggered
- added a resettable shock counter to shock units (also included with the shock clamps/dildo/prison suit/plug since they share the same code)
- improved the menu to be more in line with newer items (the buttons are greyed out instead of disappearing now)

I did not include any change to the automatic shock unit in this PR since Ada added that one and its logic is vaguely different. Letting her decide if she wants to include those changes


Ben987:
Very nice!  Let's try it in the next beta.  Thanks :)

PR #2714, opened 2021-07-11
Fixes issues with the training belt

Ada18980:
-Fixes issue where the current mode as it appears in the access denied screen is wrong, and the setting won't apply
-Fixes a crash when the player is in a dialog when it triggers


Ben987:
Absolutely!  Thanks for the fixes :)

PR #2717, opened 2021-07-11
Beta-fix: AssetCheck: respect CopyConfig

jomshir98:
Makes AssetCheck follow extended item's `CopyConfig`, so it can catch more errors


Ben987:
Sure!  Thanks for that one.

PR #2719, opened 2021-07-12
Fix intensity = 0 in training belt

Ada18980:
Should be -1 so that it doesn't arouse when off


ace-1331 (approved):

Ben987:

PR #2721, opened 2021-07-12
fix: finished collar rope split

ace-1331:
these are long outstanding issues with the collar ropes:

- fixed the collar ropes being invisible in any poses besides the main one as they werent split into layers yet
- added a allfours version for the short collar rope to better support the pose
- fixed missing locks for the collar ropes
- removed an old leftover from the sybian PR (unused image in itemlegs folder)


Ben987:
Yes!  I've seen weird things with that one before.  Thanks a lot :)

PR #2722, opened 2021-07-12
[asset] [beta fix] [Natsuki] Bracelet fix...again

Codefreak64:
Bracelets were not able to be edited or removed from another player, this fix makes it possible for it to happen


Codefreak64:
forgot to name the PR, only added tags to it

ace-1331 (changes requested):
this pr re-introduces duplicate images with the incorrect lowercase names (As well as adding the missing one for OverTheHead pose with the same lower case issue in some instances)

Codefreak64:
DO NOT MERGE THIS PLEASE

I found out this breaks a few things that were fixed so making a New PR

PR #2723, opened 2021-07-12
Futuristic Bra Extra Type

Ada18980:
Adds an extra alternate image for the futuristic chastity bra.

No extra features, tested with all body types. Validated to make sure the text/heart still show up.


Ben987:
Very cute, I love it!  Thanks a lot :)

PR #2725, opened 2021-07-12
Shop Sell Group Fix

Nina-1474:
A small fix to ensure the two new groups appear in the shop's sell options in the rare case they're the only owned items on that page.


ace-1331 (approved):

Ben987:
Absolutely!  Thanks for that quick fix :)

PR #2726, opened 2021-07-12
[Asset] [Beta Fix] [Natsuki] R70 Bracelet fix

Codefreak64:
Fixed the issue with bracelets not being able to be interacted with by other players


Ben987:
Yes!  Thanks for that fix :)

PR #2730, opened 2021-07-13
missing lock

ace-1331:
- added a missing lock for the pony bits after the new type was added


Ben987:
Yes!  Thanks for the fix :)

PR #2731, opened 2021-07-13
removed unused files

ace-1331:
- removed unused files in folder ItemMouth2 & 3 since those are no longer needed (somehow one item still had files in there)


Ben987:
Absolutely!  Thanks again.

PR #2732, opened 2021-07-13
hover priority changes

ace-1331:
- reworked hovers to be rendered above everything on the canvas (except beeps/alerts)

Currently, it's tricky to make proper hovering items because the order for which items are rendered matters. This is why on some pages people change the order of buttons (making the code harder to follow). I've seen many cases of this, but here is a recent example:

before:

after:

It should also allow us to make more complex and bigger hovering items without it being rendered poorly


Ben987:
Sure!  Thanks.

PR #2733, opened 2021-07-13
Colour Entry Fix

Nina-1474:
When typing in a hex code for a colour, the step to refresh everything with the selected colour always included the input box as well, which resulted in the player being interrupted halfway through with a "completed" value. Now when the change comes from the input box, the box's value will not be changed.
There was also an issue where the colour was converted from hex to HSV then back to hex which could change the value.


Ben987:
Yes!  That one is annoying, thanks :)

PR #2734, opened 2021-07-13
Blind Examine setting change

Nina-1474:
In the Immersion preferences menu the 'Sensory deprivation' and 'Disable examining when blind' settings affect each other.
When the deprivation is at the minimum setting of Light, then activating the Disable setting does nothing because the player can never be fully blind and therefore never be unable to see others.
When the deprivation is at the maximum setting of Total, then deactivating the Disable setting does nothing because the player will be shown a 1-person room and can't select characters to examine them anyway.
The Disable checkbox will now be disabled and automatically set in these two cases to help let the player know that changing it would have no effect.


Ada18980:
This is good, especially since the virtual world now allows examining even while blind

Ben987:
Sure!  Thanks for that change :)

PR #2736, opened 2021-07-13
Bugfix: Ensures characters refresh properly after calling `InventoryWearRandom`

elliesec:
## Summary

As part of #2649, the `InventoryWearRandom` function was changed such that calling it without a `Refresh` argument would result in the character not refreshing. This meant that a subset of `InventoryWearRandom` calls in the game would not refresh the character after an item had been applied when they should have, causing the added items not to render.


Ben987:
Yes!  I've seen that bug before, it was annoying.  Thanks a lot :)

PR #2737, opened 2021-07-13
Permissions Mode Resetting

Nina-1474:
At the moment if the player is in item permissions mode in the inventory menu and clicks to change focus to a different zone, the inventory jumps back to normal mode. Now it will stay in permissions mode until the player exits or clicks the mode toggle button again.


Ben987:

PR #2738, opened 2021-07-13
Fix bug with training belt not updating if another player changes it

Ada18980:
Fixes:
-Public mode not updating if another player changes it
-Crash when it triggers an orgasm


Ben987:
Sure!  Thanks for that fix :)

PR #2742, opened 2021-07-14
Random Item From Lists improvement

Nina-1474:
When an NPC applied a random item from a predefined list, it was not using the existing functionality that avoids the player's blocked/limited/hidden items. These cases (Pandora chastity/Daily puppy job/Private room chastity) have been changed to use this.

The old method also didn't care if the target group was blocked by another item or the asset was not normally applicable due to prerequisites, which has the danger of creating invalid combinations. The random functions have been edited to support this, since otherwise NPCs will act like they've applied something when they haven't. But long-term it would be better to handle this in the NPC's dialog/checks instead.

This will conflict with #2736.


Ben987:
That's a good idea!  I'm curious to try it.  Thanks :)

PR #2746, opened 2021-07-15
lock change belts

ace-1331:
- fixed a missing property after converting the belts


Ben987:
Yeeees!  Thanks for that fix :)

PR #2747, opened 2021-07-15
Spanking toys get type issue

ace-1331:
- fixed an issue with the get type function of handhelds returning an improper type when wearing an item with a type in the ItemHands slot


Ben987:
Absolutely!  Thanks for that one also :)

PR #2749, opened 2021-07-16
Fix crash when futuristic training belt triggers orgasm

Ada18980:
Fixed and tested. Forgot to push this commit last time. >.<


Ben987:
Sure!  Thanks for that fix :)

PR #2751, opened 2021-07-16
[Fix] Blouse TapedHands Missing

ItsJustMeVerity:
Fixes an issue with assets for the TapedHands pose not existing for either version of the Blouse.


Ben987:
Yes, yes, yes.  Right on time for the full release :)

PR #2752, opened 2021-07-17
fix leotard behavior

ace-1331:
#2585 caused a few side effects to the item which many have expressed dissatisfaction for, including the fact that categories that used to be hidden by it are no longer hidden.

This aims to fix the current behavior to allow the new ability to show items above it, while still allowing the same behavior as before (hiding most things). 

Because it wouldn't make sense to show piercings and hide belts (and so on), I renamed the groups and grouped up the related ones in the same module


Ben987:

PR #2753, opened 2021-07-17
fix chat room name garble

ace-1331:
- fixed the chatroom name garble immersion setting not garbling the room name when a chatroom is being updated. 

Before it would show "ace updated the room. Name: test room. Limit: 10.  public.  unlocked."
Now it shows "ace updated the room. Name: mmmm mmmm. Limit: 10.  public.  unlocked."

To avoid "cheating/ruining the immersion" of that setting. Seems like that was the only case left of a room name not getting garbled


Ben987:
Sure, this makes perfect sense.  Thanks :)

PR #2754, opened 2021-07-18
Fix ruined orgasms

Ada18980:
Fixes ruined orgasms so it does not randomly fail when there are others in the room.


Ben987:
Sure!  I will prepare a hotfix for that one :)

PR #2755, opened 2021-07-18
fix random typed items

ace-1331:
- fixed some typed item not having Random: false which caused problems in areas where npcs would apply a random type that blocks the player within the room (such as serving drinks in online rooms)
- fixed the asset check to include proper verification for the new Random property on typed items

This covers all hogtie + allfours + frogtie types currently available in the typed items which is in line with how normal items are blocked from being randomly applied.


Ben987:
Yes!  Thanks, I will prepare a hotfix right now.

PR #2756, opened 2021-07-18
multicolor witch hat

ace-1331:
- split the witch hat for multicolor

credit goes to Dordimar for the split, I only tested&submitted it


Ben987:
Yes!  Thanks you two :)

PR #2757, opened 2021-07-18
Split PetTape into separate arm and leg slots

wildsj:
Rather than all or nothing, the PetTape can also be applied to just arms or legs and be worn with clothes.

It works like the HempRope and ZipTie Hogties. A known issue with these is that they don't allow crawling around the club, unlike some hogties e.g. Petsuits, FrogtieStraps and the orgional PetTape.


Ben987:
Yes, yes, yes!  Thanks a lot :)

PR #2758, opened 2021-07-18
Futuristic Training Belt Punishment Fix

Ada18980:
The futuristic training belt took gags into account correctly but not stuttering. This fixes that.

Re-tested with vibrator and gag.


Ben987:
Alright!  I will hotfix that one also :)

PR #2759, opened 2021-07-19
fix incorrect room blocks persistance

ace-1331:
- fixed the block category UI menu toggle not being properly canceled

Currently, going in chatadmin to update room blocks will keep them the way you left them even if you cancel out, which can cause desyncs (for example, you have or dont have a room block others have). This was in part due to the fact that ChatRoomData.BlockCategory, ChatAdminBlockCategory and ChatBlockItemCategory were all referencing the same array so the "reset" logic wasnt working as intended (and was logically incorrect).

This fix properly instates the expected (and intended) behavior where the menu will always reflect the current state of the room when you first open the chatadmin screen, then will keep the last edits until you quit by saving or cancelling


Ben987:
Of course, thanks for that one :)

PR #2760, opened 2021-07-19
Misc. things

EmilyR42:
- Added a "no shine" variant to Swimsuit1
- Added a non-functional handheld key item as a roleplaying prop
- Added Lingerie Shop and Red Sci-Fi Room backgrounds


Ben987:
Absolutely!  Thanks for adding that stuff, I will test it soon :)

PR #2761, opened 2021-07-19
fix missing dialog

ace-1331:
- replaced old references to dialog which no longer exists


Ben987:
Alirght!  I will prepare another hotfix for this :)

PR #2763, opened 2021-07-20
type fixes

ace-1331:
- fixed various JSDoc & typedef issues
- fixed cases of references to non-existing group properties in Asset.js
- fixed cases of KeepNaked being used while it no longer exists
- fixed some cases of DrawText being called with extra arguments
- fixed `StruggleLockPickProgressGetOperation` not returning the proper dialog text for certain locks because the wrong property was accessed (item.Name instead of item.Asset.Name)


Ben987:
Sure!  I'm not sure what it does in your IDE but if it helps you to code it's fine for me :)

PR #2764, opened 2021-07-20
Random Groups fix

ace-1331:
- fixed a long outstanding bug where groups did not have `Random` defined which prevented the intended behavior from occurring
- enabled `Random: false` on clothes and added it to the Flippers and diapers.

This fix should add more variety to randomly generated appearances. The issue was that the FullRandom appearance was intended to apply non-default groups (bracelets, accessories, corsets, garters, etc.) at a 20% rate, but all groups had no random property so it never happened. Since this "bug" is many years old now, what was once the intended behavior is now possibly an undesired change so I approached it this way:

- Reduced the odds of a non-default group to 8% (20% gave too many extra items, too often)
- Disabled groups that don't fit well/rarely match with anything (suits, cloth accessories, etc.)
- Disabled cosplay groups (Ears/Tails/Wings)

Feel free to test it out and let me know what tweaks you prefer. We can always allow less/more, or just fix the old code to remove any  references to `AssetGroup.Random` . I think it's nice to add a bit of diversity. This and the recent addition of extended item support for randomization is a big boost in appearance diversity.


Ben987:
Sure!  But is 8% enough to make NPCs unique?  We can always raise it again later if it isn't, thanks :)

PR #2765, opened 2021-07-20
Fix: Private ActivityTarget being replaced only once

jomshir98:
Fixes problem with dialog reported on Discord:


Ben987:
Sure!  Thanks for that fix :)

PR #2766, opened 2021-07-20
Sensory Deprivation Changes

Ada18980:
Changes total sensory deprivation so that if you can see the player in VR then their name shows up when doing actions/emotes/chat even in total sensedep.

Also makes it so if you have the immersion option enabled that hides non adjacent players, it will replace some names with "someone" if you can't see who's doing the action or receiving it.

NOT a hotfix, makes some significant changes so do not merge till beta.


Ben987:
Lets hope that one won't create any issue.  Thanks!

PR #2767, opened 2021-07-20
Arousal Glow Options

Ada18980:
Adds two more options for the arousal screen filter (from the graphics menu):
1) Gradual filter (increases over time based on arousal, visible even at 50% arousal)
2) Strong Gradient (a very powerful visual effect with a gradient flowing from the edges of the screen and pulsating gently every 2s)

The old code was refactored to be compatible and reduce code waste. Applies in both Chatrooms and Private room. The original behavior (light glow at >90 arousal) is still available as the default option.


Ben987:
Very good, I'm curious to test this one :)

PR #2768, opened 2021-07-20
bracelet fixes

ace-1331:
- fixed bracelets being invisible with TapedHands
- removed unused unnecessary empty images from the Bracelet folder (because the pose already hides it anyway)


Ben987:
Yes!  Thanks for cleaning that up :)

PR #2771, opened 2021-07-21
Moves the $ sign to the beginning of the price rather than the end

Ada18980:
This helps reduce some confusion with items that end with numbers, e.g. Swimsuit Type 3 35$ becomes Swimsuit Type 3 $35.


ItsJustMeVerity:
That was quick; thank you! This should really help with readability. I did try to find where it was stored and try myself before making the post in #suggestions yesterday, but got lost, so I'm glad someone picked it up :P

Ada18980:
> That was quick; thank you! This should really help with readability. I did try to find where it was stored and try myself before making the post in #suggestions yesterday, but got lost, so I'm glad someone picked it up :P

Glad to help ^^

Ben987:
Sure!  Thanks you two :)

PR #2772, opened 2021-07-21
minor tweaks to items

ace-1331:
- configured the Lace Babydoll to not hide the new piercing
- fixed where you could move around room in suspension cuffs
- fixed an improperly named img for the new piercing


Ben987:
Yes!  Thanks for these fixes :)

PR #2773, opened 2021-07-21
hideitem tweaks

ace-1331:
this as a minor conflict with #2772 on the babydoll line, will resolve it on either PR when one gets merged.

- fixed items still hiding an item that no longer exists after it was merged a while ago in #1343 (`WeightedPiercing`)
- fixed the wedding dress not hiding the DiaperHarness as intended because it was improperly labelled (`ItemDevices` -> `ItemPelvis`)
- fixed PetCrawler hiding a non-existing tail (that tail renders fine with it now either way)
- fixed the X-Cross having incorrect hideitems (that did not need to be there now anyway)


Ben987:
This one causes a conflict.  Can you resolve it and test again?  Thanks :)

ace-1331:
all set

Ben987:
That was quick!  Thanks again :)

PR #2776, opened 2021-07-21
Improvement: Use TypeScript for assetcheck

jomshir98:
**~~Depends on and includes #2764, #2775~~**

**Actual size of this PR is: +338 -390** (the rest is autogenerated `package-lock.json`)

Changes assetcheck such that it uses TypeScript for the typechecking part.
This carries many improvements, for example:
- Clearer and more verbose error messages
- Less code around typechecking, because it is fully handled by TypeScript
- IDEs able to use the types to both hint users creating assets and show errors right as they are made without having to run any test first
- Ability to share defition types with asset loading code for more robust error detection

*Note, that this PR doesn't change any code of the actual game (except for small fixes), all changes are in tooling around it.*


ace-1331 (approved):

Ben987:
Since I have no idea what it does, I'll leave the gurus to review that one.  

@Nina-1474 , @elliesec , is this something that you guys need?  Will it help you to code the club without making it too complicated for new contributors?

Thanks a lot!

BondageClub/Tools/Node/AssetCheck.js
```diff
@@ -180,95 +125,70 @@ function validateArray(definition, obj, description, allowMissing=false) {
...
 		localError = false;
-		validateObject(types.AssetGroupType, Group, `Group ${Group.Group}`);
 
-		// Don't validate group further, if it had bad data
-		if (localError) {
-			continue;
-		}
 		Groups.push(Group);
+		/** @type {AssetDefinition[]} */
 		const GroupAssets = (Assets[Group.Group] = []);
```
elliesec, line 139: Question: do we ever actually do anything with this? Kind of looks like it's unused to me - we just push things to it and then never actually use it. Not entirely related to this PR, but if it's not being used, might as well clean it up while we're here.

elliesec:
@Ben987 - this will be beneficial to anyone working with an IDE, and should improve the flexibility of the type-checking system in general, because TypeScript is much more expressive for defining types. It shouldn't impact contributors any more than the current asset checking script does.

BondageClub/Tools/Node/AssetCheck.js
jomshir98: @elliesec `GroupAssets` is used as shortcut for pushing only, however `Assets` is later used on line 227 and is planned to be used by functions for checking referenced assets exist

elliesec: In that case, is there any reason not to simply use `GroupAssets` on line 227? It feels somewhat odd to define it as a reference for writing and then to use the long form for reading.

elliesec (approved):

PR #2780, opened 2021-07-22
Use string.replace rather than string.replaceAll in futuristic vibe

Ada18980:
Fixes the futuristic vibe on older browsers.
Also sneaking in a minor change that removes missing dialog tag when switching vibrator types.


Ben987:
Yes!  Thanks for that fix :)

PR #2782, opened 2021-07-22
New clothing

gatetrek:
needed 2 clothing with one able to remove the lace that's around the belly
Screenshot_17


Ben987:
These are lovely!  Thanks a lot, this is great :)

PR #2783, opened 2021-07-22
Spiked Wristbands

EmilyR42:
- Added new Spiked Wristbands item with options to wear one on either arm
- Tweaked a few existing items to hide bracelets

PR #2786, opened 2021-07-23
Bracelet and some hiding

gatetrek:
this pr adds a bracelet that can go on either arm
hiding of the itemnipples of the last thing I added as clipped to much or looked odd


ace-1331:
note that this PR will conflict with #2783 and whichever gets merged last will need to regroup their item under the same Bracelet group in the extended file

ace-1331 (approved):

Ben987:
Absolutely!  Thanks a lot for that one :)

PR #2787, opened 2021-07-23
Modular Items and Custom Tag improvements

ace-1331:
- Added support for subscreens in modular items
- Improved AllowModuleTypes handling to properly allow the "null" type which will make modular item more compatible with NPCs and scripts (aka the basic type is no longer "invisible", all current modular items benefit from this)
- Improved AllowTypes handling for modular items (the generated values from AllowModuleTypes are now appended instead of overwriting AllowTypes, for better customization of the AllowTypes)
- Changed the custom collar tag into a modular item to allow for various tag styles + text (uses the new feature to use as an example)
- Minor improvements to the custom collar tag style


Ben987:
Absolutely, this is a good idea.  Thanks for that one.

PR #2788, opened 2021-07-24
Futuristicbelt fix

ace-1331:
- fixed a property protection that didn't work because of `CommonTime` instead of `CommonTime()` which ended up adding a number to code (!)
- fixed getting kicked out of the character screen when clicking on other groups


Ada18980:
Thanks for locating the issue >.<

Ben987:
Yes!  Thanks for that fix :)

PR #2789, opened 2021-07-24
Futuristic training belt now uses original chat message

Ada18980:
Now it doesn't pay attention to stutter or try to reverse-engineer the gag level at all, which in hindsight was what I originally intended to do anyway. This way is also much less bug-prone. Still doesn't shock people if they are gagged past gag level 8 (mmmmmm!)


ace-1331 (approved):

Ben987:
Sure!  Thanks for that one :)

PR #2790, opened 2021-07-24
Fix/Improvement: Hogtied pose support for ItemNeckRestraints

ace-1331:
- added proper support for the Hogtied pose for ItemNeckRestraints items (similar to allfours, so the item wouldnt look like its way offscreen)


Ben987:
Yes, yes, yes and yes!  Thanks for that one :)

PR #2791, opened 2021-07-24
rejoin room fix

ace-1331:
- fixed a soft lock where you were stuck in an empty chat search

Due to how "environments" and the "return to room" works, it didn't account for the fact that the last chatroom you rejoined may have been in a different environment (or may now be), meaning that if it was attempting to rejoin a room that is now in the opposite "environment" (Test vs live), you would be blocked in chatsearch without being able to leave.

Its not something that happens regularly for the average player, but it has quite the griefing potential


Ben987:
That's probably a rare one for regular players, but thanks for that fix :)

PR #2792, opened 2021-07-24
Feature: Different Currencies

fleisch11:
allows one to select a currency 
(i chose euro and roubel, cause those are the currencies of the supported languages)
(i chose the chinese Yuan/Renminbi, because i frequently see chinese club rooms)

the idea is, that you get an idea, of how much something costs in your currency

the exchange rates are intentionally not exact, because they constantly change, decimal's would be a problem, and because it'd be less usefull

i tested it in every situation, with no problems
what i tested was:
-that it doesn't change stuff like $PONY, in the HorseWalk.csv
-that every price in the $999 format gets changed correctly
-that every price in the $ReplaceMoney format get's changed correctly
-that the prices of items in the shop get changed correctly
-that the money the player possesses gets changed correctly in the informationsheet and the shop

if i forgot a situation, that i should test, please tell me (or test it yourself, if you want)

i think, i explained and commented enough in the currency.js, but i can comment more, if it's not enough


ace-1331:
I feel like that's unecessary complexity..

fleisch11:
@ace-1331 
first: it is not complicated at all to add anything money related
second: it's not fitional exchange rate, it's just not exact, because as i said, the exact exchange rate constantly changes and it's not usefull to know that $1=130,12Yen.

ace-1331:
Remember that when you PR something, someone will have to work around it and might forget about it eventually. Will you be the one to PR fixes whenever someone forgets to use this instead of simply putting up the price in dollars?

Yes calling a method is simple, but it doesnt mean everyone will think about it when adding/changing stuff and it's only going to be a nuisance.

Giving people more things to think about when adding new stuff is a roadblock or a source of "errors", we've had multiple examples of that in the past.

The controller code isn't maintained by anyone and is rather buggy since there's already been a few changes that havent been adapted to it for example.

Sorry to be blunt, but this is pure code bloat

jomshir98:
Exchange rates tend to change quite a lot over the years, so they would need to be updated for this to make sense... However:
The main problem is, that currency in Club is not aligned with RL costs at all, so converting them doesn't even make sense.

From my view adding new ~125 line script that on top of it **modifies dialogs** (and I would argue noone making dialogs would expect them to be changed) adds a lot of complexity.

Also did you test how this works combined with translation?

fleisch11:
yeah, i didn't think, this would end up being added. was fun to programm though, so no regrets here

PR #2793, opened 2021-07-24
Improve visibility for nipple piercings under bras

Fbdsman:
This pull request tweaks the 'Hide' and 'Expose' properties for bras in Female3DCG.js, so the nipple piercings do not appear/disappear when removing/putting on a bra,  when they should clearly be visible underneath. 

For reference, some screencaptures of the result with different types of bras. From left to right, the models are wearing the chain necklace piercings, the nipple rings with chains attached, and the nipple chastity guards :

bras01
bras02
bras03
bras04
bras05


ace-1331 (approved):

Ben987:
It looks good!  Thanks a lot :)

PR #2794, opened 2021-07-24
Favourite Icon Improvements

Nina-1474:
- Inventory items will now check if the player has favourited an item when examining a group on another character. The ordering will prioritise items both have favourited, then items the target only favourited, then items the player only favourited, then the rest as usual.
- Items in the shop will now display the icons it can. Which is currently handhelds, owner-only and lover-only items.
- Small fixes to allow the kidnapping inventory menu and lock lists also prioritise the order for favourites.


ace-1331:
quick note that #2776 's `DialogSortOrder` type will need to be updated before/after depending on which PR gets merged first

ace-1331 (approved):

Ben987:
Very nice, I like it.  Thanks a lot :)

PR #2797, opened 2021-07-25
fix extended pony bit

ace-1331:
- fixed the harness bit gag in slot 2&3 that were not correctly labeled as extended after #2495


Ben987:

PR #2799, opened 2021-07-25
more bit gag fixes

ace-1331:
- fixed a bug where the default option on the harness bit gag was incorrect due to improper type ordering
see this: https://discord.com/channels/554377975714414605/554378725916147722/868934482223333396


Ben987:
Sure!  Thanks for that small fix :)

PR #2803, opened 2021-07-28
Changed ColorGroup lookup tags and fixed misc ColorGroup names

ace-1331:
- changed how color groups are looked up so the group tag will be GroupName-AssetName-ColorGroup instead of just ColorGroup to avoid easily conflicting tags and make them easier to look through just like layer names & asset names.
- added missing color group names to the file (notably those of the coffin, cryo capsule, vacbed and posture collar)
- removed useless color groups from the crib and unused group tags (upper/lower)
- updated the french translation for color groups
- (other) fixed the shoulderless wedding dress not using the intended pose when in OverTheHead + LegsClosed


Ben987:
Absolutely!  Et merci pour la traduction en francais :)

PR #2804, opened 2021-07-28
fix timer lover lock

ace-1331:
- fixed owners being able to apply lover locks, but unable to edit them.

A change in the recent past allowed owners to apply lover locks on their subs, this change was not reflected to the timer lock. Which caused the odd ability to be able to add the lock, but not modify it. (which turns it into another 5min lock)
The rules to prevent owners from using lover locks will work with this as it wont allow modifications.

On that note, while testing, I found a bug where the UI still allowed the option to unlock (but validation catches it and re-locks the item)
Therefore, it means that the block owner from using lover locks allows lovers to lock a sub away from their owner. I'm not 100% what the intended (or desired) behavior is, the fix is different depending on the answer. (Personally, I find it odd that an owner cannot unlock lover locks). We'll talk about it and post a fix for that problem at a later date.


Ben987:
Yes!  I've seen that one live before, thanks a lot :)

PR #2805, opened 2021-07-29
Fix // comand

NoneNoname:
`//` will be handled the same way as normal message after removing the first `/`

Gagged OOC check is refactored into a function to be more readable


ace-1331 (approved):

Ben987:
Sure!  Thanks for that fix :)

PR #2808, opened 2021-07-29
Add LayerVisibility and Visibility to kigu mask

Ada18980:
This makes it so that if you view the kigurumi mask from the player dialog, you can see what you (the player) has on under it. Otherwise, the mask displays normally (including to other players).


Ada18980:
WIP

PR #2810, opened 2021-07-30
fix - make ceiling chain unchangeable when locked

kimei-nishimura:
Corrected an oversight when making the ceiling chain lockable, by making the position fixed when the chain is locked.


ace-1331 (approved):

Ben987:

PR #2811, opened 2021-07-30
[Fix] [Natsuki] Kneel wrap addition

Codefreak64:
small fix for the plastic wrap
-Adds in knelling pose assets for plastic wrap


Ben987:

Codefreak64:
turns out i messed something up since github did that side by side thing with this so....give me one sec to fix the fix

PR #2813, opened 2021-07-30
Misc fixes

ace-1331:
- added use strict to fix the lint check in PokerHandValue.js
- fixed a typo
- fixed login maid items validation missing items that would result in the items with buygroups never getting removed.

natsuki will do a separate pr for the chain mistake that makes the asset check fail that you just merged.

Also note:
PokerHandValue.js line 262 to 266 -> I know its a WIP, but if you didnt see, `X` and `pos` are never declared so that code will likely cause bugs


Nina-1474 (approved):

Ben987:
Absolutely!  Thanks for these fixes :)

Yes, PokerHandValue is still a work in progress, lot of that stuff will change.

PR #2818, opened 2021-08-01
Cat-Eye Glasses, Antennae, HairBack52

EmilyR42:
- Added Cat-Eye Glasses and Antennae items
- Added HairBack52
- Added Unicode arrows to Bracelet1's options


Ben987:
Very nice!  Thanks a lot :)

PR #2821, opened 2021-08-01
Development improvement: Automatic changelog pregeneration

jomshir98:
Adds ability to pre-generate changelog from commits to cut on required work when making commit by not having to write every single changelog entry manually.

Discussed with @elliesec


Ben987:
Thanks!  I'm sure that Ellie will appreciate :)

PR #2822, opened 2021-08-01
KinkyDungeon 1.51 fixes

Ada18980:
Some bugfixes for Kinky Dungeon 1.51, as well as minor balance changes and a new QoL feature to show the damage of the default attack


Ben987:
Yes!  Thanks for all these fixes :)

PR #2824, opened 2021-08-02
Kinky Dungeon 1.52

Ada18980:
Made some balance changes to struggling in Kinky Dungeon. Should be less random now, leading to less frustration.


Ben987:
You're on fire!  Thanks a lot :)

PR #2829, opened 2021-08-02
Kinky dungeon bugfix + linting

Ada18980:
Fixes a bug in Kinky Dungeon where weapons do nothing even when equipped.

Also fixes many linting errors in Kinky Dungeon.


Ben987:
That's a huge clean-up, thanks!

PR #2831, opened 2021-08-03
[Asset] [Natsuki] Latex collars

Codefreak64:
Adding in a latex collar because you can never have too much of it~

-Adds in 2 Latex collars to the shops
-Moderate price at $40
-Simple look makes it great for outfit making


Codefreak64:
DO NOT MERGE! HAS FAILED CHECKS

Codefreak64:
Nevermind on that. the reason it failed is due to an extra semicolon (which breaks code), HOWEVER its not effecting the main game, as its in the prototype bondage poker. I'll make a fix on that colon issue, in the mean time this can be merged in

Ben987:
Yes!  Thanks for these new collars :)

PR #2833, opened 2021-08-03
Lace Armbands & HairFront19 Revamp

EmilyR42:
- Added Lace Armbands item in bracelet slot
- Allowed BackCuffs pose for bracelet slot
- Updated the graphic for HairFront19


Ben987:

PR #2834, opened 2021-08-03
Futuristic Training Belt Update

Ada18980:
1) Uses PersistentData() to store state info so it doesn't cause too many refreshes.
2) Adds new word banning functions
3) Allows multiple words for both word bans and required speech


Ben987:
Very nice!  Lots of Dommes will be happy, maybe not that many subs :P

PR #2836, opened 2021-08-06
Another nipple piercing fix

Fbdsman:
When applying duct tape to the arms, the nipple piercings disappear. This fixes that. 

Previews : 
tape02
tape01
tape03


Ben987:
Sure!  Thanks for that fix :)

PR #2837, opened 2021-08-10
fix non-dismountable fuckmachine

ace-1331:


Ben987:
That was quick!  Thanks a lot :)

PR #2838, opened 2021-08-10
fix mu emotes

ace-1331:
- fixed /me not working with the MU style emotes 
- fixed MU style emotes not being able to contain any `:`


Ben987:

PR #2840, opened 2021-08-10
Removed some potentially offensive dialog

Ayesha-678:
Removed a couple lines of dialog, that could be seen as offensive, to be safe.


Ben987:
Sure, why not.  Thanks :)

PR #2841, opened 2021-08-10
fixed vulva filling items prerequisites + fixed allowactivepose

ace-1331:
- re-instated the old vulva-empty and instead opted to just outright block leg items with the fuck machine which fixes stuff like dildo bars and the training belt being unremovable.
- fixed a bug with `CharacterItemsHavePoseAvailable` that always returned true if one activepose is allowed instead of properly validating if the checked pose is OK. 

retested the fuck machine, dildos, training belt, spreader bar, etc and everything appeared to work as intended
retested stuff like leg items, skirt, etc. to make sure the `CharacterItemsHavePoseAvailable` still behaves as expected. (It actually properly shows which pose is allowed in the menu now, which was the originally intended behavior)


Ben987:
Sure!  Thanks for that fix :)

PR #2842, opened 2021-08-10
[beta fix] [Natsuki] Inflating collar head fix

Codefreak64:
a small over sight while making the collar, I forgot to make it so you couldn't nod your head (as it would be difficult to do so IRL with it)
this PR adds in the code i overlooked for the collar


Ben987:
Sure!  Thanks for that fix :)

PR #2843, opened 2021-08-10
fixed undefined text for VFXFilter setting

ace-1331:
- fixed undefined text for VFXFilter setting
the text displayed in the menu said "undefined" instead of the default value for someone who changes this setting for the first time


Ben987:
Absolutely.  Thanks for that one.

PR #2844, opened 2021-08-10
New Item Face Paint

Shiranui-Izayoi:
Made a new item to allow people to "change" their skin colour. I have used the base mask from the Kirgumi mask with Ada's permission.

All coded up and tested all seems to work. I am allowing it to be locked so that others can stop people messing with it if they want to have a different "skin colour".

I may in the future add this to the wardrobe so it can be used with hoods. But currently other than `Mask` I do not see any slots that really fit.

The preview image was designed to look like a geisha. In order to get across the idea of it being a face covering that can be used to change skin colour.

The default colour of the item is actually the same as the Kirgumi Mask, and thus white skin. In fact when applied it can be difficult to tell it is there. It is then up to the played to choose if they want to change it.


BondageClub/Assets/Female3DCG/Female3DCG.js
```diff
@@ -4213,6 +4213,7 @@ var AssetFemale3DCG = [
 				],
 				Extended: true
 			},
+			{ Name: "FacePaint", Fetish: ["Latex"], Value: 50, Priority: 7, Difficulty: 1, Time: 1, Random: false, AllowLock: true, DrawLocks: false, DefaultColor: ["#9A7F76"], Hide: ["Head"] },
```
ace-1331, line 4216: Is `Fetish: ["Latex"]` actually meant to be there? If it's just face paint, I don't think that's needed

Shiranui-Izayoi: You are right I forgot to take it out. Also because I did change the name during testing. Origanlly I was going to name it something more fitting. Since it's original idea was to be used basically to change "skin" colour. But decided on face paint to keep it simple. I will remove it.

Shiranui-Izayoi:
Just a quick thing if It would make it better I can change the name to something like "Synth Skin" or maybe "Nano Paint" or something if "Face Paint" doesn't fit. I can also change the colour to white to match the image if that is prefered. But unless requested I don't see a reason to.

Shiranui-Izayoi:
I forgot to put in the `Block: []` so that it doesn't block slots. And I also reduced the price since honestly it didn't need to be 50. As it isn't that special. If possible could this be added in Beta 3/before the next full release? I know typically things aren't added in the Beta but I feel that this isn't to big a thing and would be fun.

Ruilov3:
an ItemHood that hides the head, occupying a slot that is practically the kigurumi mask on a low priority to allow people to color their face sounds like a really roundabout way to approach this. 

Ellie and I have been meaning to do a body rework for a while now that will enable players to color their bodies beyond the 3 current presets and Ellie's been meaning to do bodypaints that can be applied to the body as part of that rework.

Once we get around to doing that it'd render this item pretty much obsolete. My recommendation would be to just have a little more patience and body+face paint will eventually come in a pretty neat and flexible way that will also not eat up a slot.


Cheers ♥

Shiranui-Izayoi:
Honestly I was not aware a body rework was in the works. Is there any idea of when it might come out? 

Since if it is in say 4 or more months that is still a lot of time that this item would be useful. If I lower the price to like 1 then when the body rework is released then the item can just be deleted with no problem.

As it is though until then this can be of quite a lot of use. Even more so since you said "Once we get around to doing that", which heavily implies it isn't high on the list of priorities and thus might be a long time before it is done.

Shiranui-Izayoi:
After consideration I have moved the item to the mask, cloth accessory and hat slots so it does not take up a restraint slot. And since those slots are much rarer to be used especially all together, this gives more freedom.

While I do now know of the plan to rework the body it is unknown when this will be done. Or if it has been started. So I prepose to add this as a temporary item. As this will serve as both a teaser for what the full rework would be able to do as well as allow those that want to do it to do so sooner.

The moment that the body rework is accepted/done this can be removed and I will do so personally.

ace-1331 (changes requested):
FYI, your recent changes to this PR break the asset check because you kept the old (not needed) description (asset name), and forgot to add the asset name to the 3 slots.

But yeah.. I understand Rui's point, this is a strange solution for something that could have been solved by making the kigu mask colorable, or working on allowing more body colors than the basic 3

Shiranui-Izayoi:
> 
> 
> FYI, your recent changes to this PR break the asset check because you kept the old (not needed) description (asset name), and forgot to add the asset name to the 3 slots.
> 
> But yeah.. I understand Rui's point, this is a strange solution for something that could have been solved by making the kigu mask colorable, or working on allowing more body colors than the basic 3

whoops will fix that. As for why I have not done either of those. Is first off I don't have the artistic ability to make the body colour and I have a feeling that simply allowing people to change the bodies colour as it is wouldn't work well.

While for not altering the Kigu mask Ada thought it would be better to just make it a seperate asset instead of adding blank eyes, blush, mouth to the mask then adding a final part to put it at the right level. So thus I made this item at their suggestion.

Ben987:
@elliesec , can you help us here?  Do you have a different solution that's coming up that could be in conflict with this pull request?  Thanks a lot.

elliesec:
> @elliesec , can you help us here? Do you have a different solution that's coming up that could be in conflict with this pull request? Thanks a lot.

As Rui mentioned, we do have some plans for improved body colouring, and I had some ideas for body/face paint (body "decals" in general). I've not done much work on it yet though. I'm not particularly against this going in, but if you're happy to wait, I think that's probably a slightly cleaner solution than adding an item temporarily if we know it'll be removed eventually.

Ben987:
Thanks everyone, let's include it for now and we can always improve it later.

@Shiranui-Izayoi, the pull request has a few conflicts with other merged pull requests.  Can you resolve the conflicts and test to make sure everything works?   Thanks a lot.

Shiranui-Izayoi:
Ok I think I have got everything. Should be ready.

Ben987:
Thanks a lot!  I will test it soon :)

PR #2850, opened 2021-08-13
Fix wrong tileset when spectating kinky dungeon

Ada18980:


Ben987:

PR #2851, opened 2021-08-14
fix old gas mask missing color group

ace-1331:


Ben987:
Sure!  Thanks for that fix :)

PR #2853, opened 2021-08-17
fix 404s for custom collar tag lock

ace-1331:
this only removes unnecessary 404s, the locks still rendered properly regardless (because of the lock layer)

Its not necessary to create a hotfix just for this


Ben987:
Thanks a lot!  I might prepare a hotfix for all assets bugs.  Cheers :)

PR #2777, opened 2021-07-22
move the $ in all .csv files

fleisch11:
for consistency, every $ sign in every .csv file in the screens folder has been moved to the beginning of the price

(do not worry about game breaking mistakes, for everything has been done by hand and double checked)

PR #2857, opened 2021-08-18
fix typo in private room dialog 2$50 to $250

ace-1331:
caused by the unnecessary change of dollar sign position in #2777 
no other changed line appears to have a typo after reviewing it

regex&replace is safer than manual editing when doing bulk changes with a predictable pattern


Ben987:
Oops!  Thanks for that fix :)

PR #2858, opened 2021-08-19
Fix crash when spectating Kinky Dungeeon

Ada18980:


Ben987:
Thanks!  I will prepare a new hot-fix :)

PR #2860, opened 2021-08-22
Poker Text Fixes

Nina-1474:
When the hand ends and no players had a pair or higher, the result still said "One Pair". It now says "High Card"
Also made some spelling corrections in the dialogue.


Ben987:
Absolutely!  Thanks a lot :)

PR #2861, opened 2021-08-24
[Asset Addition] [Natsuki] Tape Mummy Hood Cut-out

Codefreak64:
With all the tape being put on the heads of members we never asked "is their face alright?".
Maybe you have but never knew, either way the fix for this (non-existant) question is answered, now we can see how the tape captives are fairing up.

-Adds a new "open mummy" to the selection of "ItemHeadDuctTape"
-Removes Adding "Effects" on the item twice


Ben987:
It looks nice, I'm curious to try it, thanks :)

PR #2865, opened 2021-08-27
Vibrating Clit Egg Update

Ayesha-678:
Made the Clit egg bigger, and gave it multi coloring


ace-1331 (changes requested):
this is unfinished and has missing assets:
the kneel pose wasnt split/copied for the multicoloring split (so it becomes invisible)

Ayesha-678:
Fixed

Ben987:

PR #2866, opened 2021-08-29
[Asset improvement] [Natsuki] Lower Cutout for tape

Codefreak64:
just a simple addition of giving access to the "fun zone" while wearing the Duct Tape on the upper legs.

-Adds new option to ItemLegsDuctTape (Cut Out)


Ben987:
I like it, that's a good idea.  Thanks :)

PR #2867, opened 2021-08-29
New Item: Super Thick ABDL Diapers

Shiranui-Izayoi:
2 new ABDL diapers made by @QueenTita these are very thick and were requested by a friend of hers on club. I have fixed as much of the clipping and stuff as I can find to the point I am pretty sure they are done.

The items come in both the `Panties` slot and also the `ItemPelvis` slot. The ones in the Pelvis slot are considered to be "Chastity Diapers" and are thus lockable. Though no locks are drawn just imagine they are inside.

During this I did alter the priorities of a few items mainly the various leotards and a couple of the strait jacket style items. They still work as normal. I only reduced their priorities to make them in line with the futuristic one as well as the original strait leotard.

One of them is extended as it can change size. These items do hide most of the `ClothLower` items as they clip badly along with the diapers just being to big. I have also made the needed adjustments to some items to properly hide the diapers.

If you have any questions feel free to ask me or @QueenTita. The credit for the assets which were apparently a pain to make goes to @QueenTita. While the credit for the coding which was also a pain with how much needed to be hidden and the custom left and top alignment goes to me.


ace-1331:
Quick note that this will conflict with #2866 for the complete legs extended option changes and both changes will need to be kept.

Also, playing priority ping pong is never good, this is affecting a LOT of existing combinations that **will** frustrate players and lead to a bunch of "error" reports and then "fixes" that lead to more unnecessary priority changes without every going back to the existing behavior. The recent priority changes with the leotard were a sore point for many and showed its not good to mess with priority like that, especially just for one item and for many old frequently used items like that... 

Have you considered just allowing options for the diaper to be on two separate layers? That way people can choose to wear it over/under, or simply lowering the layer of the diaper itself (nothing else should require priority changes for this)

Ruilov3:
My question is (unless i'm misunderstanding something), why is the extended extra padding hiding a bunch of Cloth items?

`Property: { Type: "Poofy", HideItem: ["ClothMaidOutfit1", "ClothMaidOutfit2", "ClothStudentOutfit1", "ClothBabydollDress1", "ClothChineseDress2", "ClothSummerDress", "ClothLowerPleatedSkirt", "ClothLowerSkirt2", "ClothLowerSkirt3", "ClothLowerTennisSkirt1"], },`

Wouldn't it make more sense for clipping purposes for the cloth items that are usually above to hide the diaper? Or even use alpha masks to make it work together?

Shiranui-Izayoi:
> 
> 
> Quick note that this will conflict with #2866 for the complete legs extended option changes and both changes will need to be kept.
> 
> Also, playing priority ping pong is never good, this is affecting a LOT of existing combinations that **will** frustrate players and lead to a bunch of "error" reports and then "fixes" that lead to more unnecessary priority changes without every going back to the existing behavior. The recent priority changes with the leotard were a sore point for many and showed its not good to mess with priority like that, especially just for one item and for many old frequently used items like that...
> 
> Have you considered just allowing options for the diaper to be on two separate layers? That way people can choose to wear it over/under, or simply lowering the layer of the diaper itself (nothing else should require priority changes for this).
> 
> Edit: alpha masks would also be a good solution instead of changing priorities (literally anything else than changing priorities is a better solution)

I had actually forgotten that Alpha masks existed my bad I will work on that in a bit. As for separate layers the main issue is that to put them over the leotards I would need to increase the layer to easily over 30 as most arm restraints have a Priority of 30. Which would then cause a lot of clipping with the few items it doesn't clip with.

> 
> 
> My question is (unless i'm misunderstanding something), why is the extended extra padding hiding a bunch of Cloth items?
> 
> `Property: { Type: "Poofy", HideItem: ["ClothMaidOutfit1", "ClothMaidOutfit2", "ClothStudentOutfit1", "ClothBabydollDress1", "ClothChineseDress2", "ClothSummerDress", "ClothLowerPleatedSkirt", "ClothLowerSkirt2", "ClothLowerSkirt3", "ClothLowerTennisSkirt1"], },`
> 
> Wouldn't it make more sense for clipping purposes for the cloth items that are usually above to hide the diaper? Or even use alpha masks to make it work together?

The reason the extended one is hiding the padding is that while the normal version doesn't clip the extra padding is wider and causes it to clip with those items. As it is wider than they are.

While I could have made the items hide the diapers I felt that it was better for the diaper to hide the items as that would have meant editting a lot of other items. And as the diapers are super thick it made sense in my head that these would prevent wearing of other clothing. I can potentially add alpha masks but that would result in editting nearly all of the `ClothLower` items with alpha masks as well as some of the Cloth items.

ace-1331:
Yes, if it avoids changing a bunch of priorities for existing items, its much better to make an extended option for under vs over. Mixed with alpha masks, this would solve all the issues.

For the hiding, I agree it's extremely strange that the dresses are hidden by the diaper, it should definitely be the opposite. Logically speaking, the dresses are over the diaper anyway, so it shouldnt show the diaper and someone who want to wear a diaper + a maid dress wouldnt be able to do so with the current implementation. A group mask (if possible) or just hiding the diaper will help be more futureproof because if anyone adds similar items to those dresses, they could forget to add the hide properties

Shiranui-Izayoi:
> 
> 
> Yes, if it avoids changing a bunch of priorities for existing items, its much better to make an extended option for under vs over. Mixed with alpha masks, this would solve all the issues.
> 
> For the hiding, I agree it's extremely strange that the dresses are hidden by the diaper, it should definitely be the opposite. Logically speaking, the dresses are over the diaper anyway, so it shouldnt show the diaper and someone who want to wear a diaper + a maid dress wouldnt be able to do so with the current implementation. A group mask (if possible) or just hiding the diaper will help be more futureproof because if anyone adds similar items to those dresses, they could forget to add the hide properties

Ok I will start work on adding some `AlphaMasks` and extended to certain items.

As for the hidden items with the exception of I think 2 items (One being the Long Skirt) every single one I have hidden actually has the diaper visible under it so logically at least would be visible. 

I will see what I can do, as I have used a group hide for `ClothLower` since nearly everything clips. I then used `HideItemExclude` to exclude the few items that it doesn't work with. But I didn't feel like there was enough items to hide the entire `Cloth` group then exclude the rest. As most of the non dresses or outfits with skirts don't clip.

Question is it possible to hide categories in the asset level of items? If so I could potentially add a category like "NarrowSkirt" or something to the items and Hide that. Or simply a "Dress" category and exclude the ones that fit. 

Then if someone added a dress and remembered to assign the category it would future proof it. Or get caught in the beta. Alternatively I can simply hide `Cloth` and exclude everything that fits, though it would cause the item entry to be longer. Let me know which would be better.

I am trying to think

ace-1331:
For tags, it's something that's been on our mind for a while, it would be nice to group up items with "Tags" to reduce duplication of large groups of items in stuff like hide arrays. (for example, hide "toys", etc.), but no, for now it's just copy paste sadly. (which is still the safest at the moment to prevent forgetting about certain items)

There's more than one ways that's been thought, I know right is not ideal for items that need to be hidden by lots of stuff.. but several additions to hideitem & such arent risky compared to layer changes.


For the question of hiding dresses, if alpha masks aren't a solution to your problem, I would still question the need to hide the dresses, because it prevents someone from wearing a diaper with those dresses which is a significant annoyance if one has to swap clothes more often because of it. I assume the perfect scenario (for example, if using masks) is having the bottom of the diaper visible at the bottom of the dress? The difference between the "perfect scenario" and just hiding the diaper would be fairly minor. 
This point is more of an implementation choice, than a problem. Im not into ABDL and wouldnt use these so I dont care about the final result, but I think its worth asking/thinking about because we've seen what happens with unrealistic/illogical hides... people keep reporting "errors" on and on even if its intended (talking about the new thigh high boots with hidden ankle restraints)

elliesec:
I've not been keeping up with the conversation, but @SandrineBC did add the `Attribute` and `HideItemAttribute` properties (currently there's only a `Skirt` attribute, which is attached to skirts). `HideItemAttribute: ["Skirt"]` will hide all assets marked with `Attribute: ["Skirt"]`.

That said, I agree with Ace that logically, it makes sense for overwear items to be hiding the diapers rather than the other way around. Realistically dresses/skirts etc. would go over them and hide them, and as Ace said, having the diapers hide overwear feel more likely to result in players reporting bugs, even if it's intentional.

Shiranui-Izayoi:
Ok so yer I will start work on getting Alpha masks in to hide the `Panties` layer on the hidden items. Some items will just stay hidden such as the Chinese Skirt simply because it is to open and it won't look good. That said it should significantly reduce the amount of items that need to be hidden. 

Hopefully the few items that are hidden won't be an issue. The biggest issue is when the diapers would instantly "bulge" ouwards when reaching the bottom of the dress. But hopefully due to them being diapers and them being able to "squish" it shouldn't look to bad.

That said a fair amount of the `ClothLower` category will still be hidden simply because they don't fit. For example the pants/trousers/jeans just won't work no matter what I do so will likely need to be hidden, That said with Alpha maks most "skirted" items can be saved. It will just take a fair amount of time.

Hopefully when I am done the tape pull request will have been merged letting me quickly fix the issues there. I will also work on some `AlphaMasks` and `Extended` for the items I changed the priorities for. 

Thanks for all the advice. If possible can this be put on-hold until I update it? @ace-1331

elliesec:
I've converted it to a draft for now. I'm pretty sure you should be able to switch it between draft and ready states without needing any special permissions.

ace-1331:
sure, and for reference, you can convert your PR to a draft (option is in the panel on the right of your PR) which prevents anyone from merging it till its ready. (which is especially useful given how Ben tends to be trigger happy with merges)

Once you are ready to PR it, you can mark it as "ready for review" from here

ace-1331:
sure, and for reference, you can convert your PR to a draft (option is in the panel on the right of your PR) which prevents anyone from merging it till its ready. (which is especially useful given how Ben tends to be trigger happy with merges)

Once you are ready to PR it, you can mark it as "ready for review" from here

Shiranui-Izayoi:
Thanks. Looks like I have a lot of work to do as I am going to be adding `AlphaMasks` to a bunch of stuff. But then again I kinda knew what I was in for when I agreed to code this.

Shiranui-Izayoi:
Ok think I have done everything. If you have any more questions please let me know.

ace-1331 (changes requested):
- The dialog messages for the chatroom text are missing when swapping types (ItemPelvisPoofyDiaperSetExtraPadding for example)

My other comment is more a question than a change request, was it intended that you block closing legs with those diapers even though they support that pose? If so, then it's alright, otherwise 
`SetPose: ["LegsOpen"], AllowActivePose: ["Kneel"], WhitelistActivePose: ["BaseLower", "Kneel"]` is not necessary

Shiranui-Izayoi:
> 
> 
>     * The dialog messages for the chatroom text are missing when swapping types (ItemPelvisPoofyDiaperSetExtraPadding for example)
> 
> 
> My other comment is more a question than a change request, was it intended that you block closing legs with those diapers even though they support that pose? If so, then it's alright, otherwise
> `SetPose: ["LegsOpen"], AllowActivePose: ["Kneel"], WhitelistActivePose: ["BaseLower", "Kneel"]` is not necessary

Ahhhh sorry I had missed the dialog, when I finished the masks and other things due to excitment my bad. Will add it now.

As for the blocking of the closed leg pose yes that is actually intended. The thought process is that the diapers are so thick you won't be able to close your legs by yourself. But since I couldn't find a way to prevent people from using items that could close the legs such as cuffs or many of the others. So I simply made it so you couldn't choose to close your legs and it would have to be bound.

Also since `Kneel` is the only pose that current can be used to sit down, and the assets for that match the assets for `LegsClosed` we decided to support it but "soft ban" it so to speak.

TLDR: Yes banning `LegsClosed` is intended even though it is supported. As the diapers are "too thick" to allow someone to close their legs on their own. I didn't want to add a bunch of prerequisites to items. So instead we reasoned that while hard someones legs could be forced closed with bondage.

ace-1331 (approved):

Ben987:
Sorry for the delay, there was lots of server stuff to test.  It's time to try this one, cheers!

PR #2868, opened 2021-08-30
Feature: Extended Item Alpha Masks

fleisch11:
Allows one to specify Alphamasks for a single option of an extended item


Ben987:
Sure, alphas on extended items would be useful.  I will try it soon.  Cheers :)

PR #2869, opened 2021-08-30
Type Alpha Masks

ace-1331:
This reverts #2868 entirely because it was simply not the right implementation on top of introducing a bunch of exploits

It keeps the added alpha mask values added to the duct tape in itemhead.


elliesec (approved):

Ben987:
Sure!  It can be done another way later.  Thanks!

PR #2871, opened 2021-09-01
Futuristic Item Additions

Ada18980:
-Added new modular options to the futuristic crate. Tested the animation + made sure it restrains and works with cuffs/no cuffs to avoid softlocking.
-Added a new full-face mask similar to the futuristic visor. Tested all four modes.
-Added new futuristic muzzle options (harness straps and symbol)
-Added Automatic to the training belt's messages to reduce unwanted spam

EDIT: Argh... some kinky dungeon balance changes are in here too. Bleh.


Ben987:
Very nice!  Thanks for all of these changes, I will check it soon :)

Shiranui-Izayoi:
Heya was looking through and found that a paint.net file slipped through. It is the following.

 BondageClub/Assets/Female3DCG/ItemHead/Preview/FuturisticMask.pdn

PR #2877, opened 2021-09-02
New items: 2 new panties

ace-1331:
all credit goes to lunamoon for the assets, i just did the code

thanks to ada for the multicolor split on the bow panties


Ada18980 (dismissed):
I think these should be split for multicoloring. We get a lot of requests from players to split items before multicolor and for an item like this it might be worthwhile to do it from the start rather than add the work later.

Ben987:
This looks good.  Thanks a lot for the items :)

PR #2878, opened 2021-09-02
Online Suitcase Minigame

Ada18980:
Adds a new way of earning money:

1) Talk to the kidnapper's league. You must be a member of the league, and ask "Do you have other jobs?"
2) You receive a suitcase and are told to go to chatrooms
3) Players that are in the kidnap league see a '$' sign over your head. While you are carrying the suitcase (it is in ItemMisc) they can start picking the lock on the suitcase after binding you. It takes 10 minutes and the holding player must be bound and there must be at least one player in the room that's still unlocking.
4) Other players can join in the process but it does not speed things up.
5) Once the process is over, everyone who participated in the steal gets up to 45$ depending on when they joined in. If you joined halfway through you get half the money, etc. You can repeat this once you find out that the suitcase contains a laptop that can be hacked for more data.
6) The player gets an Empty Suitcase they can take back to the kidnapper's league and get paid 65$ as consolation, plus 70$ for each time more data got stolen.
7) You can also return the suitcase for no pay.

The idea is to allow a player to spend 15 minutes tied up by another player, and it gives money to both players. However, the fact that there is a specific timer and that the timer is 15 minutes means that it is not competitive compared to, say, the maid drink serving game which can give a lot more in 15 minutes. On the other hand, this requires more effort than the club slave option for making money as you have to actively convince another player to kidnap you, hence the higher wage.

In the future, there might be more counterparts to this game: Finding and stealing the suitcase within a time limit for even more money, or being an agent and trying to swap suitcases in private for a mutual cash reward. However, those aren't in yet.


ace-1331 (changes requested):
your pushing what looks like a broken file which is residue from earlier when you added my repos as a remote? (`h origin HEADname-of-remote-branch`)

Ada18980:
1) Thanks, deleted

3) Fixed, was due to Ben merging something from a different PR

2) You don't get paid until someone steals the suitcase. So setting your permissions to Owner Only would make it impossible to get the suitcase stolen. And if you change your permissions afterward, well, if you shut the other player out then they can just leave and then the timer gets reset.

Do you think the player should be able to start the game regardless of their permissions if they are willing to get kidnapped by friends to fulfil the condition?

ace-1331 (approved):

BondageClub/Screens/Online/ChatRoom/ChatRoom.js
```diff
@@ -90,7 +90,7 @@ var ChatRoomCharacterX_Upper = 0;
 var ChatRoomCharacterX_Lower = 0;
 var ChatRoomCharacterZoom = 1;
 var ChatRoomSlideWeight = 9;
-var ChatRoomCharacterInitialize = true;
+var ChatRoomCharacterInitialize = true
```
Jomshir98, line 93:
```suggestion
var ChatRoomCharacterInitialize = true;
```

PR #2885, opened 2021-09-04
Improvement: view blocked categories from chatroom

ace-1331:
- added the ability to enter the BlockItem screen in read-only mode
- changed the admin screen to allow entering the blocked category screen even when not an admin to view what is blocked in the room (instead of having to leave/use console)


Nina-1474 (approved):

Ben987:
Absolutely!  This is a good idea.  Thanks :)

PR #2886, opened 2021-09-04
Improvement: Friendlist & beep friendly navigation

ace-1331:
- added a helper to group toggle show/hide elements for a given screen and flags on elements to know what screen generated them. (This can simplify navigation between screens. Also, one could be added for removing elements in the future, which would greatly simplify exit functions)
- modified the friendlist return functionalities to going in the friendlist from any room rather than just profile and chatsearch (it should now be easy to add a friendlist icon anywhere we want)
- added a /openfriendlist command
- added a beep click function. (At the moment, when you receive "mail", you can click to directly go to your friendlist, the function could be reused to allow different click functionalities on the beep rectangle)
- added a chat message on beep to prevent missing beeps (includes a shortcut to the friendlist on click). Comes with a setting to disable it


Ben987:
Sure!  This should be helpful.

PR #2892, opened 2021-09-04
Feature: `AllFours` pose options for Pet Suit & Pet Suit Exposed

elliesec:
New assets as always by @CeciliaBC.

This adds new `AllFours` pose options to the Pet Suit & Pet Suit Exposed (see preview image below). Unfortunately there were a few technical blockers the I needed to also address to get this working properly, which accounts for the size of the PR. Below is a list of all of the main changes in this PR:

* Reworked the `BodyUpper` assets for the `AllFours` pose so that all assets of a given body size are all the same size and shape (there were some substantial discrepancies between the sizes & shapes of the black/asian/white bodies in this pose, for whatever reason, and it became very obvious with the new assets for the Pet Suit Exposed not fitting properly for some skin tones)
* Converted the Pet Suit Exposed into a typed extended item
* Converted the Pet Suit into a modular item, with modules for zip state & pose
* Added support for dialog customisation in modular items, much like the existing dialog configuration in typed items
* Added support for chat tag customisation in modular items - behaves exactly like the existing configuration in typed items
* Added a new `AllowSelfSelect` configuration property to typed/modular item options, and modular item modules which determines whether or not the wearer can select that option/module. This enables the wearer of the Pet Suit to change poses, but not zip/unzip their suit
* Allows the locks on both pet suits to be coloured
* Removed a redundant `IsSelfBondage` argument from the `ExtendedItemRequirementCheckMessage` function which was not being used any more
* Tweaked some of the TypeScript definition to enable more strict type-checking of extended item configurations (which flagged up a redundant piece of configuration in the VR Headset's extended item definition, which I've also removed in this PR)

The new suit assets have the straps on a separate layer from the suit, which means that with a little effort to extract them from the current suit, it would be fairly straightforward to allow them to be coloured separately. I may raise a separate PR to do this in the future.

Below is a preview of the available new suit configurations:

petsuit-showcase


BondageClub/Scripts/ExtendedItem.js
```diff
@@ -379,19 +376,33 @@ function ExtendedItemHandleOptionClick(C, Options, Option, IsSelfBondage) {
...
  * @returns {string|null} null if the player meets the option requirements. Otherwise a string message informing them
  * of the requirements they do not meet
  */
-function ExtendedItemRequirementCheckMessage(Option, CurrentOption, IsSelfBondage) {
+function ExtendedItemRequirementCheckMessage(Option, CurrentOption) {
 	const C = CharacterGetCurrent() || CharacterAppearanceSelection;
 	let ValidationMessage = TypedItemValidateOption(C, DialogFocusItem, Option, CurrentOption);
 	if (!ValidationMessage) {
+		ValidationMessage = ExtendedItemCheckSelfSelect(C, Option);
+	} else if (!ValidationMessage) {
```
Nina-1474, line 387: This is the same check as above so won't it never be reached? Should it be a new `if` instead of an `else if`?

elliesec: Oops, yes. Yes it should.

elliesec: I've changed it into a short-circuiting `||`, as the `if` logic was getting a bit messy.

ace-1331 (approved):

Ben987:
This is really nice, thanks a lot.  It created some conflicts with other pull requests though, can you resolve and test again?  Thanks.

elliesec:
Conflicts have been resolved, and I've re-tested everything.

Ben987:
Awesome!  Thanks a lot you two :)

PR #2893, opened 2021-09-04
Improvement: secret beeps

ace-1331:
- added the ability to exclude the room name from the beep

Since messages were added to beeps, it's made it awkward/unclear when messaging people through beeps. Some may want to respond without showing a private room name, or simply make it clear that its not an invite to a specific room.

This adds the option to exclude the room name from a beep when inside a chatroom.

This PR requires the matching (small) server PR (https://github.com/Ben987/Bondage-Club-Server/pull/103) to function correctly. It was tested in/out of rooms, in public and private rooms.


Ben987:
I'm curious to test this one.  Thanks a lot :)

PR #2901, opened 2021-09-05
Emergancy Fix for the face paint 

Shiranui-Izayoi:
Not sure how I didn't catch it or how it wasn't caught anywhere else, but I had missed that the face paints added weren't in the right position for the assets and thus were floating around the face area. I have fixed this and it should be merged before the beta.

Ok sorry because of said forgetting to update before fixing it considered everything new. So had to pull it down and stick it back up.


ace-1331 (dismissed):
for reference, you can just push on the same branch, no need to re-do a PR when you push more commits on the same branch. (your previous PR was fine after you merged)

Shiranui-Izayoi:
> 
> 
> for reference, you can just push on the same branch, no need to re-do a PR when you push more commits on the same branch. (your previous PR was fine after you merged)

Yes I am aware that I can repush on the same branch I have done so a lot. Just when I pushed on that branch when I reloaded it it was showing that things that had already been uploaded were being reuploaded as well as other such things. Mayb have just been a bug on my end but I didn't think so at the time so I closed and reopenned as it looks like it glitched out for me.

My bad shouldn't happen again.

ace-1331 (changes requested):
you're pushing 7 extra unused assets that were removed in a previous PR (the ClothAccessory version doesnt need assets in Poses like AllFours, etc.)

Shiranui-Izayoi:
> 
> 
> you're pushing 7 extra unused assets that were removed in a previous PR (the ClothAccessory version doesnt need assets in Poses like AllFours, etc.)

Fixed

ace-1331 (approved):

Ayesha-678 (approved):

PR #2902, opened 2021-09-06
[Asset] Mobile Phone Held Item

ItsJustMeVerity:
Adds a Mobile Phone held item!
Completely from scratch, so no need to worry about infringing on existing designs.

GitPreview


BondageClub/Screens/Inventory/ItemHands/SpankingToys/SpankingToy.js
```diff
@@ -169,6 +169,11 @@ const InventoryItemHandsSpankingToysOptions = [
 		Name: "Broom",
 		Property: { Type: "Broom" },
 		ExpressionTrigger: [{ Group: "Blush", Name: "Low", Timer: 5 }, { Group: "Eyebrows", Name: "Harsh", Timer: 5 }]
+	},{
+		Name: "Phone1",
+		Property: { Type: "Phone1" },
+		ExpressionTrigger: [{ Group: "Blush", Name: "Low", Timer: 10 }, { Group: "Eyebrows", Name: "Soft", Timer: 10 }, { Group: "Mouth", Name: "Frown", Timer: 10 }]
```
ace-1331, line 175: are expression triggers needed? might be a bit strange to blush when picking up a phone

ace-1331 (approved):

BondageClub/Screens/Inventory/ItemHands/SpankingToys/SpankingToy.js
ItsJustMeVerity: Bit strange to blush when picking up a towel or a sock, too. Didn't know quite what was safe to remove so with the rest still having them I just left them in.

ace-1331: yeah i agree, would be good to remove those from the items that dont make sense at some point.
You can accept this suggested edit if you want it, otherwise you can dismiss it
```suggestion
		ExpressionTrigger: []
```

BondageClub/Screens/Inventory/ItemHands/SpankingToys/SpankingToy.js
```diff
@@ -169,6 +169,11 @@ const InventoryItemHandsSpankingToysOptions = [
...
+	},{
```
ace-1331, line 176: there is an extra empty object here that has been added and needs to be removed
```suggestion
```

ace-1331 (changes requested):

ace-1331 (approved):
also for future reference, when there are suggested edits, you can commit them in batch straight from github, saves you the need to do the work manually.

Edit: nvm, saw you did for the second 😅 all looks good

ItsJustMeVerity:
> also for future reference, when there are suggested edits, you can commit them in batch straight from github, saves you the need to do the work manually

Yep, thank you! Did that for the expression one but I wanted to test how committing my own revisions to an open PR worked on the extra brackets one. And now I know how that works! :D

ace-1331:
opening a PR is sort of like setting a flag on a branch and saying "hey I want to merge this branch" from that point, it becomes like a timeline. The PR's state is the same as your branch so whatever you push to it will be seen here, and comments/approvals etc will be in order, so you can tell when something is committed before/after something, and even view the commits separately.

That's also why it's better to not re-open new PRs when fixing stuff on a PR, it helps to maintain the history in one place to better track stuff/requested changes/etc

Ayesha-678 (approved):

PR #2903, opened 2021-09-06
New item: Bandana

ace-1331:
- added a new bandana with various pattern styles

All credits to Titania @QueenTita for the assets, I just did the code


Ben987:
Absolutely, thanks a lot you two :)

PR #2904, opened 2021-09-06
Improvement: saved expression previews

ace-1331:
- changed the saved expression screen to display the saved expression in previews instead of just 1/2/3/4/5 to make it more user friendly
- improved the type definition and JSDoc of various functions/types related to the SavedExpressions and its menu

The previews are made to not include stuff that hides the face to prevent it from being useless when wearing masks. Here's the result:

Capture


BondageClub/Scripts/Dialog.js
```diff
@@ -1002,28 +1011,66 @@ function DialogFacialExpressionsLoad(Slot) {
...
 	}
 }
-/**draws the savedexpressions menu */
+
+/**
+ * Builds the savedexpressions menu previews.
+ * @returns {void} - Nothing
+ */
+function DialogBuildSavedExpressionsMenu() {
+	const ExcludedGroups = ["Hat", "Mask", "ClothAccessory", "Glasses"];
```
Nina-1474, line 1020: I'd favour showing Hat, ClothAccessory and Glasses since most of them don't block much of the expression, and maybe exclude ItemHead and ItemHood because most of those do.

BondageClub/Scripts/Dialog.js
```diff
@@ -1002,28 +1011,66 @@ function DialogFacialExpressionsLoad(Slot) {
...
+	const BaseAppearance = Player.Appearance.filter(A => A.Asset.Group.Category === "Appearance" && !ExcludedGroups.includes(A.Asset.Group.Name));
+	Player.SavedExpressions.forEach((expression, i) => {
+		if (expression) {
+			const PreviewCharacter = CharacterLoadSimple("SavedExpressionPreview-" + i);
+			PreviewCharacter.Appearance = BaseAppearance.slice();
+			BaseAppearance.forEach(A => InventoryWear(PreviewCharacter, A.Asset.Name, A.Asset.Group.Name, A.Color));
```
Nina-1474, line 1026: Rather than starting with BaseAppearance filtered by category then readding items, why not start with everything except the excluded groups and drop this line? 
Or you could even keep the category filter and just not add items back at all.

BondageClub/Scripts/Dialog.js
```diff
@@ -1002,28 +1011,66 @@ function DialogFacialExpressionsLoad(Slot) {
...
+ * Draws the savedexpressions menu
+ * @returns {void} - Nothing
+ */
 function DialogDrawSavedExpressionsMenu() {
-	DrawText(DialogFindPlayer("SavedExpressions"), 195, 25, "White", "Black");
-	DrawText(DialogFindPlayer("SavedExpressionsSave"), 140, 180, "White", "Black");
-	DrawText(DialogFindPlayer("SavedExpressionsLoad"), 260, 180, "White", "Black");
+	DrawText(DialogFindPlayer("SavedExpressions"), 210, 90, "White", "Black");
+
+	if (!DialogSavedExpressionPreviews || !DialogSavedExpressionPreviews.length)
```
Nina-1474, line 1044: If the player has no saved expressions, this is going to try building the menu on every draw. Should this also check `Player.SavedExpressions.length` or is the difference negligible?

BondageClub/Scripts/Dialog.js
```diff
@@ -1002,28 +1011,66 @@ function DialogFacialExpressionsLoad(Slot) {
...
+		DialogBuildSavedExpressionsMenu();
+
 	for (let x = 0; x < 5; x++) {
-		DrawButton(100, 200 + (x * 100), 80, 80, x + 1, "White");
-		DrawButton(220, 200 + (x * 100), 80, 80, x + 1, "White");
+		if (Player.SavedExpressions[x] == null) {
+			DrawText(DialogFindPlayer("SavedExpressionsEmpty"), 160, 216 + (x * 170), "White", "Black");
+		} else {
+			const PreviewCanvas = DrawCharacterSegment(DialogSavedExpressionPreviews[x], 100, 0, 300, 220);
+			MainCanvas.drawImage(PreviewCanvas, 0, 92 + (x * 170), 300, 190);
```
Nina-1474, line 1052:
```suggestion
			MainCanvas.drawImage(PreviewCanvas, 20, 92 + (x * 170), 260, 190);
```
The expression previews are noticeably stretched horizontally. I think they look much better in the original ratio.

Nina-1474 (dismissed):

BondageClub/Scripts/Dialog.js
ace-1331: it is negligible, but to be safe & futureproof , I adjusted it accordingly

BondageClub/Scripts/Dialog.js
ace-1331: actually, there is a use for it because there is a lot of awkwardness with references when copying the appearance of a character. (and doing a deep clone is not possible due to circular references) 

Actually, I just thought of a better way so Ill include that in after re-testing to be 100% sure

BondageClub/Scripts/Dialog.js
```diff
@@ -1002,28 +1011,65 @@ function DialogFacialExpressionsLoad(Slot) {
...
+ * Draws the savedexpressions menu
+ * @returns {void} - Nothing
+ */
 function DialogDrawSavedExpressionsMenu() {
-	DrawText(DialogFindPlayer("SavedExpressions"), 195, 25, "White", "Black");
-	DrawText(DialogFindPlayer("SavedExpressionsSave"), 140, 180, "White", "Black");
-	DrawText(DialogFindPlayer("SavedExpressionsLoad"), 260, 180, "White", "Black");
+	DrawText(DialogFindPlayer("SavedExpressions"), 210, 90, "White", "Black");
+
+	if (!DialogSavedExpressionPreviews || !DialogSavedExpressionPreviews.length || Player.SavedExpressions.some(expression => expression != null))
```
Nina-1474, line 1043: The menu build is now occurring every draw, in _both_ the saved expressions being empty and non-empty cases. I was thinking something more along the lines of:
```suggestion
	if (!DialogSavedExpressionPreviews || (!DialogSavedExpressionPreviews.length && Player.SavedExpressions.length))
```

BondageClub/Scripts/Dialog.js
```diff
@@ -1002,28 +1011,65 @@ function DialogFacialExpressionsLoad(Slot) {
...
 }
-/**draws the savedexpressions menu */
+
+/**
+ * Builds the savedexpressions menu previews.
+ * @returns {void} - Nothing
+ */
+function DialogBuildSavedExpressionsMenu() {
+	const ExcludedGroups = ["Mask"];
+	const BaseAppearance = Player.Appearance.filter(A => A.Asset.Group.Category === "Appearance" && !ExcludedGroups.includes(A.Asset.Group.Name) &&  !A.Asset.Group.AllowExpression);
```
Nina-1474, line 1021: The previews aren't rendering properly for me anymore. I think it's because `CharacterSetFacialExpression()` needs the AllowExpression groups to be in Appearance and now they aren't.

Nina-1474 (changes requested):

BondageClub/Scripts/Dialog.js
```diff
@@ -1002,28 +1011,70 @@ function DialogFacialExpressionsLoad(Slot) {
...
+
+/** Handles clicks in the savedexpressions menu
+ * @returns {void} - Nothing
+ */
 function DialogClickSavedExpressionsMenu() {
-	if (MouseXIn(100, 80)) {
+	if (MouseXIn(290, 120)) {
 		for (let x = 0; x < 5; x++) {
-			if (MouseYIn(200 + (x * 100), 80)) {
+			if (MouseYIn(160 + (x * 180), 50)) {
```
Nina-1474, line 1070: The click and draw co-ordinates are misaligned. One has * 170 and the other * 180.

Nina-1474 (dismissed):

Nina-1474 (approved):

Ben987:
That's a nice concept, I'm curious to test it.  Thanks!

PR #2873, opened 2021-09-01
Add script hooks for typed/modular items, refactor futuristic items

Ada18980:
Refactors the futuristic functions, putting them in `Screens/Inventory/Futuristic/Futuristic.js`, and converts the following items to Typed/Modular:

Futuristic Crate is now Modular
VR Headset is now Modular
Interactive Visor is now Typed
Futuristic Mask is now Typed
Futuristic Heels are now Typed
Futuristic Shoes are now Typed
Futuristic Leg Cuffs are now Typed
Futuristic Cuffs are now Typed
Futuristic Ankle Cuffs are now Typed
Futuristic Armbinder is now Typed
Futuristic Straitjacket is now Modular
Futuristic Earphones are now Typed
Futuristic Mittens are now Typed
Futuristic Bra is now Modular

PR #2909, opened 2021-09-06
fix: polluted modular object properties

ace-1331:
- fixed a bug introduced in #2873 where `CustomBlindBackground` was added to every modular item's Property unnecessarily.


Ben987:
Alright!  Thanks for that fix.

PR #2910, opened 2021-09-06
fix: activities not hiding names in sensdep

ace-1331:
- fixed a bug where activities wouldnt replace the source name by "someone" due to the condition being wrong (one always yields false when it shouldnt, this uses the same logic as the rest) (caused by #2766 )


Ada18980 (approved):

Ben987:
Absolutely, thanks for that fix.

PR #2911, opened 2021-09-06
fix: color picker crash & errors

ace-1331:
- fixed a rare bug where the fired change handler would run after the color picker was exited (causing an error, but apparently no crash)
- fixed a more frequent crash related to character previews, caused by the removal of an item via an appearance sync when the color was being changed.


Ben987:
Yes, I've heard of that crash.  Thanks!

PR #2913, opened 2021-09-07
Improvement: Modular Item ModuleType

ace-1331:
- adds an option for modular items to support partial types to reduce the amount of necessary layers and simplify modular item layer code

Ive converted the custom tag to show a before/after. For those who have made modular item, they know the pain of needing several layers when only one would be required with this change.

Im leaving it as a draft to get input from others before merging. 
Let me know what you think and if there are areas of improvement


Shiranui-Izayoi:
So if I am understanding this right instead of having it be `Tag_Tag1` it is now `Tagt1_Tag`. With the layer part that being the `_Tag` part now being a sorta group of layers of it's own. With the `t1` part being a sort of group identifier? So you are saying "I want t1 of group tag" to the game. Is that right?

Just seeing if I understand it correct. Because if so I think that it looks quite nice. And could be really useful. 

My only real feedback would be to put somewhere what it does, how it works and where it is useful. I mean without digging through closed pull requests to find it. 

Since recently when I was adding `Alpha` to a bunch of stuff I had to trial and error it for a while. Before I figured out what each input in the masks corresponded to. Since I could see people making an item and not knowing if they should be using this or the normal. Which would help reduce times where things are included because the person isn't sure if it applies to this situation.

ace-1331:
> My only real feedback would be to put somewhere what it does, how it works and where it is useful. I mean without digging through closed pull requests to find it.

Yeah making a detail asset creation guide is long overdue by now, especially with the more complex stuff added recently. Besides reading the code/reverse engineering, you can always ask us for specific stuff or things you need, but I agree just a detailed explanation of everything available in general would help get people started.

As for this specifically, the goal is to replicate the behavior of normal typed items but for modules, which currently require extra layers for no reason. For example, a normal typed item with the type "open" and "closed":
- you can have a layer with `HasType: false` so it pulls the same image for that layer regardless of the type
- or, by default itll search for a different image for each type. You can have `AllowTypes` to define which types that layer is active for. (This is commonly used for a layer which is only present on some types, like the padding for the kennel)

The problem with modular item is that you cannot have this type of behavior for 1 module. You have to set `HasType: false` and allow one module type for each layer, which means a layer is only rendered for one option of a module. (otherwise it blows up exponentially and you need to have the same images for each possible type combination). 

This is like a inbetween HasType: true and HasType: false (which is only necessary on modular items) so you can have one layer that is labelled as having a type for one or many module. Let's say you want your layer `door` to have types for the door module with a key labelled `d` , you would just need to set `ModuleType: ["d"]` and then expect the typical behavior of needing an image for each type of that module (let's say you have 3 types, it would then mean you can have a d0_Door, d1_Door, d2_Door image without needing 3 layers)

Most modular items have a ton of layers because they use layers to bypass the current "limitations"

Shiranui-Izayoi:
> Yeah making a detail asset creation guide is long overdue by now, especially with the more complex stuff added recently. Besides reading the code/reverse engineering, you can always ask us for specific stuff or things you need, but I agree just a detailed explanation of everything available in general would help get people started.
> 
> As for this specifically, the goal is to replicate the behavior of normal typed items but for modules, which currently require extra layers for no reason. For example, a normal typed item with the type "open" and "closed":
> 
>     * you can have a layer with `HasType: false` so it pulls the same image for that layer regardless of the type
> 
>     * or, by default itll search for a different image for each type. You can have `AllowTypes` to define which types that layer is active for. (This is commonly used for a layer which is only present on some types, like the padding for the kennel)
> 
> 
> The problem with modular item is that you cannot have this type of behavior for 1 module. You have to set `HasType: false` and allow one module type for each layer, which means a layer is only rendered for one option of a module. (otherwise it blows up exponentially and you need to have the same images for each possible type combination).
> 
> This is like a inbetween HasType: true and HasType: false (which is only necessary on modular items) so you can have one layer that is labelled as having a type for one or many module. Let's say you want your layer `door` to have types for the door module with a key labelled `d` , you would just need to set `ModuleType: ["d"]` and then expect the typical behavior of needing an image for each type of that module (let's say you have 3 types, it would then mean you can have a d0_Door, d1_Door, d2_Door image without needing 3 layers)
> 
> Most modular items have a ton of layers because they use layers to bypass the current "limitations"

Makes sense. Just wanted to see if I understood it. Seems like a good change. And while obviously this shouldn't include the whole guide maybe just an explanation of this feature alone? As a start I think that would work.

Ada18980 (approved):
Great change!

Ben987:
Sure!  It might simplify item creation.  Thanks :)

PR #2914, opened 2021-09-07
support banlist for RoomRecreate

ace-1331:
- improved room recreation to include the banlist (particularly useful if the room is recreated by a non-admin.)

Note that I have submitted a PR for the server (https://github.com/Ben987/Bondage-Club-Server/pull/106), it's not required for this, but if it can be merged, it would allow refactoring autoban and room recreation in a way that would require to create the room once instead of creating it + updating it twice


Ben987:
Yes!  Time to test that one to go with the new server :)

PR #2915, opened 2021-09-07
Extended DoubleEndDildo

Shiranui-Izayoi:
- Extended the Double Ended Dildo to have a normal size which is the default and a large size.
- Greyscaled and Lightened up the item to allow bright colours to be used. (Note every colour previously possible is still there just have to go a bit darker.)
- Set a default colour for it to keep it's original look.
- Added Dialog to fit the change.

So yer really quick request I basically just grabbed the item into an edit program and enlarged it by a bit. It is the same by default but for those that want it, it can now be made bigger. I am sure the size queens will love it. Also as said greyscaled and brightened it up so that it can take extra colours.

Really short PR that I put together in like 30 minutes. Hope to see it in.


BondageClub/Assets/Female3DCG/Female3DCG.js
```diff
@@ -1906,7 +1906,14 @@ var AssetFemale3DCG = [
 			},
 			{ Name: "LoversVibrator", Value: 75, Time: 5, LoverOnly: true, DefaultColor: "#790c0c", Prerequisite: ["AccessVulva"], Effect: ["Egged", "FillVulva"], AllowEffect: ["Egged", "Vibrating", "Edged", "FillVulva"], ExpressionTrigger: [{ Name: "Low", Group: "Blush", Timer: 10 }], AllowRemoveExclusive: true, CharacterRestricted: true, DynamicScriptDraw: true},
 			{ Name: "LoversVibratorRemote", Value: 75, Wear: false, LoverOnly: true, Prerequisite: ["RemotesAllowed"], Effect: ["Remote"] },
-			{ Name: "DoubleEndDildo", Value: 15, Time: 10, Effect: ["FillVulva"], Prerequisite: ["AccessVulva"], ExpressionTrigger: [{ Name: "Medium", Group: "Blush", Timer: 20 }], AllowActivity: ["Penetrate"] },
+			{ 
+				Name: "DoubleEndDildo", Value: 15, Time: 10, DefaultColor: ["#624662"], Effect: ["FillVulva"], Prerequisite: ["AccessVulva"], ExpressionTrigger: [{ Name: "Medium", Group: "Blush", Timer: 20 }], AllowActivity: ["Penetrate"], Extended: true,
+				Layer: [
+					{ Name: "Normal", HasType: false, AllowTypes: [""] },
```
ace-1331, line 1912: For your info, for this, you dont need to define any layer.
Without layer it will be based on the type, the code will either look for DoubleEndDildo.png or DoubleEndDildoLarge.png

(so removing the layer definition and renaming the two files would be enough and makes it a bit cleaner)

Shiranui-Izayoi: Sure will do that in 2 secs.

Shiranui-Izayoi:
Changes have been made.

Ben987:
Absolutely!  Right on time for the next beta :)

PR #2916, opened 2021-09-07
Autoban refactor

ace-1331:
- refactored autoban to be included on room creation to remove the unnecessary update.
- also includes an extra label in case https://github.com/Ben987/Bondage-Club-Server/pull/107 's new response message needs to be displayed (should never happen in a natural way)

No compatibility issue, relies on the merged https://github.com/Ben987/Bondage-Club-Server/pull/106 to function.


Ben987:
Absolutely!  Thanks a lot :)

PR #2918, opened 2021-09-07
Shock visual effect

Ada18980:
Adds a new screen flash visual effect when shocked in a chatroom. This uses the existing screen flash setting for stimulation effects, which is default off.

Also fixes a bug where the stimulation flash setting was getting initialized as StimulationFlashes, which is wrong. It should be StimulationFlash.


Ben987:
Alright!  I'm curious to see this one in action.

PR #2919, opened 2021-09-07
Screen glow when vibrator is on

Ada18980:
This adds a screen glow (with option configurable in graphics setting) that happens when you have a vibrating item on. This is meant to make it more clear that your character is being vibrated in a chat room.

Also works and tested in private room.


Ben987:
I'm curious to see that one also.  Thanks!

PR #2920, opened 2021-09-08
Feature: Improved console warnings

elliesec:
> ⚠️ Accompanying server PR: Ben987/Bondage-Club-Server#109 - not required for this change to work, but should be merged after this one to remove the old server code.

With input from several other contributors, I've put together a [Player Safety wiki page](https://github.com/Ben987/Bondage-College/wiki/Player-Safety), with some safety and security advice for the club. This PR modifies the initial "Scripts can break your account" message to make it a bit more eye-catching, and to link to the new wiki page. As part of this, I've moved the messages from the server to the `ServerConnect` function on the client, to allow for easier customisation of the messages, which should have the same net effect.

Below is a preview of the new message. The formatting has been tested and confirmed working in Chrome, Firefox and Edge. I don't have access to a device running Safari, so the new message formatting is only displayed in those browsers. Other browsers will receive the message without the new formatting.

image


Ada18980 (approved):

Ben987:
Absolutely!  Let's wait for the server changes anyway, I want to make sure there are no more crashes.

PR #2921, opened 2021-09-08
Feature: Updates Pet Suit to utilise the new `ModuleType` property

elliesec:
This is a small follow-up to #2892 which makes use of the new `ModuleType` property introduced in #2913 to simplify the layer definitions for the Pet Suit. It doesn't change any of the original behaviour (and all asset changes are just renames).


Ben987:
Very good!  Thanks :)

PR #2924, opened 2021-09-09
New Fur items + Alpha mask fix

ace-1331:
- added a fur straitjacket and a fur blanket wrap to go along with the other fur items
- removed the alpha mask on the body bag introduced by the new diapers PR which causes stuff like back hair to be hidden when it previously wasnt (instead Ive made it so the diapers' groups are hidden)

All credits to Stacey for the items, I just did the code


Ben987:
Very nice!  Thanks you two :)

PR #2925, opened 2021-09-09
New Futuristic Crate Options/Features

Ada18980:
Splits arm restraints and leg restraints, and adds new styles (minimalist cuffs, a harness, and also structural background elements).

Adds the harness itself as a separate item for those who want to wear it aside from the crate.

Also adds color groups.


Ben987:
Right on time for the next beta!  Thanks a lot :)

PR #2927, opened 2021-09-10
Made lock picking only take one try when false set pins reset

Ada18980:
This is based on suggestions from discord. Should make lock picking less frustrating and makes it more meaningful to have high Lockpicking stat.


ace-1331 (approved):

Ben987:
Sure!  The lockpicking might become easier for some, thanks :)

PR #2931, opened 2021-09-11
Only include commits from upstream/master in `changelogPrepare` script

elliesec:
This is just a minor tweak to the `changelogPrepare` script so that it will only include commits on `upstream/master` when generating the change list & last recorded commit hash - useful when it's being run from a separate branch.


ace-1331 (approved):

Ben987:
Alright!

PR #2932, opened 2021-09-11
Misc Fixes

ace-1331:
- fixed a bug with `HideForPose` breaking with items that inherit their `AllowPose` from their group
- removed duplicate assets for AdmiralSkirt
- removed unnecessary images from StudentOutfit1 by using HideForPose
- removed an unused file in the asset folder
- fixed a mistake in `TextLoad` (use of `=` instead of `==` that can cause bugs in the future, but doesnt occur right now as `TextLoad` is always called without arguments.)
- style cleanup across a few files


elliesec (approved):

Ben987:
Absolutely!  Time to test these for the next beta :)

PR #2933, opened 2021-09-11
Feature: Pose mapping

ace-1331:
- added a new `PoseMapping` property for assets & layers, so that poses can reuse another pose's assets to reduce image duplication.
- adapted the tutu to use the new `PoseMapping` property as an example & fixed a missing layer

We have the intention of reducing duplicate assets as much as we logically can, and this was the last tool we didnt have to avoid duplicates. (and this type of problem is the source of a lot of dupes!)

By the end of the optimisations, it should greatly reduce total club file size + reduce bandwidth usage + less images, means smaller WebGL texture cache which is better for client performances


elliesec (approved):

Ben987:
Anything that can reduce the number of files is welcome.  Thanks!

PR #2935, opened 2021-09-12
fix color picker crash with no item in wardrobe

ace-1331:
- fixed a crash occurring when a sync removes an item while in the wardrobe


elliesec (approved):

Ben987:
Absolutely!  Thanks again :)

PR #2936, opened 2021-09-13
Feature: Updates cyclic block validation to take prerequisites into account

elliesec:
This PR improves upon #2618 to take prerequisites into account when validating for block cycles. This addition to the system should catch and fix any item combinations that would be impossible to remove when they're added by another player (and fix any such combinations on login).

Example item combination/reproduction that this fixes:
1. Equip a wooden horse on a player
2. Use the console to equip a Sybian on the same player (this would usually be disallowed due to the `NotHorse` prerequisite):
```js
InventoryWear(C, "Sybian", "ItemDevices");
ChatRoomCharacterUpdate(C); // or ChatRoomCharacterItemUpdate(C, "ItemDevices");
```
3. Note that the player now has an impossible item combination equipped, and neither item can be removed manually


ace-1331 (approved):

Ada18980 (approved):

Ben987:
Absolutely!  The less combo traps there are, the better we are :)

PR #2939, opened 2021-09-13
fix `CharacterGetCurrent` while in wardrobe

ace-1331:
- fixed `CharacterGetCurrent` returning null while in the wardrobe which can crash certain extended items like the bib


elliesec (approved):

Ben987:
Yes!  That one was annoying.  Thanks :)

PR #2942, opened 2021-09-13
missing double ended dildo large image

ace-1331:
- removed the unecessary allowpose and duplicate image from the doubleended dildo which fixes a missing image while kneeling at the same time


Shiranui-Izayoi:
Ahhhh damn I forgot to check that. Thanks Ace.

Ben987:
Absolutely!  Thanks :)

PR #2944, opened 2021-09-13
fix grandmage hogtied pose with small body

ace-1331:
- fixed grandmage hogtied pose with small body


elliesec (approved):

Ben987:
Sure!  Thanks a lot :)

PR #2949, opened 2021-09-13
Pet Suit fixes & tweaks

elliesec:
This makes a few fixes and tweaks to the new Pet Suit and Pet Suit Exposed options:
* Fixes an alignment issue where the suit was a few pixels too high, causing some minor clipping issues around the knees and face
* Changes both suits to apply the `Slow` effect
* Shifts priority down to 21. Because the suit hides most groups, this should be a relatively safe change, and it allows nipple items to render above the suit
* Hides `ItemNipples` when the Pet Suit is zipped up
* Hides the `Hands` and `ItemHands` groups on both suits


Ben987:

PR #2951, opened 2021-09-13
fix wardrobe sending a server sync on every update

ace-1331:
- fix wardrobe sending a server sync on every update

This was highlighted by an issue with recent server changes which made the appearance sync to everyone while in the wardrobe.
This PR fixes the problem by removing unnecessary syncs with the server, which at the same time:  
1) prevents appearance from being broken on login if you DC/crash while in the wardrobe (since the server is never "aware" of ongoing changes)
2) Reduces server load by only triggering a single sync when leaving (and applying changes) instead of 1 per strip/applied item (which is overkill and not needed)


Ben987:
Absolutely!  Anything that eases the server burden is welcome :)

PR #2954, opened 2021-09-14
Fix ChatRoomCarryingBounty

Ada18980:
Would cause a crash otherwise.


Ben987:
Thanks for fixing that crash!  Cheers :)

PR #2955, opened 2021-09-14
Fix picture function in chat room when only 1 player is present

Ada18980:


Ben987:
Time for beta fixes!  Thanks a lot :)

PR #2956, opened 2021-09-14
fix click on room name while in chatsearch in friends mode

ace-1331:
- fix click on room name while in chatsearch in "friends" mode


Ben987:
Thanks for that fix!

PR #2958, opened 2021-09-14
Feature: added lower quality animation settings

ace-1331:
- added a lower & lowest animation quality for slow devices

The current options didn't seem to be enough for people on slow devices who didnt want to disable animations or hide items altogether, I've adapted the fuck machine/crate animation to still render without breaking (uses a random value every x seconds instead), and added a maxWait to the function call limiter. 
Effects are:
- The fuck machine/crate on lowest settings are on par with the performance of the shock collar's blinking light which is not demanding (1 refresh every few seconds)
- Other limited functions such as the sliders for suspension/opacity on various items and the color picker are now 5x slower to refresh on lower/lowest setting to help reduce the amount of refreshes and overall lag while interacting with those.


ace-1331:
adjusted to include the same anim for the crate as #2965

Ada18980 (approved):

Ben987:
That one creates a conflict with other pull requests, can you resolve and test again?  Thanks!

Ben987:
Yes!  Some players are struggling with the new animations, it will probably help them.  Thanks :)

PR #2962, opened 2021-09-14
Improvement: tweaked the image retry strategy to not be instantaneous

ace-1331:
- changed the retry strategy from a burst of 3 retries to 6 increasingly spaced out retries which is a commonly used strategy

Should reduce query spam when there's is a momentary hiccup in serving assets, and increases the chances of successfully loading assets by properly waiting a few seconds (up to a few minutes).

More than that felt unnecessary because of how often retries are about missing assets, and spacing the retries more than that felt useless cause people will likely end up reloading after a few minutes anyway. 

I've extracted the retry logic into a variable, so adding/removing/changing the values in `GLDrawRetryTimers` will adjust the retry logic accordingly without having to change the `GLDrawLoadImage` code


Ada18980 (approved):

Ben987:
Absolutely!  It should help when we release new versions :)

PR #2965, opened 2021-09-14
Fix to futuristic crate intensity

Ada18980:
Setting the intensity to Low now does not make it disabled.

Also, I swapped out the animation for the pleasure module from a linear one to a sinusoidal one. I prefer it this way cause it looks more gentle.


Ben987:
Thanks for these extra fixes :)

PR #2970, opened 2021-09-14
[Change] Shopkeeper Hand Item Dialogue

ItsJustMeVerity:
Changes the (English) dialogue for the shopkeeper from "Hand restraints" to "Items for hands" (consistent with "Items for breasts" directly below), since the majority of the items in that category aren't even restraints.

image


Ben987:
Sure!  Thanks for that one :)

PR #2972, opened 2021-09-15
Extended Suits with Gloved Options

Shiranui-Izayoi:
- Extended all but the Blouse in the `Suit` slot. Giving them a gloved option in either opaque or see-through options.
- Removed the `Hide` for `"ItemNipplePiercings"` in the opaque suits since they are already hidden by layers. See below for more details.

I extended the various catsuits in the `Suit` slot to have a option to allow them to have gloves. This was discussed in the past but likely wasn't done because we lacked the ability to set individual layer priorities at the time.

I have offered both opaque and see-through glove options for all suits including the `ReverseBunnySuit` in the `Cloth` slot. This was done to allow better customization. So if someone wants a see-through suit with solid gloves they can.

This also restores something that was originally possible to do when the Catsuits were released. That being combining them with other gloves notably the Fingerless gloves and the Mistress gloves.

While doing this I did remove the `Hide` for `"ItemNipplePiercings"` from the suits. As these are already hidden by being on a lower layer than the suit and it isn't needed. There is no clipping and without modifying the assets no-one will be able to tell it has changed.

If this is an issue I can remove it. I did want to do the same for the `SuitLower` suits, but some items clipped, and while personally I thought it was fine since the same happens on the see-through suits I decided against it.


Ben987:
This is huge!  Thanks a lot, I will test it soon :)

PR #2977, opened 2021-09-16
Apply remove item logic when swapping

Nina-1474:
When one item is swapped for another, InventoryRemove is not called meaning certain steps are skipped. In particular:
- RemoveItemOnRemove is not checked to remove linked items at the same time. For example apply the Full LockDown Latex Suit, add the wand, swap for another arms item, and the wand will still be left behind.
- The screen flash when going from blinded to full sight doesn't occur. For example any blinding head item to Open Eyes Mouth Slim Mask has no effect.

InventoryRemove is now called in these cases. I'm keeping this in draft for now as it's too central a process to change this late into a beta.


Ada18980 (approved):

Ben987:
Absolutely!  It might solve a few bugs, thanks :)

PR #2978, opened 2021-09-16
Help kneel maintains upper pose

Nina-1474:
When helping another character kneel or stand up, their selected upper body pose would be reset to the default arms-together-in-front. Now their arms position is maintained.


ace-1331 (approved):

Ben987:
Sure!  Thanks for that one :)

PR #2980, opened 2021-09-16
Fix the crotch rope effect chat labels

zorgjbe:
This renames the effect to something more generic, and passes the affecting item name down the text resolution system, allowing the messages to change depending on the item.


Ada18980:
More work to add a crotch rope item but I think this is the better way to do it vs. Adding the item name, so items can have customized messages.

Ada18980 (approved):

zorgjbe:
Merged as 3fac9a2ee3e4e8d22fdae4f3126b1fae4b722e2d. Thank you!~

PR #2982, opened 2021-09-16
Fix: tutu image updates

ace-1331:


Ben987:
Nice!  Thanks for the update.

PR #2984, opened 2021-09-16
Fix a crash when using the techno collar shock module

zorgjbe:
Right now, when the shock collar is installed, there are 2 Draw() calls for the subscreen with the properties initialized correctly, but after a whole message processing loop, DialogFocusItem loses its additional properties, and thus the Draw() call raises.

After looking around with the debugger, trying to add the properties to the module ExtendedDefinition, and not changing anything, I applied a fix from the futuristic crate.


elliesec (approved):

Ada18980 (approved):

Ada18980:
Merging bugfix

PR #2986, opened 2021-09-17
minor assets fix

Atasly:
ocd got triggered
spread transparent lower suits and safeword padlock assets fix


Ben987:
Sure!  Thanks for these fixes :)

PR #2987, opened 2021-09-20
Overhaul image caching

zorgjbe:
I've been investigating the memory usage of the game, and as a figure, right now my Chrome GPU process sits around ~1.16GB. Running this: `GLDrawCanvas.GL.textureCache.clear(); GLDrawResetCanvas()` gets it down to ~350Mb.

This PR consolidates the caching done in `DrawLoadImage` and `GLDrawLoadImage` to a shared cache, and also move the texture cache inline with each cached image. Also adds more retries with exponential backoff (since that was what I was looking at before seeing the two caches), so an image failing to load because of a transient issue should get retried (search for `backoff` in `ImageCache.js` for its calculation).

Notes/bugs/possible enhancements:
- I haven't actually tested it in the face of an error, so it might be buggy if that happens. I know it can display the login screen, and chatrooms.
- I got a bit confused as to why the Drawing cache would behave differently from the one in GLDraw w.r.t to "Assets". I've removed that. The only leftover is `CachedImage.IsAsset()`.
- [ ] The code triggering a redraw after a load completes looks wrong to me now (the `loaded == total`). This is why there's an unused `loadCallback` parameter, but I wanted a POC first.
- [x] It seems there's an issue if you toggle the aliasing on and off. You end up drawing black squares, and `WebGL: INVALID_OPERATION: bindTexture: object does not belong to this context` (GLDraw.js:401). A quick `GLDrawResetCanvas()` fixes it.
- [x] A time-based eviction mechanism would be good.


BondageClub/Scripts/ImageCache.js
```diff
@@ -0,0 +1,217 @@
...
+};
+
+/**
+ * Cached image data
+ */
+class CachedImage {
+	/**
+	 * Create a new cached image.
+	 *
+	 * @param {ImageCache} cache - The cache holding the image
```
elliesec, line 22: I think this type should be `ImageCacheClass`
```suggestion
	 * @param {ImageCacheClass} cache - The cache holding the image
```

BondageClub/Scripts/ImageCache.js
```diff
@@ -0,0 +1,217 @@
...
+		// Get a new image element to perform the loading for us
+		this.element = new Image();
+
+		// Set up listeners to keep track of the loading process
+		let image = this;
+		this.element.addEventListener("load", function() {
+			image.errorCount = 0;
+			image.state = CachedImageState.LOADED;
+			image.cache._imageStateDidChange(image);
+		});
```
elliesec, line 63: I'd drop the `this` aliasing completely and just use arrows in the callback. Similarly for the listener below.
```suggestion
		this.element.addEventListener("load", () => {
			this.errorCount = 0;
			this.state = CachedImageState.LOADED;
			this.cache._imageStateDidChange(image);
		});
```

BondageClub/Scripts/ImageCache.js
```diff
@@ -0,0 +1,217 @@
...
+		this.cache = new Map();
+		this.loadedCount = 0;
+	}
+
+	/**
+	 * Get a cached image from the cache.
+	 *
+	 * @param {string} url - The URL of the image to lookup
+	 * @param {function(CachedImage):void} loadCallback - A callback that will be called when the load completes
+	 * @returns {CachedImage?} The cached image, or null
```
elliesec, line 119: This doesn't look like it ever returns `null` (or `undefined` for that matter, which is what the `Map.prototype.get` call would return without the provided key). Either the type notation should be `CachedImage | undefined`, or it should just be `CachedImage` (I suspect the former?)
```suggestion
	 * @returns {CachedImage} The cached image
```

BondageClub/Scripts/ImageCache.js
```diff
@@ -0,0 +1,217 @@
...
+	 */
+	set(url, image) {
+		var img = new CachedImage(this, url);
+		var realImage = null;
+		if (image instanceof HTMLCanvasElement) {
+			realImage = new Image();
+			realImage.src = image.toDataURL();
+		} else {
+			realImage = image;
+		}
```
elliesec, line 148: Could simplify this a little:
```suggestion
		let realImage = image;
		if (image instanceof HTMLCanvasElement) {
			realImage = new Image();
			realImage.src = image.toDataURL();
		}
```

elliesec (changes requested):
A handful of minor comments/questions, but otherwise looks good - about to pull it down and give it a try.

BondageClub/Scripts/ImageCache.js
```diff
@@ -0,0 +1,217 @@
...
+	_imageStateDidChange(image) {
+		if (image.state == CachedImageState.LOADED) {
+			this.loadedCount += 1;
+			if (image.loadCallback)
+				image.loadCallback(image);
+
+			if (this.loadedImages() == this.totalImages()) CharacterLoadCanvasAll();
+		} else if (image.state == CachedImageState.ERRORED) {
+			if (image.errorCount < 6) {
+				let backoff = (10 * Math.max(image.errorCount - 2, 0)) * 1000;
```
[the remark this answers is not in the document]
zorgjbe: True, I made it more linear in the end, because I couldn't find a formula that wouldn't wander into crazy long times after a while. Arguably, I could remove the `- 2` to make it start backing off sooner; right now it does 3 bursts before backing off (as the previous code did).

BondageClub/Scripts/ImageCache.js
```diff
@@ -0,0 +1,217 @@
...
+
+	/**
+	 * Get a cached image from the cache.
+	 *
+	 * @param {string} url - The URL of the image to lookup
+	 * @param {function(CachedImage):void} loadCallback - A callback that will be called when the load completes
+	 * @returns {CachedImage?} The cached image, or null
+	 */
+	get(url, loadCallback = null) {
+		var image = this.cache.get(url);
```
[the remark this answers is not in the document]
zorgjbe: Oh, it's `let` for "can change" and `const` for "cannot change", and `var` is the old keyword? I messed up those because of Swift. Will adapt, then, thanks for the hint!

BondageClub/Scripts/ImageCache.js
zorgjbe: I think I checked, and `this` wasn't the correct object (`window`, IIRC), hence the "weak" reference. I'll check again and fix if needed.

BondageClub/Scripts/ImageCache.js
zorgjbe: Yep, came from a previous iteration where you could ask the cache for an object without loading it (had a `load(url)` method). That didn't seem helpful, so both methods got merged.

BondageClub/Scripts/ImageCache.js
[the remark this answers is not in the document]
Nina-1474: #2987 implemented a backoff concept using hardcoded times that's been merged into master already. You could always have something similar to that.

BondageClub/Scripts/ImageCache.js
elliesec: `this` won't be the correct object if you pass the callback in as a `function` (i.e. `function() {}`, since functions have their own scope. But if you pass the callback in as an arrow function, as arrows inherit the ambient scope. So:
```js
const outsideThis = this;

// This function won't have access to this
const func = function() {
    this === outsideThis; // false
}

// But this function will
const arrow = () => {
    this === outsideThis; // true
}
```

BondageClub/Scripts/ImageCache.js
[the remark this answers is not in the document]
elliesec: Yep, correct - `var` is the outdated one. `let` for variables and `const` for constants are generally the modern ways of doing things.

BondageClub/Scripts/ImageCache.js
zorgjbe: I did that because the (my? I'm using Atom) linter sees that a straight `let realImage = image` assignment could still be both types after the type check, and gives me a warning for it on `img.setElement(realImage)`. Arguably, the linter seems too clever for its own good 😉.

BondageClub/Scripts/ImageCache.js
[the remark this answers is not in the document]
zorgjbe: Switched to the simpler array method.

BondageClub/Scripts/ImageCache.js
zorgjbe: Fixed, thanks for the pointer ❤️

zorgjbe:
Okay, review comments fixed. As a heads up, I suspect something fishy is going on in b2d72b42b74a195023e0b03d007bebdf3a5d090a, because I'm seeing 404 errors on that PR branch, but if I rebase all of it on top of 82f6486237 there's none. The code seems to try to load assets without the size in (so no `_Large` etc.).

BondageClub/Scripts/ImageCache.js
elliesec: Ah, I see. That'll be the passive type-checking in your IDE rather than the linter, but same effect. Happy to leave that as is in that case - it's not hurting anyone.

Jomshir98: @zorgjbe You can add the type manually above the variable:
```ts
/** @type {HTMLCanvasElement | Image} */
let realImage = image;
```
*(not sure if those exact types are what you want)*

zorgjbe:
I added the time-based eviction on top of it, and tested the backoff by yanking my webserver from under it. Seemed to work fine.

zorgjbe:
Updated, and fixed a couple more issues. It's working great, with the only caveat that it has a tendency to fire up my laptop's fans when I run that, so there might be some performance issue in there, and I don't have the tools to really debug it, so if someone with better graphics stack knowledge could help, that would be great.

zorgjbe:
Okay, this one is good to go, and I've gave it a spin by roaming around in the club with it. Sorry it got a bit out of hand commit-wise.

I don't think I'll tackle implementing a better refresh mechanism for now, since I'm not even sure it's possible to cause a redraw from only one group (I'd take out the duplicate CanvasBlink as a first performance step anyway 😆).

Ben987:
Alright!  It's time to test this big one.  I'm curious to see how the game will react.  Thanks a lot for your hard work :)

PR #2991, opened 2021-09-23
Layer the round piercing

zorgjbe:
Split the round piercing into layers. Each part can now be colored separately (Piercings, Weights, and Chains).
Also stops it from blocking the collar, since it prevents most activities on it.
Make the round piercings show up under the lace babydoll


BondageClub/Assets/Female3DCG/Female3DCGExtended.js
```diff
@@ -2677,7 +2684,7 @@ var AssetFemale3DCGExtended = {
 						BondageLevel: 0,
 						Prerequisite: ["AccessBreast", "AccessBreastSuitZip"],
 						Property: {
-							Type: null,
+							Type: "Base",
```
elliesec, line 2687: I'd recommend keeping the base type as `null`, as that's the way typed items are intended to work, and there are a couple of places in the game code that assume that an extended typed item will have a `null`  type, so removing it _could_ cause issues in some places. If you need to reference the `null` type in `AllowTypes` properties on layers, you can use an empty string (e.g. `AllowTypes: ["", "Weighted"]`).

zorgjbe: Sure, I was mostly trying to get the layer toggles right. Fixed up and rebased.

zorgjbe:
No problem. Anything adding some color flavors to the game is a 👍. Thank you for merging!

PR #2997, opened 2021-10-02
Prettify source

zorgjbe:
This a proposal at using some automatic indenter check as part of CI, as well as finding some settings for it that don't mush the code too much (I've purposefully added `Female3DCG{,Extended}.js` to the ignore list for now.

Opinions on that are welcome 😛.


elliesec:
We've discussed using formatters on the project before, and largely concluded that as things stand, there's little point in enforcing a formatter right now, for a couple of reasons:
* This kind of change needs to come in with the active blessing of the project maintainer (i.e. Ben), and tooling like this is not something that he's likely to introduce at the present time
* The biggest benefit of using prettier comes when it's run _before_ PR time, and in a project where contributors' experience levels range from very experienced down to zero (and Ben has expressed a wish not to enforce any kind of tooling that contributors must use), it's very unlikely that we're going to be able to enforce that.

I'll quote what I said the last time this was mentioned:
> I'm a huge fan of prettier, and use it in personal projects (and mandate its use on my project teams at work). However, whilst it's easy to declare that a project should use a certain tool or follow certain standards, it's important to take into consideration where the project is currently, how much disruption making those changes would cause, and whether the maintainer (Ben) is going to effectively hold the project to those standards.
>
> As @Jomshir (Clare) [23664] has said, running prettier over the codebase would result in a major reformatting of practically every file (not to mention that the last time I ran it over Female3DCG.js probably 9 months or so ago, the file ended up 15k+ lines long). Moreover, in my opinion the biggest benefit of having a consistent formatter on a project is the difference it makes when it comes to reading code diffs; to realise that benefit fully, you want everyone that contributes to be running the formatter over their code before raising a PR (through git hooks or otherwise). Currently, knowing how to use Node/npm is not essential for contributors to the game (and Ben has previously stated that he doesn't want to impose this requirement on people - he wants the game to be accessible to contributors with next to no development experience), and enforcing something like prettier on people would require extra technical knowledge and steps as a barrier to entry for people to contribute.

I understand (and largely support) the principle, but this project has proven on numerous occasions that introducing tooling like this is going to have very limited benefit until Ben is actually strict about enforcing it.

zorgjbe:
Thanks @elliesec for the history. I'm not wed to having an automated linter, and if there's interest, I'm pretty sure it's doable to wire it as a pre-commit hook, if that's something that would make it easier.

On the subject of Female3DCG, I really think splitting assets out of that file and moving their defs inside either their respective .js files, or in a new `Assets3DCG.js` would be something that helps. But as you say, that's for @Ben987 to decide that.

elliesec:
> Thanks @elliesec for the history. I'm not wed to having an automated linter, and if there's interest, I'm pretty sure it's doable to wire it as a pre-commit hook, if that's something that would make it easier.
> 
> On the subject of Female3DCG, I really think splitting assets out of that file and moving their defs inside either their respective .js files, or in a new `Assets3DCG.js` would be something that helps. But as you say, that's for @Ben987 to decide that.

The only issue with setting it up as a precommit hook would be that it enforces an "installation step" for contributors - people would at the simplest run an `npm i` themselves, which Ben has said he'd rather not enforce.

And on splitting up the `Female3DCG` files, I fully agree. Splitting them out (at least into groups) would be nice and make things much more readable. The only issue right now is that we're not using any bundling mechanism, which means we'd be loading up 50+ extra files on page load, and page load is already quite janky. I'd _like_ to get to a place where we can properly use modules and transpilation/bundling, but again, that's something that needs to come from Ben, and I think we're still a way away from that.

zorgjbe:
As a stop-gap measure, would it be okay if I resubmit that PR without the prettify-everything step, so at least IDEs can do the right thing? I had a hard time making Atom and TextMate respect the current indent without rewriting all files.

That means keeping the first 2 commits, dropping the 3rd and 5th, and changing the 4th so it only has the other existing commit in it.

Ben987:
Let's not add anything that would require contributors to run extra programs or do extra steps to be able to contribute.  We already lost a few contributors because the project got too complicated to understand for them.

I'm really not sure about that change.  For example:

for (
		let i = BackgroundSelectionOffset;
		i < BackgroundSelectionView.length &&
		i - BackgroundSelectionOffset < BackgroundSelectionSize;
		++i
	)

To me, it makes it harder to read the "for" structure by having it on six different lines instead of one.  Maybe it's the new way to structure code and I'm not used to it.

zorgjbe:
Hence why I'm proposing to only do "passive" things, ie. standardizing `.editorconfig` and stuff like that so at least IDEs don't do weird stuff like mine did (and I loose time discarding unrelated whitespace changes in my commits). I'm fine with dropping the Prettier addition, I'll just look for a way to run it only on lines I change so at least my code is consistent.

I do agree with that weird `for`, but the point of Prettier is to just kill the style war once and for all, so it has no configuration whatsoever — only simple stuff. If I understand that `for` loop correctly, I'd `stuff.slice(offset, size).forEach(…)` because a 80+ long `for` statement is already unreadable for me, as it hampers readability when you have another pane open on another file (and I work like that, so 🤷).

I'll close this PR and reopen one with only the added configuration files and no Prettier run.

PR #3001, opened 2021-10-06
Consolidate package config, bump eslint, and add editorconfig

zorgjbe:
A rejig of #2997, with only the `package.json` consolidation, merging the eslint configs, addition of `.editorconfig` and `.git-blame-ignore-revs`.


BondageClub/Tools/Node/package.json
```diff
@@ -6,9 +6,9 @@
     "changelog": "node GenerateChangelog generate",
     "changelogPrepare": "node GenerateChangelog prepare",
     "checks": "npm run lintscripts && npm run assetcheck-typescript && npm run assetcheck",
-    "lint": "eslint --config ./.eslintrc.js ../../Assets/Female3DCG/Female3DCG.js",
-    "lint:fix": "eslint --fix --config ./.eslintrc.js ../../Assets/Female3DCG/Female3DCG.js",
-    "lintscripts": "eslint --quiet \"../../Scripts/**/*.js\" \"../../Screens/**/*.js\" \"../../Assets/**/*.js\" \"../../Tools/**/*.js\" \"../../Backgrounds/Backgrounds.js\"",
+    "lint": "eslint --resolve-plugins-relative-to Tools/Node ../../Assets/Female3DCG/Female3DCG.js",
+    "lint:fix": "eslint --resolve-plugins-relative-to Tools/Node --fix ../../Assets/Female3DCG/Female3DCG.js",
+    "lintscripts": "eslint --resolve-plugins-relative-to Tools/Node \"../../Scripts/**/*.js\" \"../../Screens/**/*.js\" \"../../Assets/**/*.js\" \"../../Tools/**/*.js\" \"../../Backgrounds/Backgrounds.js\"",
```
elliesec, line 11: Shouldn't need the `--resolve-plugins-relative-to` now.

elliesec: Hmmm, I remain unconvinced. I'm fine with keeping it if we know exactly _why_ it's needed.

zorgjbe: The docs are right. A vanilla `eslint` from the shell will use the current directory, as specified. The rest of the docs say "This option should be used when plugins were installed by someone other than the end user. It should be set to the project directory of the project that has a dependency on the necessary plugins.". `npm`, in that case, is the "someone other than the end user", thus rule 2 "If an integration has dependencies on ESLint and a set of plugins, and the tool invokes ESLint on behalf of the user with a preset configuration, the tool should set --resolve-plugins-relative-to to the top-level directory of the tool." needs to apply, and the option must be set.

Arguably, I find that just having the `package.json` at the root would be easier than to jump through hoops like that.

PR #3009, opened 2021-10-10
Beta fix: Fixes 404s caused by validation dummy characters

elliesec:
## Summary

Fixes 404s that can occur when entering a chatroom whilst wearing certain item combinations. This is caused by the new checks in #2936 using `CharacterRefresh` on a dummy character to validate prerequisites, which triggers a canvas rebuild and subsequently 404s when the game tries to load the character's assets. This replaces the `CharacterRefresh` calls with a combination of `CharacterLoadEffect` and `CharacterLoadPose`, which should be sufficient for most prerequisites. In the future a slightly more comprehensive solution might be to create a different type of "in-memory-only" character whose assets never get loaded/rendered rather than using `CharacterLoadSimple`.


Ben987:

PR #3016, opened 2021-10-13
Describe the round piercings as not drawing locks

zorgjbe:
Silence 404 errors when trying to get the lock image


Ben987:
Yes!  Thanks for that fix :)

PR #3017, opened 2021-10-13
Image Rendering Fix

Nina-1474:
Now that #2962 spaces out how often we try to retrieve a missing image, the amount of time before we give up and consider that image "loaded" has gone from instant to 5 minutes. This means that depending on the order images are loaded, we can end up with images successfully loaded that don't appear on screen because the final canvas-rebuild is held up waiting for the missing image retries to complete (or more likely a character refresh from elsewhere).
This fixes the error by considering the image "loaded" as soon as the first error occurs, allowing a canvas rebuild to occur quickly once all other images are loaded.

Keeping this as a draft for now as we may have a much better solution instead soon.


Nina-1474:
Closing in favour of #3022

PR #3022, opened 2021-10-14
Beta Fix: Revert #2962

elliesec:
This reverts #2962. After several attempts at putting together fixes for the draw delay issue mentioned in #3017 and #3018, @Nina-1474 and I have decided that the safest thing to do for now is to revert #2962, as we've ended up running into several edge cases along the way that are making us question whether a bigger restructure of the code would be needed to fully solve the issue, which wouldn't be sensible to include as a beta fix.

Given that #2987 is also on the horizon, which reworks more or less all of the code around image loading & caching, this might be something we can revisit at a later date (or we simply might not need to).


Nina-1474 (approved):

Ben987:
Thanks a lot you two, I'm sure you'll find a solution later :)

PR #3042, opened 2021-10-26
Fix KneelingSpread pose

zorgjbe:
This redraws the base assets from the kneeling spread pose, rotating the legs.

- Most of the current BodyLower clothing assets have been/are being redone on top of the new pose.
- There are two "technical" changes: WoodenHorse, which is ParentGroup null since it doesn't need size variants, and HaremPants, which had pose-specific folders it didn't really need, and full size 500x1000 assets, which I cropped down to something more standard (needed for it to work, see the Tutu below).
- Kudos to gatetrek for sending me its work on the ClothLower group, which means everything from that group is supported with the following exceptions:
  - The Waspies. They don't look spread-compatible, so I've hidden them instead.
  - The Long Skirt. It was marked as hidden beforehand.
  - ~Long Leggings. Seems doable though~.
  - The Tutu, because I didn't see a way to have it work with the `X: -90` from the pose. Ideas welcome.
  - There are some (minor) clipping issues (~Jeans~, ~LatexPants~, ~Pajamas~, ~RippedJeans~).
- SuitLower is complete, ~apart from what comes from Socks~.
- Socks still misses a couple assets.

⚠️ This adds the pose to the menu, the goal being to have it usable in-game, but since there are assets missing from Socks still, maybe that should wait; alternatively I could make a pass and `HideForPose` anything that's not done yet.


Ben987:
@zorgjbe , thanks!  So will there be any errors or 404 message in the console log left with that version?  If it's only invisible assets, we can always add them later.  Cheers :)

Ben987:
There are more conflicts with other changes.  Can you make sure there are no 404 errors left?  You can put blank images to start with, until we get the real images.  Thanks!

zorgjbe:
@Ben987 There you go, rebased and all shoes are now hidden in the pose.

Ben987:
This one is huge!  It's a good thing the beta is coming up to test it.  Thanks a lot :)

zorgjbe:
Thanks for the merge!

PR #3043, opened 2021-10-29
Feature: 3 new wing assets by Titania

elliesec:
Adds three new wing assets by @QueenTita: Bee Wings, Pixie Wings and Cyber Wings. Previews below.

Preview


Ben987:
It looks wonderful!  Thanks a lot you two :)

PR #3044, opened 2021-10-30
Bugfix: Wardrobe crash

elliesec:
## Summary

This fixes an issue introduced by #2995 which could occur when a `CharacterRefresh` (or anything else triggering a `DialogBuildActivities` call) was called on a character whilst they had a group focused that did not have an associated `Activity` array. In practice this only happens when in the wardrobe (as clothing groups don't have associated activities), or when equipping an item from one of the non-interactive groups (i.e. `ItemMisc`, `ItemDevices` or `ItemAddon`), but can cause the game to crash.

Rather than adding a null-check here, I've modified `Asset.js` to ensure that groups always have an associated `Activity` array, which defaults to empty if they have no activities defined in `Female3DCG.js`.

## Steps to reproduce

There are two main reproduction methods

### Item groups

1. Click on a character
2. Click on the `ItemDevices` group - note that a stack trace is printed to console - this doesn't crash the game
3. Equip an item in the `ItemDevices` group - the game crashes and cannot be recovered by normal means

### Wardrobe groups

1. Click on a character and enter the wardrobe to change their clothes
2. Click the centre of one of the clothing groups to enter the item selection subscreen
3. Click the cancel button
4. Note that a stack trace is printed to console - this doesn't crash the game, but the only way to exit the subscreen is by accepting the changes via the tick button

The associated stack trace generally looks something like this:

```
Activity.js:133 Uncaught TypeError: Cannot read properties of undefined (reading 'includes')
    at ActivityHasValidTarget (Activity.js:133)
    at Activity.js:257
    at Array.filter (<anonymous>)
    at ActivityAllowedForGroup (Activity.js:251)
    at DialogBuildActivities (Dialog.js:781)
    at CharacterRefreshDialog (Character.js:1018)
    at CharacterRefresh (Character.js:978)
    at CharacterAppearanceSetItem (Appearance.js:860)
    at AppearanceMenuClick (Appearance.js:1201)
    at Object.AppearanceClick [as Click] (Appearance.js:988)
```


Ben987:
Absolutely!  Thanks a lot for that fix :)

PR #3045, opened 2021-11-02
Bugfix: Reverts changes to the `Hide` property made in #2972

elliesec:
Reverts the changes to the `Hide` property for the Catsuit, Seamless Catsuit and Pilot Suit that were made as part of #2972, meaning that the `BodyUpper` group would be hidden, resulting in characters' hands disappearing in the `Yoked`, `OverTheHead` and `BackCuffs` poses (similar to the issue for heads prior to #3027).

This should fix the most serious graphical issues, although there are still some imperfections that should eventually be addressed around the gloves for these suits. Eventually, we should make official assets for the `Hands` group in these poses, which would alleviate most of the issues.


Ben987:

PR #3046, opened 2021-11-03
Removed Floating Pixels on Steampunk Wings

Shiranui-Izayoi:
I am not sure if they were meant to be there. Like if they were meant to be a tiny gear or something.

But since they are only visible in very certain backgrounds I am assuming they were just an accident. So have removed them.

The files got slightly bigger due to the program I saved them in.


elliesec:
Probably worth getting confirmation from @QueenTita on this one, as the original asset creator.

Shiranui-Izayoi:
> 
> 
> Probably worth getting confirmation from @QueenTita on this one, as the original asset creator.

I will try to contact her but from what I know she is currently busy with collage or work one or the other. And thus may not have time to respond.

QueenTita:
I think the pixels Aeren has mentioned can be removed. I just checked on them and i probably left them there by mistake.

elliesec (approved):

Ben987:
Very good!  Thanks a lot everyone :)

PR #3047, opened 2021-11-03
Added Steampunk Corset & Heavy Latex Corset into the Corset Slot

Shiranui-Izayoi:
Copied the needed assets over and added the Steampunk Corset & Heavy Latex Corset to the `Corset` slot. Since the Leather Corset Top has been in the `Corset` Slot for a while I felt that this made sense to do. As to my knowledge there is nothing stopping the Steampunk one from being in the `Corset` slot.

And also I am not aware of anything that was said about the Heavy Latex that would limit it to the ItemTorso slot. As at least personally I consider the `ItemTorso` slot to be the lockable version to the `Corset` slot. If the change is liked. I may look into doing this for other similar items in the slot. Like the Futuristic Harness with the permission of Ada and the Baby Harness.


elliesec (changes requested):
Rather than duplicating the assets across, this should be making use of `DynamicGroupName` to share assets between groups. As currently, the precedent is for `ItemTorso` items to dynamically reference items in the `Corset` group, I'd recommend doing the following:

* Remove the assets for the Heavy Latex Corset from the `ItemTorso` group (with the end result being a straight move of the assets from `ItemTorso` to `Corset`)
* Similarly, remove the assets for the Steampunk Corset from the `Cloth` group (with the end result being a straight move of the assets from `Cloth` to `Corset`)
* Add `DynamicGroupName: "Corset"` to the `ItemTorso` version of the Heavy Latex Corset, along with any other adjustments (e.g. positioning) that might be needed.
* Add `DynamicGroupName: "Corset"` to the `ItemTorso` version of the Steampunk Corset, along with any other adjustments that might be needed.

This avoids us having to store (and consequently cache in memory) two identical sets of assets.

Shiranui-Izayoi:
> 
> 
> Rather than duplicating the assets across, this should be making use of `DynamicGroupName` to share assets between groups. As currently, the precedent is for `ItemTorso` items to dynamically reference items in the `Corset` group, I'd recommend doing the following:
> 
>     * Remove the assets for the Heavy Latex Corset from the `ItemTorso` group (with the end result being a straight move of the assets from `ItemTorso` to `Corset`)
> 
>     * Similarly, remove the assets for the Steampunk Corset from the `Cloth` group (with the end result being a straight move of the assets from `Cloth` to `Corset`)
> 
>     * Add `DynamicGroupName: "Corset"` to the `ItemTorso` version of the Heavy Latex Corset, along with any other adjustments (e.g. positioning) that might be needed.
> 
>     * Add `DynamicGroupName: "Corset"` to the `ItemTorso` version of the Steampunk Corset, along with any other adjustments that might be needed.
> 
> 
> This avoids us having to store (and consequently cache in memory) two identical sets of assets.

Wait this is a thing? Since when was this a thing? I guess though it only works on the full asset and not layers as if so I could have made a previous pull request way better. But yes I will get right on it. I doubt it will need any changes since the `Left` and `Top` Will be the same for all.

Shiranui-Izayoi:
I just realised an issue with using the `DynamicGroupName` thing mainly on the Steampunk Corset. The main issue is that the asset for the default pose of the cloth slot has hand cut outs. While the default for all corsets doesn't. Also is that `Corset` slot doesn't have folders for some of the poses that `Cloth` does.

Should I just copy over those folders whole sale or what?

Shiranui-Izayoi:
Ok so I have added `DynamicGroupName` to the assets. Then moved and deleted the no longer needed assets. While doing this I have also done this for the `LeatherCorsetTop` as I felt it fell within this pull request. As it was about adding the missing "corsets" to the corset slot.

Also I have added `DrawLocks: false` to the `HeavyLatexCorset`. Since the assets for it were blank and locks were invisible. meaning their was no need to keep them like that when moving them over.

I hope this was everything needed.

BondageClub/Assets/Female3DCG/Female3DCGExtended.js
```diff
@@ -2933,6 +2933,27 @@ var AssetFemale3DCGExtended = {
...
+					},
+				],
+				Dialog: {
+					Load: "SelectHeavyLatexCorsetType",
+					TypePrefix: "HeavyLatexCorsetType",
+					ChatPrefix: "HeavyLatexCorsetSet",
+					NpcPrefix: "HeavyLatexCorset",
+				},
+			}
+		}, // HeavyLatexCorset
```
elliesec, line 2956: If this config is identical to the existing Heavy Latex Corset, you can use the `CopyConfig` property to duplicate configuration between groups:
```suggestion
		HeavyLatexCorset: {
			Archetype: ExtendedArchetype.TYPED,
			CopyConfig: { GroupName: "ItemTorso", AssetName: "HeavyLatexCorset" },
		}, // HeavyLatexCorset
```

Shiranui-Izayoi: It isn't completely identical mainly I removed the chat tags from the corset version since it doesn't need them. As cloth type items don't send a chat message. I am not sure if it matters. If it doesn't then I can commit what you are suggesting.

elliesec: I don't believe that matters off the top of my head - I'd recommend trying it out. If we can get away with trimming the config down, I think we should.

Shiranui-Izayoi: Ok I am going to commit it and test it out. I doubt it will matter. Since if it isn't sending the message in the first place I doubt the chat tags will matter. Which means I am now also going to end up doing this for the Suits. Since then I only need 2.

Shiranui-Izayoi: Ok tested it now there is no difference I can see it can stay merged. This should be all done.

elliesec (changes requested):
There's currently an issue with the Leather Corset Top in the `Corset` group in the `OverTheHead`, `BackBoxtie`, `BackElbowTouch` and `BackCuffs` poses. Looks like the you'll want to adjust the assets and/or `AllowPose` entries.

Shiranui-Izayoi:
> 
> 
> There's currently an issue with the Leather Corset Top in the `Corset` group in the `OverTheHead`, `BackBoxtie`, `BackElbowTouch` and `BackCuffs` poses - they have hand cutouts. Looks like the you'll want to adjust the assets and/or `AllowPose` entries.

Ahhhh my bad yes that makes sense. I had forgotten that the corset group doesn't tend to change assets for those poses. I will get it done give me like 10 minutes.

Shiranui-Izayoi:
Done! Should work fine now. Don't know how I missed it.

elliesec (approved):
LGTM

Shiranui-Izayoi:
What failed? I can't tell what it's issue is. It won't tell me.

Shiranui-Izayoi:
Ok great... All that went wrong was that github managed to time out and screw the checks.

Ben987:
Thanks again!  It's time to test this one :)

EmilyR42:
Please ask me next time you use the assets I made, okay?
I would have preferred the Heavy Latex Corset to stay as a restraint because it's too heavy and restricting to simply be a clothes item and I would have preferred it to always be lockable to emphasize that.

Shiranui-Izayoi:
> 
> 
> Please ask me next time you use the assets I made, okay? I would have preferred the Heavy Latex Corset to stay only as a restraint because it's too heavy and restricting to simply be a clothes item, always be lockable to emphasize that. Thank you.

I am very sorry. I did try to find who made it. But I guess I didn't look in the right place. Won't happen again. Originally this was only going to be the steampunk corset. But I ended up expanding it. If you wish I can put in a second request to remove it from the slot.

elliesec:
@Shiranui-Izayoi - if you're not sure who made something, I'd recommend asking one of the devs, or mentioning that you're looking for the original creator in the PR description. If we don't know who originally made it off the top of our heads, we can usually track them down via the Git history.

EmilyR42:
@Shiranui-Izayoi Thanks a bunch for understanding! Sure, that'd be wonderful

Shiranui-Izayoi:
> 
> 
> @Shiranui-Izayoi - if you're not sure who made something, I'd recommend asking one of the devs, or mentioning that you're looking for the original creator in the PR description. If we don't know who originally made it off the top of our heads, we can usually track them down via the Git history.

I had at the time looked through the time looked through the commit history I just must have missed it. Just a reason to look harder next time. I can always undo heavy corset commit. Which it seems like is wanted. Should just have to move the assets back and that should be it.

Shiranui-Izayoi:
> 
> 
> @Shiranui-Izayoi Thanks a bunch for understanding! Sure, that'd be wonderful

Sure will move them back. Honestly part of the reason for adding it to the Corset slot was to allow it to be used with other `ItemTorso` only items such as the various harnesses like the Future Harness etc. Since I thought it would look great with it.

Give me a few days to submit another request. If I haven't by then poke me since I may have forgotten. But I am a bit busy at this moment.

elliesec:
> > @Shiranui-Izayoi - if you're not sure who made something, I'd recommend asking one of the devs, or mentioning that you're looking for the original creator in the PR description. If we don't know who originally made it off the top of our heads, we can usually track them down via the Git history.
> 
> I had at the time looked through the time looked through the commit history I just must have missed it. Just a reason to look harder next time. I can always undo heavy corset commit. Which it seems like is wanted. Should just have to move the assets back and that should be it.

My point stands - if you couldn't figure it out on your own - ask someone. Some of us are pretty good at traversing the Git commit tree. It would have taken minutes to find out that it was introduced in commit 8367df029bc431301407c955a41b4ce59e61c624 and PR #1711, versus the amount of time and effort needed to roll things back.

PR #3048, opened 2021-11-04
Removed Blank Lock Assets from corsets and added DrawLocks: false to them

Shiranui-Izayoi:
As said in the title super quick PR. That I ended up doing when adding the corsets to the slots. I seperated for the sake of making things easier to review.

All I did was delete the less than 1kb sized invisible locks that are a hold over from when `DrawLocks` didn't exist and add it in to the `ItemTorso` version of Corset2, Corset3, Corset4 and Corset5. Honestly typing up this message took longer than the pull request.


Ben987:
Absolutely!  Thanks :)

PR #3049, opened 2021-11-04
DynamicGroupName for ReverseBunnySuit

Shiranui-Izayoi:
Added `DynamicGroupName` to the ReverseBunnySuit items in the Cloth and Socks slots. Then deleted the now not needed assets. This resulted in around 110 assets being removed as they are no longer needed which is great.

Now if only I could dynamically call assets that weren't named the same. Then I could remove the like 600 assets that I had to add to make the extended suits work :) But alas that is a dream for now.


Shiranui-Izayoi:
With the new knowledge that `CopyConfig` is a thing and since this is basically a PR for streamlining and improving the Suits I decided to add in the `CopyConfig` to the various extended suits. Meaning it went from around 5 that are the same and 2 that are the same. To only 2 total. 

Since their is a difference between the See Through suits and the normal ones. Mainly cause I decided that the alt gloves should be the ones not matching their "style".

If I hadn't have done that then I could have put them all in a single config. But well I am a little dumb and don't want to mess with peoples wardrobes. Just to rename them. Nothing seems to have changed for the user when I tested it by going through all the menus.

Ben987:
@elliesec , is this something you'd like to review?   I haven't used CopyConfig myself.  Thanks.

elliesec (approved):
@Ben987 - This all looks fine to me. `CopyConfig` just allows extended configurations to be copied between items in `Female3DCGExtended.js`, and since a lot of the suits used the same extended configuration, it makes sense to use here. I've tested this PR and it all appears to be working as far as I can see.

Ben987:
Perfect!  Thanks a lot then :)

PR #3050, opened 2021-11-04
Fixed Fur Gloves and Socks Colouring

Shiranui-Izayoi:
I finally had enough of how the Fur Gloves and Fur Socks coloured so strangely, compared to their non fur counterparts and decided to fix it. Considering that they are basically the same asset with fur on them. With the base asset coloured differently. 

So with the fact that they are multi colour and multi layer now I swapped over the `Fabric` layer to their original asset in order to fix the colouring issue.

As far as I can tell from zooming in on the assets about 1000% they look the exact same by default before and after. I have included an image showing this as well as how at full white they look different.

FurItemDifference


Ben987:
That looks fine, thanks a lot!  Cheers :)

PR #3052, opened 2021-11-06
Fix: Updated masks for Super Thick ABDL Diapers

Shiranui-Izayoi:
Apparently when making it I had somehow missed a few masks. Mainly I had forgotten to apply them to the `ItemPelvis` versions in a few cases.

While doing this I also added a mask for the Hula Skirt. Which I left mostly intact except for the top part. Since it is just leaves it made sense that it could "push them aside" so I felt that it showing through the sides fit.

I also made the `ItemPelvis` versions of the diapers hide two problem items. Those being the Mistress Bottoms and the normal Shorts. As these normally hide the `Panties` version. But since the `ItemPelvis` version is meant to be a chastity device as well. Not to mention be able to be put on on top of clothing it makes sense to not have those 2 items hide them.

But due to those 2 items being on a higher layer they would render over the top and look quite strange. So I had the chastity variants hide them.


BondageClub/Assets/Female3DCG/Female3DCG.js
```diff
@@ -474,7 +474,7 @@ var AssetFemale3DCG = [
 				],
 			},
 			{ Name: "MistressBottom", Fetish: ["Leather"], Value: -1, Hide: ["Panties"], HideItem: ["ItemButtAnalBeads2", "ItemVulvaVibratingDildo", "ItemVulvaInflatableVibeDildo", "ItemVulvaClitSuctionCup", "ItemVulvaHeavyWeightClamp", "ItemVulvaClitAndDildoVibratorbelt", "ItemVulvaShockDildo", "ItemVulvaPenisDildo", "ItemVulvaPiercingsVibeHeartClitPiercing", "ItemVulvaPiercingsClitRing"] },
-			{ Name: "Waspie1", Fetish: ["Leather"], Value: 60, Alpha: [{ Group: ["Panties", "ItemPelvis"], Masks: [[0, 0, 165, 600]] }, { Group: ["Panties"], Masks: [[500, 0, -165, 600]] }, { Group: ["Panties"], Masks: [[0, 0, 500, 480]] }], HideItem: ["ItemDevicesStrapOnSmooth", "ItemDevicesStrapOnStuds"], Expose: ["ItemVulva", "ItemVulvaPiercings", "ItemButt"], Attribute: ["Skirt"], },
+			{ Name: "Waspie1", Fetish: ["Leather"], Value: 60, Alpha: [{ Group: ["Panties", "ItemPelvis"], Masks: [[0, 0, 165, 600]] }, { Group: ["Panties", "ItemPelvis"], Masks: [[500, 0, -165, 600]] }, { Group: ["Panties", "ItemPelvis"], Masks: [[0, 0, 500, 480]] }], HideItem: ["ItemDevicesStrapOnSmooth", "ItemDevicesStrapOnStuds"], Expose: ["ItemVulva", "ItemVulvaPiercings", "ItemButt"], Attribute: ["Skirt"], },
```
elliesec, line 477: I hadn't noticed this one before, but masks can (and should, because it's less memory-intensive) be combined into a single declaration if they apply to the same groups. I'd also recommend using positive values for width and height in masks, as negative values are somewhat counterintuitive.

```suggestion
			{ Name: "Waspie1", Fetish: ["Leather"], Value: 60, Alpha: [{ Group: ["Panties", "ItemPelvis"], Masks: [[0, 0, 165, 600], [335, 0, 165, 600], [0, 0, 500, 480]] }], HideItem: ["ItemDevicesStrapOnSmooth", "ItemDevicesStrapOnStuds"], Expose: ["ItemVulva", "ItemVulvaPiercings", "ItemButt"], Attribute: ["Skirt"], },
```

Same comment applies to the Short Pencil Skirt below and any other items which might have multiple masks defined for the same groups.

Shiranui-Izayoi: Huh never knew you could do this. Thought you had to redefine it each time. Makes sense though. I will get on that since their is a lot of them. Other than that, I will redefine it sure. Honestly I did think that to others negative values were counterintuitive. Give me an hour or so to fix stuff.

elliesec: Yep, that's why it's an array of arrays, rather than just a single array.

Shiranui-Izayoi:
Ok done should be fixed. I managed to work up the motivation to do it right away so yay. I just hope it works the same. But it should no reason for it to not.

Ben987:
This one has a conflict with another pull request.  Can you resolve the conflict and test again?  Thanks :)

Shiranui-Izayoi:
> 
> 
> This one has a conflict with another pull request. Can you resolve the conflict and test again? Thanks :)

Done.

Ben987:
Thanks!  I will test it soon.

PR #3053, opened 2021-11-06
Revert Heavy Latex

Shiranui-Izayoi:
On request of the items author this item is staying a restraint only. Will miss some of the looks I created in testing :)


Shiranui-Izayoi:
@EmilyR42 this should all be back to normal. I still kept the streamlining I did like `DrawLocks: false`

Ben987:
Thanks!  @EmilyR42 , let me know if it works as intended now.

PR #3054, opened 2021-11-07
Feature: Hood clipping improvements

elliesec:
## Summary

This makes changes to several hoods to address several clipping-related issues. I've also made several changes to the `AllFours` and `Hogtied` poses to enable removal of the `CanUseAlphaHood` and `CannotBeHogtiedWithAlphaHood` prerequisites, so that the Sensory Deprivation Hood, Sealed Leather Hood and Open Hair Latex Hood can all now finally be used in the `Hogtied` and `AllFours` poses without clipping problems.

* Adds `ItemHead` to the `Hide` array for most fully-enclosing hoods, to prevent items clipping out of these hoods (the VR Headset currently has a few such minor issues, but the main motivation for this is an as-yet unreleased item that I'm currently working on)
* Adds alpha masks to most hoods that allow parts of the face to show through the hood. Again, this is to prevent items in the `ItemHead` group from clipping out of those hoods
* Splits the head out of the `Hogtied` and `AllFours` poses, and ensures the `Head` group is no longer hidden in the `AllFours` pose
* Extends the feet for the hogtied pose a little so that there is a portion of the pose behind the head - this is to enable the aforementioned hoods to work without leaving a conspicuous gap between the feet and the head
* Fixes an issue caused by #2892 where the hands for the `AllFours` pose on the `Asian` and `Black` bodies weren't lined up properly.
* Removes the `CanUseAlphaHood` and `CannotBeHogtiedWithAlphaHood` prerequisites, as all hoods are now fully compatible with the `Hogtied` and `AllFours` poses

Below shows a before/after comparison for the all fours/hogtied poses on the asian body in combination with the Open Hair Latex Hood, which illustrates a lot of the changes fairly well. The top images are before and the bottom images are after.

preview


Ben987:
It looks much better like that.  Thanks a lot :)

PR #3055, opened 2021-11-08
Refactor/Bugfix: Prevents self-blocking items

elliesec:
## Summary

Currently, it is possible for an item to effectively block itself due to some combination of its prerequisites and its `Effect`, `Block`, `SetPose` or some other property. This leads to items that can be added but never removed.

* A real example: the Chastity Vulva Shield, which includes `Block: ["ItemVulva"]` and also has the prerequisite `AccessVulva`
* A hypothetical example: an item that includes `SetPose: ["Suspension"]` and also has the prerequisite `NotSuspended`

This PR modifies the behaviour of the `InventoryAllow` function so that when checking prerequisites for an item, the currently equipped item in that item's group will be ignored. It does this in a similar way to the cyclic validation checks from #2936, and the way self-block checks were handled in `ExtendedItem.js`: by loading a simple character into memory, copying the character's appearance across minus the group being checked, and then running the prerequisite checks on the simple character.

This fixes an existing bug with most chastity piercings blocking themselves and being impossible to remove, which was introduced in #3032 when the `ItemVulvaPiercings` group was added to the conditions for the `AccessVulva` prerequisite. This will also likely allow us to clean up some of the gag prerequisite code, although I've left that as is for now, and will likely come back to it in a future PR. It will also permit items like gags that require the character's mouth not to be blocked, but block the mouth once worn (which don't exist yet, but there is one such item on the way).

## Changes

* Refactored `InventoryAllow` - it now takes an `asset` parameter, and the `prerequisites` parameter is now optional (defaulting to `asset.Prerequisite` if undefined). The currently worn item in the same group as the `asset` will be ignored for the purposes of the prerequisite checks
* Removed the `SelfBlockCheck` property from extended items, and removed the `ExtendedItemSelfProofRequirementCheck` function entirely, as this is now the default behaviour for any `InventoryAllow` call.

## Steps to reproduce
1. Equip a chastity piercing item (Chastity Vulva Piercings, Chastity Vulva Shield or High Security Vulva Shield)
2. Note that the item is impossible to remove


Ben987:
Very nice fix!  Thanks :)

PR #3059, opened 2021-11-09
Fix futuristic collar bug

Ada18980:
Fixes a bug where various things like the VR headset and futuristic earphones could be modified if a collar with properties was worn that wasnt the futuristic collar.


Ben987:
Thanks for that fix!

PR #3060, opened 2021-11-10
Bugfix: Pilot Panties 404s

elliesec:
## Summary

This fixes an issue where the new Pilot Panties in the `SuitLower` group would not be displayed in the several poses due to 404 errors. This was down to a discrepancy between the `AllowPose` entry on the `SuitLower` version of the item and that on the `Panties` version. I also removed the `HideForPose: ["KneelingSpread"]` property on the item introduced in #3042 as it wasn't needed in this case.


Ben987:
Sure!  Right on time for the beta :)

Ada18980:
Thanks Ellie for noticing this

PR #3062, opened 2021-11-10
Bugfix: Prevents 404s on `SuitLower` items in the `Horse` pose

elliesec:
## Summary

The items in the `SuitLower` group that duplicate items from the `Socks` group (pantyhose & stockings) don't have assets for the `Horse` pose, meaning that wearing them on the Wooden Horse will result in 404 errors. This PR adds `HideForPose: ["Horse"]` on those items to suppress the 404s. Eventually, we may want to merge the `Horse` and `KneelingSpread` poses, at which point, these items should be compatible with the Wooden Horse again.


Ben987:
Good catch!  Thanks for that fix :)

PR #3057, opened 2021-11-09
Feature: New Lampshade Hood item

elliesec:
Adds a new Lampshade Hood item made @QueenTita. This also adds a new Forniphilia (human furniture) fetish, which has been added to the new item. Preview below:

pr-preview

PR #3063, opened 2021-11-10
Bugfix: Display Frame graphical artefacts

elliesec:
This tidies up a couple of minor graphical artefacts on the Display Frame asset that I noticed in the preview image for PR #3057. They're generally visible when the colour of the display frame is set in high contrast to the background colour of a chatroom. I've highlighted the artefacts on the original asset on a black background to illustrate below.

image


Ben987:
Sure!  Thanks for that one :)

PR #3064, opened 2021-11-10
Beta Fix: Open-back chastity belts preventing `ItemButt` items from being used

elliesec:
## Summary

In #3032, `ItemPelvis` was added to the list of groups that are checked in the `AccessVulva` prerequisite. Since all items in the `ItemButt` group use the `AccessVulva` prerequisite, this means that any item in `ItemPelvis` (i.e. chastity belts) that block `ItemVulva` (which is unsurprisingly pretty common among chastity belts) also prevent `ItemButt` items from being used, _even_ if they have an "Closed Back" option which isn't active.

This PR addresses this by creating a new `AccessButt` prerequisite, which behaves in the same way as the `AccessVulva` prerequisite, except it checks the `ItemButt` group instead of `ItemVulva`. This new prerequisite has been added to all items in the `ItemButt` group to restore proper chastity belt functionality.

## Steps to reproduce

1. Equip a chastity belt with a "Closed Back" option (e.g. Metal Chastity Belt)
2. _Do not_ select the "Closed Back" option
3. Note that no `ItemButt` items can be used


Ben987:

PR #3065, opened 2021-11-10
Sleeveless catsuit fixes

Ada18980:
Moves some pixels around.


Ben987:
There's more stuff than only pixels in that pull request, but it's probably pull requests that we've already merged.  Thanks!

PR #3066, opened 2021-11-10
Catsuit panties added to panties slot

Ada18980:
Someone asked why the pilot panties were in the panties slot but not catsuit panties, so this fixes the conundrum.

Not necessarily a beta fix, but came from beta feedback.


Ben987:
This one creates a conflict, and there's other stuff in that pull request.  Can you double-check?  Thanks.

PR #3067, opened 2021-11-10
Increase difficulty of Irish 8 Cuffs

Ada18980:
Makes the Irish 8 Ankle Cuffs much more difficult to escape since they are steel.


Ben987:
Sure!  Thanks :)

PR #3068, opened 2021-11-10
Fix HeavyLatexCorset after being broken by @Shiranui-Izayoi

Jomshir98:
The heavy Latex Corset was first added to Corset slot without permission of item creator and after being asked to revert said change failed to revert it properly, causing the item to become broken, clearly not testing the revert at all.

Broken state

This PR fixes the revert so the item works like before.

( cc @EmilyR42 )
Relevant PRs: #3047, #3053


Ben987:
Nice!  Thanks for that fix :)

PR #3069, opened 2021-11-10
Beta Fix: Prevents Kigurumi Mask from hiding `ItemHead` when in latex mode

elliesec:
Fixes an oversight in the changes from #3054, which added `Hide: ["ItemHead"]` to the base Kigurumi Mask asset, which meant that it would hide blindfolds when in latex mode as well, which was not the previous behaviour. This should retain the previous behaviour whilst addressing the clipping issues when the mask is not in latex mode.


Ben987:
Absolutely.  Thanks for that one :)

PR #3070, opened 2021-11-11
Fix timer lock duration with futuristic collar

Ada18980:
Fixed a bug where the max time allowed for a lock from the futuristic collar was interpreted as 60x the actual allowed amount.


Ben987:
Absolutely.  Thanks!

PR #3072, opened 2021-11-11
Feature: Image compression script

elliesec:
## Summary

> ℹ️ I recommend merging this PR _after_ the R74 release and hotfix period

This adds a script designed to run image compression tools over all images in the game (specifically, just images under the `BondageClub` directory). This PR only contains the script itself (plus changes to the `package.json`, `package-lock.json` and `.gitignore` files required for it to work) - no game files have been modified as part of this PR and this will have no direct effect on the game itself.

Running this script over all assets resulted in an average decrease in file sizes of around 15% (and almost 18% across PNG files), with total space savings of about 100 MB (see image below), which should yield notable improvements in load times and memory consumption over the course of a long browser session.

I will raise a separate PR which actually modifies assets so that it can easily rolled back if needed.

unknown

## Things to note:

* Importantly, as assets are compressed in place, all asset compression is **lossless**, so there will be no change to asset quality, and running the script multiple times will not cause assets to deteriorate over time
* The script performs compression on PNG, JPG and SVG files
* All images under the `BondageClub/Tools` directory are ignored (notably this means anything in the `node_modules` directory)
* The 3DCG pose files in `BondageClub/Assets/Female3DCG/BodyUpper/3DCGPose` have been ignored, as they appear to contain metadata chunks specific to 3DCG, so minifying them might not be a good idea (and is not really necessary anyway) - if there are any other files that might contain important metadata, let me know, and I can exclude them as well, although everything else appears to be fairly straightforward as far as I can see.
* As there are ~20,000 image files in the `BondageClub` directory, the script performs minification in batches, as attempting to minify all files at once unsurprisingly resulted in memory issues


Ben987:
Time to test that compression!  Cheers :)

PR #3073, opened 2021-11-11
Feature: Image compression first pass

elliesec:
## Summary

> ⚠️ This PR depends upon and includes #3072 and should be merged after it

This PR runs the image compression script from #3072, but only includes changes to files in the `BondageClub/Backgrounds` and `BondageClub/Icons` directories. I chose those directories as a first proof of concept pass because they don't contain too many files, are ubiquitous in the game, and also cover all three file types (PNG, JPG and SVG).

Once we've verified that there are no issues in this initial sample, I'll raise a PR for everything else, which is likely to be very large.


Ben987:
Some of the images are getting bigger.  Could you put a validation to skip replacing the image if it would result in a bigger file?  Thanks.

elliesec:
@Ben987 - seems this was largely a quirk of the minification for JPEGs, which by their nature don't benefit quite as well from lossless minification as the other file types. I've added a filter to the script to stop it from touching files if it would increase their size, and those files have now been omitted from the PR.

Ben987:
Awesome!  We'll see if players can notice a difference in the next release.  Thanks :)

PR #3076, opened 2021-11-12
Add some kneeling spread variants to the restraints

zorgjbe:
This fixes the leg cuffs (leather, ornate and futuristic) definitions to allow spreading in detached chain mode. Note that, though the futuristic cuffs work as intended, they're marked as hidden for now.


Ben987:
Yes!  Thanks for that one :)

PR #3078, opened 2021-11-13
Beta fix: `ItemBoots` being displayed in the `KneelingSpread` pose

elliesec:
This fixes an issue where items from the `ItemBoots` group are visible in the `KneelingSpread` pose despite not supporting it. This was previously not an issue since the Sybian hid that group, but has become apparent now that the pose is available through the pose menu. For now, I've hidden the `ItemBoots` group in that pose, as I don't believe that we have any assets for the pose. If/when @zorgjbe or someone else decides to work on pose support for the `Shoes`/`ItemBoots` group, this blanket hide can be removed and assets can be more selectively hidden with `HideForPose`.


zorgjbe:
Thanks @elliesec! I've fixed ItemBoots in #3079 (had been working on those).

elliesec:
Closing in favour of #3079.

PR #3079, opened 2021-11-13
Do ItemBoots from the kneeling spread pose

zorgjbe:
Hide all boots that don't show for the pose, and add assets for the two thigh-high boots that do.

Supersedes #3078


Ben987:
Yes!  Thanks again :)

PR #3080, opened 2021-11-13
Beta fix: Fixes position of large Sleeveless Catsuit in `BackBoxtie` pose

elliesec:
The asset for the large Sleeveless Catsuit in the `BackBoxtie` pose was one pixel too high, causing it to clip with characters' chins. This repositions the asset and cleans up a few stray pixels.


Ben987:
Sure!  Thanks for that one :)

PR #3081, opened 2021-11-14
Feature: Vibrating item archetype

elliesec:
## Summary

This adds a vibrating item archetype to extended item framework, allowing items currently using the `VibratorMode.js` script to be configured entirely through `Female3DCGExtended.js` without the need for a separate script file. This PR also migrates any such items in the `ItemVulva` group to the archetype system. I'll raise follow up PRs to deal with different groups.

I also have several enhancements & refactors to the existing `VibratorMode` script that I'd like to make, which I'll raise later in separate PRs. One of these is to reinstate the potential for vibrators to have their own custom audio, which was something that got lost/forgotten when they were migrated over to the current `VibratorMode` script, as well as improving the typings for the file and related type definitions.

This change migrates the following items and removes their script files:
* Vibrating Dildo
* Clitoral Stimulator
* Vibrating Egg
* Vibrating Latex Panties
* Wand Belt
* Full LockDown Latex Suit Wand
* Hemp Rope Wand Belt
* Wired Vibrating Egg

I've tested each affected item and verified that they each work as before.


Ben987:
This is huge!  But we have plenty of time to test it before the next release.  Thanks :)

PR #3082, opened 2021-11-14
Player List Update Changes

Nina-1474:
When a player ghosts someone else using the chat command (i.e. by typing /ghostadd 6969), it was not adding the number to their blacklist like the dialog option does, meaning the ghosted character could still silently interact with the player.
This has been corrected by reworking the dialog options and commands to use a common function.

An array has been defined to easier manage when one change to a list should trigger another. This now covers a few unlikely-but-possible scenarios where odd combinations could arise. For example a number could have been added to both the whitelist and ghostlist but now ghosting will automatically de-whitelist. 
Similar triggers could be set up for the friend list but since that was never linked to other lists before, I've left it out for now.


Ben987:

PR #3086, opened 2021-11-16
Bugfix: Techno Collar default intensity

elliesec:
This fixes an error introduced by #2984 where the Techno Collar's intensity could get set to -1, which is not a valid shock unit intensity, resulting in the message `(MISSING PLAYER DIALOG: TriggerShock-1)` being printed to chat. This PR sets it to 0 instead, and also changes the null-check to the more standard `== null` rather than `== undefined`.

This has always been a bit of a stopgap fix, and at some point we'll need to look a bit deeper into it (and a similar issue with the Futuristic Crate) and identify a more appropriate solution, but this should suffice for now.


Ben987:
Absolutely!  Thanks :)

PR #3087, opened 2021-11-16
Bugfix: Correctly sets pose on the Metal Leg Spreader

elliesec:
Not 100% whether this is a beta error or not - haven't tested it on R73. Adds a `SetPose` to the Metal Leg Spreader. Because this is a typed item, the `null` type should match the default asset properties, otherwise the type won't be applied correctly when first equipping the item. In this case, the base asset is missing a `SetPose` entry.

## Steps to reproduce

1. Manually set a leg pose other than the default legs open pose via the pose menu (i.e. legs closed, kneeling spread or kneeling)
2. Equip the Metal Leg Spreader
3. Note that the character's pose does not change
4. After selecting a type from the extended item menu, the pose works correctly


Ben987:
Sure!  Thanks again :)

PR #3088, opened 2021-11-16
Beta fix: Only applies the `Chaste` effect on the Futuristic Chastity Belt when the front is closed

elliesec:
This fixes an issue with the Futuristic Chastity Belt where penetration/masturbation/vulva-related activities are unavailable even when the front is open. The PR moves the `Chaste` effect off of the base asset and applies it to the closed option for the `Front` module.


Ben987:
Yes!  Thanks for that one :)

PR #3090, opened 2021-11-17
Beta fix: Arousal preferences for mirrored groups

elliesec:
## Summary

It's been noted ([see Discord report](https://discord.com/channels/554377975714414605/909840420123582464/910318439166525481) and image below) that the names for groups with mirrored activities don't show up properly in the arousal preferences. However, rather than add dialog for those groups, I've modified the preference behaviour so that a mirrored group also mirrors arousal factor and orgasm settings, as I felt this was the most sensible approach here (it doesn't make much sense for `ItemMouth2` and `ItemMouth3` to have separate arousal settings from `ItemMouth`, or similarly for `ItemNeck`, `ItemNeckAccessories` and `ItemNeckRestraints`).

This PR edits `PreferenceSubscreenArousalRun` and `PreferenceSubscreenArousalClick` so that these mirrored zones are no longer displayed or interactable, and modifies `PreferenceGetZoneFactor` and `PreferenceGetZoneOrgasm` so that mirrored zones copy the settings of the zone they're mirroring.

image


Ben987:
Right on time for the full release.

PR #3092, opened 2021-11-17
Fix training belt dialogue

Ada18980:
Re-adds FuturisticChastityBeltSetGeneric after it was mistakenly deleted


Ben987:
Thanks!  Right on time :)

PR #3093, opened 2021-11-19
Adding BlindHeavy back onto Ventless type for Locker

Nina-1474:
When the normal and ventless lockers were merged into one item in #2431, the ventless version lost the BlindHeavy effect. This has been added into the type.


Ben987:
Sure!  Thanks for that one :)

PR #3094, opened 2021-11-19
Changed how gagged kisses work

Manilla32:
Allows two gagged people to do the "gagged kiss" action on the mouth.


Ben987:
Absolutely.  Thanks :)

PR #3096, opened 2021-11-20
New Item: Empty and filled glass

JulesPapillon:
Drinking glasses added for RP purposes.

Yuki_glass_empty
Yuki_glass_whiskey


Ben987:
Great idea!  Maybe one day we could change the maid serving drink game to automatically add these items.  Cheers :)

PR #3097, opened 2021-11-20
Opaque Catsuit Remove Hide: ["ItemNipplesPiercings"]

Shiranui-Izayoi:
Removed the `Hide: ["ItemNipplesPiercings"]` from the Catsuit, Seamless Catsuit and Pilot Suit.

These were accidentally readded when `Hide: ["BodyUpper"]`  was readded to fix the bug. They don't need to be there as all items in that slot are already hidden by default due to layers.

This also brings it in line with the See-Through suits and the newly added SleevelessCatsuit.


elliesec:
I actually reverted that hide entry for a reason: it makes some sense for the see-through suits not to hide the group because, well, they're see-through, whereas logically it doesn't make much sense for piercings to be rendered on top of an opaque suit, because that would imply that the piercing is _through_ the suit, which doesn't make much sense in my eyes. I think this one should probably be up to @gatetrek and/or @Ben987 to make the decision on.

Shiranui-Izayoi:
> I actually reverted that hide entry for a reason: it makes some sense for the see-through suits not to hide the group because, well, they're see-through, whereas logically it doesn't make much sense for piercings to be rendered on top of an opaque suit, because that would imply that the piercing is _through_ the suit, which doesn't make much sense in my eyes. I think this one should probably be up to @gatetrek and/or @Ben987 to make the decision on.

Actually IIRC the piercings are not rended on top of the suits by default. Since as I recall the suit is on a higher layer than the nipple piercings. Meaning hiding them wouldn't matter. As they are already hidden by layer.

elliesec:
Sorry, slight correction: I hid it because it was originally hidden, and as no items in that group currently render above suits, there's no reason to draw that group: it's just another run through unnecessary run through the render loop each time.

Shiranui-Izayoi:
Sorry wrong buttons.

Ben987:
Sure, showing these assets make sense.  Thanks!

PR #3098, opened 2021-11-20
Fix TargetMouthOpen prerequisite check

zorgjbe:
The condition wasn't rewritten correctly, resulting in not being able to use a penetration item on an ungagged mouth.


Ben987:
Sure!  Thanks for that fix :)

PR #3099, opened 2021-11-22
Bugfix: `TapedHands` 404s on the Sleeveless Catsuit

elliesec:
This fixes 404 errors and missing assets on the Sleeveless Catsuit. I removed `TapedHands` from the `AllowPose` entries on the suit, and replaced the existing default assets with hand cutouts with assets without hand cutouts, as the cutouts are no longer necessary since hands were split apart from `BodyUpper`.


Ben987:

PR #3100, opened 2021-11-24
Bugfix: Lock validation fixes

elliesec:
## Summary

This fixes a couple of minor issues with the validation around locks:
* The original intention was to always allow owners to remove lover locks, but not to add them if the associated lover rule was active. However, this functionality wasn't working, as the `remove` parameter was never passed into the `ValidationIsLockChangePermitted` function
* Successfully removing a log would trigger a validation warning message in the console and a pointless rollback (which didn't actually do anything) due to `ValidationResolveLockModification` returning `undefined` if none of the conditions for the final `if`/`else if` block were met. This PR adds a fallback return value to ensure that such changes are now considered valid

### Issue 1

1. Equip any lockable item
2. Ensure the lover rule for preventing owners from using lover locks is active
3. As the character's owner, remove or unlock the item
4. Note that the change is caught by validation and rolled back

### Issue 2
1. Equip any lockable item
2. Lock the item with any lock
3. As another player, remove the lock
4. Note that in the locked player's console, there is a validation warning, despite the change being valid


Ben987:
Sure!  Thanks for that fix :)

PR #3101, opened 2021-11-25
Bugfix: Adds lover rules to the rules menu

elliesec:
## Summary

When lover rules were introduced, they were never added to the active rules menu, leading players to be confused over whether or not lover rules were currently active on them. This change renames the owner rules menu (internal naming only) to the rules menu, and adds the two lover rules to the menu. I also slightly reworked the spacing and positioning of the text, and enabled wrapping for longer rules, as some rules could go off the edge of the screen. There are also a few cosmetic improvements I'd like to make to the menu, but that would require more extensive changes, so I'll save that for another PR.


Ben987:
Makes perfect sense.  Thanks for that one :)

PR #3102, opened 2021-11-25
Chat Search - Filter out rooms using keywords

Nina-1474:
When the "Filter Rooms" button on the chatroom search screen is clicked, the bottom-left text input now has the label "Filter out terms". Instead of its usual function of narrowing the listed rooms to rooms where the name contains the entered text, the user can enter a list of words separated by commas. Any rooms with any of these terms in the room name will be automatically hidden from the normal view, the same way they would if the player had toggled their visibility off manually. For example entering "afk" will hide any rooms with "afk" in the name. This comma-separated list is saved to the account so it's automatically applied against future room queries.


Ben987:
That's a good idea!  I'm curious to test it.  Thanks!

PR #3103, opened 2021-11-25
Typescript fixes

zorgjbe:
This is the result of running `tsc` against the code base. Reported error count go from 598 down to 414.

Most of the changes are either whitespace, comments (ie. missing `@param`, copy-paste, etc.), changes to the TypeScript definitions or variable initialization types, so mostly non-functional changes. A nice side effect is that all NPC-related functions are a little more type-safe, since they now take or return an "explicit" `NPCCharacter` type instead of a `Character`.

The only real code-related change is c53dc01f2ac3b66e9e2e1c910dd25e63fbb5dc0a, since I'm merging Pandora archetype management with the base game ones.

Most of those changes should be freestanding, so can be cherry-picked out, and I've tested that the infiltration minigame still works fine (I got an USB key 😉).


Ben987:
These automatic changes makes me nervous, we had many bugs because of them before.  

Can you split them up into small changes so we can roll them back if they cause problems?

Thanks!

zorgjbe:
That's not automated at all. I went through the list of errors and fixed the ones I could (or tried to, at least). CI being red is because I messed up the syntax in the TS file, and those would show up, but in a sea of 400+ errors it's just noise.

BondageClub/Screens/Room/MaidQuarters/MaidQuarters.js
```diff
@@ -430,8 +430,8 @@ function MaidQuartersOnlineDrinkStart() {
 
 /**
  * On online player picks a drink from the plyers tray. She only gets credited, if it was a new customer
- * @param {string} MemberNumber - The member ID of the customer
- * @param {string} DrinkValue - The value of the picked drink
+ * @param {number} MemberNumber - The member ID of the customer
+ * @param {number} DrinkValue - The value of the picked drink
```
Jomshir98, line 434: Because how `CommonDynamicFunctionParams` is made, all arguments to dynamic functions are always `string`.

BondageClub/Screens/Room/Shibari/Shibari.js
```diff
@@ -9,7 +9,8 @@ var ShibariSubCommentDone = false;
 var ShibariDomCommentDone = false;
 var ShibariSurrenderDone = false;
 var ShibariSpankDone = false;
-var ShibariTeacherReleaseTimer = false;
+/** @type number */
```
Jomshir98, line 12:
```suggestion
/** @type {number} */
```

BondageClub/Scripts/Common.js
```diff
@@ -9,7 +9,7 @@ var CurrentModule;
 var CurrentScreen;
 /** @type {ScreenFunctions} */
 var CurrentScreenFunctions;
-/** @type {Character|null} */
+/** @type {Character|NPCCharacter|null} */
```
Jomshir98, line 12: `NPCCharacter` won't do anything here, as it already extends `Character`.
This means, that because `null` is ignored (we are not in strict mode), the largest intersection of all types here is still `Character`.
```suggestion
/** @type {Character|null} */
```

BondageClub/Scripts/Dialog.js
```diff
@@ -117,32 +117,32 @@ var DialogSelfMenuOptions = [
 /**
  * Compares the player's reputation with a given value
  * @param {string} RepType - The name of the reputation to check
- * @param {string} Value - The value to compare
+ * @param {number} Value - The value to compare
```
Jomshir98, line 120: `CommonDynamicFunctionParams` (see above)
```suggestion
 * @param {string} Value - The value to compare
```

BondageClub/Scripts/Dialog.js
```diff
@@ -117,32 +117,32 @@ var DialogSelfMenuOptions = [
...
  * @returns {boolean} - Returns TRUE if a specific reputation type is less or equal than a given value
  */
 function DialogReputationLess(RepType, Value) { return (ReputationGet(RepType) <= Value); }
 
 /**
  * Compares the player's reputation with a given value
  * @param {string} RepType - The name of the reputation to check
- * @param {string} Value - The value to compare
+ * @param {number|string} Value - The value to compare
```
Jomshir98, line 128: `CommonDynamicFunctionParams` (see above)
```suggestion
 * @param {string} Value - The value to compare
```

BondageClub/Scripts/Dialog.js
```diff
@@ -117,32 +117,32 @@ var DialogSelfMenuOptions = [
...
  * @returns {boolean} - Returns TRUE if a specific reputation type is greater or equal than a given value
  */
 function DialogReputationGreater(RepType, Value) { return (ReputationGet(RepType) >= Value); }
 
 /**
  * Compares the player's money with a given amount
- * @param {string} Amount - The amount of money that must be met
+ * @param {number|string} Amount - The amount of money that must be met
```
Jomshir98, line 135: `CommonDynamicFunctionParams` (see above)
```suggestion
 * @param {string} Amount - The amount of money that must be met
```

BondageClub/Scripts/Dialog.js
```diff
@@ -117,32 +117,32 @@ var DialogSelfMenuOptions = [
...
  * @returns {boolean} - Returns TRUE if the player has enough money
  */
 function DialogMoneyGreater(Amount) { return (parseInt(Player.Money) >= parseInt(Amount)); }
 
 /**
  * Changes a given player's account by a given amount
- * @param {string} Amount - The amount that should be charged or added to the player's account
+ * @param {number|string} Amount - The amount that should be charged or added to the player's account
```
Jomshir98, line 142: `CommonDynamicFunctionParams` (see above)
```suggestion
 * @param {string} Amount - The amount that should be charged or added to the player's account
```

BondageClub/Scripts/Timer.js
```diff
@@ -117,7 +117,7 @@ function TimerPrivateOwnerBeep() {
 			ServerBeep.Message = DialogFindPlayer("BeepFromOwner");
 			LogAdd("OwnerBeepActive", "PrivateRoom");
 			LogAdd("OwnerBeepTimer", "PrivateRoom", CurrentTime + 120000);
-			FriendListBeepLog.push({ MemberName: Player.Owner, ChatRoomName: DialogFindPlayer("YourRoom"), Sent: false, Time: new Date() });
+			FriendListBeepLog.push({ MemberNumber: 0, MemberName: Player.Owner, ChatRoomName: DialogFindPlayer("YourRoom"), Private: false, Sent: false, Time: new Date() });
```
Jomshir98, line 120: `MemberNumber` should be kept as `null`ish value (`undefined`/`null`). There are checks elsewhere in code that rely on this (see ex. `FriendListLoadFriendList`)
```suggestion
			FriendListBeepLog.push({ MemberNumber: null, MemberName: Player.Owner, ChatRoomName: DialogFindPlayer("YourRoom"), Private: false, Sent: false, Time: new Date() });
```

BondageClub/Scripts/Typedef.d.ts
```diff
@@ -123,6 +123,73 @@ declare function LoseFocus(event: MouseEvent): void;
...
 
+//#region Chat
+
+interface ChatRoom {
+	Name: string;
+	Description: string;
+	Admin: number[];
+	Ban: number[];
+	Limit: number;
+	Game: any;
```
Jomshir98, line 134: Assured by server code
```suggestion
	Game: string;
```

BondageClub/Scripts/Typedef.d.ts
```diff
@@ -123,6 +123,73 @@ declare function LoseFocus(event: MouseEvent): void;
...
+	Background: string;
+	Private: boolean;
+	Locked: boolean;
+	BlockCategory: "Photos";
```
Jomshir98, line 138: This looks wrong
```suggestion
	BlockCategory: string[];
```

BondageClub/Scripts/Typedef.d.ts
```diff
@@ -123,6 +123,73 @@ declare function LoseFocus(event: MouseEvent): void;
...
+	Character?: Character[];
```
Jomshir98, line 139: The type here is actually character data from server, not created character. I don't think we have types for that
```suggestion
	Character?: any[];
```

BondageClub/Scripts/Typedef.d.ts
```diff
@@ -123,6 +123,73 @@ declare function LoseFocus(event: MouseEvent): void;
...
+}
+
+interface IChatRoomSyncMessage extends IChatRoomSyncBasic, ChatRoom {}
+
+//#endregion
+
+//#region FriendList
+
+interface IFriendListBeepLogMessage {
+	MemberNumber: number;
```
Jomshir98, line 181: `null` is valid, if from NPC
```suggestion
	MemberNumber: number | null;
```

BondageClub/Scripts/Typedef.d.ts
```diff
@@ -618,14 +674,74 @@ interface Character {
...
+	TrainingCountLow?: number;
+	TrainingCountHigh?: number;
+	TrainingCountPerfect?: number;
+}
+
+/** Kinky Dungeon */
+interface Character {
+	KinkyDungeonKeybindings?: any;
+	KinkyDungeonExploredLore: any[];
 }
```
Jomshir98, line 745: I personally think it is better to keep `Character` as one interface and only use comments as separators inside, but just opinion/personal style.
Reasoning: It makes it easier to see what actually is definition of `Character` if one looks for it, also avoiding a property being defined twice with same type, but different meaning. (all those properties shouldn't be randomly assigned to `Character` in the first place, but it is what it is)

Jomshir98:
> These automatic changes makes me nervous, we had many bugs because of them before.
> 
> Can you split them up into small changes so we can roll them back if they cause problems?
> 
> Thanks!

@Ben987 Could you please point me to a specific bug caused by automated change? I'm curious to learn. Thank you!

Ben987:
Please split this one into small pull requests guys.   It will be much easier to review and debug if we need to roll back stuff.  We've had bad experiences with pull requests that target too many screens at once in the past, I don't have an example on hand though.  Thanks!

BondageClub/Scripts/Typedef.d.ts
zorgjbe: Yeah, I'm not sure what's best. I'm going along those things being type extensions from the base `Character` type, so it makes more sense to have them as separate interfaces. It just feels cleaner than just dumping all this under the "true" `Character` interface.

BondageClub/Scripts/Common.js
zorgjbe: Huh, but you can assign a more specific `NPCCharacter` instance in here, and then all of its subclass-specific property accesses will warn because the "compiler" think its a `Character` and doesn't see them?

zorgjbe:
Rebased. I've yanked anything that would cause major changes, so this is mostly non-functional/TypeScript-related changes only.

Ben987:
Thanks for the cleanup, I'll test it soon, it shouldn't impact the game.  Cheers :)

PR #3104, opened 2021-11-26
Make the Frogtie Strap spreadable

zorgjbe:
As it says on the tin 😉.


Ben987:
Yes, yes, yes.  Thanks!

PR #3106, opened 2021-11-27
Added new Item: Pet Nose (Mask)

JulesPapillon:
The mask can be completely hidden and blocked via "Item Visibility".

NoMouth
cat1
cat2
dog1
dog2
dog3


Ben987:
Sure!  And if there's too many complaints we can always add a new filter like the ABDL filter.  Thanks!

PR #3110, opened 2021-12-01
Bugfix: `ActivityCheckPrerequisite` logic fixes

elliesec:
## Summary

As part of #2995, a large portion of the `ActivityDialogBuild` function was refactored out into a `switch` statement in the new `ActivityCheckPrerequisite` function. As it turns out, a few of the pieces of logic for prerequisite checks weren't quite translated correctly, which resulted in a change in behaviour, notably for the `UseArms` and `CantUseArms` prerequisites.

This PR corrects these logic issues, and also makes the checking for the `SelfOnly` prerequisite a little more robust, since the function now operates on generic `acting` and `acted` characters, rather than the `acting` character always being the player, so keeping the `C.ID === 0` would likely have resulted in bugs further down the line if/when someone exercised the new genericism of the function.

I've tested all behaviour affected by this this PR against the behaviour before #2995 was merged and verified consistency.

## Bugs fixed

* The `Struggle` activity (or any other activity requiring the `CantUseArms` prerequisite) is currently available when _only_ wearing mittens
* The `Pet` activity (or any other activity requiring the `UseArms` prerequisite) is currently unavailable when only wearing mittens


Ben987:
Sure!  Thanks for these fixes.  Cheers :)

PR #3111, opened 2021-12-01
Feature: `ZoneAccessible` prerequisite expansion

elliesec:
## Summary

This is something that occurred to me when looking at #3110. This PR arranges asset groups into equivalence classes based on the `MirrorActivitiesFrom` property, and modifies the `ZoneAccessible` prerequisite so that the prerequisite returns `true` if _any_ of the equivalent groups satisfy the accessibility condition.

In particular, this means the following are now possible:

* The `GagKiss` activity is now available even if wearing gags in `ItemMouth2` or `ItemMouth3`, but _not_ if the mouth groups are all blocked by, say, a hood
* Neck activities (e.g. `Kiss`, `Caress` etc.) are now available when wearing neck-blocking items like the Collar Cuffs or gag version of the corset collars, without needing to make explicit exceptions using `AllowActivityOn`

Notably, this does also mean that activities like `Kiss`, `Lick`, etc. will now be available when a character has a mouth-covering gag (e.g. the corset gags) in _any_ `ItemMouth` group, but this has been a longstanding thing when wearing items in the original `ItemMouth` group anyway, and I think is something we could address in a future PR by adding an opposite prerequisite to the existing `TargetMouthBlocked` one or something along those lines.


Ben987:
Absolutely!  Thanks for these fixes :)

PR #3118, opened 2021-12-09
Fixed crash when spectating kinky dungeon

Ada18980:
Should hopefully fix a crash when spectating kinky dungeon by preventing bad data from going through.


Ben987:
Thanks a lot for that fix.

PR #3123, opened 2021-12-11
Chat Room Filter - Text limit increase

Nina-1474:
The chatroom name filtering from #3102 was still applying the room-name 20-character limit to the player's entered terms, which meant only a few words would fit. 
The textbox now allows many more characters and therefore more words as intended, while still limiting the standard room search mode to 20 characters.


Ben987:
Absolutely!  Thanks :)

PR #3124, opened 2021-12-11
Beta Fix: Missing Futuristic Training Belt chat messages

elliesec:
## Summary

Looks like #3092 didn't quite fix the issue of the missing chat messages when exiting the extended item menu for the Futuristic Training Belt. This PR fixes that, and also changes the message key to reference the training belt rather than the chastity belt. I've also removed a couple of messages from `Audio.js` that were either duplicated or not being used.


Ben987:
Nice extra fix.  Thanks again!

PR #3125, opened 2021-12-11
Beta Fix: Snowman difficulty tweaks

elliesec:
## Summary

It was pointed out that there was a bit of a discrepancy between the difficulty level of the Snowman and its self-bondage requirement. This PR brings the difficulty of the item down, and introduces a (low) self-bondage level requirement to it, which should hopefully make it a little more forgiving.


Ben987:
I agree.  Thanks :)

PR #3126, opened 2021-12-12
Medical Bed

Ayesha-678:
Adds a medical bed, always available in the asylum, but you can get it in the normal chatrooms with the restraints.


Ben987:

PR #3128, opened 2021-12-12
Thin Leather Strap Difficulty Update

Ayesha-678:
Added proper difficulties to the thin leather straps types.


Ben987:
Sure!  Let's slide it in the next beta.  Thanks :)

PR #3129, opened 2021-12-12
Better identification of players in the Suitcase's Outgoing messages

zorgjbe:
This wires the 2nd outgoing message you get when checking on someone's steal status, correctly picking up whether their suitcase's open or not. Also uses the actual player name in those, and changes it to be from the actual sender.

Note that I initially wanted the name in, and then proceeded to also switch the sender around. One issue I can think of is that if you are ghosting your target, you won't get any status message back…


Ben987:
@Ada18980 , would you like to review this one?  It modifies some of your code.  Thanks.

Ada18980:
@Ben987 This looks good, as long as it's tested. I think it's good to add a specific Type for the chat room game data.

Ben987:
Lets try it in the next release then!  Thanks you two :)

zorgjbe:
Thanks ❤️

PR #3130, opened 2021-12-12
Clamp background selection offset to avoid a crash with no search results

zorgjbe:
Crashes if you click the previous page button when searching for backgrounds with 0 results visible

Reported-by: Sidious


Ben987:
Sure!  Thanks for that fix :)

PR #3133, opened 2021-12-13
Thin Leather Strap Multi-Color + Lock assets

Ayesha-678:
Adds multi-color, and lock assets to the thin leather straps!


Ben987:
You're on fire!  Thank you very much :)

PR #3134, opened 2021-12-13
Admiral Shirt Fix

Nina-1474:
Fixing the image for the admiral top's small size shirt layer in the over-the-head pose.


Ben987:

PR #3136, opened 2021-12-14
College Skirt Split

Nina-1474:
This splits the skirt layer of the College uniform in Cloth into a separate asset in ClothClower. All instances of the asset being added/worn/checked directly in the game have been updated to account for both items.


Ben987:
Yes, yes, yes and yes!  Thanks :)

PR #3137, opened 2021-12-15
Fix conflicting pose issue between the HempRope & FrogtieStrap

zorgjbe:
The pose system seems to expect all pose arrays to be ordered the same,
as it implicitly relies on the last pose being the same no matter which
item allowed it. Change the pose ordering on the HempRope asset so it
matches the one from FrogtieStraps, squashing a bug where the
legs and frogtie would match the kneeling spread pose, but the hemp rope
would display the kneeling one.


Ben987:
Sure!  I'll slide it in beta 3.  Thanks :)

zorgjbe:
Thanks Ben!

PR #3139, opened 2021-12-16
Kinky Dungeon Balance Overhaul

Ada18980:
Gameplay Overhaul
- Removed willpower. Stamina is the primary resource now for determining whether the player gets captured
- New consequences when an enemy closes the gap and you have 0 stamina
- Stamina is overhauled. It does not regen except when sleeping. You can click a button to sleep for 20 turns to regain 20 stamina. If you are attacked then you wake up automatically, but you are blinded while sleeping so you won't be able to see enemies coming. You can cancel sleep by hitting any key.
- Mana is now a separate resource. It does not regenerate, but mana potions can restore it.
- Each shrine now has a mana pool. You can drink mana from any of the mana pools once per floor without repercussions. This restores your mana fully. Drinking more than once on the same floor (from any number of shrines), however, has a chance to backfire.

UI Changes
- Changed the stat bars visually
- Added Buttons to drink stamina/mana potions without going to the inventory
- Added restart button and confirmation screen

Content
- New Spell Point system: You can learn new spells by spending points. Find spell points in chests and on special orbs, but be wary of angering the gods
- New curse system: Items which are Cursed can not be struggled out of, but can be unlocked if you meet their condition
- Added shortcuts, on floor 3 there is a stairway to a new (unfinished) tomb area with a different tileset
- Reputation system that affects certain enemy spawn rates
- New rope themed enemies that appear more often if you annoy the rope goddess
- New spell: Ice Breath, creates icicles that deal damage
- Added a new spooky illusion Enemy
- New enemies: samurai, ninjas, mummies
- new bolt cutters weapon
- Some doors have a chance to be locked
- Added helpful (or unhelpful) adventurer ghosts that may assist you with struggling. Careful not to rely on them too much as they might enjoy seeing you bound!
- New special pressure-plate traps (that are often the only way through if you dont have a key to unlock a door)
- Enemies now have pathfinding! They should be a bit harder to evade now.
- Some enemies now hunt along defined patrols rather than at random
- Enemies now have a 2 turn delay before they start chasing you
- Added grates you can squeeze through

Other Features
- Knives now use the same random-number-of-uses mechanic as lockpicks which gives some guaranteed uses
- Picks, Knives, and Red Keys are now in the shop
- Gagged mumbling while moving around
- More flavorful text when waiting
- Made the player character more expressive based on stamina, mana and arousal
- Removed green and yellow locks (too much bloat)
- Only one shrine of each type per floor
- If you are bound with shrine-able restraints, one random restraint type will be guaranteed to have a shrine on the next level
- Removed enemy friendly-fire in most cases
- Button to toggle automatic door-closing
- Walking around with a plug inside of you will arouse you

Bugfix:
- Enemies with high attack width will now hit you properly even if you sidestep (as long as you are within range)
- Enemies that attack while moving will no longer hit you if they move at the same time that you do such that you are still in range, even if you dodged last turn's warning tiles
- Multiple enemies will no longer be spawned on the same tile
- Blue locks can no longer be picked, as all things should be
- The quick potion buttons no longer bypass being bound or gagged.

This version is probably quite imbalanced, so playtesting will be needed, which is why I'm PR'ing it now to bring attention to anyone interested. Keeping as WIP in order to fine tune the experience before pushing to the players.


Ada18980:
Okay... that should be it for now. I still have a big TODO list but I don't want to burn myself out so I'll probably leave it at this for the beta.

Maaaaaybe a new area and increased max floor to 30, but I'd need some inspiration

Ada18980:
Ready to go.

Ben987:
Wow!  So many changes to try.  Thanks a lot :)

PR #3140, opened 2021-12-17
Beta-fix: Allow `BlindHeavy` effect for `SmallLocker`

Jomshir98:
#3093 Added `BlindHeavy` effect to `Locker`, but `SmallLocker` uses same options.
This PR allows `SmallLocker` to correctly copy behaviour of `Locker`


Ben987:
Sure!  Thanks for that one :)

PR #3141, opened 2021-12-17
Allow KneelingSpread pose on the ItemNeckRestraints

zorgjbe:
This adds the needed stanza to the short chain, rope, and post.


Ben987:
I think these were meant to prevent kneeling.  I'm not 100% sure though, anyone here has the answer?

Jomshir98:
> I think these were meant to prevent kneeling. I'm not 100% sure though, anyone here has the answer?

They actually force kneeling (short rope leading to the ground).
This PR only allows sub to select whether they kneel normally (current behaviour) or kneeling with spread legs (new option).

Ben987:

zorgjbe:
Yup, thanks for the clarification @Jomshir98, and for the merge @Ben987 👍.

PR #3142, opened 2021-12-17
Fix crash in `DialogMenuButtonClick`

Jomshir98:
Fixes a possible crash in `DialogMenuButtonClick`, if you click the button to switch to activities screen twice between two frames.


Ben987:

PR #3143, opened 2021-12-17
Prevent crash if creation of notification fails

Jomshir98:
Prevent crash when creation of notification fails, for example if notification API is not available. This happens mainly in Android Chrome, which forbids creation of notifications unless using service worker.


Ben987:
Right on time for the full release.  Thanks!

PR #3144, opened 2021-12-18
Erase stray pixel from the scarf

zorgjbe:
As it says on the tin.


EmilyR42:
Thanks.

Ben987:
Sure!  Thanks for that one :)

PR #3145, opened 2021-12-19
Bugfix: Removes legs from the `KneelingSpread` pose for `Stockings1`

elliesec:
Fixes an issue with `Stockings1` where the `BodyLower` asset was accidentally included in the `KneelingSpread` asset for the large size, resulting in graphical issues when the stockings get coloured.


Ben987:
Very good.  Thanks :)

PR #3146, opened 2021-12-19
Bugfix: "behind you back" typo

elliesec:
Fixes a few typos that someone brought to my attention in the Asylum where "behind you back" is used instead of "behind your back". The large diff in `Dialog_NPC_AsylumTherapy_Patient_RU.txt` is because the original file was checked in with CRLF line endings rather than just LF - the only actual text changes in that file are on lines 143 and 169.


Ben987:

PR #3148, opened 2021-12-23
Remove most hide-specifiers from the Sybian

zorgjbe:
Now that (most) assets are working, allow them to show when sitting on the sybian; the only strange one is the teddy bear, which looks like it shouldn't have been hidden in the first place.


Ben987:
Alright!  Let's hope there's no missing assets for it.  Thanks :)

PR #3151, opened 2021-12-23
next/prev button for character dialog

fleisch11:
(i don't know, if this is a good idea or not)

adds a prev and next button to the character dialog, if there are more than 8 options.
this would allow one to easily add as many options to a single stage as one wants.

(for testing since currently the amount of options is always at most 8, i added a bunch of test options to the player.csv .
i, or someone else will remove them if this will get added)


Ben987:
Probably not needed for now.  We try to keep the dialog short so it's easy to navigate and find options.  Thanks anyway :)

PR #3152, opened 2021-12-23
Fix ServerBeep random timings

Jomshir98:
`ServerBeep` (indicator of received beep) was using server time, which can be very inconsistent and cause the beep to disappear quickly or stay on for very long time.
This PR changes all logic to use `CommonTime()` instead, which is always consistent.
Also cleans up some code around it.


Ben987:
Never seen that bug before, but thanks for fixing it :)

PR #3154, opened 2021-12-26
Fix typing of vibrator modes and intensity

Jomshir98:
Fixes the declared types to be more strict, reducing count of reported errors.


Ben987:
Not sure what it does, but if it helps you it's all good :)

PR #3155, opened 2021-12-26
Fix possible crash in `CharacterAppearanceFullRandom`

Jomshir98:
Crash can happen if there is group without valid wearable item.


Ben987:
Absolutely!  Thanks :)

PR #3159, opened 2021-12-29
Ziptie Message Fix

Nina-1474:
Fix to messages like "(Nina uses the zipties on TargetCharacter to half bind her feet.)" when changing the type for zipties on feet.


Ben987:
Good catch!  Thanks :)

PR #3161, opened 2021-12-29
Cleanup nullchecks

Jomshir98:
Remove unnecessary nullchecks following this pattern:
```js
(A != null) && (A == "something")
```
Replacing it with this:
```js
A === "something"
```
For this condition to be satisfied `A` must be `"something"`. This however means, that it can't be null, making the right side sufficient and the `!= null` check unnecessary.

( similar logic applies to `(A == null) || (A != "something")` )


Ben987:
This one should be fine but it will be hard to test all of it.  We can always fix in the beta.  Thanks!

PR #3162, opened 2021-12-29
Leg Spreader Prerequisite Fix

Nina-1474:
There are a few item combinations where arms restraints that hold the legs closed can be applied while there are leg restraints that hold the legs open. E.g. Pantyhose Fullbody or Strait Dress when wearing Wooden Cuffs or Tentacles spread.
This has been fixed by tweaking the NoFeetSpreader prerequisite to check for forced leg-open poses instead of a few specific items. This covers all the old cases, some appropriate new cases, and the rest are combinations that were already impossible.


Ben987:
Absolutely!  Thanks :)

PR #3164, opened 2021-12-30
Missing Effect Fix

Nina-1474:
Unlike many other properties, the `Effect` array in `Property` is added to the asset/group `Effect` rather than replacing it. This means an item option can't remove or replace an effect from the base asset.
Some items (e.g. HarnessPonyBits and DentalGag) worked around this by changing the asset's Effect to empty and moving those effects to the `null` type option instead. However this creates an issue where applying the item without changing the type will not apply those effects, unless the type is changed then changed back.

This fix introduces a new `OverrideAssetEffect` field on Property and when true only the property's effects will apply, not the asset/group effects. This opens up option configurations with fewer/different effects to the base asset.


Ben987:
Sure!  Thanks for that fix :)

PR #3165, opened 2021-12-30
Hair & Hat Duplicate Removal

Nina-1474:
Many of the pose-specific images for HairBack and Hat are duplicates of the default image. Their configurations have been changed to allow the duplicates to be deleted, saving a few MBs.


Ben987:
Very nice!  Thanks a lot :)

PR #3166, opened 2021-12-30
Empty FilterTerms Fix

Nina-1474:
A small fix to an error that's thrown when a newly created character searches for a room.


Ben987:
Absolutely!  Cheers :)

PR #3167, opened 2021-12-30
Room Recreate Loop Fix

Nina-1474:
When a player tries to automatically rejoin a room they've been banned from, they get stuck in a search loop.

On reconnecting, the initial search for the room returns nothing since rooms the player is banned from don't appear. The attempt to automatically recreate and join the room then fails because the original room still exists, causing the variable `ChatCreateMessage` to be set to "ResponseRoomAlreadyExist"`.
On then searching for a room or trying to exit the search screen, the room rejoin code starts a loop where a number is appended to the previous room name, but the "room already exists" variable makes it think the previous name failed and it must try a new name. This repeats until the maximum retry count of 50 is reached, resulting in a final name like "RoomName 2 3 4 5 50".

This can also happen when switching between environments, similar to the issue fixed by #2791 which is what introduced this bug. It could conceivably happen in any situation where the player can't find a room but can't recreate it either.
The fix reduces this from searching 50 times to just once (though it doesn't resolve the odd moment of the player reconnecting into the search screen instead of a recreated room).


Ada18980 (approved):

Ben987:
Yes, yes, yes!  Thanks a lot :)

PR #3170, opened 2022-01-04
Kinky Dungeon Pre-Beta fixes/changes

Ada18980:
Some fixes
- Arousal now regens while moving again (was only supposed to not regen if you have a plug)

- Chests don't appear while there are jailers
- New spell: Invisibility
- Balance changes
- Buffs to the bast clerics so they stun
- Buffs have been changed and evasion works differently (diminishing returns)
- New area: Underground jungle


Ada18980:
Here we go!

Ben987:
Wonderful!  I love the new sprites.  And if there are bugs we can always patch them in the beta versions.

PR #3171, opened 2022-01-05
Activity Null Check Fix

Nina-1474:
The default value of `Activity` on asset groups was changed from null to an empty array under #3044. The few cases where this is compared to null were then returning `true` when they used to return `false`. In particular it makes all zones appear in the Arousal preferences screen even if they have no activities.
These cases have been changed to check for a non-zero length to restore the old behaviour.


Ben987:
Sure!  Thanks for that fix :)

PR #3173, opened 2022-01-07
Kinky Dungeon Saving

Ada18980:
New experimental feature I would like to test...


Ben987:
How much space will it roughly take in the database?  If it's a lot, is there any way to compress the data?  Thanks :)

Ada18980:
> How much space will it roughly take in the database? If it's a lot, is there any way to compress the data? Thanks :)

As of the latest update, about 1kb uncompressed with full restraints. For comparison my player's appearance is about 1.4kb with no restraints on.
In previous versions it was 10kb uncompressed, which was an order of magnitude higher than player appearance.

Let me know if you are not comfortable with this and I can disable the account storage. Instead I will have a key-based export where the player can back up their player as a string.

Ben987:
Could you save it in the browser local storage instead of the database?  Would that be lots of work on your side or would it create bugs?  I don't want you to have to rewrite the full thing.

Ada18980:
Done, it saves in the localstorage now.

That was easier than I thought it would be.

Ben987:
That was quick!  Thanks a lot, I'm curious to try it :)

PR #3180, opened 2022-01-12
Cleaning and Serving self and dust sprites

karame1:
- Update the dust sprites in the cleaning minigames for a more consistent look between dust levels
- Allow to run the cleaning and serving games with self-bondage

When correctly bound while wearing a maid outfit (same pattern as when the maid ties you up), a new dialog option is available in the Maid Quarters to start the minigame. So you can choose your outfit including the bindings and increase your submissive rate by the way.

Cleaning requires self-bondage (because you can't reach the Maid Quarters while bound), with feet, legs, arms restrained and the duster gag. The player must not be blindfolded.

Serving can be achieved either by someone binding you or through self-bondage. It requires bound legs and arms, a gag and a service trail (no matter the variant). The player must be able to move and not being blindfolded.

In both cases the maid releases you at the end of the job, except for locks.

You can't ask the maid "what's going on here" while wearing a maid uniform, because you should know at that time and that would add too many dialog options.


Ben987:
Thanks for that one :)  I'll review it in the next release though, we are debugging the current beta.

Ben987:
Time to test this one!  Thanks a lot.  I'll probably reduce the reputation change though, -6 can be a huge hit on a reputation.  Cheers :)

PR #3188, opened 2022-01-15
Pet Bowl Fixed Position

Nina-1474:
Configuring the pet bowl to be a fixed position asset so that it always appears on the "ground", including when suspended.


Jomshir98 (approved):

Ben987:
Thanks for that fix :)

PR #3190, opened 2022-01-17
Remove app-level HTTPS redirector

zorgjbe:
This removes the Javascript-based HTTPS redirector, since it breaks developer configuration and is better handled at the webserver level.


zorgjbe:
No, we don't need that, and at least someone has a broken installation because of the `location` shenanigans. If you want to customize the `ServerURL` parameter, move it in a not git-included `config.js` file and have the "production" and "test" environment be different. The application shouldn't care where it's hosted, or if its a throwaway test.

PR #3191, opened 2022-01-22
Low-level activity arousal helpers

zorgjbe:
Cleanup arousal-related checks by adding helper functions to accessing Arousal settings and runtime values.

Split out from #3036.


Ben987:
This touches a lot of code from many contributors.  Does it change anything for players?  Does it have any in-game impact?

Also, this will create bugs with the new GGTS system since it can now override the arousal preferences.  Thanks.

zorgjbe:
It's only a cleanup PR, aimed at having a better interface between the arousal system and the rest of the codebase; that's why it touches contributed assets, changing them to the helper functions instead of the chains of checks it needed before.

I've checked that the arousal still behaves as it did, so manual clicking works, orgasm resist and surrender work as well.

BondageClub/Screens/Character/Preference/Preference.js
```diff
@@ -63,24 +63,196 @@ var PreferenceGraphicsWebGLOptions = null;
...
+
+/**
+ * Check a character's arousal mode against a list.
+ * @param {Character|Player} character - The character or player to read arousal from
+ * @param {Array<ArousalMeterMode>} modes - The arousal modes
+ * @returns {boolean|null}
+ */
+function PreferenceArousalIsInMode(character, modes) {
+	// Special case for the Asylum GGTS rooms
+	if (CurrentModule === "Online" && CurrentScreen === "ChatRoom" && ChatRoomGame === "GGTS" && ChatRoomSpace === "Asylum" && AsylumGGTSGetLevel(character) >= 4) {
```
zorgjbe, line 120: And the GGTS override is still here, copy/pasted from the previous function.

PR #3192, opened 2022-01-24
bugfix: add missing chat settings to PreferenceInitPlayer()

Sidiousious:
ColorTheme, EnterLeave, and MemberNumbers were missing from the chatsettings object for players who never modified these settings. For the latter two this results in the default behavior not getting restored after sensory depravation hides the enter/leave messages and member numbers on messages.

Steps to repro the original bug:
- Create a new account
- Go to a chatroom
- Equip sensory depravation hood (or equivalent full sensory depravation binds)
- Notice how member numbers next to messages disappear, players joining/leaving can no longer be seen
- Remove the sensory depravation hood (or equivalent binds)

Expected: Member numbers and join/leave messages would be visible again

Actual: Member numbers and join/leave messages remain hidden.

This PR fixes the above by adding the default settings to the chatsettings object so that ChatRoomRefreshChatSettings() is aware of their existence and will correctly restore them back to their defaults after sensory depravation.

Additionally sorts the settings alphabetically so they are easier to read and maintain.


Ben987:
Sure!  Thanks for the fix and the sorting :)

PR #3194, opened 2022-01-28
bugfix: dental gag in open mode has gag effect when locked

Sidiousious:
Locking the Dental gag adds an Effect array of one item, "Lock", which overrides the asset's default Effect array removing the gag effect. Adding the asset's default gag effect to the default type in extended item configuration fixes this behavior, and the character has gagged speech even while locked.


Ben987:

PR #3195, opened 2022-01-28
Color layers for maid uniforms

karame1:
Separate layers for MaidUniform1 (regular) and MaidUniform2 (exposed), with Dress, Frills and Apron.

The dresses were greyscaled to allow bright and dark colors, while keeping the default dark color.

Separate layers for FrilledSleepMask (Frills and Mask) and MaidCollar (Collar, Frills and Brooch)


Ben987:
Wow!  This is huge!  Thank you so much, I'm sure players will love it :)

PR #3193, opened 2022-01-27
Allow changing your own clothes from the dialog

zorgjbe:
Adds an option to access your own wardrobe from the dialog, to make it consistent with changing other people's wardrobe.

PR #3196, opened 2022-01-29
More online menu changes

zorgjbe:
So, I know #3193 just went in, but looking more closely, I found a few missing things w.r.t to wardrobe access and also wanted both the online menu and the player's own menu to be more consistent, for muscle-memory reasons.

This does the following:
- clarifies the `CanChange()` function behavior, renaming it to `CanChangeOwnClothes()` globally and merges it in a new `CanChangesClothesOn(C: Character)`, making it the game-wide handler for when you can touch someone's (or your own) clothes. That allows cleaning up a few conditions across the code base.
- changes the player menu again by making it match the online menu, adding "View profile" and "Character actions".
- most character-related actions got moved to the new submenu, so changing skills, calling the maid, safewording, and taking a picture are now in there. The "guideline" I used would be that anything impacted by the player's state (eg. skills) or restraints, and not needing quick access (so the "Prevent from leaving" action, though that's from the online menu) gets to live down there.
- tweaked the new cloth-changeable check so it's available in both chatrooms and SP rooms which have a wardrobe button.

I did move @Ada18980's Kinky Dungeon options here as well for consistency, I hope you don't mind.
@Jomshir98 The method name change breaks one of BCX's patches, so this is a heads-up.

Here's a screenshot of the new root:
top-level

And the new Character actions submenu:
char-actions


Ben987:
I'm curious to test this one.  Have you done a quick tour of most single player screens to make sure it doesn't cause any bug or crashes?  Thanks :)

zorgjbe:
I did try all rooms — though didn't test all options each time — when looking for places where changing clothes would be appropriate.

Ben987:
This one has a few conflicts with the latest changes I've did for GGTS.  Could you resolve the conflicts and tests again?  I'll try to test it soon on my side.  Cheers :)

zorgjbe:
Updated.

To be honest, I've been mightily confused at the new "Help and information" menu, since it clashed with my previous changes, so I took the liberty to remove it.

Was that menu added to prevent safewording and maid-calling from GGTS rooms? Maid-calling explicitly requires no game to be running (as per `DialogCanCallMaids()`), so I've added a new `AsylumGGTSIsEnabled()` and wired that into the safeword machinery.

Ben987:
The help and information section is made to make sure we don't overflow the first screen and go beyond 8 options.  We can use it later to add more stuff to help the current player.

zorgjbe:
Okay, then new Character actions submenu serves the same purpose. Since this is about making both the Online and Player menu hierarchy look the same, is that okay with you?

Ben987:
Sure, as long as the options don't overflow the max size of the menu it should be fine.  Some sub menus will help.  I'll test this one right now :)

PR #3197, opened 2022-01-29
feature: chat notification option for mentions of your character name

Sidiousious:
image

new Name Mentioned option for chat message notifications affecting whispers and normal messages that include the player's character name


Ben987:
Sure!  That sounds useful :)

PR #3198, opened 2022-01-29
College Chess Strip Fix

Nina-1474:
The chess game's strip function wasn't updated to account for the uniform split from #3136


Ben987:
Yes!  I've had a few reports on that one.  Cheers :)

PR #3200, opened 2022-01-29
Bed & Crib Legs Open fix

Nina-1474:
When a fix was made to the spreader prerequisite in #3162 it resulted in the Bed and Crib only being allowed if the character can close their legs, which is an unnecessary check that ends up blocking common combinations like the poofy diaper + crib.
The only item that looks odd is the floor shackles, but this is already an issue with other top-down items and deserves a separate fix.


Ben987:

PR #3201, opened 2022-01-29
bugfix: remember private room becoming public on recreate

Sidiousious:
A private room changing to public did not properly save the change on Player.LastChatRoomPrivate leading to a scenario where the room getting recreated by the player would lead to the room being private again. Also changed the corresponding check on other properties where false-y values are valid updates.


Ben987:
Sure.  Thanks for that fix!

PR #3202, opened 2022-01-29
Blouse Alpha Masks

Nina-1474:
The alpha masks on the Blouse have been extended (and in the Upper Suit case, added) to cover bras and corsets which are under it but often poke out at the sides. They've also been tweaked to allow items that poke up at the neck to appear there.


Ben987:
Very nice!  I've seen that one before, thanks for the fix :)

PR #3203, opened 2022-01-30
Fix missing Bondage Harness for the LegsClosed pose

zorgjbe:
When used in the panties slot, the files are missing. Copy them from the ItemPelvis version.


Nina-1474:
Assuming the images are identical across both groups, the panties version should have a `DynamicGroupName` pointing to ItemPelvis, which would allow it to reuse the images instead of needing copies.

zorgjbe:
Yes, Jomshir pointed that out to me. Here's an updated version that use `DynamicGroupName` on both harnesses and fixups the assets (turns out the full harness `ItemPelvis` had some missing ones compared to `Panties`). Thanks for the advice though @Nina-1474.

Ben987:
Absolutely,  Removing some duplicated assets cannot hurt :)

PR #3204, opened 2022-01-30
Effect Retrieval Rework

Nina-1474:
In #3164 the retrieval of effects for a character was changed to _add_ effects from Property onto effects from Asset or Group instead of overriding them by default. The function to check how gagged a character is had a separate retrieval of effects that was not updated to behave in a similar way.
This has been fixed by creating a common function that both `CharacterLoadEffect` and `SpeechGetGagLevel` call, ensuring asset configurations work consistently for both. This removes the need for the fix from #3194.


Ben987:
Thanks!  I'll check it soon.  Cheers :)

PR #3206, opened 2022-01-31
[Change/Asset] Kitty Gear Remaster + New Muzzle Gag

ItsJustMeVerity:
Replaces all my original Kitty headwear with new scratch-made assets that are better in far too many ways to list concisely.
No more crappy flat colours and a horribly pixellated shrunken image!
Also adds a new Muzzle Gag variant!
image

Also makes an internal change, with the Kitty Harness Gag (which was mistakenly called _Kitten_ HarnessGag) now being referred to as KittyHarnessGag in line with the others, which unfortunately means people will need to re-buy, but they'll need to visit the shop to grab the new muzzle variant anyway :P


Ben987:
Absolutely!  There are double DefaultColor tags in the assets though, but I can fix it later.  Cheers :)

PR #3207, opened 2022-01-31
Nicer labels for activities

zorgjbe:
This changes the activity labels to be more descriptive, allowing them to be fully customizable depending on where the activity takes place. That allows things like `Nose > Pet` to show up as "Boop nose", instead of being "Pet". The old names have to be kept for the Arousal Preference screen.

This comes from one of BCE's QoL features, so hat-tip to @Sidiousious for doing the grunt of the writing work.


Ben987:
That's a fun one!  Thanks you two :)

PR #3208, opened 2022-02-01
[Fix] Emote Edge Artifacting Fix

ItsJustMeVerity:
Fixes the artifacting around the edges of the 3 emoticons that suffer from it.
Also resizes the preview icon for the Reading emoticon ever so slightly so that it's consistent with all the others.

image


Ben987:
Cute!  I like it :)

PR #3209, opened 2022-02-01
Remove the "immediate" action items

zorgjbe:
This removes the 3 immediate actions items, namely "Leather Whip", "Leather Crop" and "Vibrating Wand", on the basis that they break immersion and duplicate actions that are possible using the handheld items.

Since this is a removal, I've asked the community about it [here](https://discord.com/channels/554377975714414605/554378904308023297/938113437429542982), if anyone was having hard feelings about that.


Ben987:
Have you checked in the single player rooms code to see if these items are still used?  Maybe there are custom scripts that run with them.

zorgjbe:
Yeeeah, there's a bunch of references in the dialogs. Let me fix up those as well.

zorgjbe:
Yeaah no. The SpankingToys can't be handled correctly in there, so scraping that for now.

PR #3210, opened 2022-02-01
Descriptive remote availability

Sidiousious:
Instead of hiding the remote control button the game will now tell you why it is unavailable so the player is no longer forced to figure it out on their own.

image


Ben987:
That's a good idea!  Thanks :)

PR #3213, opened 2022-02-02
[Change] Kitty Gag Tweak

ItsJustMeVerity:
Removes the speech garbling effect from the basic kitty mask "gag", since the preview image only presents itself as a regular facemask, and changes the display name to reflect this.

I may revisit this down the line with a similar, dedicated gag variant, but if people want the current effect then they can still do it by using a light gag underneath (e.g. duct tape or just the regular cloth stuffing).

Don't think I need a preview image for this one given there's no visual changes other than the name that shows up in-game :P


Ben987:
Absolutely!  Thanks :)

PR #3214, opened 2022-02-02
Fix bug where you would always be able to ask for a Pandora Padlock

zorgjbe:


Ben987:
Good catch!  Thanks :)

zorgjbe:
Thank the typescript compiler for that one 😉. I'm down to 180 errors with another batch of tweaks, most coming from the items.

PR #3215, opened 2022-02-03
[Change/Asset] Kitty Blindfold Remaster

ItsJustMeVerity:
Entirely new artwork for the Kitty Blindfold, bringing it in line with both the style and quality level of my other assets, allowing significantly better customisation, and finally respecting the fact that cats do not have eyebrows.

With no duplicated Default Colour setting this time!
image


Ben987:
I love it!  Thanks :)

PR #3216, opened 2022-02-03
More Typescript tweaks

zorgjbe:
Another batch of Typescript-related fixes. The only "functional" changes are added `toString()` method calls when `DrawText`-ing numeric variables to make the type coercion explicit.


BondageClub/Scripts/Typedef.d.ts
```diff
@@ -40,6 +40,10 @@ interface HTMLImageElement {
 	errorcount?: number;
 }
 
+interface HTMLElement {
+	setAttribute(string, number): void;
+}
+
```
Jomshir98, line 46: This is wrong - first you are missing names of arguments, so here you describe named arguments with type `any` (implicit any) and on top `setAttribute` takes a `DOMString` as second argument - not number. It does implicitly convert numbers to strings, but this is extra help from JS engine - not from standard.

BondageClub/Scripts/Typedef.d.ts
```diff
@@ -713,6 +717,8 @@ interface Character {
 	ArousalZoom?: boolean;
 	FixedImage?: string;
 	Rule?: LogRecord[];
+	Status?: unknown;
```
Jomshir98, line 720:
```suggestion
	Status?: string | null;
```

BondageClub/Screens/Room/Gambling/Gambling.js
```diff
@@ -335,7 +335,7 @@ function GamblingToothpickController (ToothpickState) {
 	}
 
 	else {
-		GamblingToothpickCount -= ToothpickState;
+		GamblingToothpickCount -= parseInt(ToothpickState);
```
Sidiousious, line 338: think it would be better to change the type and callers (in the dialog csv) to pass in a number instead of a string

BondageClub/Scripts/Typedef.d.ts
Sidiousious: Why is this interface needed here at all instead of making use of the standard HTMLElement distributed with Typescript in its lib.dom.d.ts?

zorgjbe: I'm attempting to silence stuff like this:

```
BondageClub/Screens/Online/ChatRoom/ChatRoom.js:2458:36 - error TS2345: Argument of type 'number' is not assignable to parameter of type 'string'.

2458    div.setAttribute('data-sender', data.Sender);
```
```
BondageClub/Screens/Online/ChatAdmin/ChatAdmin.js:39:70 - error TS2345: Argument of type 'number' is not assignable to parameter of type 'string'.

39  document.getElementById("InputAdminList").setAttribute("maxLength", 250);
```

BondageClub/Scripts/Typedef.d.ts
zorgjbe: Missed that one, thanks. Fixup applied.

BondageClub/Screens/Room/Gambling/Gambling.js
zorgjbe: Unclear what would be better. Right now (as I've been told), anything called as a Dialog action is stringly-typed (because of `CommonDynamicFunctionParams`), so I usually let those work like this and coerce with `parseInt`. The only exceptions are when non-dialog code calls stringly-typed dialog functions (like `DialogReputationLess`). Those I'll usually change to `{string|number}`.

BondageClub/Scripts/Typedef.d.ts
zorgjbe: Thanks @Jomshir98, I've updated the signature with the one from `lib.dom.d.ts`. Granted, this class of "errors" (13 in the codebase) makes me wonder, since I wasn't expecting issues with DOM-level functions/implicit conversions.

Ben987:
Sure!  Let's hope it won't create any bug in the gambling halls.

PR #3219, opened 2022-02-07
Fix: framerate-consistent smooth zoom in rooms

Sidiousious:
The animation calculation for "Zoom chatrooms smoothly" graphics setting did not take framerate into account. This changes the calculation so that the animation finishes in the same amount of time regardless of the current fps. Improves perceived responsiveness in low fps scenarios (<30). Previously with particularly high fps (200+) the animation would be near-instantaneous and its benefits mostly lost.


Ben987:
Thanks a lot!  I've never seen that one, but I'm sure that players with slower devices will appreciate :)

PR #3220, opened 2022-02-08
Kinky Dungeon 2.7.0 stuff

Ada18980:
Added more stuff from the current KD indev, mainly because there's a lot of key bugfixes in there and I'm not sure I want to separate them out.


Ben987:
You're on fire!  Thanks again :)

PR #3221, opened 2022-02-08
Add previous button to both wardrobe views

Sidiousious:
This also moves WardrobeClick to use MouseIn and fixes some inconsistencies in the preview character click areas. While the previous button is not that useful in the 2-page wardrobe with previews, it is useful in the 4-page wardrobe listing under appearance menus, and paves way for possible upgrades to wardrobe size.

image

image


Ben987:
Sure, that looks useful.  Thanks :)

PR #3225, opened 2022-02-10
Fix suit lower spread pose

zorgjbe:
This goes through the lower part clothes and realigns them with the Spread pose legs. The only one which I'm not satisfied with is the two pantyhoses, because the graphic hit the top of the "viewport", so I can't move it upward enough, and I think getting some pixels back would entail setting its Top property and reediting all its assets to be lower.


Ben987:
Very nice!  Thanks a lot, I'll include it in the next beta :)

PR #3226, opened 2022-02-11
fix: save block categories for room recreation

Sidiousious:
Blocked categories are currently lost when a room has to be recreated.


Ben987:
Sure!  It's a nice to little upgrade.  Thanks :)

PR #3227, opened 2022-02-11
Kinky Dungeon Update

Ada18980:
Add in some indev fixes to fix the issue with newgame+ entering in lost temple


Ben987:
Absolutely!  Thanks for that update :)

PR #3228, opened 2022-02-12
Add missing sprites for witch skirt

Ada18980:
Fixes missing sprites in different leg poses


Ada18980:
> I might be wrong, or there's a difference I'm not seeing, but if those are the same assets, a PoseMapping would cut down on the duplicates.

Sounds like it would, but I havent used a PoseMapping before. Ill have to look into it when I stop being too depressed to work on things

PR #3229, opened 2022-02-12
beta fix: mismatched status and character in vr

Sidiousious:
Status drawing was using ChatRoomCharacter instead of ChatRoomCharacterDrawlist resulting in the wrong player's status being drawn under certain circumstances, such as being in VR while the first person in the room is not.


Ben987:
Very well.  Thanks for that fix :)

PR #3231, opened 2022-02-12
fix: set wardrobe status when accessed via dialog

Sidiousious:
Currently the wardrobe status is only set when clicking the button above the chat. This expands it to be set when accessing clothes via a player's (your own or another's) dialog menu.


Ben987:
Good catch.  Absolutely :)

PR #3232, opened 2022-02-12
New Rope Necklaces and Tattoo Chokers

TheGnarp:
This PR adds a rope necklace (short and long variant) and two tattoo chokers (necklaces) to the game.
Since all those necklace items are worn rather tightly, I decided to tuck them underneath the upper clothes.

Feel free to contact me for any related assets if somebody wishes to add other variants or items as I always create those with vector based graphics which can easily be scaled or re-shaped.


Ben987:
Thanks for that one.  Have you checked to make sure that these cannot be used in violent ways?  There were a few scandals with players being hanged by the neck.  Cheers :)

zorgjbe:
@Ben987 IIRC the plan was to have the ceiling rope changed to have more than one suspension point visually instead of having to take out the neck rope (or, in that case, stop that one from getting in). Arguably, it's still a work-in-progress, so up to you.

TheGnarp:
Dangit! Thoughts like that don't even come to me. I could totally understand if you don't want to use the rope-portion, but if the suspension ropes change to a more comfortable variant, that would be really great! It would allow decorative ropes at the neck that don't cause harm AND look more realistic at the same time.

If however is working on the suspension ropes needs any assistance, I'm happy to help!
I can offer vector based ropes and suspension experience with ZERO casualties so far ^_-

Nina-1474:
The existing Neck Rope looks much more like a noose would and it's still in the game, so adding this one wouldn't be any more of an issue imo.

TheGnarp:
In case it is still need or wanted, I could turn this into a pull request:
grafik

Ben987:
@TheGnarp, this is a great idea.  Could you prepare a pull request for that?  Cheers :)

TheGnarp:
-> #3238

Ben987:
Thanks!  But let's wait until the next release to add that new stuff :)  I'll merge it next week.

Ben987:
It's time to test these changes!  Thanks again for your patience :)

PR #3235, opened 2022-02-12
Fix the witch skirt

zorgjbe:
This supersedes #3228 by redeclaring correct pose data for the skirt. The only assets needed are for the kneeling spread pose, because of the different width and for visual purposes.


Ben987:
Great!  Thanks for that one :)

PR #3237, opened 2022-02-13
Empty Chatroom Data bug

Nina-1474:
It's possible for the leash check to be run outside chatrooms when no "last room" data is loaded, which crashes out when it tries to check the block category and freezes the game. Now fixed.


Ben987:
Good catch!  Thanks for that fix :)

PR #3238, opened 2022-02-13
Update ceiling rope

TheGnarp:
Updated the look of the ceiling rope to make it use multiple attachment points and be more comfortable / less dangerous for the wearer.

This change we done very quickly and I haven't used the ceiling rope myself a lot, so please help me out if there are any scenarios where the new look might cause weird effects that I have not encountered or tested.


TheGnarp:
Sorry, didn't know who to ask apart from my comments in the previous PR.
Please feel free to use your or any other version. I was just trying to react on Ben's comment about preventing a scenario that could look like somebody was hanging from the neck.

PR #3240, opened 2022-02-14
Fix the Witch Skirt's other poses

zorgjbe:
It seems I pushed the wrong version of what I attempted to do to not copy assets around. That one's correct, and actually working. Sorry about that :/


Ben987:
No problem, that's why we have beta versions :)

zorgjbe:
Thanks!

PR #3241, opened 2022-02-14
Fix a couple of bugs in Stables

zorgjbe:
This fixes the following issues:
- not accounting for the player's amount of money before subtracting.
- the player's initial outfit getting overwritten in some conditions.
- updating the dialog correctly so you can't join the Trainer's Guild ($500) multiple times anymore.
- the dialog talking about a bridle when you stopped the training as a Trainer.


Ben987:
@GIT4NICK , would you like to review and test these changes?  Please let me know if you're ok with them.  Thanks you two :)

Ben987:
Thanks for these fixes, I will try it soon.  @GIT4NICK , feel free to review, test and apply patches on your side if you're still around.  Cheers!

PR #3242, opened 2022-02-15
Shock remote shenanigans

zorgjbe:
There's a couple of edge cases when using the shock remote:
- the automated collar had no text for it
- the automated unit could have an undefined intensity when it's applied but not saved yet (the other player still has the UI open).


Ben987:

PR #3243, opened 2022-02-15
fix: college crash in non-default colored skirt

Sidiousious:
Fixes ItemColorIsDefault to not crash when a DefaultColor has not been specified for the given item asset.


Ben987:
Sure!  Thanks for that fix :)

PR #3244, opened 2022-02-15
[Fix] ArmsCuffed Pose Random Line Removal

ItsJustMeVerity:
Fixes some random line on the character's left shoulder that differs in opacity depending on the body size for some reason (Doesn't affect XLarge), in the ArmsCuffed pose.

image


Ben987:
You have pretty good eyes, I've never noticed that one.  Thanks :)

PR #3246, opened 2022-02-17
Invalid tag for the Chain Attached button on the feet/ankles

zorgjbe:


Ben987:
Sure!  Thanks for that one :)

PR #3247, opened 2022-02-18
Bugfix: Fixes max length variable reference in the Transport Jacket script

elliesec:
Just a tiny bug that came to my attention recently. The Transport Jacket's script was referencing the max length variable from the Wooden Box when redrawing the preview text, probably due to a copy/paste oversight.


Ben987:
Sure!  Thanks for the fix :)

PR #3248, opened 2022-02-18
Fix issue with the talk status and dead keys

zorgjbe:
Some keys should never be reported as "talking". Right now, if you type a few characters in the field and start tabbing in and out, the status will show, because you'll keyup on Alt or Ctrl with a non-empty field.

Likewise if you start moving around your post with the arrow keys. This one's actually a bit far-fetched, but the less we divulge about non-talking, the better I'll feel.


Ben987:
The less unused statuses we send over the better.  Thanks!

PR #3249, opened 2022-02-19
Update CN Translation

pjarayuzu:
Update CN Translation


Ben987:
Wow!  This is huge!  Thanks a lot, I'm sure that Chinese players will appreciate.

PR #3250, opened 2022-02-19
fix: Cannot read properties of null (reading 'GL')

Sidiousious:
Fixes the GL errors posted on Discord in recent weeks, e.g.

```
----- UNHANDLED ERROR (OUTSIDE OF BCX) -----
Message: Uncaught TypeError: Cannot read properties of null (reading 'GL')
Source: <url>/Scripts/GLDraw.js:591:31
TypeError: Cannot read properties of null (reading 'GL')
    at GLDrawAppearanceBuild (<url>/Scripts/GLDraw.js:591:31)
    at GLDrawRebuildCharacters (<url>/Scripts/GLDraw.js:150:3)
    at GLDrawResetCanvas (<url>/Scripts/GLDraw.js:141:2)
    at <url>/Scripts/GLDraw.js:99:3

----- Current context (most recent first) -----
[None]

----- Debug report -----
Location: https://www.bondage-europe.com/R77/BondageClub/
UA: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/97.0.4692.99 Safari/537.36 OPR/83.0.4254.46
BC Version: R77
BCX Version: 0.7.9-dc6bf72f
No other mods detected.

----- BC state report -----
Mouse position: 825 9
Connected to server: true
Screen: Online/ChatRoom
In chatroom: true
GLVersion: No WebGL
Last received message: ServerInfo
Last sent message: ChatRoomChat
```

The error occurs when WebGL context is lost for more than 10 seconds or restored, but GLDrawLoad finds no WebGL support and sets the canvas as null. The Load function is not aware of the previous load and so does not restore the `CharacterAppearanceBuildCanvas` function from the backup when it sets `GLDrawCanvas` to null. `GLDrawResetCanvas` or `GLDrawOnContextRestored` ends up calling `GLDrawRebuildCharacters`, which ends up calling the GL version of `CharacterAppearanceBuildCanvas` (`GLDrawAppearanceBuild`), which no longer has the GLDrawCanvas to work with and errors.

This proposed fix does away with the confusing (and clearly error-prone) function swap and instead checks the state of the system to see which draw method should be used inside `CharacterAppearanceBuildCanvas`.

GL context loss can be tested with this console snippet:
```
{
  let gl = GLDrawCanvas.getContext("webgl2");
  let ext = gl.getExtension("WEBGL_lose_context");
  ext.loseContext();
}
```


Ben987:
Absolutely!  Thanks a lot of that one.  I'm not sure how to recreate and test it though.  We'll see when beta comes.  Cheers :)

PR #3252, opened 2022-02-22
Small ts tweaks

zorgjbe:
Another batch of TypeScript fixes from the TSC compiler, of the smaller kind, the more "invasive" ones being the switch to `.setAttribute()` for Input fields placeholders, and `ItemProperties` dictionaries typing in the item scripts.


Ben987:
These changes affect lots of different screens.  It makes it much harder for me to test them individually.  Have you tested all of them one by one manually?  Thanks :)

zorgjbe:
Most of those changes are just about fixing Typescript errors, so only a few of them have any functional impact — like using `toString()` or `parseInt()` instead of relying on implicit JS type coercions (5092ca8, 941f852, 036fddb, 06d7ffb, 2e3adb1, 33cb482, 7c97f6a, a2e9dff), or removing dead arguments to functions (7469955, d22ddcc, 806b977, a447ac5). A good chunk of the rest is about typing variables/parameters correctly, so only jsdoc comments or changes to the typing files (46fbf64, 5d5168b, 8c7fb32, 0ef7580, eb75e72, 150c8da).

Which means the only "real" changes are 920b915 (and supplement 0cc1c10 thanks to Jomshir), which I've tested ; d1ce3f9, which is obviously wrong and never worked correctly in the first place ; 6166d88, which I've tested one of its instances on a password lock and worked ; a3b10f6, fixing parameters so they match the function's typing even though it doesn't care at runtime (there's a `.toUppercase()` in there) ; and 39f9c9a/4ab8720, a variable typing fix and not passing a required argument to a function, and resilencing ESLint because that file is a complete mess anyway.

Maybe one I should point out though is 2e3adb1, because once you unpack the function's indirection, if your Dominant stat is -49, you end up with a case where both those functions can be false for the first two (a bug), as well as the comments saying "less than 23 dominant" being wrong (since the helpers do a `<=`) meaning the threshold is actually 22, which is why I changed the operator in that case.

Ben987:
The typescript stuff is fine, but this has impact in lots of places throughout the game.  Have you double-tested to make sure that this won't create any bug?  Next time, try to split these changes in smaller pieces, it makes it much easier for me to test.

PR #3253, opened 2022-02-22
Nursery TS bugs

zorgjbe:
Two minor fixes to the nursery, the first is a syntax error in an unused function, the 2nd looks like a leftover from times before Item being a thing, causing some items to not be accounted for when checking for the ability to open the nursery's gate.


Ben987:
Sure!  Thanks :)

zorgjbe:
You're very welcome 👍

PR #3257, opened 2022-02-24
Fix a MagicBattle bug

zorgjbe:
There's an issue where `GameMagicBattleNewTurn` would be called before `GameMagicBattlePuzzleEnd`, resulting in an invalid property access on the focused character, since it wouldn't be available anymore.

Workaround that by checking if the puzzle minigame actually ended before yanking the focused character, since its callback also handles it.

To reproduce, you need two players, start casting with the non-admin character, put your mouse on the green spot, switch back to the admin character tab and wait for the turn to end. When coming back to the non-admin character, you'll hit the bug.

Report and discussion [here](https://discord.com/channels/554377975714414605/554378725916147722/946405935058911352).


Ben987:
Good catch!  Thanks for the fix :)

PR #3259, opened 2022-02-25
Fix renamed function from the Photo room dialog

zorgjbe:
I missed that one when renaming, since it's in a CSV file. Sorry about that 😟.


Ben987:
No problem, thanks for the fix!

PR #3261, opened 2022-03-03
Maid room fix (serving drinks)

karame1:
A few fixes in maid quarters and cafe while bound serving drinks.

When "Cannot enter singleplayer rooms while bound" is activated, a few actions are impossible:
- Access the maid quarters to get a new tray if you lost it
- Access the maid cafe to refill your tray
The exception check was updated to allow it.

Some dialog are fixed in the maid cafe:
- Unexperienced maid check is fixed
- Use the right title when not a headmaid after changing your bounds


Ben987:
That sounds like a good idea, thanks a lot.  @wildsj , would you like to review the changes to the maid cafe?

Ben987:
Time to test this one!  @wildsj , let me know if there's something you'd like to change in your pieces of the code.

PR #3262, opened 2022-03-04
Clothes NoFeetSpreader Prerequisite Fix

Nina-1474:
It's possible to wear an item that requires the legs to be closed, and then a piece of clothing that should force the legs to be open (e.g. Bulky Diaper), since clothes don't have prerequisite checks like items do. Since such a combination is assumed impossible, the item then can't be removed (without removing the clothing first).
This change excludes clothing from that can-open-legs check, though long-term it would probably be better if clothing had prerequisite checking.


Ben987:
Sure!  Thanks for that fix :)

PR #3264, opened 2022-03-07
Sepi asset glitter

SepiaOulomenohn:
New typed 'Glitter' _Cloth Accessory_ and _Mask_ cosplay item:

- Freckles (Across face, larger dots)
- Mid Freckles (Nose only, larger dots)
- Split Freckles (Inner sides of eyes, larger dots)
- Small Freckles (Across face, smaller dots)
- Small Nose Freckles (Nose only, smaller dots)
- Small Cheek Freckles (Cheeks only, smaller dots)
- Stars Both (Star-like glitter, outer sides of eyes)
- Stars Left
- Stars Right
- Dots Both (Dot-like glitter, outer sides of eyes)
- Dots Left
- Dots Right

This item is meant not only to give players the option of having a number of freckles or glittery makeup styles, but also to enhance certain cosplay looks, such as dragon-like or alien-like (especially when combined with facepaint)


Ben987:
This is great!  I'm curious to try it :)

I'm not sure if "Cloth Accessory" is the best place for it though, but we can always relocate it later.  Thanks!

PR #3265, opened 2022-03-09
[Natsuki] [Backgrounds]: Restaurants

Codefreak64:
This adds in 2 new restaurant backgrounds
(due to invalid room data showing up on every room when trying to make a test room, I was unable to test it fully, but it seems to have no errors so if things do happen I'll fix it durring beta)

- credit goes to discord user Frenchurch#0683 for the backgrounds


Ben987:
That's a good idea!  Thanks a lot :)

PR #3267, opened 2022-03-12
Drone mask fixes

SepiaOulomenohn:
Fixed dialogue glitch when changing patterns that read 'AssetName' instead of drone mask.

Fixed inability for Onahole and Fleshlight mouth options to be used as if mouth were open.

Added additional module: Sight (AKA Lenses)
- Contains 'Opaque' and 'One-Way' options, in order to account for a common player request for blank masks to allow sight. Also adds additional blindfold functionality to all other eye options.

Renamed Drone Mask to 'Smooth Latex Mask' onscreen as per player request, may revert name depending on reaction.

Added 'Sci-Fi' category temporarily, may revert. (Although the mask is very sci-fi-ish, it can be used in fantasy settings if set up right.)


Shiranui-Izayoi:
Might I also suggest slightly modifying the base asset for the mask to better work with light colours? Currently while dark colours work great light colours get a very distracting black parts around the centre of the face and other spots. Here is an image as an example.
Drone
As you can see in bright colours the black parts can look quite strange. I know it may be asking a lot but would it be possible to fix that?

Ben987:
Sorry for the delay, it's time to test these changes!  Thanks a lot :)

SepiaOulomenohn:
To answer the above request about the darkened colors, I happen to really like the metallic look the masks have with lighter colors a _lot_.

That said, I agree that the colors don't look particularly like latex-it looks more like painted steel or gold. What I'd probably attempt to do (Even though I personally don't have the scripting skill for the last step yet), is:
- Separate out the shade layer from the base layer (for a min of 3 instead of 2, and a max of 10 layers when in full use instead of 9)
- Add a new module, or fit it into a current module as a 'material' change (So that the more metallic look is still an option)
- Create a new part of the script that, when the appropriate option is selected, modifies the color of the shade layer a bit more than the base layer--that is to say, it brings up both the lightness and saturation a little more (I'd definitely have to play with the math on this.)

PR #3268, opened 2022-03-12
Slim Latex Leotard Missing Asset Fix

Ayesha-678:
Just fixes a missing asset in the slim latex leotard in the suit section


Shiranui-Izayoi:
Hey do you think it would be possible to allow colouring of the two shade layers? That way it would allow more customization?

Currently some colours can look really strange when the shading is always the same. Like bright whites for example have quite dark spots. While dark colours can have quite light spots.

Maybe also change the breast shade to a lighter colour and then set the colour to default to dark in the file? Since when I tested out multi colour support while the shine could change the breast shading couldn't.

I hope it isn't to much to ask.

Ben987:
Absolutely!  Sorry for the delay.  Thanks a lot :)

PR #3269, opened 2022-03-14
Half shaved hairstyle Asset Fixes

MoonlightGleam:
- Added a new layer to have the posibility to color the hair and shaved parts differently
- Added a front layer in order to partially cover some head items


Ben987:
Very nice!  Sorry for the delay, thanks a lot :)

PR #3270, opened 2022-03-16
Fix cafe maid removing the player's multi-layered gags

karame1:
Ungagging at the cafe assumed the gag was only the first layer when serving drinks. This breaks the dialog flow and remove only the deepest layer when multi-layered (it can happen now that self-bondage is accepted to serve drinks or when being regagged online).

The fix uses almost the same code as being ungagged at the maid quarters, without changing conditions to be ungagged at the cafe.


Ben987:
Good catch!  Sorry for the delay, thanks a lot :)

PR #3271, opened 2022-03-17
Fix positioning of the steel ankle cuffs

zorgjbe:
Those cuffs were set as "absolute" (ie. origin is set via Top/Left). This resulted in a flying chain when used on the cross. Crop and realign the images so they don't need the origin set.


Ben987:
Yes, yes, yes and yes!  Thanks a lot, sorry for the delay.

PR #3273, opened 2022-03-18
Corset Cutout & Misc Fixes

Nina-1474:
LeatherCorsetTop1 and SteampunkCorsetTop1 both have hand cutouts in the default images which are unnecessary now that hands are a separate group, and means there are lots of duplicates without cutouts in all the other poses. These have been cleaned up.
The configuration for the Steampunk one was also wrong meaning the cutouts always appeared anyway, which is resolved now.
AllFours was hiding the corset meaning even these two assets which have images for it weren't appearing. They will now.


Ben987:
Very nice!  Right on time for second beta, thanks a lot :)

PR #3274, opened 2022-03-20
Drone mask fixes 2

SepiaOulomenohn:
Changed all 'Mouth', 'Eyes', and 'Pattern' icons for visual clarity.

As a result, swapped 'Mouth' and 'Eye' module order, for an upwards mouth-eye-pattern visual progression instead of a wonky up-down-up visual progression.

Fixed two dialog lines that incorrectly reference the item.


Ben987:
Absolutely!  Right on time for the full release :)

PR #3275, opened 2022-03-21
Allow Activity on extended types

Nina-1474:
This allows an AllowActivity (e.g. "Penetrate") to be available only for a specific extended type on an asset, instead of the entire asset. This was requested for an upcoming item.


Ben987:
Great idea!  Thanks a lot :)

PR #3276, opened 2022-03-21
fix: combo lock blindness updating crash

Sidiousious:
When inspecting your own locks, the Load function is called every time your appearance updates. This leads to a bug where the input fields do not update to reflect current blindness.

When inspecting another player's lock that is in a blocked slot (e.g. lock on ankles while also wearing a hobble skirt), the inspecting player's blindness changing crashes the game, when the Draw function attempts to change the type on an input that does not exist.

This change moves to use the Load function as the canonical way to update the state of the blindness, instead of the small patch in Draw, fixing both issues.


Ben987:
Nice fix!  Thanks a lot :)

PR #3277, opened 2022-03-21
R78 changelog

Sidiousious:
also re-enables the version change maid


Ben987:
Thanks a lot!  Would you like to build the log again for the next version, before April 10th ?  Cheers :)

PR #3278, opened 2022-03-22
Expression commands

karame1:
Add commands in chatrooms for fast expression changes.

- /blush [0-5|b|+|-] to set blush level with 0-5 and b (the blue one), + and - to increase/decrease blush.
- /expr [0-5] to load a saved expression with 1-5, 0 to clear current expression.
- /pose [skoc][frhbeu] to change pose.
- /wink [rlb] to open/close an eye or both. Right and left is user-centric (instead of character-centric), like for eyes style and colors.

/pose manages legs and arms position and both can be set with a single command. Stand/kneel uses the same code that clicking on the button in chat (triggers minigame and add a status message). The other acts like clicking on a button in the pose menu. It can be used with the full pose name (for readability) or the one character shortcut.


BondageClub/Screens/Online/ChatRoom/Commands.js
```diff
@@ -368,6 +368,140 @@ const CommonCommands = [
...
+				if (args[1] == "l" || args[1] == "b")
+					TargetLeft = true;
+				Close = (args[0] == "c");
+			}
+			if (AcceptCmd) {
+				let Expression = null;
+				if (Close)
+					Expression = "Closed";
+				if (TargetLeft && TargetRight)
+					CharacterSetFacialExpression(Player, "Eyes", Expression);
```
Sidiousious, line 452: I would add a duration for these since winking is usually understood to be different from closing eyes.

BondageClub/Screens/Online/ChatRoom/Commands.js
```diff
@@ -368,6 +368,140 @@ const CommonCommands = [
...
+			let BlushLevels = [null, "Low", "Medium", "High", "VeryHigh", "Extreme"];
+			if (/(^[0-5]$)/.test(args)) {
+				let BlushNum = parseInt(args);
+				CharacterSetFacialExpression(Player, "Blush", BlushLevels[BlushNum]);
+			} else if (/(^(b|blue)$)/.test(args)) {
+				CharacterSetFacialExpression(Player, "Blush", "ShortBreath");
+			} else if (/(^(\+|-)$)/.test(args)) {
+				let CurrentBlush = InventoryGet(Player, "Blush")
+				if (CurrentBlush == null || CurrentBlush.Property == null || CurrentBlush.Property.Expression == null)
+					CurrentBlush = null;
```
Sidiousious, line 395: improve maintainability by either
1) keeping the conditional statement on the same line with the `if`, or
2) adding {}, even if it is a single line

Lots of bugs occur when people mistakenly add additional statements with the same indentation forgetting about the {}.

BondageClub/Screens/Online/ChatRoom/Commands.js
karame1: I ended up splitting /wink and /eyes, because a timer could not be set to reclose eyes as it should do without a timer set. And it ended being a mess between opening/closing eyes and winking.

And so /wink winks, and /eyes controls eyes status. There is no /peek, but that may add too much commands for too specific things. /winks may already be a little more specific, I don't know...

BondageClub/Scripts/Character.js
```diff
@@ -255,6 +255,9 @@ function CharacterReset(CharacterID, CharacterAssetFamily, Type = CharacterType.
 		IsKneeling: function () {
 			return (this.Pose != null && (this.Pose.includes("Kneel") || this.Pose.includes("KneelingSpread")));
 		},
+		IsClosingLegs: function() {
```
zorgjbe, line 258: `HasLegsClosed` ? Reads as less of a temporary, in-progress thing. It looks unused, too, so you might want to remove it.

karame1:
Thanks for the reviews, I think I've taken all the remarks in account.

But after some more testing the /pose command is a bit messy and doesn't work well when mixing kneel/stand up with open/close legs, and all other commands are ignored when the minigame is run. I'll open a new request with only the facial expressions and think a bit more about the body poses.

PR #3279, opened 2022-03-22
Changed Blouse1 to use DynamicGroupName

Shiranui-Izayoi:
Added DynamicGroupName to the Blouse item in the Suit slot. So it now uses the assets from the Cloth version. The only difference being that the Cloth version has hand cutouts.

As far as I can tell the lack of hand cutouts was the reason for the Suit versions original 25 priority. With that fixed I have increased it's priority to 29 allowing gloves to layered under it. As well as make it so that the hand asset no longer cuts off the asset. While also keeping it below the Cloth slot items.

Removed the Blouse1 assets from the Suit slot as they are no longer needed.


Ben987:
Absolutely!  Less files to load is always welcome.  Thanks :)

PR #3280, opened 2022-03-23
Fix suspension and hemp rope to allow kneeling spread when appropriate

zorgjbe:
This changes the ceiling rope and chain to allow kneeling and spreading when suspended (legs wiggling's a thing now), the KneelingHogtie as well, and fixes the leg zipties preventing someone from kneeling.


Ayesha-678:
The purpose of the ceiling rope is to NOT allow kneeling

zorgjbe:
…right, I was trying to enable it when raised only, but yeah, that make sense. I'll try to make it do that.

zorgjbe:
Here, that works correctly now, no more kneeling while still touching the ground.

Ben987:
Thanks a lot!  I'm curious to test this one :)

PR #3282, opened 2022-03-24
Add command /expr [0-5] to clear/load facial expression

karame1:
Add a quick way to get saved facial expressions in chatrooms.

- /expr 0 clears the current expression
- /expr 1-5 loads a saved expression
- /expr opens the saved expressions menu

The "focus on a character" code was extracted to a dedicated function and DialogFindSubMenu was added to simulate clicking on a player and navigating to a submenu by code.


Ben987:
This is a good idea.  Thanks!

PR #3285, opened 2022-03-24
Accepted characters for the name when creating a new character

tetris245:
The LS variable controls which characters can be accepted for the name of a character when creating a new character. It is too restrictive as it accepts only the normal characters corresponding to the English alphabet. I suggest to extend it to 
* the other alphabets, because players can use their own language in the chat rooms, so it would be coherent to allow that also for the name of their characters in the game
* the numbers, because numbers are actually used in the GGTS part of the game, so it would be coherent to allow them for the name of the characters outside GGTS
* the spaces to allow more creativity when choosing a name for a character
What I suggest excludes all the special characters, so there is no risk for the readabiliy and the security of the chat rooms.


Sidiousious:
Would need to be updated in the server first: https://github.com/Ben987/Bondage-Club-Server/blob/9dffea6705909d2255c8ca2fe9440e66e14b1799/app.js#L235

Ben987:
Sorry but no, lets keep the basic letters for now.  We will get lots of crazy names.

PR #3288, opened 2022-03-26
Audio system improvements

zorgjbe:
This improves the game's audio system with the following features:
- merge the code paths so the volume is calculated consistently (ie. for ropes, no more loud apply, quiet changing)
- change the volume modifier curve so it's more perceptible and less wacky
- wire activities through the "sound effect detector", and enable slapping and spanking sounds
- more than one sample per sound effect, chosen randomly, giving it a bit of variety


Ben987:
Absolutely!  I'm curious to try that one.  Thanks :)

PR #3291, opened 2022-03-31
fix: stop message receiving from popping afk bubble

Sidiousious:
The `ElementScrollToEnd` when a new message is received while the chat is scrolled to the bottom triggers a scroll event (at least on Firefox) on the chat log and the code in AfkTimer.js detects it as activity, when in reality the player has not done anything. I don't think listening to the scroll event on top of mousemove brings much value, so proposing we simply remove it instead of adding complexity to check for manual vs. code-triggered scroll.


Ben987:
I've never that one, but thanks!

PR #3293, opened 2022-04-02
Fix an invalid property access on messages that do not have a sender

zorgjbe:
It turns out there are messages sent with no Sender information.

Move the variable inside the block to fix the crash

Introduced by #3288


Ben987:
Sure!  Thanks for that fix :)

PR #3294, opened 2022-04-02
Cleanup the way activities are defined

zorgjbe:
This changes the data so that activities declare their targets instead of the other way around, mostly because checking them requires jumping around that 6000+ line file, which is a complete hassle.

It also makes activities somewhat independent from the AssetFamily, in the off chance we want to support more than one of those in the future.

A few more activities get enabled (mostly on yourself), since checking for discrepancies between the target lists is so much easier now.

See the individual commits' messages for more information.


Ben987:
This one is pretty big, I'll need to find some time to test it.  

It also has a few conflicts with other pull request.  Can you resolve and test again?  Thanks :)

zorgjbe:
Rebased. I redid the ordering completely because there were some cases where it was all over the place, so it's now more consistent — they go mouth, hand, feet, whole-body, head, item, in increasing order of harshness when applicable.
🤞 I won't get pitch-forked into oblivion for changing that.

Ben987:
Wow!  This is huge.  I hope you'll be available next week when beta is released.  Yes, lets hope there won't be too many criticism or bugs.  Thanks a lot :)

PR #3295, opened 2022-04-02
Make the ItemProperties interface explicit to catch issues

zorgjbe:
This makes the `ItemProperties` interface more fleshed out, allowing TS to catch more problems (and it did, there where a couple typoes, or misused of `Hide`/`HideItem` keys).


BondageClub/Scripts/Typedef.d.ts
```diff
@@ -916,10 +1006,150 @@ interface NPCTrait {
...
+ * Custom properties for extended items
+ *
+ * Those are properties that are asset-specific, so the handling might be done
+ * per-item.
+ */
+interface ItemPropertiesCustom {
+	ItemMemberNumber?: number;
+
+	Mode?: unknown;
+	Intensity?: (() => number) | number;
```
zorgjbe, line 1056: As a heads up, this one got me worried. There are many places where the code expects it to be a number, but it can actually be a lambda (the advanced vibrator modes use that) arithmetic doesn't work correctly.

`Effect` (at line 1034 above) suffers from that as well as it can also be a lambda, taking intensity as a parameter.

Ben987:
Good catches!  I won't be able to test these myself though, we'll wait for the beta to see if it creates bugs.  Thanks :)

PR #3296, opened 2022-04-02
Make the PonyGag effects work correctly

zorgjbe:
This tweaks the PonyGag definition so it doesn't rely on `OverrideAssetEffect`, which doesn't seem to work (at least I haven't been able to make `CharacterGetEffects` take it into account, as `Property` doesn't seem to contain that key)


Nina-1474:
Removing the asset's `Effect` entirely means it's inherting the group's `Effect` (["BlockMouth", "GagNormal"]) and applying this to every configuration. Changing it to [] instead should make things work.

`OverrideAssetEffect` was never tried on modular items before but looks like they don't need it. Typed extended assets need the default effects on the base asset since applying the item then escaping without making any selections results in only the base asset's effects being added. The Override then allows configurations to have fewer effects than the base/default asset has.
Modular assets don't have this issue because applying the item and selecting nothing still sends an additional update with the default configs.

zorgjbe:
Yeah, I had a hunch modular items weren't handling that, so thanks for clarifying and for the explanation for the group's default Effect inheritance. I've set the Asset's effects to `[]` and it seems to work.

Ben987:
Very good!  Thanks for the fix and the whitespace cleanup :)

PR #3298, opened 2022-04-03
Add minigame-specific success messages when struggling out

zorgjbe:
There was no way previously of knowing when and how a player did manage
to struggle out of some bindings.

Pass a specific action when the player succeeds in the struggling
minigame, and provide chat messages accordingly.


Ben987:
That's a good idea!  Thanks :)

PR #3299, opened 2022-04-03
/blush and /eyes commands

karame1:
Add commands to control blush and eyes expressions. Both commands without arguments open the expression menu.

- Blush accepts [0-5] and +, - or b to set blush level, increase or decrease it, or use set the fear (blue) one
- Eyes accepts either [o|open|c|close] with an optional [l|r|left|right] to open/close eyes or only one, or a set of expressions based on codes or emoji-like ”default”, ”dazed”, ”shy”, ”sad”, ”horny”, ”lewd”, ”verylewd”, ”heart”, ”heartpink”, ”lewdheart”, ”lewdheartpink”, ”dizzy” (or ”@@”), ”daydream” (or ”><”), ”shylyhappy” (or ”^^”), ”angry” (or ”èé”), ”surprised” (or ”éè”) or ”scared”

A new function was added in Dialog to open a submenu. I'm not very happy with the way the commands update the GUI as the code is split in Dialog and in commands. But a neat solution may require a lot of refactoring to tie the current expression and GUI a bit differently.

The help can grow a lot with all the available arguments. If the GUI handling and the help "wall of text" is fine, I can continue with status, eyebrows, fluids and mouth.

Also the eyes command does not have the same behaviour as the GUI buttons: setting an expression doesn't open eyes.

The /wink command suggested earlier was removed because the timer on eye1 reverts both eyes and I haven't found a way to prevent this without messing with targets (and include probably unexpected changes to triggered expressions).


BondageClub/Scripts/Dialog.js
```diff
@@ -1790,6 +1790,24 @@ function DialogFindSubMenu(MenuName) {
...
+/**
+ * Finds and sets a facial expression group. The expression sub menu has to be already opened.
+ * @param {string} The name of the expression group, see XXX.
+ * @returns {boolean} True, when the expression group was found and opened. False otherwise and nothing happens.
+ */
+function DialogFindFacialExpressionMenuGroup(ExpressionGroup) {
+	if (DialogSelfMenuSelected.Name != "Expression") return false;
+	if (!DialogFacialExpressions || !DialogFacialExpressions.length) DialogFacialExpressionsBuild();
+	for (let I = 0; I < DialogFacialExpressions.length; I++) {
+console.info(DialogFacialExpressions[I].Group);
```
zorgjbe, line 1802: It looks like a debugging leftover.

You could also use `.findIndex` instead of an old-style for loop, maybe?

BondageClub/Screens/Online/ChatRoom/Commands.js
```diff
@@ -384,6 +384,151 @@ const CommonCommands = [
...
+			if (/^(right|left|both|r|l|b)$/.test(args)) {
+				AcceptCmd = true;
+				if (args[0] == "r" || args[0] == "b") TargetRight = true;
+				if (args[0] == "l" || args[0] == "b") TargetLeft = true;
+				let LeftClosed = InventoryGet(Player, "Eyes").Property.Expression == "Closed";
+				let RightClosed = InventoryGet(Player, "Eyes2").Property.Expression == "Closed";
+				let Close = (TargetLeft && !LeftClosed);
+				Close = Close || (TargetRight && !RightClosed);
+				NewExpression = Close ? "Closed" : "Opened";
+			} else if (/(?:^(close|open|c|o)) *((?:right|left|both|r|l|b)?)$/.test(args)) {
```
zorgjbe, line 462: `/^(c(?:lose)?|o(?:pen)?)\s*(r(?:ight)?|l(?:eft)?|b(?:oth)?)?$/` ?

I'm just confused at the `^` being inside a non-capturing group, as well as the non-capturing groups themselves, I guess.

karame1: `/^(close|open|c|o) *(right|left|both|r|l|b)?$/` makes /eyes o crash, by returning ["o", "o", undefined]. And I just prefer "" for strings instead of null or undefined for crashing reasons.
`/^(close|open|c|o) *((?:right|left|both|r|l|b)?)$/` works fine, maybe because the last ? must be inside a group so that it can catch an empty string that is checked later for the default value. It looks like it just adds "" in the r|l|b choices.
`/^(close|open|c|o) *(right|left|both|r|l|b|)$/` seems to be equivalent and looks to me easier to read, but I don't know if that the ending | is something that is not to discreet to be used, or if r(?:ight)? is considered a more elegant way than r|right (it is technically, but regex syntax is somehow not very intuitive).

Anyway, if there are no more details, I'll add `/^(c(?:lose)?|o(?:pen)?) *(r(?:ight)?|l(?:eft)?|b(?:oth)?)$/` which works without too much nesting and use !string instead of string == "" for the default value.

Ben987:
Thanks a lot!  I'm curious to try that one.  Cheers :)

PR #3301, opened 2022-04-05
Revert "Add minigame-specific success messages when struggling out"

zorgjbe:
Reverts Ben987/Bondage-College#3298

It turns out this causes equipping items on yourself to trigger those new messages, so you always "snap out" of your item 😞. I need to take a closer look at how those minigames are started and clarify normal equipping from struggling.


Ben987:
No problem :)

PR #3303, opened 2022-04-06
Chinese translation for NPC Collaring and NPC Slave Auction

WWWWWWWWWWWWWWWWWWWWang:
Adding Chinese translation for NPC Collaring and NPC Slave Auction


Ben987:
Awesome!  Thanks for the transaction.  Don't you also need to change the translation main file to include these changes?  Cheers :)

PR #3308, opened 2022-04-07
Misc: Code restructure.

remiliacn:
### What is this change

This change is more a refactoring code to improve the code structure. In case of new languages garbling checks that is needed to be added, the new functions added in this commit should help to improve the code's maintainability and readability.
This change has three main goals:
* Make the if statements more readable by using `includes()` feature instead of using nested if or statement.
* Group same checks to a function to improve readability and maintainability.
* Move for loop to top level, which could improve performance.

### Testing
Local testing with several string tests are conducted. The below are the functions that used to test:
```
const testString = 'abcdefghijklmnopqrstuvwyz123456789áàãâéèêíìîõóòôúùûçñaeиоуюляcйхжклрузсгйбрвыдфгнм';
> SpeechGarbleByGagLevel(21, testString, true)
'm   m   m   mmm   mm     mmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmm mmmmmmmmmmmmmmmmmmmm'

> SpeechGarbleByGagLevel(17, testString, true)
'm mmm  mm  mmmm  mmmm    mmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmm'

> SpeechGarbleByGagLevel(15, testString, true)
'mmmmmmmmm  mmmmm mmmm mm mmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmm'

> SpeechGarbleByGagLevel(8, testString, true)
'mmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmm'

> SpeechGarbleByGagLevel(7, testString, true)
'eebmкmeemmheaaammemmahmemmeheeeeeeeeeeeeeeeeeheeeeeeeaeeaemккгкaaeaaфeaгггмкгфaфффммгммм'

> SpeechGarbleByGagLevel(6, testString, true)
'eefkкmeemmheaaammefkaheefaeheeeeeeeeeeeeeeeeehmeeeeeeeaeeaekккгкaaeaaфeaгггмкгфaфффммгммм'

> SpeechGarbleByGagLevel(5, testString, true)
'aefkкmeemmheakamnefkaheefaeheeeeeeeeeeeeeeeeehmaeeeeeeaeeaekккгкaaeaaфeaгггмкгфaфффммгммм'

> SpeechGarbleByGagLevel(4, testString, true)
'aeememnhiakamnofkaheueaahaaaeeeiiiooouuusnaeаахeкакаафахкнфыаффыамфымкн'

> SpeechGarbleByGagLevel(3, testString, true)
'aeemefhhiakamnofkaseueaasaaaeeeiiiooouuusnaeаахeкакаафахкнфыаффыамфымкн'

> SpeechGarbleByGagLevel(2, testString, true)
'abemehnmhikmnofpkzeuvwysaaaeeeiiiooouuusnaeиоaaeккaрcзмкркaмкммнмм'

> SpeechGarbleByGagLevel(1, testString, true)
'abkmefmhiakamnopkaseuvwyzaaaeeeiiiooouuusnaeиоaakккaрcзмкркaмкммнмм'

> SpeechGarbleByGagLevel(0, testString, true)
'abcdefghijklmnopqrstuvwyz123456789áàãâéèêíìîõóòôúùûçñaeиоуюляcйхжклрузсгйбрвыдфгнм'

> SpeechGarbleByGagLevel(0, testString, true) === testString
true

```


BondageClub/Scripts/Speech.js
```diff
@@ -95,6 +95,36 @@ function SpeechGarble(C, CD, NoDeaf=false) {
...
+ * See below: 
+ * - includes(): https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes
+ * - indexOf(): https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/indexOf
+ * @param {string} stringSet - The string set that needs to be checked against to determine if characterNeedsToCheck is included.
+ * @param {string} characterNeedsToCheck - The character that needs to be checked that if is included in the stringset.
+ * @returns 
+ */
+function dialogContainsCharacter(stringSet, characterNeedsToCheck) {
+	return stringSet.indexOf(characterNeedsToCheck) > -1;
+}
```
Jomshir98, line 110: The argument about IE browser compatibility is not valid, as we use soooo many different features (some of which don't even have easy equivalent to transpile manually to - like `async` functions) that already aren't supported by IE - IE is just not supported in BC.
Also we heavily use `includes` in the code already so no point not using it here.

BondageClub/Scripts/Speech.js
```diff
@@ -528,11 +557,11 @@ function SpeechBabyTalk(C, CD) {
...
 			if (!Par) {
-				if (H == "k" || H == "l") NS = NS + "w";
-				if (H == "s") NS = NS + "sh";
-				if (H == "t") NS = NS + "st";
-				if (H == "a" || H == "b" || H == "c" || H == "d" || H == "e" || H == "f" || H == "g" || H == "h" || H == "i" || H == "j" || H == "m" || H == "n" || H == "o" || H == "p" || H == "q" || H == "r" || H == "u" || H == "v" || H == "w" || H == "x" || H == "y" || H == "z" || H == " " || H == "'" || H == "?" || H == "!" || H == "." || H == ",") NS = NS + H;
-			} else NS = NS + CD.charAt(L);
+				if (dialogContainsCharacter('kl', H)) NS += 'w';
+				if (H == "s") NS = NS += 'sh';
+				if (H == "t") NS = NS += 'st';
+				if (H.match('[a-z]') || dialogContainsCharacter(" '?!.,", H)) NS += H;
```
Jomshir98, line 563: Honestly... why not just use regexes for everything? It would look much cleaner and you don't need any extra functions...
```suggestion
				if (/[a-z '?!.,]/.test(H)) NS += H;
```

Jomshir98 (changes requested):

BondageClub/Scripts/Speech.js
Sidiousious: I believe old Edge support is still relevant, but `includes` is supported there too. Async/await is not, but it's only used in the chess minigame and would only break that one screen.

remiliacn: Sounds good, In that case, the mozilla doc might be outdated then.

BondageClub/Scripts/Speech.js
remiliacn: But definitely let me know what you think. In general, If we want to replace all of the if statements that using includes with the regex, as stated in the previous comment, the string that we want to look for is not always be consistent, that means, regex needs to be compiled multiple times during a chat room session. This may cause a significant performance impact especially for older PCs.

remiliacn: Sorry I think I mis-read the comment lol. Nice catch.

remiliacn:
This force push revision above contains the fix of the two comments above.

remiliacn:
The force push revision about removed comments of why used `indexOf()` instead of `includes()` since we decided to stick with `includes()`

Jomshir98:
In this case, why not just directly use includes instead of the indirection?
```
dialogContainsCharacter(' .?!~-', character)
```
vs
```
' .?!~-'.includes(character)
```
Also some of the functions don't have annotated the actual type of return, should be `@returns {boolean}`

remiliacn:
Revision above addressed the comment made by Jomshir98@

BondageClub/Scripts/Speech.js
```diff
@@ -149,11 +166,11 @@ function SpeechGarbleByGagLevel(GagEffect, CD, IgnoreOOC) {
 		for (let L = 0; L < CD.length; L++) {
 			let H = CD.charAt(L).toLowerCase();
 			if (H == "(" && !IgnoreOOC) Par = true;
-			if (Par) NS = NS + CD.charAt(L);
+			if (Par) NS = NS += CD.charAt(L);
```
zorgjbe, line 169: Duplicate assignment here.

BondageClub/Scripts/Speech.js
```diff
@@ -329,34 +344,34 @@ function SpeechGarbleByGagLevel(GagEffect, CD, IgnoreOOC) {
...
-				if (H == "í" || H == "î" || H == "ì" || H == "ï") NS = NS + "i";
-				if (H == "ó" || H == "ô" || H == "ò") NS = NS + "o";
-				if (H == "ú" || H == "û" || H == "ù" || H == "ü") NS = NS + "u";
-				if (H == "ç") NS = NS + "s";
-				if (H == "ñ") NS = NS + "n";
+				if ('áâà'.includes(H)) NS += 'a';
+				if ('éêèë'.includes(H)) NS += 'e';
+				if ('íîìï'.includes(H)) NS += 'i';
+				if ('óôò'.includes(H)) NS += 'o';
+				if ('úûùü'.includes(H)) NS += 'u';
```
zorgjbe, line 361: I'd move that block to a helper too, say `stripDiacriticsFromCharacter(char: string) => string | false` or something, since that's repeated a couple of times in various gag levels. Then it becomes
```js
if ((strip = stripDiacriticsFromCharacter(H)) NS += strip;
```

zorgjbe (commented):
A couple small comments, and thanks for the refactor work.

The only major surgery I'd do would be to hoist the for-loop/parenthesis-detection handling at the top-level, and sink the GagLevel checks and substitution inside it, instead of having a repeated for-loop for each level.

BondageClub/Scripts/Speech.js
remiliacn: Nice catch.

remiliacn:
Two above force push was to address @zorgjbe comments.

* Add stripDiacriticsFromCharacter() to strip out the dialects from the character
* Move the for loop to the top level to make the code tidy.

Ben987:
Thanks for that one.  Are you 100% sure it doesn't change anything for the in-game experience?  Cheers.

remiliacn:
> Thanks for that one. Are you 100% sure it doesn't change anything for the in-game experience? Cheers.

I have tested using different gagEffect levels to make sure that from 0 - 21, the garbled text is displayed as expected for each if check. So there should be no in-game experience changed.

Ben987:
Very well, let's test it in the next beta.  Thanks again!

Ben987:
There seems to be a small bug in your change.  The right parentheses are missing.  For example:  (test) shows (test

Can you fix it?   Thanks.

remiliacn:
Sure I will look into it.

remiliacn:
#3312

PR #3310, opened 2022-04-07
fix #3309: nursery dialog error causing accountupdate flood

### LINKED ISSUE #3309: Nursery disconnection bug
When asking nurse for additional restraints, connection rejected error is happening. It is not possible to fill the password and relogin, the button just doesn’t become active, so you must first leave to the main login menu and repeat your sign in. Error is new, something should be broken is the last two weeks commits.
Nursery

Sidiousious:
A comma in an unquoted string pushed the DialogWearItem from outcome to prerequisite causing it to be called once per frame, flooding the connection with AccountUpdates. This also makes it so you see the ", honest!" part in the dialog options as intended.

Fixes #3309


Ben987:
Sure!  Thanks for that fix :)

PR #3311, opened 2022-04-08
Fix a bug with the cheat key disabling cheat mode

zorgjbe:


Ben987:
No, some minigames are meant to have only one cheat per play.

PR #3313, opened 2022-04-09
Fix the command help being shown twice

zorgjbe:
Since there can be more than one element triggering keydown events to the same chain of handlers (namely, the one on document at Drawing.js:91 and the one in Element.js:44), CommandAutoComplete ends up called twice.

Fix the issue by stopping the event from propagating.


Ben987:
Sure!  Thanks for the fix :)

PR #3314, opened 2022-04-09
Drone mask fullhead

SepiaOulomenohn:
Added a new module to the Smooth Latex Mask (drone mask): 'coverage'
- Mask (as is)
- Hood (Covers whole head, alpha mask for more slick head shape, allows certain accessories)
- Total hood (Same as hood, but blocks all ears/accessories/hats)

Added new eye variant to the mask: 'Concave'

Fixed inability for onahole and pussy mouth variants of the mask to be penetrated. (Again)


Ben987:
Very nice!  Right on time for the next beta :)

PR #3317, opened 2022-04-10
Cancel Private Room Security Service fix

Nina-1474:
Trying to select No in the confirmation to cancel security service does nothing, leaving Yes as the only exit.


Ben987:

PR #3319, opened 2022-04-10
Fix item application playing sounds

### LINKED ISSUE #3318: Sound is playing even audio setting is turned off.
So in my understanding, if the sound is disabled, the sound should not be played in any occasion right? Several item sounds seem to be able to play even the audio setting is completely disabled.

After digging through the code base, which mainly the Audio.js file. It seems like the `AudioPlayInstantSound()` is heavily used, which lacking the audio setting check (mainly on if the play sound is enabled.).
https://github.com/Ben987/Bondage-College/blob/68c62181434045f34b99d8245b7b90e2a3a9ce1b/BondageClub/Scripts/Audio.js#L302-L310

And the method used that contains an audio check is never used. `AudioDialogStart()`
https://github.com/Ben987/Bondage-College/blob/68c62181434045f34b99d8245b7b90e2a3a9ce1b/BondageClub/Scripts/Audio.js#L317-L324

Would like to ask everyone for some opinions before making some changes.

zorgjbe:
After the rework in #3288, asset sounds go through `AudioPlaySoundForAsset` instead of `AudioDialogPlaySound`, and that function didn't account for `.PlayItem`.

Move the usually reason for silencing sounds in a helper, and update the high-level functions to check it before proceeding.

Closes #3318.


BondageClub/Scripts/Audio.js
```diff
@@ -424,6 +436,8 @@ function AudioPlaySoundEffect(soundEffect, volumeModifier) {
  * @returns {boolean} Whether a sound was played.
  */
 function AudioPlaySoundForAsset(character, asset) {
```
remiliacn, line 438: I think this method is only used in `struggle.js`, can you double check?

zorgjbe: Yes? I didn't want any external user of the low-level playing functions around.

remiliacn: Oh okay, this commit is related to components that uses the `AudioPlaySoundForAsset()` method, got it. In that case I think we are good.

But we might want to change how `AudioPlayInstantSound()` work, too. As this method handles audio play for most of the components in the code base.

zorgjbe: That's the low-level function, yes. Given it's low-level, it should have no business knowing if it's an item sound or not, and can't even tell. It just cares about playing a sound, or not if the volume is actually muted.

zorgjbe: I should have removed `AudioDialogStart/Stop` as well, since those were only used by the Struggle minigame. I kept it because it had some notion of "channel" (1, to be clear), which makes it immune to the sound effect stacking like crazy as the other function can. The notification sound getting played 30 times instantly when your computer gets out of sleep is something I wanted to fix, for my ear's sake.

remiliacn: Sorry I may be a little bit unclear of what I meant. The futuristic chastity belt's sound play is using this function, so I believe this is the root cause of why the sound plays even the audio setting is off.

zorgjbe: Yep, I forgot to move the check over, which is why I filed that PR 😉.

remiliacn: Yass! Maybe check here, too? :P

https://github.com/Ben987/Bondage-College/blob/bfff0e4277b2836b6fe23b98a1795edbda1cff1d/BondageClub/Screens/Inventory/ItemPelvis/FuturisticChastityBelt/FuturisticChastityBelt.js#L63-L67

https://github.com/Ben987/Bondage-College/blob/bfff0e4277b2836b6fe23b98a1795edbda1cff1d/BondageClub/Screens/Inventory/Futuristic/Futuristic.js#L154-L162

Ben987:

PR #3322, opened 2022-04-10
Fix: Draw fix.

remiliacn:
## What is this change?

Fix the issue where the text does not wrap correctly when using Chinese language.

### Before (English)
3 - origin

### Before (Chinese)
4 - origin

### After (Chinese)
image


Nina-1474:
This will cause English words to end up split across two lines.

remiliacn:
> This will cause English words to end up split across two lines.

Yes you are right. I was thinking how to fix that just now.

zorgjbe:
You're carrying around a bunch of unrelated commits (the ones from the OOC refactor) as well. I guess you might want to make sure your master is in sync with upstream, then rebase that draw_fix branch on top of it.

remiliacn:
Making it to a draft for now until I found something feasible to handle both English and Chinese.

PR #3286, opened 2022-03-25
Shion hoods 1

Ayesha-678:
credit to Shion for the assets, i only coded them

Adds 4 hoods

PR #3323, opened 2022-04-11
Restore the Vacbed Hairback hiding

zorgjbe:
That was removed in #3286, but as far as I can see this thing is opaque, so the hiding might have been for drawing optimisation reasons.


Ben987:
Alright, thanks :)

PR #3324, opened 2022-04-11
Pony harness gag

MoonlightGleam:
Modified assets for better coloring.
Enabled Metal, Reins and Panel color layers


BondageClub/Assets/Female3DCG/Female3DCG.js
```diff
@@ -4450,7 +4450,7 @@ var AssetFemale3DCG = [
 					{ Name: "Horn2Shine", HasType: false, AllowModuleTypes: ["h2"], AllowColorize: false, Priority: 58 },
 
 					// All Panels
-					{ Name: "PanelBase", HasType: false, AllowModuleTypes: ["p1", "p2","p3" ,"p4" ,"p5" ,"p6" ,"p7" ,"p8"], InheritLayerColor: "BaseStraps", Priority: 55  },
+					{ Name: "PanelBase", HasType: false, AllowModuleTypes: ["p1", "p2","p3" ,"p4" ,"p5" ,"p6" ,"p7" ,"p8"], InheritColor: "BaseStraps", Priority: 55  },
```
Nina-1474, line 4453: `InheritColor` can only reference asset groups. I'm guessing you wanted `CopyLayerColor` here, and on the Reins below.

MoonlightGleam: Copy layer color makes it unable to color them separately. Unless I'm missing something. DId have any issues testing this, but if its wrong Ill delete that part and use the default color, as its already there.

MoonlightGleam: MMh...The color group might actually be a good idea.
Im still learning my way around all the code so all of this is very valuable. Thanks ^^
Since there are a bunch of fixes, Ill add this once everything is fixed.

MoonlightGleam: Question. Making the color group shouldn't mess with the default colors right?

Nina-1474: It might actually. Easiest way to find out is to make the change and experiment I'd say. Worst case scenario you'll just need to rearrange the DefaultColor setting.

Ben987:
Sure, we can slide it in the next beta.  Thanks a lot :)

PR #3325, opened 2022-04-11
Fix: Chinese language text not wrapping properly.

remiliacn:
## What is this change

This commit is for fixing Chinese language(or any language desirably that does not use half-width space) not wrapping correctly, and also make sure that English, and other languages are not affected by this change.

### Before - Chinese
Before - CN

### After - Chinese
After - CN

### Before - English
Before - EN

### After - English
After - EN

### English - Extreme long word test case
Extreme case EN test

nese


Sidiousious:
Since wrapping rules vary by language and not necessarily by the alphabet used, it could make sense to check the current language and change logic depending on that instead of attempting to write a generic solution.

remiliacn:
> Since wrapping rules vary by language and not necessarily by the alphabet used, it could make sense to check the current language and change logic depending on that instead of attempting to write a generic solution.

I’m afraid that may generate too many rules for languages if more players in different regions come into the game. I had some more time testing this morning, and this generic rule seems to be able to handle most of the languages including Arabic.

Ben987:
As long as other languages are not impacted, it's all good with me.  Thanks!

PR #3326, opened 2022-04-11
Transparency in Preview Images

Nina-1474:
Changing the preview images for two new items to have transparent backgrounds instead of white.


MoonlightGleam:
Since I have the originals, It would have taken me very little time to fix this. But thank you nonetheless

Ben987:
Absolutely!  Thanks :)

PR #3327, opened 2022-04-11
Enable FixedPosition on layers

zorgjbe:
This adds the same property from assets down to an individual layer.


Ben987:
Good idea.  Thanks :)

PR #3329, opened 2022-04-12
Remove invalid block entries from pony gag reins

Sidiousious:
#3324 added invalid entries to the Block array on pony gag's reins. Removing them as they have no effect and produce typescript errors.


Ben987:
Absolutely, thanks!

PR #3320, opened 2022-04-10
Missing message on the Pony gag

zorgjbe:
The Blinders module lacks its dialog message

PR #3330, opened 2022-04-12
Typo on the PonyGag Blinders key

zorgjbe:
Followup to #3320 to fix the obvious typo.


Ben987:
Thanks for that fix :)

PR #3335, opened 2022-04-13
Fix UseRemote effect

zorgjbe:
I was investigating the effects as part of writing some documentation for them, and noticed that effect doesn't work anymore. Looking closer, I found a couple more issues, so here's a few fixes in that area:

- A bit of code cleanup, because those repeated conditions are hard to parse.
- We were not checking what is the actual return value from the `DialogCanUseRemoteState` helper, so clicking on the "item" remotes always opened the vibrator UI, regardless of things like the owner's "Prevent using remote on yourself" rule.
- The `UseRemote` effect itself would be considered by the helper, but only accounted for if a Use button wasn't added previously. Since any "extended" (a.k.a usable) asset qualified, the Remote button would never show up.
- With the effect added, it became very easy to lock oneself out of the Futuristic Collar interface using its "Block remotes on the wearer". Removing the effect (in RP, requiring the player to fiddle with the collar) prevents the deadlock from happening.

Right now, only items from the futuristic set use that effect, so if @Ada18980 would kindly take a look, I'd be glad 😄.


Ada18980 (approved):

Ben987:
Right on time for the next beta.  Thanks!

PR #3340, opened 2022-04-14
Typescript tweaks

zorgjbe:
Some more small fixes to staticly-typing stuff. All those are mostly comments changes and I don't expect them to cause any issues. The only "real" change is the type-confusion bug, and I've checked that it still functions correctly: I forcibly tried to set an advanced vibrator mode on an alt with those disabled, and I wasn't able to make the bugged version propagate outside in any case, so I suspect *something* *somewhere* was fixing it up properly anyway, and I don't think it was the validation system.


Ben987:
Sure, thanks for the tweaks.

PR #3341, opened 2022-04-14
Pony harness gag

MoonlightGleam:
Some fixes
Pentacle is now working correctly again
Detached gag no longer gags speech
Post is colorable
Color Layers are now correctly named


Ben987:
Thanks for the fixes, right on time for the next beta :)

PR #3345, opened 2022-04-15
Fix latest asset errors

zorgjbe:
With the recent usage of `HideAs`, the feature was extended slightly to allow an asset to partly work as if it was drawn under a different group. Since the original required both a group & an asset name, CI is now failing.

Provide an interface to please the Typescript goddesses, and rename the property to something more explicit. After a bit of clarification, @elliesec and I agreed the name wasn't very clear. As it just makes parts(?) of the drawing pipeline identifying it as another asset entirely, so `GraftOnto` feels like a better name.

…also, whitespace.


Nina-1474:
`GraftOnto` is a more confusing name to me as it implies the mentioned asset is actually present on the character and this one is joined on somehow. `TreatAs` is what I'd have suggested. Just my 2 cents though, not a huge deal.

elliesec:
I'm inclined to agree with @Nina-1474. Whilst I do think `HideAs` is a confusing name, I'm not convinced that `GraftOnto` is particularly more intuitive as Nina said, it feels like you're trying to somehow attach the asset itself to another one. Something like `TreatAs` or even `MimicAsset` feel a bit more descriptive about the behaviour that the property is providing.

Ben987:
Lets keep "HideAs" then.  Thanks everyone :)

zorgjbe:
As you wish. Graft comes from git terminology, so I'm using that meaning as in, "a non-existent commit making something act as if it had different parents that it really has". Sounds pretty accurate.

elliesec:
I get that, but I still think there's a difference in nuance between what we're doing with `HideAs` and what grafted commits in Git do. In Git, you intrinsically think about commits in terms of the commit tree (in terms of the data structure), whereas we don't tend to think of assets in that same way. Sure, the family/group/asset hierarchy forms a tree, but when we think about an asset, we tend to think about it as an object in its own right, rather than within the context of the data structure. So `GraftOnto` makes it sound like you're trying to stick two actual assets together, rather than trying to conflate them within the family/group/asset data structure.

That aside, there's also the angle that grafted commits in Git are probably obscure enough that it's not what the vast majority of people (even devs) would think of upon reading the word "graft". It's certainly not what came to mind for me, and ideally we want property names to be intuitive in the sense of "you understand it as soon as you read it", rather than "it makes sense once you understand where the name came from". I do think we could do better than `HideAs`, but I personally don't feel that `GraftOnto` is an improvement on `HideAs` in that sense.

PR #3346, opened 2022-04-15
Pony harness gag

MoonlightGleam:
Added some more fixes from the list.
Added HideAs properties on some modules
Dialog Tweak


Nina-1474:
A heads up, this will conflict with #3345 so whichever is merged second will have a conflict to resolve.

Ben987:
Thanks for that one!

MoonlightGleam:
The HideAs changes where done by Ellie, probably at the same time as zorgjbe did #3345

PR #3347, opened 2022-04-15
Type definition fixes for `HideAs` changes

elliesec:
This PR contains only the type definition fixes for #3341 to get the asset check passing again, similar to #3345, but smaller in scope for the purpose of getting the fix in during the beta, without any refactors.


Ben987:
Sure, thanks for that one :)

PR #3351, opened 2022-04-16
Pony harness gag

MoonlightGleam:
Fixed an Issue with the Blinders coloring and their default color


Ben987:
That should be safe for the full release.  Thanks a lot :)

PR #3354, opened 2022-04-16
The Chinese translation of chapters 10 and 11 and the repair of some …

estuiguang:
The Chinese translation of chapters 10 and 11 and the repair of some problems in Chapter 4


remiliacn (approved):
好哎！辛苦了！

Ben987:
Wow!  This is great, thanks a lot.  I'll slide it for the next full release :)

PR #3357, opened 2022-04-16
Variable Height Archetype

Nina-1474:
This converts the variable height suspension hogtie for hemp rope into a new Archetype, making it easy to add the same functionality to other items through the Extended configurations. It's currently only in use by subscreens for extended options, but it can be enabled for non-typed assets and with custom functions could even be applied to a specific layers only in future.
The hemp rope item itself has been converted into a typed item.
The suspended hogtie for chains now allows variable heights.


Ben987:
Absolutely!  I'm curious to test that one.  Cheers :)

PR #3358, opened 2022-04-16
Fix dereferencing null DOM elements

zorgjbe:
Most callers of ElementValue expect it to return a string, and usually run away doing a toUpperCase or some String method call on it. But if the element isn't in the DOM, you get back a null, and the code crashes.

Log an error but return an empty string, so the code can at least continue safely.

Fixes the crash reported here: https://discord.com/channels/554377975714414605/962770716141453342/964980248355680327


BondageClub/Scripts/Element.js
```diff
@@ -7,14 +7,18 @@
...
+		return "";
+	}
 
-// Returns the current HTML content of an element
+	if (Value == null)
+		return e.getAttribute("value").trim();
+
+	e.setAttribute("value", Value);
+	return "";
+}
```
Jomshir98, line 21: Might be nicer to always return a string rather than `string | undefined`; right now it isn't noticable, but if we ever manager to get to "strict null checks" level, it can help.

zorgjbe: Right, I saw the comment and forgot to update it. Let me fix that up…

Ben987:
Sure! Thanks for that fix :)

PR #3361, opened 2022-04-17
Improved types for assets and lock properties

Jomshir98:
Mostly typings updates.

The only code changes are in `BondageClub/Scripts/Asset.js` file:
 - In function `AssetGroupAdd` renamed argument `NewAsset` -> `NewGroup`
 - Added error instead of crash when CopyConfig points to non-existent entry
 - Avoids using `null` or `undefined` to index object in `AssetAssignColorIndices`
 - Refactors `AssetLoad` to use for-of
 - Changes `AssetActivitiesForGroup` to use simpler `forEach` instead of reduce


Ben987:
@elliesec , @Nina-1474 , is this a change that will be useful for you devs?  Thanks a lot :)

Nina-1474:
Hey Ben, yes these changes are useful to have. It all looks good to me.

PR #3362, opened 2022-04-17
Fix: Temporary fix for Chinese dialog issue.

remiliacn:
### what is this change?

This change is to fix the display bug for Chinese dialog buttons. Ideally, Chinese garbling should be determined by its tone and [Chinese pinyin](https://en.wikipedia.org/wiki/Pinyin), but we might need a fast fix instead of deep diving on how to garble each word properly. For now, it will just use a PRNG(pseudo random number generator) to generate a sentence that kinda looks like it is being garbled.

### Testing
Local override to test if Chinese garbled as expected, and other languages are not broken

#### Before
image

#### After
after


BondageClub/Scripts/Speech.js
```diff
@@ -96,12 +99,126 @@ function SpeechGarble(C, CD, NoDeaf=false) {
...
+			garbleRandomChance = .25;
+			break;
+		case 9: case 10: case 11: case 12:
+			garbleRandomChance = .2;
+			break;
+		default:
+			garbleRandomChance = .1;
+	}
+
+	const seed = randomSeeding(character + garbleRandomChance.toString());
```
zorgjbe, line 206: I'm a bit worried that doing that doing the seeding on every character is a bit too much…

How about initializing the seed with the first character from CD in `SpeechGarbleByGagLevel` and pass that down in each `doChineseGarbling` call?

zorgjbe (commented):
A few comments from me 😉.

BondageClub/Scripts/Dialog.js
```diff
@@ -406,7 +406,7 @@ function DialogPrerequisite(D) {
 		return CurrentCharacter[CurrentCharacter.Dialog[D].Prerequisite.substring(17, 250).replace("()", "").trim()]();
 	else if (CurrentCharacter.Dialog[D].Prerequisite.indexOf("!CurrentCharacter.") == 0)
 		return !CurrentCharacter[CurrentCharacter.Dialog[D].Prerequisite.substring(18, 250).replace("()", "").trim()]();
-	else if (CurrentCharacter.Dialog[D].Prerequisite.indexOf("(") >= 0)
+	else if (CurrentCharacter.Dialog[D].Prerequisite.indexOf("(") >= 0 || CurrentCharacter.Dialog[D].Prerequisite.indexOf("（") >= 0)
```
zorgjbe, line 409: Is that something that's already true? The "Prerequisite" column is not user-translatable and ideally shouldn't need to be (it's code), so in my opinion it'd be better to revert those that are so you don't have to teach `CommonDynamicFunctionParams` and friends how to handle those.

BondageClub/Scripts/Speech.js
remiliacn: This seeding should be super fast actually since it is using the bitwise calculation.

Using first character is kinda weird since it will make all the sentences have same gagging result since most of the Chinese sentences, like English starts with only a few selected characters.

BondageClub/Scripts/Dialog.js
remiliacn: Huh, after re-checking the code, I think it might be safe to remove this. Looks like the fullEmote() is where needed the change and handles NPC interaction.

remiliacn: Confirmed, with or without this line does not matter lol. Removing it.

BondageClub/Scripts/Speech.js
remiliacn: Also something that I found that is interesting.

I have ran a benchmark against this random function with the built in `Math.random()`, it seems like this PRNG is even faster than `Math.random()` :P

https://jsben.ch/9iksw

zorgjbe: It's faster. If you don't recreate the PRNG on each loop. Otherwise it's 5 times slower.

https://jsben.ch/9uDys

zorgjbe: Apart from the benchmark-bickering (sorry about that), I do see the point of using the current character for seeding, so I'm fine if it stays like this. I just hope it won't impact performance too much, because the low-end mobile users might notice.

remiliacn: Got it. But it should be okay since dialog shows only when you need to click button and all that, even performance is impacted, it shouldn’t be the end of the world. But I can see that could be an issue if someone spamming in chat room.

Nina-1474: It doesn't only run on clicking a button. The dialog option garbling gets recalculated on every frame drawn, so this seeding will be too.

remiliacn: For now, I think it might be okay. But I will work on an elaborate solution to actually properly garble the text. As I mentioned this is just a hot fix :P

Nina-1474: That's fair. Just wanted to make sure you're aware!

remiliacn: I know this is going to suck (no offense to other contributors) when I see every frame it calls the speechGarble() function lol XD.

VCode28629 (approved):
This fits the Chinese way of speaking

BondageClub/Scripts/Speech.js
```diff
@@ -395,11 +522,12 @@ function SpeechGarbleByGagLevel(GagEffect, CD, IgnoreOOC) {
 				if (H == "с") NS += 'з';
 				if ('дфмг'.includes(H)) NS += 'м';
 				if ('апрокенмит'.includes(H)) NS += H;
+				if (isChineseCharacter(H)) NS += doChineseGarbling(H, GagEffect);
 
 			} else NS += CD.charAt(L);
 		}
 
-		if (H == ")") Par = false;
+		if (H == ")" || H == ")") Par = false;
```
Sidiousious, line 530: same character checked twice (both U+0029)

remiliacn: Nice catch!

PR #3363, opened 2022-04-17
Improve LoginValidCollar so it handles lost ownership

zorgjbe:
When doing the collar check on login, if the player isn't owned clear all owner-related rules.

This fix an issue where players would have stale entries for those rules, and the fix made to remotes in #3335 would prevent them from using remotes on themselves.

Error report: https://discord.com/channels/554377975714414605/554378725916147722/965368521972391946


zorgjbe:
Closing in favor of #3365.

PR #3364, opened 2022-04-17
Move Wooden horse and remove its special pose

zorgjbe:
This PR does a couple things at the same time (yeah, sorry about that, but as always, I start working on something and then end up with a stack of small changes 😞).

Initially I wanted to move the Wooden Horse out of the leg slot to the devices, but I didn't want to break people's inventories, so I added a way to perform inventory fixups on login, allowing items to be moved easily from one group to another. It's a bit of a one-shot, so I don't know how people feel about that, and can take it out if it's not deemed useful. There's also some removed special-cases for some items, since RemoveAtLogin performs the same function.

The meat of the PR changes the Horse to use the "KneelingSpread" pose, removes all of its assets, and updates the rest of the leg assets with a new prerequisite that needs the ability to close legs (via the pose system), superseding "ToeTied" which kinda did the same thing but with an hardcoded list of items.


remiliacn:
`Question - non blocking`

Just a curious question, will this change break players that currently on the wooden horse? 🤔

remiliacn:
`Question - non blocking`

Just a curious question, will this change break players that currently on the wooden horse? 🤔

zorgjbe:
No, the fixup takes care of replacing the old item with the new one, both in the player's inventory and appearance. The only thing they do not handle is the wardrobe, but that could be added (the horse is a restraint, so it isn't concerned, hence why I didn't write that part). The only case where that would fail is if the device slot is already in use, but I think the horse blocks that.

Ben987:
This one has conflicts with other pull requests.  Can you resolve and test again?  Thanks :)

zorgjbe:
There you go! Since the merge was just the rack getting in the way and I had already checked the fixup application thoroughly, I only gave it a quick try, and seen nothing wrong.

Ben987:
It's time to test this huge one.  Thanks again :)

zorgjbe:
Sure hope it goes well, that'd be one less pose to design assets for 🤞.

PR #3365, opened 2022-04-18
Ownership Break Fixes

Nina-1474:
The data cleanup after removing an owner had some issues: owner only items and owner rules were not removed. Also breaking a trial vs breaking full ownership was inconsistent about what was reset and when.
Now breaking from an online owner (at both stages) clears out all relevant data, and does so *after* receiving confirmation from the server that the breakup was completed.


zorgjbe (approved):
Looks good to me 👍 . The only thing I might do is to move `ManagementClearOwnership()` to Character.js, since it's not really related to the Management screen.

And now I know why there's both .Owner and .Ownership, so yay~

Nina-1474:
I think you can only prompt an owner break from the Management room so it is at least a little related. I do see the Character reasoning, but that file is already so massive that avoiding adding more there is desirable~

zorgjbe:
I'm not sure I fully agree but hey, it's not like trying to find something in here doesn't already involve a project-wide find because so much stuff is scattered everywhere 🤣. At least you didn't base64-encode stuff 😏.

Nina-1474:
Alright you convinved me! I suspect the NPC slave releasing could make use of it too.

Ben987:

PR #3366, opened 2022-04-18
Assetcheck powerup

zorgjbe:
This is an attempt at making assetcheck filesystem-aware, in that it will build a tree of all .png files, perform the — dare I say — hairy transformations needed to turn an asset's layering into a list of its filenames, and compare that with the tree, reporting any "missing" or "exists" files ("matched" is when the layer matches a file, and is silent).

It's still rough around the edges, and there are definitely things that are not processed correctly (like, unsurprisingly, modular items). But at least it's easy for someone to go over the list of reported issues and be cleverer.

Known issues:
- modular items as well as some simpler ones like the coffin.
- files that are managed outside of the asset's definition. The kennel's animation files get flagged.
- files that are prefixes of another set, like FuturisticHeels vs. FuturisticHeels2. The former will get the latter's files reported
as "exists".


zorgjbe:
…it was supposed to be a draft, and I misclicked 😢. Don't merge that, that's just for posterity, or until someone makes it work correctly.

Nina-1474:
There's actually a 'Convert to draft' button that looks awfully like plain text under the Reviewers section in the top-right. (Or there was til I clicked it!)

zorgjbe:
Thanks! I completely missed it.

zorgjbe:
Updated that with more good stuff. There's still a few things that aren't handled properly though, but it's much better at checking complex things. 154 assets with an issue right now (a couple were handled by #3388, but like, 5?):
[assets.log](https://github.com/Ben987/Bondage-College/files/8577173/assets.log). There are some where the checker is just confused, but most of those seem legit, from assets that claim to support a pose but do not, or that support a pose but don't claim to.

PR #3368, opened 2022-04-18
PonyGag added to slots 2 and 3

MoonlightGleam:
Added the missing code for MouthSlots 2 and 3


Ben987:

PR #3369, opened 2022-04-18
Additional garbling fixes

elliesec:
Similar to #3331, this fixes some additional garbling issues that were introduced in #3308 where Latin & Cyrillic homoglyphs had gotten mixed up, or where some Cyrillic characters had been omitted entirely (presumably due to them looking like Latin characters). The changes in garbling behaviour were notably throwing off my gagspeak interpretation skills!

Specifically, after the changes in #3308 (and taking into account the fixes in #3331):
* At gag level 7, the Cyrillic characters `а`, `е` and `с` were no longer being garbled
* At gag level 6, the Cyrillic characters `а` and `е` were no longer being garbled
* At gag level 5, the Cyrillic characters `а` and `е` were no longer being garbled, and the Latin character `a` was being double-garbled
* At gag level 2, the Latin characters `j`, `l` and `r` were no longer being garbled, and the Latin characters `p` and `g` were being double-garbled

I've verified on a selection of Latin and Cyrillic strings that this restores the garbling behaviour so that it is identical to that before the refactor in #3308. The remainder of the changes in this PR are removal of trailing whitespace.


remiliacn:
> notably throwing off my gagspeak interpretation skills!

Nice, I guess some of then Latin & Cyrillic homoglyphs do look like being garbled twice, which I thought was an oversight as first. But I guess it is intended ╮(￣▽￣")╭ .

elliesec:
I don't read or write Cyrillic (beyond roughly knowing the alphabet), but had always assumed that the double-garbling of certain Cyrillic characters was deliberate (in the same way that you might phonetically soften "t" to "ch", or "p" to "ph" in English, even though BC doesn't do that in its garbling code). I don't know enough about the pronunciation to know whether that's the case, but it's the assumption I've been working on.

Aside from that, I'm a strong proponent for keeping functionality changes separate from straight refactors, and since #3308 was largely just a refactor, I don't think it should change the behaviour of the garbling functionality - just the readability/performance, so this PR restores the functionality from before #3308, whilst keeping the improvements introduced. If a Russian (or other language that uses the Cyrillic alphabet) speaker believes that the phonetics of Cyrillic garbling could be improved, then I'm all for that, but am of the opinion that it should be in a separate behaviour-focused PR.

remiliacn:
> Aside from that, I'm a strong proponent for keeping functionality changes separate from straight refactors

Since it is a refactor, and chained if statements are prompted for disaster in general, I assumed that this is just a coding mistake so I handled it during the refactor. But you are right that we shouldn't touch garblings for languages that we are not familiar with.

Ben987:
Thanks a lot!  I will prepare a hotfix :)

PR #3370, opened 2022-04-18
Implement protruding effect for assets

zorgjbe:
As there are quite a few items that should prevent other things from being applied because they "stick out", add a new protruding effect to keep track of those, and (one of) its counterparts "NotProtrudingFromMouth" prerequisite.

For now, only add the effect to the eligible mouth items, and use it to block hoods.


BondageClub/Scripts/Inventory.js
```diff
@@ -250,6 +250,12 @@ function InventoryPrerequisiteMessage(C, Prerequisite) {
 		case "GagUnique": return InventoryPrerequisiteConflictingGags(C, ["GagFlat", "GagCorset", "GagUnique"]);
 		case "GagCorset": return InventoryPrerequisiteConflictingGags(C, ["GagCorset"]);
 
+		// There's something in the mouth that's too large to allow that item on
+		case "NotProtrudingFromMouth":
+			return ["ItemMouth", "ItemMouth2", "ItemMouth3"].some(g =>
```
Nina-1474, line 255: Effects from worn items are all loaded into `C.Effect`, so it would be much simpler to just check that instead.

Jomshir98: But what if this effect is on non-mouth item? It could make sense for example for strapons or other toys to block things like belts.

Nina-1474: I feel this would more typically be done with separate effects like "MouthProtruding" and "VulvaProtruding" with the different prerequisites checking for the relevant one. Would there be any benefit to reusing the same "Protruding" effect if you need a different prerequisite to check if anyway?

zorgjbe: I'm not sure I'm fond of `C.Effect`, given the recent report about mittens + cuffs loosing the Boop activity while those independently should not. And yes, as @Jomshir98 says, it's likely I'll add more of those prerequisites over any needed group.

For the full picture, the initial work was about attempting to remove the alpha mask from the gwen hood entirely, given it impacted the PonyGag's mane and post pretty badly. It turns out there are a few assets that would work visually, but need it (Gwen Hood + Scarf), but anything that would be "solidly" in the way gets handled by that prereq. And yes, it would clearly benefit other assets like the strapon with a `NotProtrudingFromPelvis` or something, but as a first pass I feel it's enough.

zorgjbe: Simplicity? To me Protruding is a property of the asset, so it doesn't matter where it actually goes; in contrast, anything "wrapping" one group (or more, like the hoods do for most of the head's slots) would be interested to know about that.

elliesec: This is _mostly_ (but not entirely) thought experiment land now, but one benefit to Nina's suggested approach of separating the "type" of protrusion would be items that might protrude from multiple places. Consider for example the new Bridle Gag, which can protrude from the mouth (the dildo gag) and the forehead (the horn options). A muzzle might care about mouth protrusions, but not about forehead protrusions.

Or a neck corset might have a built-in dildo gag option (doesn't exist right now of course, but it's not implausible), and the check against the 3 mouth zones misses this.

Separating the type of protrusion allows finer granularity for handling cases like this, and doesn't really impact the prerequisite checks in any way. You'd just pair up `NotProtrudingFromMouth` against the `ProtrudingMouth` effect, `NotProtrudingFromForehead` against the `ProtrudingForehead` effect, etc. At the end of the day, prerequisites _are_ just strings - having that level of granularity isn't really going to have a major impact.

Nina-1474: I suppose but it's simplicity in one place for the trade-off of complexity in another. The asset can simply say "I protrude" but then the hood has to find out "But do you protrude from the mouth or the waist?". Instead of the asset saying "I protrude from the face" and the hood seeing the character has "Some asset somewhere protrudes from the face".
Both obviously work fine. It's just less consistent with what's already being done. But if y'all think it's more logical this way I don't actually have any major issue with it.

Sidiousious: As things stand I think the slot-specific effects would be simpler to handle for items that might protrude from multiple different places depending on what type is active, but I also think many of those items would be better served by splitting them into addon items (like the blanket is for the bed) that go in the proper slots, which would solve some asset conflicts and avoid having to do more complicated effects logic here.

elliesec: I'm inclined to agree with that. The introduction of the modular item system opened up the doors to a new level of flexibility in asset creation, but I feel like we've stepped into the realm where "just because you can, doesn't mean you should" applies, with some items using the system to massively overstep the bounds of their assigned group, or cramming loads of assets into a single item. It's not all on the modular item system - people have been pushing the limits of what the game allows in this way for ages - but the system has definitely made it easier.

I think the new Bridle Gag is an example of this usage of the system in ways it wasn't intended, and would be much more widely compatible (not to mention customisable) split out into a few separate items. That said, I think that the desire asset creators have to do this kind of thing does point at a separate need which the asset archetype system is lacking, which is the ability for items to interact with other items/groups (e.g. have a modular option whose state depends on the equip state of an item/items in another group, and where selecting an option in the module is a shortcut for equipping/unequipping said items). Having observed the way people are using the system now that it's more widely-used, and the feature requests I've had for it, I have plans for overhauling modular items and expanding the archetype system further, which should hopefully address some of the limitations that we're now running into in ways which will (hopefully) mean less need for such complex interdependence between items.

zorgjbe: I've renamed the effect, since that seems preferable.

Ben987:
Absolutely!  Thanks a lot :)

PR #3374, opened 2022-04-19
Add Chinese translations for some minigames

Anonymous-WghrYkBGUjBt:


elliesec:
There appear to be some issues with the naming of `Text_DojoStruggle_CN.txt` - you've included the contents of the CSV file in the filename, which is the cause of the failing checks.

It's additionally worth noting that these new translations won't automatically become available in the game unless you also add them to the CN translation dictionary in [`BondageClub/Scripts/Translation.js`](https://github.com/Ben987/Bondage-College/blob/master/BondageClub/Scripts/Translation.js#L288).

elliesec (changes requested):
There appear to be some issues with the naming of `Text_DojoStruggle_CN.txt` - you've included the contents of the CSV file in the filename, which is the cause of the failing checks.

It's additionally worth noting that these new translations won't automatically become available in the game unless you also add them to the CN translation dictionary in [`BondageClub/Scripts/Translation.js`](https://github.com/Ben987/Bondage-College/blob/master/BondageClub/Scripts/Translation.js#L288).

Anonymous-WghrYkBGUjBt:
Oh, one thing of note: There are a few instances of "Starts in" that doesn't have a placeholder for the number, these can't be properly translated due to different word order ("Starts in 5" translates to "5秒后开始"). I used something else that kinda works in this way, but it would be better if this can be changed to have a placeholder replacement.

Also, I didn't translate either of the auction minigames, partly there is one untranslatable string due to the same thing as above ("Do I hear $10"), partly I don't know some of the contexts of the strings.

elliesec:
> There are a few instances of "Starts in" that doesn't have a placeholder for the number, these can't be properly translated due to different word order

Thanks for that - it's something I'd suspected might be the case in certain languages. I'll make a note to look into a more general way of permitting variable substitution in dialogue.

elliesec (approved):
LGTM

Ben987:
You're on fire!  This is great, thanks a lot :)

PR #3375, opened 2022-04-19
Revert "Fix dereferencing null DOM elements"

Ben987:
Reverts Ben987/Bondage-College#3358


Ben987:
ElementValue() doesn't work anymore with that fix.

zorgjbe:
Huh? I tested using the password lock field, and the first two actual uses are on the login screen, and it worked correctly. Which screen was failing?

PR #3376, opened 2022-04-19
Fix dereferencing null DOM elements, take 2

zorgjbe:
This reapplies the fix in #3358, taking into account that Chrome seems to work differently and automatically proxy internal state to the element's attribute (and vice-versa).

I've tested Safari and Firefox on macOS, and the login form worked with both of them.


Ben987:
Thanks for the fix for the fix.  I'll try it right now :)

PR #3380, opened 2022-04-21
Medieval rack

MoonlightGleam:
New Device item. A medieval rack/torture rack/stretching rack.
Can be used with ropes, needs cuffs to be used with chains.


Ben987:
Wow!  This is awesome!  Thanks a lot, I'm curious to try it :)

PR #3382, opened 2022-04-22
Use a PRNG to build the lock pin combination

zorgjbe:
As we were saving the lockpick combination onto the item, but not sending the update, a concurrent update from the character being lockpicked caused it to be reset and thus regenerated, apparently randomly (since they wouldn't have had the property set if it hadn't opened the lockpicking UI and then caused an update, and doing that would also have frozen the combination to something non-lockpicker specific). Given the combination is hidden, you'd have no visual cue that you needed to start over from scratch.

Fix the bug by building the lockpicking order using the PRNG, seeded with a combination of the lockpicked character's ID (so every character gets a different combo), the group being lockpicked (so every group on a character gets a different combo), and the difficulty of the lock and number of pins (which themselves are mostly dependent on the lockpicker's skill and blocking, since the rest of the properties are common between lockpickers). This means that, unless you level up (or the lock properties are changed via the console), the combination doesn't change but is different for everyone.

This allows removing the validation handling for that property, since there's no need to share it anymore.


remiliacn:
Nice, I like this change. 👍

Nina-1474:
The only drawback is that if Player A wants to repeatedly lock Player B on Group C, the pick combination will always be the same. You could add the locked asset's name for further variation, but there's still potential for predictability. Perhaps that isn't a common play scenario but it's a change in behaviour to consider.

zorgjbe:
@Nina-1474 How about using the player's skill value itself then? I'm aware the way I'm hashing lock properties for the Difficulty calculation is a bit shoddy, so I could use the full Skill + progress points + locked asset name, and that would make sure that the combination is always the same, unless you fail or succeed (which would cause a skill update), or swap the restraint. Would that work better?

Additionally, the difficulty calculation is… strange. I tried to clean that up by making it like the usual difficulties, as in higher is harder, but decided against that because fixing that pesky bug was deemed higher-priority. But if you look at the "has key" handling, having them actually makes it harder (as it subtracts) ? Also, that doesn't seem to account to you having your keys taken away by your owner.

Nina-1474:
Wouldn't using Skill mean the combination is different for two different people trying to unlock it? And it wouldn't help for those who have their skill maxed out.
Generating and saving a random seed onto the property when first applying it is the only surefire way I can see of keeping the old behaviour. Is that possible to do that without reintroducing the issue you were fixing in the first place?

zorgjbe:
I've been leaning toward attaching ~some random crap~ reusing the `LockPickSeed` on the lock when the restraint get locked, so that is different for each lock, but changes when you lock/unlock repeatedly. Then when lockpicking you use that seed and the player's skill level as part of the combo calculation. For the maxed skill players, I could seed using the lockpicker's member number as well, that would make everything always look different.

I'm not sure what where the parameters for the initial design of that minigame, but if I were to improve on that, I'd generate the combination when the lock is added, setting that as `LockPickSeed`, then when the minigame starts you do a calculation (something that would be 0 if the player's lockpicking skill is greater than the lock's difficulty), make n shuffles on top of the "true" combo, taking into account the impossibility and the number of tries, and that would be what each player plays against. What I don't like is that the combo is now "public", though you can't see your own "skilllessness".

I kinda liked how the PRNG makes it way less obvious what the combination is, while still allowing it to be deterministic. Also, I haven't read the arousal/fail handling, but a cursory search tells me nothing ever resets the combo, apart from that bug, which kinda makes it too easy? It just makes you wait 30s and try again.

Nina-1474:
My argument about using skill was that the combination _shouldn't_ be different based on who's picking the lock. For an individual picker it makes no difference but supposing two people are collaboratively picking a lock taking turns and sharing information, they'd expect it to be the same. That's probably a very edge-case but even so, if we have a saved seed then I'd have thought we no longer need something like skill level to introduce unpredictability.

PR #3383, opened 2022-04-24
Feature: Door Indicator for Bondage Brawl

fleisch11:
Draws an UpArrow whenever you are in front of a door through which you can go by pressing up.

not sure if this is a good idea, but what counts as a door isn't always obvious.


Ben987:
Sure!  I'm curious to try it.  Thanks :)

PR #3386, opened 2022-04-27
Medieval rack

MoonlightGleam:
Fixed some incorrect names.
Removed the stains module and added it as another option for the bed.


MoonlightGleam:
> @MoonlightGleam - you'll want to resolve the conflicts on this before it can be merged.

Will do, I think it has to do with the wooden horse changes

Ben987:
Absolutely!  Thanks a lot :)

PR #3388, opened 2022-04-27
Asset fixes

zorgjbe:
A small collection of easy fixes caught by the improved asset check in #3366.

The only functional ones are the FuturisticHeels, for which I actually tweaked the heeled Lock layer so it wasn't so offset, and the Short Collar Rope, which had a Hogtied variant that wasn't enabled.


Ben987:
Sure!  Thanks for that fix :)

PR #3391, opened 2022-04-28
Fix: Small fix for translation updater.

remiliacn:
### What is this change?

Fixing a small bug that found while I was trying to do translations to new texts. Apparently there is risk of `newLines[m]` being index out of bound.

### Issue screenshot

image


remiliacn:
🤔 Looks like the root cause of this issue might be human error, but I am still persist of pushing this change out just make sure the updater won't break.

zorgjbe:
I've seen it. I think it's because the updater moves unknown keys at the end of the file, adding a one line comment in the middle of the file, which makes the line count odd.

Ben987:

PR #3393, opened 2022-05-01
Bugfix: `CanCloseLegs` prerequisite

elliesec:
## Summary

This fixes an issue with the new `CanCloseLegs` prerequisite introduced in #3364, which was intended to supercede the `ToeTied` prerequisite. I'm not entirely sure what the intentions were behind the original logic for the prerequisite, but the prerequisite would fail even if the character had no pose-affecting items equipped (i.e. wearing no items at all, in the default active pose, the prerequisite would fail).

Instead of using the `CharacterItemsHavePose` function to check against the `LegsClosed` pose, I've changed the prerequisite to use the `CharacterItemsHavePoseAvailable` function, which is used by the pose menu, as this felt like the most appropriate way of checking whether a character could adopt the `LegsClosed` pose. I've kept the `CharacterItemsHavePose` check for the `Kneel` pose to permit equipping/unequipping items when kneeling.

## Steps to reproduce

1. Ensure no pose-restricting items are equipped on the character (nude & wearing no items should suffice)
2. Set the character's pose to the default, if not already
3. Note that items using the `CanCloseLegs` prerequisite cannot be equipped (e.g. toe cuffs)


Ben987:
Yes!  Thanks for that fix :)

PR #3394, opened 2022-05-02
Gambling hall fixes

karame1:
A few fixes for the gambling hall

- Make Street to Roissy more legible by discarding rolls that don't make the game progress (so you can see what is the next expected number and where each player is)
- Prevent clicking on the "Dress back" button when it is not shown
- Fix a call to undefined GamblingAnnoySub which was renamed to GamblingAnnoyFirstSub

The lastest fix restores a behavior that was removed probably unexpectedly 3 years ago (and her reaction is unexpected too). I added a new dialog option just for wording when blindfolded.


Ben987:
Absolutely!  Thanks for fixing these old bugs.  @GIT4NICK , let me know if you want to review or revert any of these fixes.  Cheers :)

PR #3396, opened 2022-05-04
Sepi assets sheath gag

SepiaOulomenohn:
New expression-matching 'Latex Mouth Sheath' latex mouth item:
- Works with every mouth expression (Although a number are blank, expressed as transparent 1x1 pixel images) based on what the blackout contacts item had already set up.
- Adds a new folder for each relevant expression in the ItemMouth asset folder.
- Intended for both kinky latex play and for certain kinds of cosplay (eg. Robotic or Alien).
- Adjusted 'Fluids' priority to 12 to account for an interaction between the Latex Mouth Sheath (Priority 11), Mouth (Priority 10), and Fluids (Previously Priority 11) layers

New expression-matching 'Mouth Feature Gag' latex mouth item:
- Works similarly to the 'Latex Mouth Sheath', but as a panel gag.

New typed 'Boned Neck Corset' latex neck item:
- Choice of Ring or No Ring


Ben987:
This is great!  Thanks a lot, I'm curious to test it :)

PR #3398, opened 2022-05-05
Stop item sounds from playing when another screen is shown

zorgjbe:
As the CurrentScreen variable gets changed to whatever screen you're looking at (be it InformationSheet or Preference), the condition would allow sounds to go through, even though we're technically still in a chatroom.

Change the condition to check for the subscreens as well, so item sounds get correctly silenced.


Ben987:
Absolutely!  Thanks :)

PR #3399, opened 2022-05-05
Feature: Blur & tint effects

elliesec:
## Summary

This PR adds two new visual effects to the game: the blur effect, and tints. I've added tints to several items (with thanks to @Ada18980, @EmilyR42, @Ayesha-678 and @gatetrek for the permission to do so):

* Futuristic Mask
* Interactive Visor (restraint version only)
* Pantyhose (`ItemHead` and `ItemHood` groups)
* GP-9 Gas Mask
* Techno Helmet

I've also added a tint and a very minor blur to the Cloth Blindfold, which is the only item using blurs right now. Cecilia & I have another item on the way which will make use of both blurs and tints, and was the inspiration for adding these effects. Both effects have toggles within the immersion preferences screen (but are toggled on by default, and forced on for extreme mode players).

### Blur

The blur effect adds a gaussian blur visual effect to the game when rendering backgrounds & characters. Blur itself is a numerical value (representing the blur radius in pixels), but I've added a set of four effects which can be used to apply varying levels of blur (multiple blur effects on different items stack in the same way as deafness): `BlurLight` (3px), `BlurNormal` (8px), `BlurHeavy` (20px) and `BlurTotal` (50px). Below is a sample of the different blur effect levels.

blur

### Tints

Tints are effects that can be applied by an item which overlay the screen with a colour (much like the blind & arousal visual effects). Tints can have a fixed colour, or can be set to follow the colour of an item (or any index within its colour array) - for example to always match the colour of an items lens/visor. Tints can be set as a separate property on asset, and the strength (opacity) of the tint is also configurable. Tints are also supported in an item's `Property` object, and can be configured in both Typed and Modular item configuration (but tints in an item's `Property` are validated against an `AllowTint` property which prevents tints being applied to items which shouldn't have them). The `Tint` property is an array of tint definitions, and tints on the player stack if multiple items have them (or one item has multiple tints for whatever reason). Below is an example of a few different tint levels (combined with blindness effects) as applied to the Futuristic Mask:

tint-examples


Ben987:
Wow!  This looks really awesome.  I'm curious to try it.  Thanks a lot :)

PR #3402, opened 2022-05-06
Woodbox hood

ace-1331:
- added a new item (one-way woodbox hood) 

credit to Gnarp for the assets

_Note: this item is an old WIP that was supposed to ship with another version, but since both me and Gnarp have be on and off in the past year, we never managed to find time to finish it, so I've come around to salvage what was complete to get it into the club_


Ben987:
Cool!  I love it.  Thanks a lot you two :)

PR #3404, opened 2022-05-07
Feat: More online dialog cn translations.

remiliacn:
### What is this change?

More Chinese translation to online dialogs.


Ben987:
Absolutely!  I'm sure that Chinese players will love it :)

PR #3407, opened 2022-05-08
Appearance Screen Merge Concurrent Changes

Nina-1474:
When someone is editing a player's appearance and there's an online character refresh, e.g. due to a third person making a separate edit of their own, all ongoing changes in the screen are lost. 
Additionally, exiting the screen restores the character to before either set of changes, leaving the appearance out of sync for a while at best and potentially deleting the missing synced change later on if further changes are made on the out-of-sync character.

Now any incoming refreshed appearances while in the appearance screen will be compared to the current set of changes, and both will be merged together. The ongoing appearance edits will be preferred wherever both sets of changes affect the same groups, whether adding/removing/changing.
Exiting without saving appearance changes will now revert to the synced appearance instead of the originally backed up one.


Ben987:
Very nice!  I'm curious to test that one.  Thanks a lot :)

PR #3408, opened 2022-05-08
Asset: Poncho

elliesec:
## Summary

Adds a new Poncho asset made by @QueenTita to the `ClothAccessory` group. Tested against most assets in the game that I could think of that might conflict. The poncho has been hidden in the case of bad clipping issues (hence the large number of `Female3DCG.js` changes in this PR). Preview of the new item below:

image


Ben987:
Yes, yes, yes and yes!  This is lovely.  Thanks a lot you two :)

PR #3409, opened 2022-05-08
Fix: Add missing links to some translation files

karame1:
A few translation files were not linked in translation.js and were not loaded. While checking if there wasn't any other unlinked files, I reordered the lists in translation.js to be able to compare them with file lists.

I also removed some empty translation files and unlinked them.

Only the last commit has effect for the game, the other two were just some cleaning.


Ben987:
Absolutely, thanks for these fixes :)

PR #3411, opened 2022-05-09
Feat: Chinese translation for activity.

remiliacn:
### What is this change?

Chinese translation for user activities.


Ben987:
Very nice!  Right on time for the next beta.  Cheers :)

PR #3412, opened 2022-05-09
Cosplay Dildo Horn

MoonlightGleam:
Adding a second cosplay horn, bundled with the first unicorn horn.


BondageClub/Assets/Female3DCG/Female3DCG.js
```diff
@@ -1466,6 +1466,12 @@ var AssetFemale3DCG = [
 					{ Name: "Shine", AllowColorize: false}
 				]
 			},
+			{ Name: "DildocornHorn", Value: -1, DefaultColor: "#Default", BuyGroup: "UnicornHorn", BodyCosplay: true, Attribute: ["PenetrateItem"], Top: -30,
```
elliesec, line 1469: `DefaultColor` should be `Default` here - you only need the `#` if you're using hex code colours. The drawing code _should_ handle `#Default` because of guards we put in against invalid colours, but it will likely have to jump through more hoops to do so than if it was just `Default`.
```suggestion
			{ Name: "DildocornHorn", Value: -1, DefaultColor: "Default", BuyGroup: "UnicornHorn", BodyCosplay: true, Attribute: ["PenetrateItem"], Top: -30,
```

BondageClub/Assets/Female3DCG/Female3DCG.js
```diff
@@ -1535,6 +1541,12 @@ var AssetFemale3DCG = [
...
```
elliesec, line 1544: Same comment regarding `DefaultColor` here.
```suggestion
			{ Name: "DildocornHorn", Value: -1, DefaultColor: "Default", BuyGroup: "UnicornHorn", BodyCosplay: true, Attribute: ["PenetrateItem"], Top: -30,
```

elliesec (changes requested):

BondageClub/Assets/Female3DCG/Female3DCG.js
MoonlightGleam: Absolutely, it must have slipped by when replacing the code with the default. Already fixed

elliesec (approved):

Ben987:
Hahaha, sure, it's a fun idea.  Thanks :)

PR #3413, opened 2022-05-09
Bugfix: Kigurumi Mask chat message grammar

elliesec:
Fixes a minor grammatical error in the chat message when changing the type of the Kigurumi Mask.


Ben987:
Sure!  Thanks for that fix :)

PR #3414, opened 2022-05-09
Ensures dictionary tags are handled in the correct order

elliesec:
## Summary

This fixes a minor bug with chat message dictionary tag handling where if a chat message dictionary contains both `TargetCharacter` and `TargetCharacterName` tags, or `DestinationCharacter` and `DestinationCharacterName` tags, then the non-`Name` versions of the tags get processed first, resulting in the `-Name` tags being partially replaced in the chat message, resulting in `herName` or `herselfName` in the final message.

This change adds a sorting step to dictionary processing which ensures that the `TargetCharacter` and `DestinationCharacter` tags are sorted to the end of the array, guaranteeing that they get processed after the `TargetCharacterName` and `DestinationCharacterName` tags.


Ben987:
I don't think I've seen that bug before, but thanks for the fix :)

PR #3415, opened 2022-05-09
Bugfix: Caps Bondage Brawl level at 10

elliesec:
## Summary

In the Bondage Brawl game, it is currently possible to reach level 11 via the `PlatformAddExperience` function. This causes the game to crash when you deal damage to another character, as the `Attack.Damage` arrays are only indexed up to 10, meaning the following line in `PlatformProcessAction` sets `Damage` to `undefined`:
```js
Damage = Attack.Damage[Source.Level];
```
The game then crashes when drawing the damage number as it calls `Damage.Value.toString()`, where `Damage.Value` is `undefined`.

This PR fixes that by removing the last item in the `PlatformExperienceForLevel` array (which currently represents the experience it takes to get from level 10 to level 11). The result is that this brings the level cap in the `PlatformAddExperience` function down to 10 (which I assume was the originally intended maximum level). This approach means that the experience it takes to level up at each level remains unchanged from the previous behaviour, but you can no longer hit level 11.

The only real change in this PR is the change to the `PlatformExperienceForLevel` array, the rest is removal of trailing whitespace.

## Steps to reproduce

1. Play the game until you reach level 11, _or_ run the following:
    ```js
    for (let i = 1; i < 11; i++) {
	    PlatformAddExperience(PlatformPlayer, 500);
	}
    ```
2. Note that `PlatformPlayer.Level` is now 11
3. Attack an enemy
4. The game will crash with the following error:
    ```txt
    Platform.js:984 Uncaught TypeError: Cannot read properties of undefined (reading 'toString')
        at PlatformDrawCharacter (Platform.js:984:27)
        at PlatformDraw (Platform.js:1293:36)
        at Object.PlatformRun [as Run] (Platform.js:1328:2)
        at DrawProcess (Drawing.js:1352:30)
        at MainRun ((index):325:2)
    ```


remiliacn:
I'm thinking if it is okay to just let the player to deal 0 damage if player level is over 10, to also handle the situation where player purposely change his/her level to, let's say 999.

So something like this:
```javascript
Damage = Attack.Damage[Source.Level] || 0;
```

elliesec:
> I'm thinking if it is okay to just let the player to deal 0 damage if player level is over 10, to also handle the situation where player purposely change his/her level to, let's say 999.

Perhaps, but I'd be inclined not to explicitly cater to situations that shouldn't happen when playing the game normally. There are plenty of things in the game you can break with console, and trying to account for every possibility in the game code is impossible, so that tends to be where I draw the line.

remiliacn:
> > I'm thinking if it is okay to just let the player to deal 0 damage if player level is over 10, to also handle the situation where player purposely change his/her level to, let's say 999.
> 
> Perhaps, but I'd be inclined not to explicitly cater to situations that shouldn't happen when playing the game normally. There are plenty of things in the game you can break with console, and trying to account for every possibility in the game code is impossible, so that tends to be where I draw the line. The reason I made this fix was that you _could_ reach level 11 through playing the game normally, which I assume was unintended.

Makes sense! Cheers! 👍

Ben987:
Hahaha, thanks, I don't think I've ever played enough to reach that level.  Cheers!

PR #3416, opened 2022-05-09
Bugfix: Moves the `AllowBlur` preference into graphics settings

elliesec:
## Summary

It's come to my attention that on some devices (notably where hardware acceleration is turned off/not available), the blur effect causes a substantial drop in the game's frame rate (down to single digit FPS). Because of this, I've moved the preference controlling the blur effect out of immersion preferences and into graphics preferences so that players on hardcore mode aren't forced into potential performance problems. I also took the opportunity to add a tip to the main hall directing players to their graphics preferences if experiencing performance issues. As this change is likely not to make it into R80 beta 1, I've made use of the `PreferenceMigrate` function here to clean up players' immersion preferences object. As it'll likely only be available for one beta, it should be fairly safe to remove the migration before next release, and I've added a TODO to that effect.


Ben987:
The main hall notice is also a good idea.  Thanks!

PR #3417, opened 2022-05-10
New Variable-Height Rope Suspensions

Nina-1474:
This adds four more options for Hemp Rope in the arms slot which put the character in different rope suspensions: Kneeling Hogtied, All-Fours, Inverted Hogtie and Inverted All-Fours.
ropes

The following closely related minor changes have also been made:
- The variable height's slider can have its position and height configured to closer correspond to the player/item change.
- The SuspensionHogtie pose has been dropped as it's effectively covered by the Hogtie pose plus an override height.
- Prerequisites relying on that removed pose now check for a "Suspended" effect instead. The NotReverseSuspended prereq has been subsumed by the NotSuspended one.
- A bug where being suspended did not increase struggle difficulty as intended has been corrected.
- Some rope previews have been redone to have consistent colours.


BondageClub/Assets/Female3DCG/Female3DCGExtended.js
```diff
@@ -598,11 +604,17 @@ var AssetFemale3DCGExtended = {
...
+							Slider: {
+								Icon: "player",
+								Top: 0,
+								Height: 100,
+							},
 							Dialog: {
 								ChatPrefix: "SuspensionChange",
 								NpcPrefix: "ChainBondage",
 							},
+							GetHeightFunction: VariableHeightGetOverrideHeight,
```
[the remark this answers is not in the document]
Jomshir98: I would just add `Scripts/VariableHeight.js` file to the list of scripts loaded by asset check - it is there for that very reason.

BondageClub/Scripts/VariableHeight.js
```diff
@@ -72,8 +72,8 @@ function VariableHeightCreateData(asset,
 			CommonChatTags.TARGET_CHAR,
 			CommonChatTags.ASSET_NAME
 		],
-		getHeight: window[GetHeightFunction] || VariableHeightDefaultGet,
-		setHeight: window[SetHeightFunction] || VariableHeightDefaultSet,
+		getHeight: GetHeightFunction,
+		setHeight: SetHeightFunction,
```
elliesec, line 76: One further thought on this - I feel like it would probably be neater to have the `VariableHeightGetOverrideHeight`/`VariableHeightSetOverrideHeight` functions be the default values for this so that people don't need to explicitly define it, as I imagine the majority of cases are going to use these default implementations.
```suggestion
		getHeight: GetHeightFunction || VariableHeightGetOverrideHeight,
		setHeight: SetHeightFunction || VariableHeightSetOverrideHeight,
```

Nina-1474: I agree and had planned on doing so as soon as we had even one different function in use, so that these could serve as a demonstration of how the config should/can be done. Realistically that mightn't happen any time soon and would probably require assistance anyway, so yeah may well make that change now.

Ben987:
I'm very curious to try this one, thanks a lot!  Cheers :)

PR #3419, opened 2022-05-10
Additional handheld items

T-BoneShark:
Added by request of a few friends.
 - Adds a potion bottle handheld to the store that uses the `"RubItem"` arousal activity.
 - Adds a medical injector handheld to the store that uses the new `"InjectItem"` arousal activity.
 - Added a new `"InjectItem"` activity that can be done on the following slots.
   - `["ItemNeck", "ItemBreast", "ItemArms", "ItemButt", "ItemLegs", "ItemFeet"]`

As before, art done by Leah! Pics below.

preview_handhelds


zorgjbe:
Use `AllowPose` so you don't need so much placeholder assets? Especially if that makes the item just vanish when switching poses, as having to manually go through the files to know if something is supported or not is both a hassle and a loss of bandwidth…

elliesec:
To add to @zorgjbe's comment, there are two properties that are really useful for controlling the way assets interact with poses: `AllowPose` and `HideForPose`. Use `AllowPose` to indicate that an item has a different set of assets for certain poses, and use `HideForPose` to indicate that the item has no assets for certain poses and should be hidden. Between those two properties, you should be able to get rid of placeholder assets completely.

As an example:
```js
{
    AllowPose: ["Yoked", "Hogtied"],
    HideForPose: ["AllFours", "BackElbowTouch"]
}
```
A configuration like above means that the item has unique assets for the `Yoked` and `Hogtied` poses, but has _no_ assets for the `AllFours` and `BackElbowTouch` poses, and that all other poses should use the default assets.

In this particular case, as it looks like there are only assets for the default pose, you'll want to add a `HideForPose` entry which covers most (if not all) of the upper body poses, which will remove the need for placeholder assets in those poses, and prevent the game from attempting to load those assets completely.

T-BoneShark:
@elliesec @zorgjbe I attempted to make the changes as brought up by both of you (thanks for that!)
I ran into issues with trying to make use of the `HideForPose` property as both of the items used the `SpankingToys` asset, so I moved the medical injector out into it's own asset to make use of this feature along with fixing the injection activity. If there is a way to make use of these fields for the `SpankingToys` sub types I'm all ears, but none of the others ones seem to make use of this.

Otherwise I'm willing to also pull the potion out as well, at least we could make use of different coloring layers in that case.

T-BoneShark:
> I took a stab at that and ended up with this: [master...zorgjbe:fix/medical-injector](https://github.com/Ben987/Bondage-College/compare/master...zorgjbe:fix/medical-injector), which seems to have no problem with the Yoked pose (if that was what you were trying to do?). Feel free to cherry-pick that one commit on top of yours.

Thanks for putting the effort into doing that, I appreciate it. I took a look at your changes which still has the same issues in that `AllowPose: ["Yoked"], HideForPose: ["OverTheHead", "BackBoxTie", "BackElbowTouch", "BackCuffs", "AllFours"]` is ignored specifically when used in the `SpankingToys....` assets (since they really are just used for type info).

I was able to successfully use the aforementioned properties when created as it's own thing, along with the yoked pose and there it made less calls for unsupported poses, so I think I'll leave it as it's own asset (and not a `SpankingToys...`) unless we specifically want to move it back in.

zorgjbe:
I don't understand. It's working okay here, if "working" means the injector has a Yoked pose and no 404 for the other ones' assets. Looking more closely, I think I see what you mean: the main asset doesn't get those properties set… Let me try something else.

I've just pushed a version with those properties under your item's "Typed" definition, and they do seem to show up under Property now, and there is no asset loading in that case.

T-BoneShark:
> I don't understand. It's working okay here, if "working" means the injector has a Yoked pose and no 404 for the other ones' assets. Looking more closely, I think I see what you mean: the main asset doesn't get those properties set… Let me try something else.
> 
> I've just pushed a version with those properties under your item's "Typed" definition, and they do seem to show up under Property now, and there is no asset loading in that case.

Thanks! I believe it's all sorted now, also had a leftover line in one of the csv files I had to remove.

Ben987:
Very nice!  I'm sure that players will love it.  Thanks a lot :)

PR #3420, opened 2022-05-10
Fluffy twin tails hairstyle

MoonlightGleam:
Added Front hair style 27 and Back Hair style 48 with lovely gradient posibilities


Ben987:
This looks great!  I'm curious to see how player will mix and match it.  Thanks a lot :)

PR #3421, opened 2022-05-10
Pony harness gag fix

MoonlightGleam:
Updated the priority for the post, as it stood behind some items


elliesec:
Presumably you'll also want to make these changes on the `ItemMouth2` and `ItemMouth3` versions?

MoonlightGleam:
> Presumably you'll also want to make these changes on the `ItemMouth2` and `ItemMouth3` versions as well?

Cant we use something like CopyConfig for those?

elliesec:
> Cant we use something like CopyConfig for those?

Afraid not at the moment. It's something I've considered, but the usage would have been limited, and we haven't really had any items that would benefit significantly from it up until now, where the sheer number of layers is obviously a lot of copy/paste code.

Ben987:
Sure!  Thanks for these changes :)

PR #3422, opened 2022-05-10
Bugfix: 404s corset items in the `Cloth` group

elliesec:
## Summary

The Leather Corset Top and the Steampunk Corset Top in the `Cloth` group cause 404s when the character is put in most upper body poses. This is because they inherit their `AllowPose` entry from the `ItemArms` group, but proxy to the `Corset` group using `DynamicGroupName`, and the `Corset` version of the items have a more restricted `AllowPose` array. This adds appropriate `AllowPose` properties to these items, fixing the 404s and making the `Cloth` version of the tops visible again in all upper body poses.


Ben987:
Sure!  Thanks for that fix :)

PR #3423, opened 2022-05-10
Catsuit Collar

EmilyR42:
- Added Catsuit Collar in both ClothAccessory and Necklace slots


Ben987:
Good idea!  Thanks a lot :)

PR #3425, opened 2022-05-10
Beta Fix: Casing of PantyLine layer in Harem Pants 2

elliesec:
## Summary

The assets all use `PantyLine` whereas the layer was named `Pantyline`, causing 404s on the actual game servers.


Ben987:

PR #3426, opened 2022-05-10
Modular Item Difficulty Fix

Nina-1474:
The base difficulty of modular assets was being double-counted in the struggle challenge difficulty.
This is because the calculation of `Property.Difficulty` was including the base asset's difficulty with added difficulty from selected modules, but it's already accounted for in the item's main `Difficulty` property.


Ben987:

PR #3427, opened 2022-05-10
Fix profile nickname breakage

zorgjbe:
The introduction of nicknames made it a bit too crowded visually, or broke it in some cases. Rework it so things can be omitted when unnecessary, and change the line height to have better visual separation.

Disregard the BCX/BCE indicators. I just had those enabled when taking the screenshots

Broken on NPC screen:
E6MuVno

New version (for someone):

New version (for yourself):

New version (for NPC):


Ben987:
That looks pretty good, thanks!

PR #3343, opened 2022-04-14
adding/uploading: Clothing from 

gatetrek:
adds a few more styles of panties, 
harem gloves and stocking and a reworked off the pants
and 3new hair styles

credits to EmilyFoxx for all the assets
Screenshot_25

PR #3428, opened 2022-05-10
Reinstates removed `AllowPose` for original Harem Pants

elliesec:
## Summary

As part of #3343, the `AllowPose` entry got removed from the original Harem Pants item, resulting in 404s in the kneeling pose. This PR reinstates the `AllowPose` entry, fixing the issue. I've also removed the redundant `Horse` directory under `ClothLower`, and several duplicated assets for the original Harem Pants (duplication verified via MD5 hash) which were added back as part of #3343.


Ben987:

PR #3430, opened 2022-05-11
Medieval rack

MoonlightGleam:
Greatly reduced the dificulty of the racks modules and, hopefully fixed the overall dificulty with the new fixes.


Ben987:
Good idea!  Thanks :)

PR #3438, opened 2022-05-11
Bugfix: `ExtendedItemLoad` fixes

elliesec:
## Summary

This fixes a couple of bugs in the `ExtendedItemLoad` function, which have probably been around for a while:

1. When a typed item whose first option has a `Type` of `null` is applied, no character refresh is applied, even if the `null` type has properties attached
2. When a typed item whose first option is blocked by a player is applied, the item's `Property` is set to the type's configuration option rather than the `Property` inside that option.

This PR fixes the second bug directly, and changes the load function so that a refresh is always called if the item has a `Property` defined (which should be in most cases) after resolving the correct initial type accounting for blocked types.

## Steps to reproduce

The first bug is most obvious using the new tint functionality, but can be seen for other effects:

1. In a single-player area, equip an item that has no tint on its asset, but has a tint on the default (`null`) type (not sure if there are any such assets yet, but the changes to the box hood in #3437 are an example)
2. Change the colour of the tint layer, and notice that the colour of the screen does not change
4. Changing the item's type or visiting a chatroom fixes the issue

The second bug looks like a long-standing one:

1. Block the default type of the Hemp Rope arms item (the Wrist Tie option) - I'm... actually not sure if there's a way to do this without console, but there's handling for the situation in the code. Any ideas @Nina-1474?
2. Equip the hemp rope. Note that it appears to have switched to the boxtie as expected (because the first option is blocked). However, on closer inspection, the item's `Property` is actually the `Option` object from the configuration in `Female3DCGExtended.js`, not the `Property` object.


Ben987:

PR #3440, opened 2022-05-11
Medieval rack: Various fixes

MoonlightGleam:
Fixed the dialog error which showed "Name" after the players name


Ben987:
Thanks for these fixes :)

PR #3441, opened 2022-05-12
Fix autoscroll

ace-1331:
- fixed the behavior of returning a chatroom from the friendlist when entering it via a beep notification to be in line with other screens that return you to the chatroom (ie: scrolls the chatroom to the end if you were already scrolled to the end)


Ben987:
Sure!  Thanks for that fix.  Cheers :)

PR #3443, opened 2022-05-12
Fix: Chinese translation in GGTS.

remiliacn:
### What is this change

A follow up fix for #3433, forgot to fix the translation in GGTS.

Also have made another commit to fix a small translation error in activity.


Ben987:
Yes!  Thanks for that fix :)

PR #3446, opened 2022-05-12
Remove the Leather Whip, Crop, and Vibrating Wand items

zorgjbe:
As those items are only action-based, and don't actually appear on the character, replace them with their SpankingToys counterparts.

This was previously submitted as #3209, but this one takes care of updating all the SP dialogs and screens to use the SpankingToys subitems.

It's now possible for dynamic items to be correctly identified in dialog states (so you can add specific lines for the crop, or the feather for that matter) using the usual `${groupname}${assetname}` syntax.


BondageClub/Scripts/Import.js
```diff
@@ -46,8 +46,7 @@ function ImportBondageCollege(C) {
 			if ((localStorage.getItem("BondageCollegeExportArmbinder") != null) && (localStorage.getItem("BondageCollegeExportArmbinder") == "true")) InventoryAdd(C, "LeatherArmbinder", "ItemArms", false);
 			if ((localStorage.getItem("BondageCollegeExportChastityBelt") != null) && (localStorage.getItem("BondageCollegeExportChastityBelt") == "true")) InventoryAdd(C, "MetalChastityBelt", "ItemPelvis", false);
 			if ((localStorage.getItem("BondageCollegeExportCollar") != null) && (localStorage.getItem("BondageCollegeExportCollar") == "true")) InventoryAdd(C, "LeatherCollar", "ItemNeck", false);
-			if ((localStorage.getItem("BondageCollegeExportCrop") != null) && (localStorage.getItem("BondageCollegeExportCrop") == "true")) InventoryAdd(C, "LeatherCrop", "ItemPelvis", false);
-			if ((localStorage.getItem("BondageCollegeExportCrop") != null) && (localStorage.getItem("BondageCollegeExportCrop") == "true")) InventoryAdd(C, "LeatherCrop", "ItemBreast", false);
+			if ((localStorage.getItem("BondageCollegeExportCrop") != null) && (localStorage.getItem("BondageCollegeExportCrop") == "true")) InventoryAdd(C, "SpankingToysCrop", "ItemHands", false);
```
Nina-1474, line 49: How does this add handle the prerequisite buy group's main "SpankingToys" item not being owned? Same question with the asylum vibrating wand, but here it's guaranteed to be the case. My guess is the crop just won't be usable until another spanking toy is bought, since the handling of buygroup items is currently only completed in the shop.
Perhaps those steps should be covered under `InventoryAdd` instead and maybe normal buy groups should too since they'll be auto-added next login anyway. If that's a lot more effort/testing, short-term this line could probably just add the generic "SpankingToys" too.

zorgjbe: I honestly have no experience with buy groups, apart that as far as I've understood, when you buy one item you get all of them. I *think* it ought to work, because the "main" SpankingToy uses `DynamicAllowInventoryAdd` to do its magic, meaning it's not even required to buy it, only the "subitems".

I'll try to give it a spin though, just in case that fails to work.

Ben987:
This one is huge!  I won't be able to test all NPCs, but we have betas for that.  Thanks a lot :)

PR #3449, opened 2022-05-13
Fix: issue with Chinese OOC not blocked in Extreme difficulty.

remiliacn:
### What is this change.

PR #3362 introduced Chinese garbling and Chinese OOC syntax with the full width left parenthesis `（`. But in extreme difficulty, OOC should be blocked. In the current code, it is lacking the check of this specific full width left parenthesis thus renders the immersion restriction useless.


Ben987:
Sure!  Thanks for that fix :)

PR #3450, opened 2022-05-13
Improvement: Standardizes `AllowLock` behaviour

elliesec:
## Summary

Currently, the behaviour of `AllowLock` and the ability to define lockable types on items is quite unintuitive to work with for asset creators. In order to make individual types lockable, the definitions _must_ be subtractive: you need to add `AllowLock: true` on the base asset, and can then define which types _don't_ permit locks using `AllowLockType` on the base asset and `AllowLock: false` in the item's property. It's currently impossible to selectively pick types which _do_ permit locks on an asset with a base `AllowLock` value of `false`.

This PR changes that, and removes the need for explicitly defining the `AllowLockType` array in items using the typed archetype. If an asset has an `AllowLock` array defined, that is exclusively used for determining whether or not a given type can be locked. If not, then the asset's base `AllowLock` value is used. This means that you can set `AllowLock: true` on an asset and `AllowLock: false` in typed item options to subtract types, or set `AllowLock: false` on the asset and `AllowLock: true` in typed item options to add types, and there is no longer a need to set `AllowLock` in an item's property to determine whether or not it can be locked.

This lays the groundwork for `AllowLock` support in modular items at the type/option level, which will come in a separate PR.

## Changes in this PR

* Removes support for `AllowLock` from item properties - instead, only the `AllowLockType` array and the `AllowLock` property at the asset level are needed
* Adds support for `AllowLock` in typed item options (not inside `Property`), which is used to automatically generate a typed item's `AllowLockType` array
* Updates the logic for lock-related buttons in `Dialog.js` and creates a new `InventoryDoesItemAllowLock` function, which can be used to determine whether an item can be locked (taking its type into account) - this has been used to replace several lockability checks across the game, and also notably simplifies the nesting in the `InventoryLock` function
* Updates the two existing items (Bondage Bench & Fur Blanket Wrap) making use of type-level locking accordingly
* Tweaks the behaviour of typed item menus to remove locks if changing from a lockable type to a non-lockable one, rather than letting validation handle it on the receiving end
* Updates the validation to correctly reflect the changes
* Updates the type definitions to reflect the above changes

## Testing

I've pretty thoroughly tested this with both non-archetyped items, typed items, modular items (although they don't currently support type-level lockability), and the Bondage Bench and Fur Blanket Wrap which make use of fine-grained locking to verify that locking still works as expected.


zorgjbe (approved):
Looks good to me 👍.

Ben987:
That looks very good.  Thanks for that one :)

PR #3453, opened 2022-05-14
Beta fix: Prevents non-tinted modular items from setting a Tint property

elliesec:
## Summary

This fixes a minor bug with modular items where items that didn't permit tints would always set an empty tint array on item properties when merging module options. This resulted in validation catching the item updates as invalid because the item didn't permit tints. This PR prevents the extraneous `Tint` property from being sent for modular items that don't permit them, and prevents these validation warnings.


Ben987:
Thanks for that fix :)

PR #3454, opened 2022-05-14
Feature: Adds `AllowLock` support to modular items

elliesec:
## Summary

This PR builds on top of #3450 and adds support for `AllowLock` to modular item options. This utilises the new `AllowLockType` functionality from #3450, with modular items building their own `AllowLockType` arrays as appropriate. As there is potential for two options in different modules on the same item to declare conflicting  `AllowLock` properties, I've implemented the following behaviour for those cases:

* If any active module option declares `AllowLock: true`, then the item will be lockable (even if another active module option declares `AllowLock: false`). The logic behind this is that if one "piece" of an item is lockable, then the item can be locked. The fact that some other piece isn't lockable shouldn't prevent the pieces that are lockable from being locked
* If no active module options declare `AllowLock: true`, but some active module option declares `AllowLock: false`, then the item won't be lockable (overriding the `AllowLock` property of the base asset)
* If no active module option declares any `AllowLock` value, then the `AllowLock` property of the base asset will be used to determine whether the item can be locked (`false` by default)

So, in other words:
Option-level `AllowLock: true`  ----->  Option-level `AllowLock: false`  ----->  Asset-level `AllowLock`.

## Changes in this PR:

* Modular item configuration now permits an `AllowLock` property to be defined at the option level
* Option-level `AllowLock` properties are now used to generate the `AllowLock` and `AllowLockType` properties on a modular item's base asset
* Refactors the `AllowLockType`/`AllowLock` reduction logic from #3450 into a common `TypedItemSetAllowLockType` function for use in both typed & modular items
* Updates the behaviour of modular item menus to remove locks if changing from a lockable type to a non-lockable one, rather than letting validation take care of it on the receiving end
* Updates the behaviour of modular item menus to rebuild the inventory and menu buttons on exit (as the item's locked state may be automatically removed as part of a module switch)
* Updates the type definitions to reflect the above changes

## Testing

Currently, no modular items make use of this, but I know of at least one (the Wooden Rack) whose creator would like to. I've tested several `AllowLock` configurations on modular items, with both `AllowLock: true` and `AllowLock: false` on the base asset, and all behaviour is as expected (as outlined above). No validation changes are needed as part of this change, as all validation is handled by the `AllowLock` and `AllowLockType` changes in #3450.


elliesec:
@MoonlightGleam - once this PR has been merged, you should be able to implement option-level locking on the Wooden Rack. Note that `AllowLock` should be defined inside the option, but _outside_ the `Property`:
```js
{
    Options: [
        {
            AllowLock: true,
            Property: {/* ... */}
        },
        // ...other options
    ]
}
```

MoonlightGleam:
This was way faster than I expected , I will draft the fix and wait for the merge.
Thanks for your work!

elliesec:
> This was way faster than I expected , I will draft the fix and wait for the merge. Thanks for your work! If the item is not lockable, I asume AllowLock: false wouldnt be neccesary. Or is it?

Sorry, I overlooked the question, but put the answer in a comment on your PR. In short, if the item's not lockable by default, then no, you shouldn't need to set `AllowLock: false` anywhere, just `AllowLock: true` on the lockable options.

elliesec:
This has now been updated and re-tested, and should now be ready for review.

Ben987:
It's time to test this one!  Thanks a lot :)

PR #3455, opened 2022-05-14
Beta fix: Updates slime girl priority to render above wings

elliesec:
## Summary

This is just a minor update to the priority of the slime girl layer for the `ItemArms` slime item so that it no longer appears to clip through wings.


Ben987:
Absolutely!  Thanks :)

PR #3457, opened 2022-05-14
Medieval rack: locks

MoonlightGleam:
ℹ️ This PR depends on https://github.com/Ben987/Bondage-College/pull/3454. Will be merged once that is ready.
Adds the option to lock certain modules on the rack


BondageClub/Assets/Female3DCG/Female3DCGExtended.js
```diff
@@ -2747,6 +2747,7 @@ var AssetFemale3DCGExtended = {
 							Options: [
 								{}, // t0 - No
 								{
+								AllowLock: false,
```
elliesec, line 2750:
```suggestion
```

Just a note here - you don't need to explicitly set `AllowLock: false` here - you can just leave it out completely. If no `AllowLock` value is defined on the option, it'll inherit the behaviour of the main asset (which is `false` if you don't explicitly set it to `true` at the asset level).

This means configure assets in either of two ways:
1. Don't set `AllowLock` at the asset level and just set `AllowLock: true` on the options that should be lockable. Useful if only a couple of options should be lockable.
2. Set `AllowLock: true` at the asset level and just set `AllowLock: false` on the options that _shouldn't_ be lockable. Useful if most options should be lockable, but a few shouldn't.

Same comment for the other places where you've set `AllowLock: false`.

elliesec:
Asset check failures should be resolved once #3454 has been merged.

BondageClub/Assets/Female3DCG/Female3DCGExtended.js
MoonlightGleam: Will do, I was unsure so Id rather put it amd ve wrong about it

Ben987:
Thanks a lot you two :)

PR #3458, opened 2022-05-14
Fix InflationAmountSetLight dialog

anniclub:
This seems to have been a typo: There is only InflationAmount Light, but no InflationAmount None, therefore in-game there is currently the message `(MISSING PLAYER DIALOG: InflationAmountSetLight)`.


Ben987:
Yes!  Thanks for that fix :)

anniclub:
@Ben987 I think we have to revert: That last commit slipped through but wasn't reviewed yet (I suppose) and breaks lint

Nina-1474:
I've pushed a fix for the issue in #3463 so that could be done instead of a revert.

elliesec:
@anniclub - I've got you down as "anniclub" in the changelog & game credits. Just let me know if you'd prefer another name and I'll update.

anniclub:
@Ben987 I just realized that my second commit reveals a bit too much information. Not super bad, but if there is a way for you to amend the commit and just remove that part from the metadata, that would be awesome.

Also, I'm really sorry for this whole confusion, didn't mean to cause that much trouble.

PR #3460, opened 2022-05-15
Refactor: Random code cleanup and simplification

Jomshir98:
~~*This is a low-priority PR and should be merged after PRs that do add features or fix things.*~~
~~*I will keep it as draft until majority of current PRs are merged.*~~

This PR has no impact on functionality.


BondageClub/Scripts/Dialog.js
```diff
@@ -969,11 +969,11 @@ function DialogMenuButtonBuild(C) {
...
-						&& (Player.CanInteract() || DialogAlwaysAllowRestraint() || Item.Asset.AlwaysInteract)
-						&& (!IsGroupBlocked || Item.Asset.AlwaysExtend)
-						&& (!Item.Asset.OwnerOnly || (C.IsOwnedByPlayer()))
-						&& (!Item.Asset.LoverOnly || (C.IsLoverOfPlayer()))
-						&& canUseRemoteState === "InvalidItem")
+					&& (Player.CanInteract() || DialogAlwaysAllowRestraint() || Item.Asset.AlwaysInteract)
+					&& (!IsGroupBlocked || Item.Asset.AlwaysExtend)
+					&& (!Item.Asset.OwnerOnly || (C.IsOwnedByPlayer()))
+					&& (!Item.Asset.LoverOnly || (C.IsLoverOfPlayer()))
+					&& canUseRemoteState === "InvalidItem")
```
zorgjbe, line 978: That comes from my personal style for multi-line conditions 😟. I don't really mind it being changed (and can stop doing that if it's kinda non-standard), but I find that extra indent makes it easier to tell the actual block of code from the sea of checks.

zorgjbe (approved):
Looks good to me, thanks for the `for...of` cleanups 👍. Just one small remark from me about style (what else 😁).

BondageClub/Scripts/Dialog.js
Jomshir98: That change came just from following ESLint indent rule.
I also refactored it to follow my personal style - is it better now?

elliesec: I think generally not being able to tell code from conditions in stuff like this is a pretty good argument for the use of braces.

Sidiousious: The more lines you start to get in an if-clause the more you should consider creating a separate function for it, with the if split into multiple ones that return true/false to replace the original clause. Even with clearest formatting these still require much more effort to parse than a function name. This particular one could be just `if (DialogPlayerCanControlItem(C, Item, IsGroupBlocked))` or similar.

zorgjbe: Can't argue with ESLint 😅. I think I'd do what Sid says, extract it as a helper function.

elliesec: I fully agree that extracting into a function is nicer (and there are hundreds of places in BC where that should probably happen), but I'd also say that it's probably outside the scope of this PR, as it's primarily just a linting pass.

BondageClub/Scripts/Character.js
```diff
@@ -1147,9 +1147,22 @@ function CharacterHasNoItem(C) {
...
+			(
+				C.IsNpc() ||
+				!(
+					C.Appearance[A].Asset.Group.BodyCosplay &&
+					C.OnlineSharedSettings &&
+					C.OnlineSharedSettings.BlockBodyCosplay
+				)
+			)
+		)
+			return false;
```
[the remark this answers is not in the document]
Jomshir98: I honestly don't like that style. I would much more prefer changing the whole thing to `return !C.Appearance.some(<predicate>)`
I only added comments and seems I also during that discovered a possible logic bug. But again, this isn't PR that should change anything, only format/cleanup PR

BondageClub/Scripts/Character.js
```diff
@@ -1153,10 +1153,14 @@ function CharacterIsNaked(C) {
...
 		if (
 			(A.Asset != null) &&
+			// Ignore items
 			(A.Asset.Group.Category == "Appearance") &&
+			// Ignore body parts
 			A.Asset.Group.AllowNone &&
+			// Always ignore all cosplay items
 			!A.Asset.BodyCosplay &&
 			!A.Asset.Group.BodyCosplay &&
+			// Ignore cosplay items if they are considered bodypart (BlockBodyCosplay)
```
zorgjbe, line 1163: 👍

zorgjbe (approved):
Last few changes look good to me. It's up to you @Ben987 now 🚢.

BondageClub/Scripts/Dialog.js
```diff
@@ -126,6 +126,17 @@ var DialogSelfMenuOptions = [
 	},
 ];
 
+/**
+ * Returns character based on argument
+ * @param {Character | string} C - The characer to get; can be `"Player"` to get player or empty to get current
+ * @returns {Character} - The actual character
+ */
+function DialogGetCharacter(C) {
+	if (typeof C === 'string')
```
elliesec, line 135: I know it's not in the lint rules (yet), but should probably aim for quote consistency (and double seems to be the overwhelming majority in this particular file).
```suggestion
	if (typeof C === "string")
```

BondageClub/Scripts/Character.js
```diff
@@ -653,7 +653,7 @@ function CharacterLoadOnline(data, SourceMemberNumber) {
 				Char = Character[C];
 
 	// Decompresses data
-	if (typeof data.Description === "string" && data.Description.startsWith("╬")) {
+	if (typeof data.Description === "string" && data.Description.startsWith(String.fromCharCode(9580))) {
```
elliesec, line 656: I feel like it's probably worth refactoring this out into a sensibly-named common constant (e.g. `CompressedDescriptionMarker` or something equally catchy). It's a little bit "magic value"-like as is. There's also one other instance of `╬` being used in `Login.js` which hasn't been touched by this PR (which would be the only remaining instance of it in string literal form).

I'd say it's probably worth fixing the instance in `Login.js` as part of this so that we don't have it existing in two forms across the codebase, but up to you whether or not you want to include the refactoring as part of this PR.

elliesec (commented):
A couple of minor comments/queries, but otherwise looks good to me.

elliesec (approved):

PR #3463, opened 2022-05-16
Asset Check ArchetypeConfig fix

Nina-1474:
Fixes the commit in #3458 which is breaking the asset checks.
That change was last-second and unrelated to the actual PR as stated but is still worth having.


elliesec (approved):

elliesec:
This and the extra commit in #3458 both look fine to me. Not too urgent if Ben doesn't catch it before he deploys beta 3 - the game will still run fine.

Ben987:
Sure!  Thanks for that fix :)

PR #3464, opened 2022-05-17
Minor Changes to Speech.js

Shiranui-Izayoi:
Changed and fixed a few things in Speech.js.

Fixed an issue with the Baby Talk function where it was both translating a character and also adding it in. So "s" would be changed to sh then because they were all `if` statements it would find the a-z part and add in resulting in it becoming "shs". This was accidentally added when the code was cleaned up and apparently wasn't found. With current behaviour it becomes nearly impossible to understand the baby talk.

Changed the if statements inside garbles to else if. It won't effect current behaviour simply make it so that it isn't having to check every statement for every letter. So as an example in the Very Light Garble the letter "t" is changed to an "e" but it then has to go through 13 more checks such as for Accents/Latin and Cyrillic characters.

Finally I added a new function called `isAnyPunctuationOrSpace` which serves a similar function to `isStandardPunctuationOrSpace`, but is meant for lower gag levels as well as baby talk. Where keeping more punctuation makes sense and only makes sentences easier to read. The new function can have any punctuation added to it. Since I didn't want to add more punctuation to the existing funtion as that would heavily effect the higher gag levels especially total.

**EDIT:** Ow yes I did forget to mention one change I made. I changed how the "t" is translated in the baby talk function.

I changed it from "st" to "th" this is because I felt that it made more sense with trying to pronounce the words. Since I could not figure out how you would pronounce "there" as "sthere" etc. So I changed it to "th" while this does result in words like "thhere" I think this makes more sense for a sort of lipsing baby talk.

Sorry I forgot to mention it in the inital posting.


Ben987:
Can you show us some "before and after" of a garbled chat after that change?  Thanks!

Shiranui-Izayoi:
> Can you show us some "before and after" of a garbled chat after that change? Thanks!

I suppose you mean me? If so sure I can show you how the regressed talk looks after the change. But it will look very different because of fixing the issue with the if statements. If you mean something else could you specify what? Since I am not sure what you are asking exactly sorry.

Also I apologise if you meant this towards Ellie.

Ben987:
@wildsj , this script changes the garbling done by the Nursery Milk.  Would you like to test and approve it?  Thanks a lot :)

elliesec:
@Ben987 - This change actually brings the baby speech back closer to how it behaved before #3308 (which modified its behaviour). Here's a quick comparison:

```
Plaintext:      The quick brown fox jumps over the lazy dog
Before #3308:   sthe quicw brown fox jumpsh over sthe wazy dog
Master:	        stthe quicwk brown fox jumpshs over stthe wlazy dog
After this PR:  thhe quicw brown fox jumpsh over thhe wazy dog
```

As you can see, the changes in #3308 caused some characters to be double-garbled. The only difference now between the behaviour in this PR and the behaviour before #3308 is the handling of the letter `T` (which now garbles to `th` instead of `st`), as mentioned in the PR description by @Shiranui-Izayoi. It's probably worth getting @wildsj's opinion on that particular change, but on the whole this is a fix to bring the garbling closer to its original behaviour.

Ben987:
Time to test this one then!  Thanks a lot you two :)

PR #3466, opened 2022-05-18
Bugfix: Stops locks from being deleted by the variable height load function

elliesec:
## Summary

This is a small fix to the variable height load function which ensures that it doesn't remove locks from items. Previously it was just cloning the config property and overwriting the item's properties completely, resulting in locks getting removed. The issue wasn't really present before the rope suspension code was moved into an archetype in #3417 because the rope wasn't lockable, so was only noticeable once the archetype was applied to the chains.

## Steps to reproduce

1. Equip chains in the `ItemArms` group
2. Lock the chains with any lock
3. Select the Suspended Hogtie option for the chains and confirm selection
4. Notice that the lock has been removed


elliesec:
@Ben987 - I've checked it against all of the items that use the variable height script, and verified that it's working. I've just added an additional null safety check for the case where the item's `Property` isn't defined, although that shouldn't happen in theory.

Ben987:
Time for hotfixes!  Cheers :)

PR #3467, opened 2022-05-18
Feature: Adds lock icons to extended item menu preview images

elliesec:
## Summary

This PR is a minor enhancement to menus for extended items using the archetype system which adds a "Locked" icon to the main preview image if the item is locked. Since some extended items remove locks automatically if switching from a lockable to a non-lockable type, I thought a visual indicator in the menu of whether or not the item is locked would be a useful addition (in the future, it might be nice as a convenience to be able to lock/unlock an item directly from its extended menu, but that's a PR for another day).

## Changes in this PR

* Updates the type definitions for `DialogInventoryItem` so that the `Icons` property now uses a string union type rather than just being typed as a `string[]`
* Adds an `InventoryIsAllowedLimited` function as a shorthand for identifying whether an item is an allowed limited one, and uses it to simplify some icon logic
* Adds a lock icon to the main preview image in extended menus when the item is locked for items using any of the four extended archetypes (the variable height case doesn't display properly right now, but will start working automatically once #3466 has been merged)


Ben987:
Sure!  Thanks for that change :)

PR #3468, opened 2022-05-18
Feature: Defaults chatroom language to translation language on room creation

elliesec:
## Summary

This PR makes a minor change to chatroom creation which defaults the chatroom language to the user's translation language if available. As is, most chatrooms seem to get created with the default "EN" language, presumably because it's the path of least resistance and players don't manually update it before creation. This change should hopefully be a more useful default for the majority (but may be slightly more work for those wishing to create chatrooms in languages other than their translation language, who I'm hoping are a relatively small contingent of players).


Ben987:
Lets leave it as it right now.  Some players will play in a foreign language but will want to create rooms in English by default.

PR #3472, opened 2022-05-20
Translation fixes

Anonymous-WghrYkBGUjBt:
Fix mistakes (one typo, two mis-translation due to lack of research into tennis terms)


Ben987:
Sure!  Thanks for that fix :)

PR #3474, opened 2022-05-21
Kitty Hood Tweaks

Ada18980:
I contacted Lucille about the kitty hood and I made it more colorable, as well as fixing up a few pixels and making the lock layer colorable.


Ben987:
Absolutely!  Thanks for the tweaks :)

PR #3477, opened 2022-05-25
Multi Colour Pleated Skirt

Shiranui-Izayoi:
Simply removed `AllowColorize: false` from the Pleated Skirt and gave them layer names.


Ben987:
Sure!  Thanks :)

PR #3481, opened 2022-05-29
Fix issue with extended items not accounting for character nicks

zorgjbe:
It turns out that part was missed, resulting in type change messages on assets letting the character original names bleed through.


Ben987:
Good catch!  Thanks a lot :)

PR #3482, opened 2022-05-30
Sturdy leather belts color and sound

karame1:
A few enhancements for sturdy leather belts.

- grayscaled layers for better colorability
- separate belt and metal layers
- separate layer for each belt (belts in a color group, metal in an other one)
- remove types to avoid duplication of belt layers that are common in each version (one, two or three belts)
- add a buckle sound

A prewiew was posted in work-in-progress on discord and received good feedback. https://cdn.discordapp.com/attachments/721493158625804298/979350768517808178/belts_comparison.png


Ben987:
Great idea!  I'm curious to try it.  Thanks a lot :)

PR #3485, opened 2022-05-31
cn translation

dynilath:
## Summary
* fix some error in activity text
  There is a translation problem that interpret "butt" as "小腿", more like a typo. And "ribs" should not be "肋下".

* cn translation for introduction class
  Dialog text of introduction class, translated into CN by me.


Ben987:
This is huge!  Thanks a lot for the translation :)

PR #3487, opened 2022-06-07
Fix some issues with the asset loading

zorgjbe:
This is a follow up to #2987, as Ben reported that your character could become invisible if you left the tab running for too long in the background. Sid suggested moving the automatic purge inside the main timer loop instead of having it in a "dangling" `setTimeout` to make both the drawing and purging happen as part of `requestAnimationFrame`, hopefully ensuring the cache won't be cleared if nothing had to be drawn.

Then I noticed that fully missing assets would cause that as well, because of how brittle the "refreshCanvas" condition is — my testing only did transient failures, not complete ones since there should be no assets with missing files (don't quote me on that). This required keeping track of both loaded assets and those currently missing (experiencing a failure of any kind), because those would prevent the condition from ever becoming true, resulting in blanks, holes, and invisibility until something else would cause a canvas redraw and actually draw those assets that were always available (the light latex bra has no yoked pose, for example, if you want to try that out).

I'll keep checking for problems until the beta because all those counters make the code paths very fiddly, I don't like it and I'm not sure I caught all cases, but hey there's a reason why caching is hard 😅.


Ben987:
Time to test this one!  Thanks a lot :)

PR #3489, opened 2022-06-08
Fix missing handheld activities on the vulva piercing slot

zorgjbe:
Since the "meta" SpankingToy item doesn't appear in that group, you couldn't use any of the handheld items there, even though some activities allow it.


Ben987:
Sure!  Thanks for that one :)

PR #3492, opened 2022-06-10
Always return exactly matching room-names

Nina-1474:
When the text a user enters into the Chat Search box fully matches a room name's, the ignore-room list and filter-out terms will not apply and that room will be included in the results. In particular this ensures a player can join a room selected from their friend list when that room would normally be filtered out.


Ben987:
Good idea!  Thanks a lot :)

PR #3494, opened 2022-06-10
Fix: Head harness Heavy Type

MoonlightGleam:
Fixed an error in the type definition


Ben987:
Sure, thanks for that fix :)

PR #3495, opened 2022-06-11
Beta fix: Increases priority of back layer for the Asymmetric Skirt

elliesec:
Increases the priority of the back layer of the Asymmetric Skirt so that it renders in front of back hair (see image below for demonstration of the bug).

image


Ben987:
Absolutely!  Thanks :)

PR #3496, opened 2022-06-11
Add a Jacket to the Cloth group

zorgjbe:
This adds a colorable jacket with its hood, potentially allowing ears to poke out of it.

Note that due to how the ears look better when behind the hood, it intentionally has a much higher priority which does cause some issues with a couple of the ears as they're too small to really poke through.

In-game preview of how it looks: 
jacket-1
jacket-2
jacket-3

Assets courtesy of RedStacey.


Ben987:
Thanks a lot your two, RedStacey seems very confident we can slide it in the third beta.  Cheers :)

PR #3497, opened 2022-06-11
Fix Poker & Platform image caching

zorgjbe:
Fix the few missing reference crashers for the cache as its name changed between the initial implementation and now.

Note that now that the cache purges itself from time to time, it could result in a bit of animation glitchiness (namely, high-speed blinking) as the missing assets get reloaded.


Ben987:
Sure, I'm curious to try that one.  Thanks :)

PR #3498, opened 2022-06-11
Hide the Hair accessory groups on the Kitty and Latex Dog Hood

zorgjbe:
Additionally, remove Ears from the typedef since it doesn't actually exist.


Ben987:
Sure!  Thanks a lot :)

PR #3500, opened 2022-06-12
Stitch gag

MoonlightGleam:
Added stitched lips:
-Colorable thread
-Lips react to expresions and are colored the same as character lips


Ben987:
This is pretty hardcore but players can always block it if it's too violent for them.  Thanks a lot!

PR #3502, opened 2022-06-12
Safeguard against VariableHeightChange getting called with no item selected

zorgjbe:
Under certain conditions (which I haven't been able to reproduce, since it looks like another stars-must-align bug), `DialogFocusItem` can end up being null, mostly likely if the function gets throttled on the binder's side as the bindee struggles out of it, causing it to execute after the restraint has been removed.


Ben987:
Sure!  Thanks for that fix :)

PR #3505, opened 2022-06-13
Improvement/fix: Lock number validation improvements

elliesec:
## Summary

This makes a few improvements & fixes to the `ValidationSanitizeLock` function, which should help prevent players from getting stuck in owner/lover locks when they shouldn't. Namely:

* Owner locks will be removed if the player is not owned
* Owner locks with the player's number on them will be removed if the player's owner has prevented them from using owner locks on themselves
* Lover locks will be removed if the player has no lovers
* Lover locks with the player's number on them will be removed if one of the player's lovers has prevented them from using lover locks on themselves
* Lover locks with the player's owner's member number on them will be removed if one of the player's lovers has prevented their owner from using lover locks on them

These new checks are in addition to the existing checks which remove lover/owner locks with non-lover/non-owner member numbers on them.

## Testing

I believe I've tested every possible permutation of the player being owner/not owner/loved/not loved in combination with the owner and/or lover rules, and varified that the behaviour is as expected.


zorgjbe (approved):
LGTM! Thanks for handling this, @elliesec 👍.

Ben987:
Sure!  Thanks a lot for that fix :)

PR #3507, opened 2022-06-13
Revert the image cache

zorgjbe:
I found a memory leak on Firefox (on macOS), which seems to have been made more important by the new cache. For now, revert the cache to its previous implementation, just so that it's not a problem for a month.

I'll rebuild the patchset and file it again after the beta is over.


NoneNoname:
I made an actual fix in #3509, no need to revert the cache

Ada18980:
Thanks for letting me know. KD on bondage club should be unaffected, this will affect the standalone though. I can just wait till next months beta to merge

zorgjbe:
@Ben987 Okay, so I've seen report of Firefox still acting up, crashing the moment you enter a room, and throwing internal errors around. As I don't think I have the time to tackle that issue right now, nor have the willingness to feel pressured into finding a solution before the final release, that one should go in.

Ben987:
Very well, we can try again another month then.  Thanks for all your hard work.  I'll prepare a special fourth beta to make sure everything works.

PR #3508, opened 2022-06-13
Assets for Kyra

T-BoneShark:
Adds several new hairstyles, a hat with two variants, and a pair of sunglasses; all designed by Kyra!

- Adds a single and double bun to the back hair (Multiple layers exist for the double bun)
- Two new front hair styles
- A pair of sunglasses, can be dipped or swapped from flat to gradient looking
- Two Baseball cap flavors, forward and back.
  - Backwards one can have the strap go under or over the front hair.
  - Forwards one has several print options, like the ones present on the shirt.

As usual, preview below...

kyras_preview


Ben987:
This looks great!  Thanks a lot you two, I'm curious to try them :)

PR #3509, opened 2022-06-14
FIX: Firefox memory leak

NoneNoname:
These changes fix the memory leak on Firefox with the new cache

no need for #3507

Leak can be tested with this patch, before this fix, it shortly goes over GBs on memory in Firefox
```diff
diff --git a/BondageClub/Screens/Character/Login/Login.js b/BondageClub/Screens/Character/Login/Login.js
index 4a68242b3..6b6f39ca5 100644
--- a/BondageClub/Screens/Character/Login/Login.js
+++ b/BondageClub/Screens/Character/Login/Login.js
@@ -144,7 +144,7 @@ function LoginRun() {
 
        // Draw the character and thank you bubble
        DrawCharacter(LoginCharacter, 1400, 100, 0.9);
-       if (LoginThankYouNext < CommonTime()) LoginDoNextThankYou();
+       LoginDoNextThankYou();
        DrawImage("Screens/" + CurrentModule + "/" + CurrentScreen + "/Bubble.png", 1400, 16);
        DrawText(TextGet("ThankYou") + " " + LoginThankYou, 1625, 53, "Black", "Gray");
 
diff --git a/BondageClub/Scripts/ImageCache.js b/BondageClub/Scripts/ImageCache.js
index 832d110dd..6a86f4154 100644
--- a/BondageClub/Scripts/ImageCache.js
+++ b/BondageClub/Scripts/ImageCache.js
@@ -155,7 +155,7 @@ class CachedImage {
        isLoaded() { return this.state == CachedImageState.LOADED; }
 }
 
-let ImageCachePurgeDelay = 30 * 60 * 1000; /* minutes */
+let ImageCachePurgeDelay = 100; /* minutes */
 
 /**
  * The class responsible for loading and caching images
  ```

PR #3510, opened 2022-06-14
More image cache work

zorgjbe:
This tweaks the new cache :
- assets currently on characters were being cleared too quickly, causing holes and missing things on some refreshes.
- as pointed out by #3509, an awful memory leak caused by not deleting textures when the cache.
- an issue with alpha masks not being deleted when the canvas resets (same underlying issue)
- bumped the purge delay to one hour, because the cache uses half of that to consider an asset unused, and 15 minutes seems a bit too low.

Kudos to @NoneNoname for the leak fix. I've just tweaked a bit the solution to keep the cache unaware of anything rendering-related by making image lifetime explicit.


NoneNoname:
I closed my PR, and give this a test, it is stable in firefox

BondageClub/Scripts/Drawing.js
```diff
@@ -1318,6 +1318,13 @@ function DrawProcess(time) {
 		DialogLeave();
 	}
 
+	// From time to time, ping the cache with the list of drawn assets for
+	// every shown character, so that it doesn't prune those.
+	if (Math.floor(time % 1000) < 20) {
+		DrawLastCharacters.forEach((C) => {
+			CommonDrawWarmDrawnAssets(C);
+		});
+	}
```
Jomshir98, line 1327: This can break on slow machines or when tab is in background - by not hitting the small window for some time continuously.
I think saving time of last trigger and doing condition "if time since last is bigger than" will be more stable

zorgjbe: Fixed, yeah I wasn't fond of this and was weary of adding yet another global, buuuut it is how it is.

Ben987:
Time to test this one!  Thanks a lot everyone :)

PR #3511, opened 2022-06-14
Fix spacing on the InformationSheet

zorgjbe:
With the modifier running, a player with all Skills would see some overlap. Fix it by making the spacing 55px instead of 75px, matching the smaller spacing from the 1st column.

Here's how it looks: 
infosheet


Ben987:
That's an old one, thanks a lot :)

PR #3513, opened 2022-06-15
Fix: SideFluff color gradient

MoonlightGleam:
Added 6 layers to the sidefluff to make it capable of having a color gradient. 

image


Ben987:
This one has a few conflicts with other pull requests.  Can you resolve the conflicts and test again?  Thanks a lot :)

PR #3514, opened 2022-06-15
Fix: Head harness priority

MoonlightGleam:
Fixes in the priority of the Head Harness maks, so the main straps lie under the hair front.
Small fix to Side Fluff hair so the shaved base still lies underneath the straps


Ben987:
Sure!  Lets slide it in beta 3.  Thanks :)

PR #3518, opened 2022-06-15
Fix trial ownership owner locks disappearing

Nina-1474:
When a player has a null/empty `Owner` property the validation considers them not owned and removes owner locks. However this is empty during ownership trials so it removes locks it shouldn't. (From what I've investigated, it seems this might even be empty for full ownership sometimes too and only be set for NPC owners, but I'm not fully sure on that.)


elliesec:
`C.Owner` is in fact populated in the case of full ownership (or at least it should be). Might be worth investigating whether the `IsOwned` method could be made a little more robust without having knock-on effects elsewhere at a later date.

elliesec (approved):

zorgjbe:
It's not updated as soon as the trial ends (or at least something isn't), which requires a relog, at least from the POV of the owner.

Nina-1474:
> It's not updated as soon as the trial ends (or at least something isn't), which requires a relog, at least from the POV of the owner.

Yes I assumed the owner name would be passed back by the server but see it's only saved to the database, which is why the relog is needed.

Nina-1474:
> `C.Owner` is in fact populated in the case of full ownership (or at least it should be). Might be worth investigating whether the `IsOwned` method could be made a little more robust without having knock-on effects elsewhere at a later date.

I thought the same but didn't want to risk breaking anything else with this fix! I do wonder if there are other `IsOwned` cases that haven't considered the trial period.

elliesec:
At a glance, `IsOwned()` is used pretty much exclusively in single player content right now (for the most part). This was definitely the right fix at the moment, as I could believe there's stuff in single player that relies on its existing behaviour.

PR #3519, opened 2022-06-15
Fixes ownership check in validation

elliesec:
The lock validation changes in #3505 were using the `Character.IsOwned()` function, which doesn't count players as owned if on a trial.


elliesec:
Closing in favour of #3519.

PR #3521, opened 2022-06-19
Enable favourite of clothing in Extreme difficulty

Nina-1474:
The clothing menu doesn't need to hide the toggle permissions-mode button for the Extreme difficulty anymore, since #2699 ensures that it can't block or limit items and accessing the button allows clothing to be favourited.


Ben987:
Good catch!  Thanks a lot :)

